PR #6, opened 2017-01-07
os_windows.odin fixes

Fyoucon:
- Handle changed to int because some os functions return -1 as a invalid handle
- fixed os.open inverse invalid handle check
- fixed os.write where win32.WriteFile was assumed to return error but only returns a BOOL, error is found via GetLastError instead
- win32.WriteFile return type changed to BOOL to match actual function
- fixed os.read inverse BOOL check

### DIFF
--- a/core/os_windows.odin
+++ b/core/os_windows.odin
@@ -2,7 +2,7 @@
 #import "fmt.odin";
 
 
-Handle    :: uint;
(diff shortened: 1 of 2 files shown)


gingerBill:
I've made the changes myself as I've had to change some other code too but thank you very much! 😄

PR #14, opened 2017-01-29
Fix incorrect version number

ThisDrunkDane:
The version number is still 0.0.5e


gingerBill:
The next update will have a different version number so I'll fix it later.

PR #24, opened 2017-02-13
small additions & fix

adekto:


gingerBill:
`[]f32` is not the same as a `rawptr`. A slice holds count information so it's actually a different type (unlike C where arrays get "demoted" to pointer automatically).
However, I will add the extra constants to `sys/windows.odin`.

PR #30, opened 2017-03-30
Allow `main` to take the standard paramaters and return an exit code

zangent:
This just modifies the proc type checking in `check_decl.c` so that the acceptable types of procedure for the entry point go from just `proc()` to `proc()`, `proc(i32, []^byte) -> i32`, `proc(i32, []^byte)` or `proc() -> i32`

This allows Odin to be used for command line applications, like an asset packager that partially shares a code base with the game that uses it.

The checking code is a little verbose, but I couldn't find a sensible and safe way to make it more concise.

### DIFF
--- a/src/check_decl.c
+++ b/src/check_decl.c
@@ -252,10 +252,27 @@ void check_proc_lit(Checker *c, Entity *e, DeclInfo *d) {
 	    str_eq(e->token.string, str_lit("main"))) {
 		if (proc_type != NULL) {
 			TypeProc *pt = &proc_type->Proc;
-			if (pt->param_count != 0 ||
-			    pt->result_count != 0) {
+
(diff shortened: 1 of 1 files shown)


zangent:
I just noticed two things about this that may not make it the perfect implementation of this.

First of all, although it works fine in practice, I should have probably not assumed that what is normally just a double pointer (`const char**`) could be converted into a slice (`[]^byte`).

Second of all, it's not very Odin-like to receive an argc and an argv. A better solution would probably be to detect whether or not the program calls for arguments in the form of a `proc(args: []string)`, and if so, map the c-style data into a slice, perhaps right before `@.__$startup_runtime()`

I'm going to work on this, and make a new pull request tomorrow that feels much more in line with the way the rest of the language operates.

PR #32, opened 2017-04-03
Update os_windows.odin

Krognol:
Added an extra check to make sure the data slice size is greater than 0.


gingerBill:
In the latest version, a slice can be compared to `nil`. 
```
if data == nil {
    return nil, false;
}
```
This is much better than checking its length and data pointer manually.

PR #73, opened 2017-06-16
Building in linux

vassvik:
Building the current version of Odin in linux results in this: https://hastebin.com/qulekonudu.lua. The diff shows the changes i had to make to get it to compile and run without segfaulting.

Each error one by one:
`In file included from src/main.cpp:5:
src/build_settings.cpp:144:2: error: C++ requires a type specifier for all declarations
        Array(char) path_buf;`
Changed to
`Array<char> path_buf;`
Next,
`In file included from src/main.cpp:9:
src/ssa.cpp:10:8: error: ISO C++ forbids forward references to 'enum' types
enum   ssaBlockKind;`
Required me to remove the forward declarations. Finally I had to explicitly cast the const C-strings in ir.cpp.

This compiled successfully, but segfaulted on the demo. Turned out `when` is broken, and it loads the sys/windows.odin even in linux. Commenting out the relevant line in os.odin made it run successfully.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,8 +2,8 @@
 
 release_mode=0
 
-warnings_to_disable="-std=c11 -Wno-switch -Wno-pointer-sign -Wno-tautological-constant-out-of-range-compare -Wno-tautological-compare -Wno-macro-redefined"
-libraries="-pthread -ldl -lm"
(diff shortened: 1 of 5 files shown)


gingerBill:
Please check if the latest commit fixes the problems you have.

PR #85, opened 2017-08-02
Automatically select the newest vcvarsall.bat and call it

bdoner:
Just select the newest `vcvarsall.bat` and call it from `shell.bat` no matter the system.
Also set the Windows `path` variable to include the containing directory. No matter where you have Odin saved.


gingerBill:
That shell.bat doesn't even work VC2010 which what the default is 😛 Could you please add that?

bdoner:
Sorry about that. Added VS 10.0 :)  

I search for the newest version first. Do you see any problems with that?

PR #99, opened 2017-09-16
Launch arguments for Linux and macOS

zangent:
These commits enable the same `os.args` that works on Windows to work on *nix-based systems, too, without affecting the Windows builds at all.

The code that adds the implicit `argc` and `argv` to `main()` isn't pretty, but it was the most sensible way I could see implementing this without creating a wrapper around the Odin program's `main` procedure.


kiljacken:
Might a cleaner solution be to read in `/proc/{PID}/cmdline`? Not sure if macOS exposes that, but on Linux its a `\0` seperated array of the args

kiljacken:
macOS has `_NSGetArgc()` and `_NSGetArgv()` ([more info here](https://unixjunkie.blogspot.dk/2006/07/access-argc-and-argv-from-anywhere.html)) in case they do not have the `/proc/{PID}/cmdline` thing

gingerBill:
I would prefer a function call solution if possible rather than "hacking" in the parameters and changing the procedure signature of `main`

zangent:
I didn't know macOS has the `_NSGetArgX()` things, but IIRC the `/proc/pid/cmdline` is somewhat unreliable. I tried to implement that back when I was maintaining the Linux and macOS port of Odin, but I never had read permissions to it. While I could chalk that up to a configuration issue and say that it should work elsewhere, it would probably be best if it worked across the board, no matter whether or not the system is security-hardened.

I don't think it's necessarily that bad to change the signature of `main`, really. It literally has no effect on the user's program, so I don't see why you shouldn't do what works. That said, looking back at this PR, it would probably be better to insert the args at the IR level instead of generating AST nodes like I did.

gingerBill:
Could ye try the latest commit to see if it does what you need. I have made the initial `main` be called in a custom entry point procedure that is C like (contains `argc` and `argv`).

zangent:
Perfect! Glad to see this gap between Windows and *nix finally closed

PR #105, opened 2017-09-21
Added missing newline in error message.

thebirk:


gingerBill:
Those other prints will also need new lines too on lines 2847 and 2849.

PR #132, opened 2017-11-02
Improve building Odin on Linux.

multigrit:
This commit adds detection of LLVM & clang in the Linux build script.

Turns out if LLVM is not installed, Odin itself will build fine, but upon compiling/running Odin programs, the following message occurs:

`sh: 1: opt: not found`

This commit prints install instructions for Ubuntu if clang or LLVM is not found.

(Tested on Ubuntu 17.04 which uses LLVM v4.0)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,8 +1,44 @@
 #!/bin/bash
+#
+# NOTE:  POSIX prefers 'printf' over 'echo'.
+
+missing_llvm=0
+missing_clang=0
+
+# Check if LLVM is installed on system
+if ! hash llvm-ar 2> /dev/null; then
+    printf "[ERROR]  LLVM component not found:  llvm-ar\n"
+    missing_llvm=1
+fi
+if ! hash llvm-as 2> /dev/null; then
+    printf "[ERROR]  LLVM component not found:  llvm-as\n"
+    missing_llvm=1
+fi
+
+# Check if clang is installed on system
+if ! hash clang 2> /dev/null; then
+    printf "[ERROR]  Clang compiler not found:  clang\n"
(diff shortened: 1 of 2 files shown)


zangent:
This is... odd.
Like, I don't see why this is really... necessary? When Odin is able to be packaged into a package manager, this stuff will be done automatically, and until then, the build instructions tell you that you need LLVM and clang. Also, I'm not a big fan of providing Ubuntu-specific instructions in a shell script meant for every Linux distro and macOS.

Aside from that, you changed this from the build script
```bash
${compiler} src/main.cpp ${warnings_to_disable} ${libraries} ${other_args} -o odin && ./odin run examples/demo.odin
```
into
```bash
printf "LLVM & clang installed on system.\n"
printf "Building Odin...\n\n"

printf "${compiler} src/main.cpp ${warnings_to_disable} ${libraries} ${other_args} -o odin\n"
${compiler} src/main.cpp ${warnings_to_disable} ${libraries} ${other_args} -o odin

printf "\nTesting Odin...\n\n"
printf "./odin run examples/demo.odin\n"
./odin run examples/demo.odin
```
but that means that the shell script will continue and still try to build the `demo.odin` file even the compiler fails to build.


Then again, I'm not the guy in charge. Just thought I'd share my two cents :D

gingerBill:
I am not going to accept this for the reasons @zangent has said.

multigrit:
Thanks @zangent for the feedback -- I didn't think far enough ahead regarding packaging dependencies.

grxnola:
You really don't think it would be helpful to check  for LLVM and tell a user they don't have it?

PR #158, opened 2017-12-03
Fix floating point rounding functions

lonjil:
These seem to be accurate.

### DIFF
--- a/core/math.odin
+++ b/core/math.odin
@@ -85,14 +85,28 @@ copy_sign :: proc(x, y: f64) -> f64 {
 	return transmute(f64)ix;
 }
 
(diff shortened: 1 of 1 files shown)


Kelimion:
It still uses old-style overloading instead of e.g.
```cpp
floor_f32 :: proc(x: f32) -> f32 { ret := f32(i32(x)); return x >= ret ? ret : ret-1; }
floor_f64 :: proc(x: f64) -> f64 { ret := f64(i64(x)); return x >= ret ? ret : ret-1; }
floor :: proc{floor_f32, floor_f64};
```

gingerBill:
I am closing this PR as I have just implemented the more accurate implementations of them in the following commit: https://github.com/odin-lang/Odin/commit/5c3dc30dc00d9505cb9d28f3eda053575cb8b7d2

PR #161, opened 2017-12-06
InnoSetup Script to create an installer

ThisDrunkDane:
This script requires InnoSetup 5.5. It creates an installer called odin-installer.exe, it modifies the PATH if chosen (ticked by default) and removes it from the path upon uninstall if you use the uninstaller. The uninstaller has weird name but that is apparently hard-coded in InnoSetup and can't be changed sadly.


create-installer.iss
```diff
@@ -0,0 +1,99 @@
...
+AppName={#MyAppName}
+AppVersion={#MyAppVersion}
+;AppVerName={#MyAppName} {#MyAppVersion}
+AppPublisher={#MyAppPublisher}
+AppPublisherURL={#MyAppURL}
+AppSupportURL={#MyAppURL}
+AppUpdatesURL={#MyAppURL}
+DefaultDirName={pf}\Odin
+DisableProgramGroupPage=yes
+LicenseFile=E:\Odin\LICENSE
```
kiljacken: Hardcoded path, seems bad?

ThisDrunkDane: Ahh that was a typo from the auto-setup I used to get started, mean to be .\LICENSE

bdoner:
The AppId starts with two curly brackets. I guess that's a typo as well? 😄

gingerBill:
Is this necessarily yet? The language is still in heavy development and what is needed will mostly change in the near future.

This will be great for when release is near, just not yet.

ThisDrunkDane:
I don't see why not have it already? It's not something that really takes effort to maintain? if you add some new files at root or a new folder, it's just 1 line to add under the `[Files]` section.

gingerBill:
I am closing this at the moment because it is outdated and for my previous reason above that this language is still in heavy development and an installer is not yet needed.

PR #190, opened 2018-02-18
Fixed a bug (segfault)?

nakst:
When I tried to compile CEL on linux, Odin segfaulted.
I messed around which the code that the debugger said was crashing and this appeared to fix it.
I'm not entirely sure why what I did "worked", so your mileage may vary, I guess?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -271,3 +271,6 @@ shared/
 # temp files
 * .ll
 *.bc
+
+# vim swap files
+.swp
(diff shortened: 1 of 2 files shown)


gingerBill:
If this does happen, that means there is something else wrong in the compiler rather than just what you have found. Could you make minimal reproduction that caused the segfault and which operating systems it happens on, please?

nakst:
Minimal reproduction:
```c
Array :: []Value;
Dict  :: map[string]Value;
Nil_Value :: struct{};

Value :: union {
	Nil_Value,
	bool, i64, f64, string,
	Array, Dict,
}
```

When compiling on 64-bit archlinux using `./odin build file.odin` with the latest commit of the Odin compiler it produces the result `Segmentation fault (core dumped)`

gingerBill:
Please try again with the current commit.

PR #260, opened 2018-09-05
[WIP] Provide llvm with more debug info (for Visual Studio debugger support)

lachsinc:
High chance this pr may break IR output so testing / review appreciated.

#### TODO
- [x] Watch/Locals support for Primitives/Enums/Strings/Arrays/Slices/DynamicArrays/Map/Any/Union/typeid/Bitset/Bitfield/Type Info/Tuple/Proc/Complex
- [x] Improved "location" info (jump between procedures, files)
- [x] Stepping

Note: Stepping now works but it still needs a lot more attention; Would be nice if we didn't step into mem.zero / context setup stuff, could step _over_ for loops, return instructions jump to closing brace, breakpoints on proc header get missed etc. I will leave that for a separate pull request though.

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -573,10 +573,23 @@ void init_build_context(void) {
 
 	bc->optimization_level = gb_clamp(bc->optimization_level, 0, 3);
 
-	gbString opt_flags = gb_string_make_reserve(heap_allocator(), 16);
+	gbString opt_flags = gb_string_make_reserve(heap_allocator(), 64);
+	opt_flags = gb_string_append_fmt(opt_flags, "-O%d ", bc->optimization_level);
 	if (bc->optimization_level != 0) {
-		opt_flags = gb_string_append_fmt(opt_flags, "-O%d", bc->optimization_level);
+		// NOTE(lachsinc): The following options were previously passed during call
+		// to opt in main.cpp:exec_llvm_opt().
+		//   -die:       Dead instruction elimination
+		//   -memcpyopt: MemCpy optimization
+		opt_flags = gb_string_appendc(opt_flags, "-memcpyopt -die ");
 	}
+
+	// NOTE(lachsinc): This optimization option was previously required to get
+	// around an issue in fmt.odin. Thank bp for tracking it down! Leaving for now until the issue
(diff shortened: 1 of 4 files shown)


gingerBill:
>This may not be the ideal way to handle collecting debug info but rather the one that required the least changes. ...

@lachsinc I will not accept this pull request until the hack is removed. Other than that, it looks good.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ThisDrunkDane:
I've marked this with [WIP] so feel free to make changes, we'll review when you feel you're "done"

lachsinc:
llc/opt hack has been removed; llvm bitcode output contains useful debug info now thanks to the removal of a few optimization options. Removal of "-mem2reg" broke some things in fmt.odin. @bpunsky helped track this down, see issue #265 for further details.

lachsinc:
cce5e59 5961a63 both introduce a temporary (?) solution for composite types string and dynamic array, treating them as a struct w/ hard-coded fields. I'll make a cleanup pass over most of this before it gets accepted. 

If there is a better way to obtain the type info (notably their [field layout](https://github.com/odin-lang/Odin/blob/master/core/mem/raw.odin)) from inside the compiler let me know.

gingerBill:
```
Type info / typeid support
```
What is meant by this point? `typeid` could simply be just an opaque integer in the debug window

ThisDrunkDane:
While `typeid` could just be that, he might mean he wants to translate the `typeid` into the full `TypeInfo` struct in the watch window.

gingerBill:
That might be a good idea, theoretically. But it may be a little weird. I am not so sure as to how much control you have in the watch window.

lachsinc:
Yeah will just expose typeid as an integer. Probably should have separated them for clarity..

ThisDrunkDane:
How far along are you @lachsinc? You seem to have checked all your boxes 👍 👍 👍

gingerBill:
I am going to merge this now and if there are any other issues, we can fix them later.

gingerBill:
One issue is that the moving of the `-memcpyopt -die`  stuff to optimizations doubles compile times. This is why I added them originally. However, if they cock up debug, it might be better to only enable then when debugging in on.

Issue #240, opened 2018-08-03
Context based Logging System


gingerBill:
Add logging system to the context

General idea (not proposal)

```
package log 

Logger_Mode :: struct {
    None,
    Minimal,
    General,
    Verbose,
}

Logger_Proc :: #type proc(
	logger_data: rawptr, mode: Logger_Mode,
	ident: string, text: string, level: int, location := #caller_location,
);

Logger :: struct {
    procedure: Logger_Proc,
    data:      rawptr,
}
```
```
Context :: struct {
	allocator:  mem.Allocator,
	logger:     log.Logger, 
	thread_id:  int,

	user_data:  any,
	user_index: int,

	parent:     ^Context,
	derived:    any, // May be used for derived data types
}
```

ThisDrunkDane:
Hasn't this been implemented or has it just been worked on a little @gingerBill?

gingerBill:
This has not been implemented yet as I do not know the best "format" for the logger. Odin does have the `assertion_failure_proc` which is similar but for assertion-like calls.

ThisDrunkDane:
Well for the different "logger levels/mode" at work we use log4net (which is a port of an apache thing). But it has `Debug, Info, Warning, Error, Fatal` and that has served us really well, they then have a xml config file to define what levels to print, that I could live without! But that would be up to the different logger implementations on how to handle that

ThisDrunkDane:
This was implemented by PR #276 a while ago, so I'm closing this

PR #276, opened 2018-10-23
Add a simple file-, console- and multi-logger to core

ThisDrunkDane:
## Actual PR
The primary interfaces to actually log by are these in `log.odin`;
```
debug :: proc(fmt_str : string, args : ..any, location := #caller_location)
info  :: proc(fmt_str : string, args : ..any, location := #caller_location)
warn  :: proc(fmt_str : string, args : ..any, location := #caller_location)
error :: proc(fmt_str : string, args : ..any, location := #caller_location)
fatal :: proc(fmt_str : string, args : ..any, location := #caller_location)
```
These will use `fmt.tprintf` first on the `fmt_str` and `args` to produce the logged text.
Then it's up to the individual logger to take the specified options and indentifier and prefix (or postfix) that the logging content.
This also includes 3 types of basic loggers. A console and file logger whose format is identical if the same options are set, you just simply either log to the console (stdout or stderr) or a `os.Handle` which points to a file. There is also an option to limit what is actually logged by defining a "lowest level".
The file and console logger will output in a format similar to this;
```
[DEBUG] --- [main.odin.main():49] [System API]  We are debug now
[INFO ] --- [main.odin.main():50] [System API]  We are info now
[WARN ] --- [main.odin.main():51] [System API]  We are warn now
[ERROR] --- [main.odin.main():52] [System API]  We are error now
[FATAL] --- [main.odin.main():53] [System API]  We are fatal now
[WARN ] --- [main.odin.main():60] [Program API]  We are warn now
[ERROR] --- [main.odin.main():61] [Program API]  We are error now
[FATAL] --- [main.odin.main():62] [Program API]  We are fatal now
[DEBUG] --- [2018-12-08 14:57:46] [main.odin.main():26]  We are debug now
[INFO ] --- [2018-12-08 14:57:46] [main.odin.main():27]  We are info now
[WARN ] --- [2018-12-08 14:57:46] [main.odin.main():28]  We are warn now
[ERROR] --- [2018-12-08 14:57:46] [main.odin.main():29]  We are error now
[FATAL] --- [2018-12-08 14:57:46] [main.odin.main():30]  We are fatal now
```
Though how it exactly looks is based on what `log.Options` you specify.
The multi logger simple carries multiple loggers in it, and calls each `Logger_Proc` whenever something is logged.

_While options are manually specified here there are default options for both [the file- and console-logger](https://github.com/odin-lang/Odin/commit/a16344d0964f656e06024466054103be64ac666c#diff-7c510ccc05d5ab19450316377f2ef214R14) so you can just do `l := log.console_logger();`, the file logger does always need a file handle_
Example; 
```
h, err := os.open("test.log", os.O_RDWR|os.O_APPEND);
fl := log.file_logger(h, log.Level.Debug,{    
    log.Option.Level, 
    log.Option.Long_File_Path, 
    log.Option.Line, 
}, "System API");
cls := log.console_logger(log.Level.Debug,{    
    log.Option.Level, 
    log.Option.Short_File_Path, 
    log.Option.Line, 
    log.Option.Procedure, 
    log.Option.Terminal_Color
}, "System API");

clp := log.console_logger(log.Level.Warning,{    
    log.Option.Level, 
    log.Option.Short_File_Path, 
    log.Option.Line, 
    log.Option.Procedure, 
    log.Option.Terminal_Color
}, "Program API");

m1 := log.multi_logger(fl, cls);
m2 := log.multi_logger(fl, clp);

{
    context.logger = m1;

    log.debug("We are debug now");
    log.info("We are info now");
    log.warn("We are warn now");
    log.error("We are error now");
    log.fatal("We are fatal now");
}
{
    context.logger = m2;

    log.debug("We are debug now");
    log.info("We are info now");
    log.warn("We are warn now");
    log.error("We are error now");
    log.fatal("We are fatal now");
}
```

### DIFF
--- a/core/log/log.odin
+++ b/core/log/log.odin
@@ -1,5 +1,8 @@
 package log
 
+import "core:fmt";
+import "core:runtime";
+
 Level :: enum {
 	Debug,
 	Info,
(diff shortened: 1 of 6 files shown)


core/log/log.odin
```diff
@@ -9,26 +12,75 @@ Level :: enum {
...
 }
 
+Multi_Logger_Data :: struct {
+    loggers : []Logger,
+}
+
+multi_logger :: proc(logs: ..Logger) -> Logger {
+    data := new(Multi_Logger_Data);
+    data.loggers = make([]Logger, len(logs));
+    for log, i in logs do data.loggers[i] = log;
```
gingerBill: This line should be `copy(data.loggers, logs);`

core/log/log.odin
```diff
@@ -9,26 +12,75 @@ Level :: enum {
...
 	data:      rawptr,
+    options:   Options,
+    ident:     string
 }
 
+Multi_Logger_Data :: struct {
+    loggers : []Logger,
+}
+
+multi_logger :: proc(logs: ..Logger) -> Logger {
```
gingerBill: I would call this `create_mulit_logger`.
It also requires an equivalent destroy procedure.

core/log/log.odin
```diff
@@ -9,26 +12,75 @@ Level :: enum {
...
+    Option.Short_File_Path,
+    Option.Long_File_Path
+};
 
 Logger_Proc :: #type proc(data: rawptr, level: Level, ident, text: string, options: Options, location := #caller_location);
 
 Logger :: struct {
 	procedure: Logger_Proc,
 	data:      rawptr,
+    options:   Options,
```
gingerBill: Please try to not add the extra data here, keep that data in the `data` component if possible. If it isn't, then fine but I do want to keep this "interface data" small.

core/log/log.odin
```diff
@@ -9,26 +12,75 @@ Level :: enum {
...
+
+debug :: proc(fmt_str : string, args : ..any, location := #caller_location) do log(level=Level.Debug,   fmt_str=fmt_str, args=args, location=location);
+info  :: proc(fmt_str : string, args : ..any, location := #caller_location) do log(level=Level.Info,    fmt_str=fmt_str, args=args, location=location);
+warn  :: proc(fmt_str : string, args : ..any, location := #caller_location) do log(level=Level.Warning, fmt_str=fmt_str, args=args, location=location);
+error :: proc(fmt_str : string, args : ..any, location := #caller_location) do log(level=Level.Error,   fmt_str=fmt_str, args=args, location=location);
+fatal :: proc(fmt_str : string, args : ..any, location := #caller_location) do log(level=Level.Fatal,   fmt_str=fmt_str, args=args, location=location);
+
+log :: proc(level : Level, fmt_str : string, args : ..any, location := #caller_location) {
+    logger := context.logger;
+    str := fmt.tprintf(fmt_str, ..args); //NOTE(Hoej): While tprint isn't thread-safe, no logging is.
```
gingerBill: The `temp_allocator` needs to be made thread safe. This is now possible with the atomic intrinsics.

core/log/log.odin
ThisDrunkDane: Options and ident will have to be here if they're going to be passed to the logger proc.

core/log/log.odin
ThisDrunkDane: Yeah, but until then, it isn't ;)

core/log/log.odin
ThisDrunkDane: Knew there was a smarter way to do that, I'll change that!

core/log/log.odin
ThisDrunkDane: Shouldn't that be the case for console and file logger too then? They all need to allocate a ptr to save their data into.

core/log/log.odin
gingerBill: 🤔 I wonder if they should be passed or let the logger decide that itself.

gingerBill: It probably is a good idea my question is whether `options` and `ident` will be modified enough that they should be there or that we always want those options on all loggers.

ThisDrunkDane:
For linking, this PR address issue #240

ThisDrunkDane:
@gingerBill Any news on the time library?

gingerBill:
@ThisDrunkDane time library will come pretty soon. It will work for windows for definite.

PR #291, opened 2018-12-05
fixed typo in 'GetMonitorInfoA' & added additional window styles

dimenus:
Self explanatory?

### DIFF
--- a/core/sys/win32/windows.odin
+++ b/core/sys/win32/windows.odin
@@ -6,6 +6,7 @@ foreign import "system:user32.lib"
 foreign import "system:gdi32.lib"
 foreign import "system:winmm.lib"
 foreign import "system:shell32.lib"
+foreign import "system:ole32.lib"
 
 Handle    :: distinct rawptr;
 Hwnd      :: distinct Handle;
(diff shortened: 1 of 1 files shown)


core/sys/win32/windows.odin
```diff
@@ -19,12 +20,14 @@ Hgdiobj   :: distinct Handle;
 Hmodule   :: distinct Handle;
 Hmonitor  :: distinct Handle;
 Hrawinput :: distinct Handle;
+Hresult   :: distinct int;
```
gingerBill: `HRESULT` is defined as a `long` which on windows is 32 bits. `i32` should be used, not `int`.

core/sys/win32/windows.odin
```diff
@@ -19,12 +20,14 @@ Hgdiobj   :: distinct Handle;
...
 HKL       :: distinct Handle;
 Wparam    :: distinct uint;
 Lparam    :: distinct int;
+Lpvoid    :: distinct rawptr;
```
gingerBill: It's better to just use a plain `rawptr` rather than having a distinct type for `Lpvoid`.

core/sys/win32/windows.odin
```diff
@@ -902,7 +1062,19 @@ foreign winmm {
 	@(link_name="timeGetTime") time_get_time :: proc() -> u32 ---;
 }
 
+//objbase.h
+Com_Init :: enum {
+	MultiThreaded = 0x0,
```
gingerBill: Prefer `Multi_Thread`, `Apartment_Threaded`, etc.

PR #299, opened 2018-12-10
Add the missing INVALID_HANDLE so that the demo will build and run on Linux

CaptainKraft:
The demo wouldn't run or build on Linux. I copied this line from the os_windows.odin file. Pretty straightforward.

### DIFF
--- a/core/os/os_linux.odin
+++ b/core/os/os_linux.odin
@@ -12,6 +12,7 @@ Handle    :: distinct i32;
 File_Time :: distinct u64;
 Errno     :: distinct i32;
 
+INVALID_HANDLE :: ~Handle(0);
 
 O_RDONLY   :: 0x00000;
 O_WRONLY   :: 0x00001;


Kelimion:
I noticed in #odin that you had this error before you installed LLVM. I had no trouble building the demo on Linux, so I'm wondering if you could check that it'll build post-LLVM-install without this fix.

CaptainKraft:
It still does not build without LLVM. The code attempts to use `os.INVALID_HANDLE` which doesn't exist in the `os_linux.odin` or the `os.odin` files.

CaptainKraft:
Here are the error messages:

``
./odin run examples/demo
/home/captainkraft/dev/Odin/core/log/file_console_logger.odin(47:28) 'INVALID_HANDLE' is not declared by 'os'
/home/captainkraft/dev/Odin/core/log/file_console_logger.odin(55:24) 'INVALID_HANDLE' is not declared by 'os'
/home/captainkraft/dev/Odin/core/log/file_console_logger.odin(55:10) Assignment count mismatch '1' = '0'
/home/captainkraft/dev/Odin/core/log/file_console_logger.odin(70:28) 'INVALID_HANDLE' is not declared by 'os'
``

CaptainKraft:
As @thebirk mentioned in his PR #301, perhaps the handle should be -1 since that is what open and create will return on failure?

PR #301, opened 2018-12-11
Added INVALID_HANDLE constant os_linux.odin.

thebirk:
`core/os/os_linux.odin` was missing a INVALID_HANDLE constant, this PR adds it with the value of `-1` as this is what `open()` and `creat()` will return on failure.

This constant was required by the new `core/log` library and without this the demo won't run on linux.


thebirk:
Welp I just realised @CaptainKraft had already posted a PR for this so I am closing this one.

PR #324, opened 2019-01-16
Fix assert when printing IR of pointer with endianness

Tetralux:
When printing out the IR for a pointer value it in `ir_print_exact_value`, `is_type_integer_endian_{little,big}`
did not correct handle being asked about pointer types.

They now reply with whether the endianness of `uintptr` matches the
endianness being asked about.

I'm not 100% sure if this is all that needs to be done or not, but this fixed the problem I had in my own code where I was getting an assert in `is_type_integer_endian_little` with the `^u8` type.
I'm not very familiar with the codebase as yet.
I debugged it and at first glance seemed like an easy fix. So... here it is...? xD


Tetralux:
P.S: Have not tested this on a big-endian machine.

Tetralux:
(Edit: Made OP more better.)

PR #328, opened 2019-01-24
Add rand.fill(^rand.Rand, []$T) for filling a slice with random data.

Tetralux:
Exactly as advertised: fills a slice with random data.

Achieves a speed of ~1.45GB/s on my machine with `-opt=2` and a 8GB slice.

### DIFF
--- a/core/math/rand/rand.odin
+++ b/core/math/rand/rand.odin
@@ -60,3 +60,21 @@ int63_max :: proc(r: ^Rand, n: i64) -> i64 {
 
 float64 :: proc(r: ^Rand) -> f64 { return f64(int63_max(r, 1<<53)) / (1 << 53); }
 float32 :: proc(r: ^Rand) -> f32 { return f32(float64(r)); }
+
+fill :: proc(rng: ^rand.Rand, buf: []$T) {
+    ptr := cast(^u8) &buf[0];
+    total_bytes := len(buf) * size_of(T);
+    count := total_bytes;
+    for i := 0; i < count-1; {
+        if total_bytes-i >= size_of(u32) {
+            ptr32 := cast(^u32) ptr;
+            ptr32^ = rand.uint32(rng);
+            ptr = mem.ptr_offset(ptr, size_of(u32));
+            i += size_of(u32);
+        } else {
+            ptr^ = cast(u8) rand.uint32(rng);
+            ptr = mem.ptr_offset(ptr, 1);
+            i += 1;
+        }
+    }
+}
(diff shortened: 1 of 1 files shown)


core/math/rand/rand.odin
```diff
@@ -60,3 +60,21 @@ int63_max :: proc(r: ^Rand, n: i64) -> i64 {
 
 float64 :: proc(r: ^Rand) -> f64 { return f64(int63_max(r, 1<<53)) / (1 << 53); }
 float32 :: proc(r: ^Rand) -> f32 { return f32(float64(r)); }
+
+fill :: proc(rng: ^rand.Rand, buf: []$T) {
+    ptr := cast(^u8) &buf[0];
+    total_bytes := len(buf) * size_of(T);
+    count := total_bytes;
+    for i := 0; i < count-1; {
+        if total_bytes-i >= size_of(u32) {
```
Kelimion: Have you tried a loop for `count/size_of(u32)` and a loop for the remainder? That'd eliminate a conditional in the loop,

Tetralux: Originally, it _was_ structured that way, but was hard to read and understand.
So I rewrote it to be simpler while maintaining performance.
This version of the code is just as fast as the original.

gingerBill: I don't understand this procedure. What if the parameter contains a pointer? Wouldn't that pointer then be just random?  Does this need to be a core procedure rather than a user one (if that makes sense)?

Tetralux: If you passed a slice of pointers, then yes they would. 
I could change it to write into the pointers instead if they're non-null, if that seems like preferred behaviour to you. However, that'd seem a bit inconsistent, albeit understandable.

Alternatively, if you'd rather pointers were not allowed, then I'm not entirely sure how best to do that here. But likewise, inconsistent.

I could also change it to only work with slice of primitive numeric types, of course, thereby side-stepping the issue.

As to being a core proc however, filling a slice with random data from a random number generator is a fairly primitive operation, just as asking for a single integer is.
I routinely find myself doing one of those things when I use RNGs.

gingerBill: I'd rather have consistency than edge cases. My issue is that if personally wanted to fill something up with random data, I would just do a for loop manually, but maybe that is just me.

Tetralux:
This was added as the `read` procedure in https://github.com/odin-lang/Odin/commit/2347dca9d98e2e1f73896e96f861b13a510a7147.

PR #336, opened 2019-02-06
adding a few COM items, window styles, etc...

dimenus:
COM I_Unknown interface has been tested using the other code I'm working on.


ThisDrunkDane:
@gingerBill Shouldn't we just get this in? We will need the COM stuff anyway at some point

PR #339, opened 2019-02-12
"core:path" Path library and "core:strings" `split` Utilities

dotbmp:
I've added a path library which can extract components of file paths (drive letter, parent directory, base, file name, file extension), get the current working directory, get the absolute version of a path relative to the current directory, convert between Windows' "short" and "long" path names (because why not), verify a path's existence, determine whether a valid path is a file or a directory, get the relative path between two paths, as well as the current directory and another path, and split a path into its constituent elements.

In order to facilitate the latter, I've also added two quite useful `string.split` variants to the `core:strings` library, which split a given string into sub-slices along one or more given sub-strings. Both the single and multiple sub-string variants have an option not to include empty sub-slices in the results.

If you have any concerns or edits you'd like me to make before pulling, just let me know. Of course, feel free to edit the code/comments as you see fit.

There are a couple of outstanding design issues that I'd be happy to discuss if they're problematic; most notably with the behavior of `path.dir`, `path.base`, `path.name`, and `path.ext` in terms of what they return in certain cases. I've tried to use [Go's path package](https://golang.org/pkg/path/) as a reference, though I opted not to follow some of its choices such as returning `"."` from `path.base` when the `path` parameter is empty. Ultimately I think the best way to figure out what design works best for the edge cases is to put `core:path` into the wild and see which choices people respond negatively to.

### DIFF
--- a/core/path/path_linux.odin
+++ b/core/path/path_linux.odin
@@ -1,0 +1,7 @@
+package path
+
+
+SEPARATOR        :: '/';
+SEPARATOR_STRING :: "/";
+
+#compile_assert("Linux is not yet supported");
--- a/core/path/path_linux.odin
+++ b/core/path/path_linux.odin
@@ -23,7 +23,7 @@ null_term :: proc(str: string) -> string {
 }
 
 
-full :: proc(path: string, allocator := context.allocator) -> string {
+full :: proc(path: string, allocator := context.temp_allocator) -> string {
     cpath := strings.clone_to_cstring(path, context.temp_allocator);
     
     foreign libc {
@@ -41,7 +41,7 @@ full :: proc(path: string, allocator := context.allocator) -> string {
     return null_term(string(buf[:]));
 }
 
-current :: proc(allocator := context.allocator) -> string {
+current :: proc(allocator := context.temp_allocator) -> string {
     foreign libc{
         getcwd :: proc(buf: ^u8, size: int) -> cstring ---;
     }
--- a/core/path/path_linux.odin
+++ b/core/path/path_linux.odin
@@ -1,7 +1,0 @@
-package path
-
-
-SEPARATOR        :: '/';
-SEPARATOR_STRING :: "/";
-
-#compile_assert("Linux is not yet supported");
--- a/core/path/path_linux.odin
+++ b/core/path/path_linux.odin
@@ -1,0 +1,80 @@
+package path
+
+foreign import libc "system:c"
+
+import "core:os"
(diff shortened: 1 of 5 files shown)


bpunsky:
Oh yeah, it only works on Windows. ¯\_(ツ)_/¯

bpunsky:
All of this functionality can be previewed with my [copy in library form](https://github.com/bpunsky/odin-path)

core/path/path_linux.odin
```diff
@@ -0,0 +1,7 @@
+package path
+
+
+SEPARATOR        :: '/';
+SEPARATOR_STRING :: "/";
+
+#compile_assert("Linux is not yet supported");
```
mattt-b: `#assert`

Also this assert doesn't get hit. Moving it to the second line does have it get hit

bpunsky: Ah, thanks! I'll push an edit tomorrow.

Also, that's super weird about the second line thing...

Maybe I should just remove `path_linux.odin`, since there's nothing really in it anyway, and that way it'll behave like any other unimplemented platform. I don't really know what would happen if you tried to use it like that though... Probably just tell you none of the things are defined

mattt-b: > I don't really know what would happen if you tried to use it like that though... Probably just tell you none of the things are defined

Yes, which it still does with the file since it doesn't define `full`, `is_dir`, or `current`. Removing the file just adds `SEPARATOR` and `SEPARATOR_STRING` to that list (and obviously removes the `#assert` error message)

bpunsky:
Okay, deleted `path_linux.odin`

bpunsky:
Any thoughts on this? Would be perfectly happy to change things to get this approved, it would just be nice not have a dependency on my own lib everywhere I want to manipulate path strings

bpunsky:
I've updated the PR to reflect some improvements/fixes to the path lib, properly placing the Win32 bindings in `core:sys/win32` and to merge recent additions to `core:strings`

bpunsky:
@Kelimion 👀

gingerBill:
@bpunsky  This just needs macOS support and updating for the new `strings.split`.

Tetralux:
@dotbmp

dotbmp:
I can do `strings.split`

Someone else will have to do macos

Afaik the linux code should just work on Mac with a modified foreign import?

powerc9000:
I can test for macos. Need these utils anyway so that works out.

PR #346, opened 2019-02-26
Fix #345: panic when using enum as map key

### LINKED ISSUE #345: Map with enum key panics the compiler when assigned to
## Context

```cpp
package minrep

Enum :: enum {
    A = 0,
    B = 1,
}

main :: proc() {
    foos : map[Enum]int;
    foos[Enum.A] = int{}; // Comment this out to remove the panic
}
```

Results in internal compiler error: `src\ir.cpp(3251): Panic: Unhandled map key type`

* Operating System:
* Odin version/Commit: master / 76a6757ee9e838326519ffb30a81dc60667d253d

## Expected Behavior

I expect an enum to be usable as a map key.

## Current Behavior

The enum as map key compiles happily as long as you don't try to add anything to the map.

Odin apparently has no problem with an enum as a map key as it doesn't complain at the declaration site. It's only when you go and use the map that it throws up the panic.

### Steps to Reproduce

1. Save fragment to minrep.odin
2. Run `odin run minrep.odin`
3. Observe failure
4. Comment out the map assignment and save
5. Run again
6. Observe the lack of crash

Kelimion:
Fix for issue #345 
```cpp
package minrep
import "core:fmt";

Enum :: enum {
    A = 0,
    B = 1,
}

main :: proc() {
    foos : map[Enum]int;
    foos[Enum.A] = int{};
    fmt.println(foos);
}

```
Without the patch it panics: `src/ir.cpp(3251): Panic: Unhandled map type`
With the patch it doesn't panic and the output is as expected: `map[A=0]`

### DIFF
--- a/examples/demo/demo.odin
+++ b/examples/demo/demo.odin
@@ -703,9 +703,9 @@ map_type :: proc() {
 		A = 0,
 		B = 1 << 64 - 1,
 	}
-	Enum_i32 :: enum i32 {
+	Enum_i8 :: enum i8 {
 		A = 0,
-		B = -(1 << 31),
+		B = -(1 << 7),
 	}
 	Enum_i64 :: enum i64 {
 		A = 0,
--- a/examples/demo/demo.odin
+++ b/examples/demo/demo.odin
@@ -530,7 +530,7 @@ parametric_polymorphism :: proc() {
 			// `I` is the type of N
 			// `T` is the type passed
 			fmt.printf("Generating an array of type %v from the value %v of type %v\n",
-			           typeid_of(type_of(res)), N, typeid_of(I));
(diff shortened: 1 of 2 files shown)


bpunsky:
Good fix... sketchy practice though 😛

gingerBill:
Don't bit cast to a `uint` as the `enum` could be a different size. You will need to convert the `enum` to it's underlying type and then the hash type.

Kelimion:
Alright, it now handles `enum`s of type `u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64` and `i64` as `map` keys.
A little demo of the map functionality was also added to `demo.odin`.

PR #384, opened 2019-06-09
Implement `get_last_error` for osx and make it return Errno

hasenj:


hasenj:
fixed compile error on windows 😅

hasenj:
since this is not being taken I guess there's no point in keeping it open.

PR #404, opened 2019-07-16
Impl time for macOS

hazeycode:
Similar to the linux implementation except macOS only has SYSTEM_CLOCK and CALENDAR_CLOCK therefore boot_time() is not implemented (like the windows implementation)


gingerBill:
I would recommend making `package time` not rely on `package os`. If you need to have foreign declarations, just declare them within `package time`, and if necessary, mark them as `@(private)`.

hazeycode:
Ah. I just followed time_linux here which also imports core:os. Maybe needs to be considered as a whole linregard to the os module v2 revision. Seems both time_linux and time_osx could share the same POSIX foreign declarations?

gingerBill:
The idea is that `package time` should be completely separate from `package os`. I think most core packages should try to be as self kept if possible.

hazeycode:
The windows implementation imports core:sys/win32. So do we want a core:sys/posix or something to that effect?

gingerBill:
The `foreign` time code needs to exist in `package time` rather than in `package os`.

PR #425, opened 2019-08-26
Fixed parser creating a new thread for each file.

thebirk:
I have done limited testing with this PR, so I would recommend trying it with some bigger projects before merging.


gingerBill:
@thebirk Does this improve performance too?

gingerBill:
@thebirk Is this ready to merge?

thebirk:
@gingerBill 

As this PR stands it compiles all the projects I have tried so I think it is ready to be merged.
As for performance it seems to be pretty equivalent to the pre-fix speeds.

I also think it would make sense to squash this PR as it doesn't make sense for it to be 5 different commits.

PR #435, opened 2019-09-08
Add bit_stream package 

Lunatoid:
Adds a package called `bs` with `Bit_Stream` functionality.
Functionality includes creating a bit stream with an allocator, writing variables and buffers into it, reading it back out, all the stuff you expect.

Example:
```
import "core:bs"

stream := bs.create();

// Writing
foo: Foo = get_foo();
bs.write(&stream, foo);
bs.write(&stream, buffer, buffer_len);

// Reading
bs.reset();

foo2 := bs.read(&stream, Foo);
bs.read(&stream, buffer2, buffer2_len);
```

Feedback is welcome!


gingerBill:
Just call the package bitstream or bit_stream.

Calling the package bs is... Male cattle poop.

Lunatoid:
Alright, done!

core/bit_stream/bitstream.odin
```diff
@@ -0,0 +1,116 @@
...
+package bs;
+
+import "core:mem"
+
+BITSTREAM_INITIAL_SIZE: int: 128;
+BITSTREAM_GROWTH_RATE:  f64: 2.0;
+
+Bit_Stream :: struct {
+    data, head: ^byte,
+    size, cap: int,
```
gingerBill: I would recommend using a slice for the data rather than a pointer. 
```
data: []byte,
head: int,
len: int,
```

gingerBill: But using pointers may not be an issue in practice if it is only meant to be used internally.

core/bit_stream/bitstream.odin
```diff
@@ -0,0 +1,116 @@
...
+// Reads and returns a type
+read :: inline proc "contextless" (bs: ^Bit_Stream, $T: typeid) -> T {
+    val: T = (cast(^T) bs.head)^;
+    bs.head = mem.ptr_offset(bs.head, size_of(T));
+    
+    return val;
+}
+
+// Reads `size` bytes into the `dest` ptr
+read_into :: inline proc "contextless" (bs: ^Bit_Stream, auto_cast dest: rawptr, size: int) {
```
gingerBill: `auto_cast` is not required here.

core/bit_stream/bitstream.odin
```diff
@@ -0,0 +1,116 @@
...
+write :: proc{ write_value, write_ptr };
+
+// Writes a variable into the stream
+write_value :: inline proc(bs: ^Bit_Stream, val: $T) {
+    tmp := val;
+    write_ptr(bs, &tmp, size_of(T));
+}
+
+// Writes `size` bytes of `ptr` into the stream
+write_ptr :: proc(bs: ^Bit_Stream, auto_cast ptr: rawptr, auto_cast size: int) {
```
gingerBill: `auto_cast` is not required here.

core/bit_stream/bitstream.odin
```diff
@@ -0,0 +1,116 @@
...
+    return Bit_Stream{
+        data, data,
+        0, cap,
+        false,
+        allocator
+    };
+}
+
+// Creates a bitstream from an existing buffer that can only be read
+create_from_ptr_readonly :: inline proc "contextless" (auto_cast data: ^byte) -> Bit_Stream {
```
gingerBill: `auto_cast` is not required here.

core/bit_stream/bitstream.odin
```diff
@@ -0,0 +1,116 @@
...
+    return Bit_Stream{
+        head, head,
+        0, size,
+        false,
+        allocator
+    };
+}
+
+// Creates a bitstream from an existing buffer, allocators must be the same
+create_from_ptr :: inline proc "contextless" (auto_cast data: ^byte, auto_cast cap: int,
```
gingerBill: I would not recommend `auto_cast` here.

core/bit_stream/bitstream.odin
Lunatoid: I chose for using pointers because it makes a lot of memory operations trivial.
The original use of this package was to construct network packets for a game so I would need the pointer and the size for that and I felt that having a slice would add unnecessary friction.

I'll update the package to remove the unnecessary `auto_cast`s and change the structure to something like you described sometime this week (probably).

Lunatoid:
I appreciate the feedback but I have decided I will no longer be updating this PR. Sorry!

PR #446, opened 2019-10-19
A "Guess The Number" demo program, plus a couple of core library additions I had to implement for it

slightlyoutofphase:
Don't feel bad at all if you don't want to merge this, it didn't take particularly long and my feelings will not be hurt, haha. I just figured better to open the PR in case any of it might be considered useful, than not.

If you do want to merge some / any / all of it but with changes, also, just let me know and I'll make them.

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -62,11 +62,10 @@ readln :: proc() -> (success: bool, result: string) {
 	len, err := os.file_size(os.stdin);
 	if len == 0 {
 		return false, "";
-	} else {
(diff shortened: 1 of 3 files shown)


gingerBill:
I am not going to merge this mainly because some of the functions need to be in different packages than the ones you have placed them in. The code is fine and the example would be a really idea for website to show others other examples.

slightlyoutofphase:
How would you want the functions moved around, exactly? Again, I'm happy to make any changes, as I was really just kind of guessing as to where I put them.

PR #458, opened 2019-11-01
Implement core:thread and core:sync on Unix using pthreads

Tetralux:
Implement core:thread and core:sync on Unix using pthreads
Also do some cleanup and refactoring of the thread, sync and time APIs.
    
- remove `semaphore_release` because `post` and `wait` is easier to understand

- change `semaphore_wait` to `*_wait_for` to match Condition

- ~~pthreads can be given a stack, but doing so requires the user to set up the guard
pages manually. BE WARNED.~~ The alignment requirements of the stack are also
platform-dependant; it may need to be page size aligned on some systems.
Unclear which systems, however. See `os.get_page_size`, and `mem.make_aligned`.
HOWEVER: I was unable to get custom stacks with guard pages working reliably,
so while you can do it, the API does not support it.

- add `os.get_page_size`, `mem.make_aligned`, and `mem.new_aligned`.
- removed thread return values because windows and linux are not consistent; windows returns `i32`
and pthreads return `void*`; besides which, if you really wanted to communicate how the
thread exited, you probably wouldn't do it with the thread's exit code.

- fixed `thread.is_done` on Windows; it didn't report true immediately after calling `thread.join`.

- moved time related stuff out of `core:os` to `core:time`.

- add `mem.align_backward`

- fixed default allocator alignment
The heap on Windows, and calloc on Linux, both have no facility to request alignment.
It's a bit of hack, but the heap_allocator now overallocates; `size + alignment` bytes,
and aligns things to at least 2.
It does both of these things to ensure that there is at least two bytes before the payload,
which it uses to store how much padding it needed to insert in order to fulfil the alignment
requested.
    
- make conditions more sane by matching the Windows behaviour.
The fact that they were signalled now lingers until a thread tries to wait,
causing them to just pass by uninterrupted, without sleeping or locking the
underlying mutex, as it would otherwise need to do.
This means that a thread no longer has to be waiting in order to be signalled, which
avoids timing bugs that causes deadlocks that are hard to debug and fix.
See the comment on the `sync.Condition.flag` field.
    
- add thread priority: `thread.create(worker_proc, .High)`


Tetralux:
I'm gonna scan read this at least once more before considering it done - but we're at the point where some extra eyes start to be useful now. 😄

core/sync/sync_windows.odin
```diff
@@ -1,51 +1,41 @@
...
 
 
-current_thread_id :: proc() -> i32 {
-	return i32(win32.get_current_thread_id());
-}
-
-semaphore_init :: proc(s: ^Semaphore) {
-	s._handle = win32.create_semaphore_w(nil, 0, 1<<31-1, nil);
+semaphore_init :: proc(s: ^Semaphore, initial_count := 0) {
+	assert(initial_count >= min(int) && initial_count <= max(int));
```
sci4me: Could you say
```odin
semaphore_init :: proc(s: ^Semaphore, initial_count : int = 0) { ...
```
instead of this assertion? If so, would this be desirable?
This could possibly be extended to the other implementations as well if desirable.

gingerBill: That assertion makes no sense either as it will always be true.

Tetralux: Quite right. I meant to assert that the count was within range of `i32` since I'm about to cast to that, and that cast doesn't check.

Tetralux:
I also have no idea why the Linux `-vet` is failing...

core/sync/sync_windows.odin
Tetralux: This now reads
```
assert(initial_count >= min(i32) && initial_count <= max(i32));
```

Tetralux: I'm open to better ways to do it though.

Tetralux: Okay yeah - I don't know how to best handle this - I want to say something like:
```
assert(is_within_range(initial_count, i32));
```

Tetralux: Although I guess we don't really want an assert on a common path; `semaphore_post` has a similar problem with `count` being an `i32`.
Sigh.

Tetralux:
Fun story: I enabled the threading demo in `examples/demo/demo.odin` for Linux.

Turned out I should have used a semaphore and not a condition for the start gate 😁 

Calling `start` very quickly after `create` sometimes hung because the condition was getting signalled before the thread had starting waiting for it. 🤣

Tetralux:
The demo is now enabled on Linux and works the same as Windows.

gingerBill:
This should work fine now with the latest release.  @Tetralux

Tetralux:
Apparently, although MacOS has it's own impl of pthreads which just forwards to IOKit (which then uses native threads ... I don't even know man) --- Point is, pthreads in theory should just work on MacOS.

HOWEVER - I can't enable `thread_linux.odin` and `sync_linux.odin` to build on both Linux _and_ MacOS -- Once @gingerBill figures out how to do that, I'll put that in and then it'll hopefully pass `-vet` on everything!

Tetralux:
There's a possibility that we might want to use IOKit directly (native threads ("Mach")) instead of pthreads on MacOS, or do the syscalls manually if we want to make `-no-crt` work on Linux?
I guess we'd wanna do the syscalls or something even on Linux...

Whatever! One thing at a time I guess.
Something to be aware of.

Tetralux:
*Added a message to the assert for custom thread stack alignment to be more ergonomic.*

Tetralux:
Err? Did the CI die? O.O

Tetralux:
Hmmm. Seems like `//+build !windows` does NOT in fact work for Mac...

Tetralux:
HERE. WE. GO.

Tetralux:
So... apparently `//+build` does  not override the implict behavior - which is to use the file suffix `_linux`, `_windows` etc, to only build on that platform, which then ignores `//+build` annotations.

I was under the impression that the specific use of those annotations overrode any implicitness about that - I was in fact mistaken. xD

Renamed `_linux` => `_unix`.

gingerBill:
@Tetralux Do you know what's causing the infinite hang on macOS?

Tetralux:
@gingerBill I'm not 100% sure. I have some guesses though.
Unfortunately, I don't currently have a Mac set up and can't easily get my hands on a VM to debug it which is making it more annoying that it need be.

Tetralux:
I have now fixed conditions such that they now match Windows' `AutoResetEvent` which is much more sane - their signalness now lingers until someone waits, thereby making them a lot more useful and a lot less hassle.

The latest failure is not this PR.

Also, @gingerBill, the compiler bug from earlier can be reproduced by toggling these lines:

https://github.com/odin-lang/Odin/blob/73d207ce8b1f19191b7395b0588295ea6e1a5510/core/thread/thread_unix.odin#L92:L93

https://github.com/odin-lang/Odin/blob/73d207ce8b1f19191b7395b0588295ea6e1a5510/core/sys/unix/pthread.odin#L8:L9

And by then running this program:
```
package test;

import "core:thread"

p :: proc(t: ^thread.Thread) {
	return;
}

main :: proc() {
	t := thread.create(p);
	thread.start(t);
	defer thread.destroy(t); // NOTE: if you comment this out, it works
}
```

Tetralux:
Well - we're back to a hang on MacOS which I shall investigate soon.
Otherwise, everything else seems to work.

Tetralux:
You did mean to remove the threading test, right? 😁

Tetralux:
Have disabled the SOA test in the demo until @gingerBill fixed comparison against `nil` for SOA arrays, since until then, `reserve_soa` will crash and cause the CI to report failure.

Tetralux:
Turns out the hang was creating a weird situation where `Thread.started` was getting set to `true` when it called `condition_init` in `Thread.create` -- even though that doesn't touch that variable!

Anyhow, this was fixed once I discovered that the declarations for the pthread structs on Darwin are very different from Linux.

For example, `pthread_t` is just a `u64` on Linux - on Darwin it's 8176 bytes!

Relevant to this bug though, `pthread_cond_t` is 4 bytes on Linux, but 8 on Darwin.
`Thread.started` is immediately after the condition in the thread struct...

HOWEVER - Still gotta fix semaphores on Darwin though. :)

Tetralux:
Fixed Darwin semaphores by calling into Mach. (Which is included in the Darwin libc.)

Tetralux:
Implemented threading priority on all platforms.

Tetralux:
Uses `Lowest`, `Normal` and `Highest` on Windows.

PR #459, opened 2019-11-01
Base32 and CSV added for core:encoding

zhibog:
CSV is pretty much done, unless you would want to add more complexity.

Base32 could probably use some refactoring, as it is based on ref stuff from Go. but it encodes the test vectors correctly.

Also provided the test vectors to the Base64 package for reference.


core/encoding/base64/base64.odin
```diff
@@ -90,4 +90,39 @@ decode :: proc(data: string, DEC_TBL := DEC_TABLE, allocator := context.allocato
...
+        delete(enc);
+    }
+
+    // Decode test
+    for v in test_vectors {
+        dec := decode(v.encoded);
+        fmt.printf("decode(\"%s\") => \"%s\" \t| want: \"%s\"\n", v.encoded, string(dec), v.plain);
+        delete(dec);
+    }
+}
```
ThisDrunkDane: These probably shouldn't stick around in the actual `core` library

zhibog: I can remove them. I had them for proof initially, since those are the test vectors from the actual RFC

ThisDrunkDane: Hmm, you could also turn it into a usage example as you have in `csv` and then do that same for `base32`?

core/encoding/csv/csv.odin
```diff
@@ -0,0 +1,106 @@
...
+write_raw :: proc(path: string, data: []byte) -> bool {
+    file, err := os.open(path, os.O_RDWR | os.O_CREATE | os.O_TRUNC);
+    if err != os.ERROR_NONE do return false;
+    defer os.close(file);
+
+    if _, err := os.write(file, data); err != os.ERROR_NONE do return false;
+    return true;
+}
+
+read :: proc(path: string, delimiter := DELIMITER, skip_header := false) -> ([]string, int, bool) {
```
ThisDrunkDane: for some self documentation, perhaps give the return values names?

ThisDrunkDane:
Have you thought about writing versions of the csv stuff that operates on strings as well? and not only on file paths. People might wanna load the csv data themselves into a string

core/encoding/csv/csv.odin
zhibog: Done

core/encoding/base32/base32.odin
```diff
@@ -0,0 +1,145 @@
...
+    15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25,  0,  0,  0,  0,  0,
+	 0,  0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 
+    15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25,  0,  0,  0,  0,  0,
+	 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  
+     0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+	 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  
+     0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+	 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  
+     0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0
+};
```
gingerBill: Pretty pattern

core/encoding/base32/base32.odin
```diff
@@ -0,0 +1,145 @@
...
+		switch len(data) {
+            case:
+                out[7] = ENC_TABLE[data[4] & 0x1f];
+                carry = data[4] >> 5;
+                fallthrough;
+            case 4:
+                out[6] = ENC_TABLE[carry | (data[3] << 3) & 0x1f];
+                out[5] = ENC_TABLE[(data[3] >> 2) & 0x1f];
+                carry = data[3] >> 7;
+                fallthrough;
```
gingerBill: Please use tabs for indentation only.

core/encoding/base32/base32.odin
zhibog: Should be ok now.

core/encoding/base32/base32.odin
zhibog: I went through and I think I caught all of them

ThisDrunkDane:
Can you open another one with just the Base32 stuff? :)

PR #473, opened 2019-11-13
Allow running Odin compiler without using visual studio command prompt or setting up environgment variables.

castano:
This change is using a slightly modified version of Jon's single file library to search the latest compiler and SDK library paths. I've modified Jon's code to user gb's defer and avoid including windows.h without WIN32_LEAN_AND_MEAN. A lot more could be done to integrate it better.

The main unfortunate side effect of this change is that it introduces various library dependencies in order to interact with COM objects. If this is not desired we could link the required entrypoints dynamically.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -271,3 +271,5 @@ shared/
 * .ll
 *.bc
 *.ll
+
+*.sublime-workspace
--- a/Odin.sublime-project
+++ b/Odin.sublime-project
@@ -1,0 +1,21 @@
+{
+	"folders":
+	[
+		{
+			"path": "."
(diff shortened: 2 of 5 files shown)


ThisDrunkDane:
Don't add a sublime project to this, developers might be using whatever editor they want

castano:
Sorry, I'm new to git and didn't realize commits created after the pull request would be automatically added to it. Let me see how can I clean up this mess.

castano:
OK, I've created a new branch, cherry picked the changes, and submitted it to #478. Closing this.

PR #487, opened 2019-12-01
add support for libc's dirent in core/os/os_linux.odin

grufwub:
Found out about and started coding in Odin yesterday evening. Already loving it! For a project I'm working on I needed bindings to dirent, but considering they're a core part of libc figured best to add this to core/os. Unless you think this should be kept in another library under core?

Brings support for: opendir(), fdopendir(), dirfd(), closedir(), readdir(), alphasort(), rewinddir(), scandir(), seekdir(), telldir()

Our equivalent for libc's DIR (typedef struct __dirstream) has been set to a rawptr due to it being opaquely typed and not used for anything other than holding onto a directory stream to be passed to other functions. Further supporting this intended usage, dirstream has a member 'fd' but this is only accessible via dirfd() without libc source.

### DIFF
--- a/core/os/os_linux.odin
+++ b/core/os/os_linux.odin
@@ -12,6 +12,7 @@ Handle    :: distinct i32;
 File_Time :: distinct u64;
 Errno     :: distinct i32;
 Syscall   :: distinct int;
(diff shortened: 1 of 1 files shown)


grufwub:
realizing now the issue here being that this is currently linux only... will work on a platform agnostic version and close this PR

PR #508, opened 2019-12-16
Add note to assert in scratch_allocator_proc to provide fix hint

Tetralux:


Tetralux:
Now panics with better message.

PR #526, opened 2019-12-31
Implement dynlib core library for unix/darwin

sci4me:
I have implemented the `dynlib` core library for unix/darwin. I added `//build` tags; not sure if this was necessary or correct. Other than that, as far as I can tell, it just works. :)

### DIFF
--- a/core/dynlib/lib_windows.odin
+++ b/core/dynlib/lib_windows.odin
@@ -1,3 +1,4 @@
+// +build windows
 package dynlib
 
 import "core:sys/win32"
--- a/core/dynlib/lib_unix.odin
+++ b/core/dynlib/lib_unix.odin
@@ -1,0 +1,21 @@
+// +build linux, darwin
+package dynlib
+
+import "core:os"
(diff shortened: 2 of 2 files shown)


Tetralux:
@sci4me You'd need the `//build` tags for `_unix` because it would otherwise compile it on all platforms, since only `_linux`, `_darwin`, `_windows`, and `_essence` are supported, I think?

At any rate, if you have a file that you want to compile on multiple--but not all--platforms, _that_'s when you _need_ to use `//build`.

Tetralux:
Note that you can also do things like `//build windows amd64, linux i386` if you really wanted to.
I forget the _exact_ syntax, but I think that's right.

core/dynlib/lib_windows.odin
```diff
@@ -1,3 +1,4 @@
+// +build windows
```
gingerBill: This is not needed if the file name has `_windows.odin`, however it does hurt to have it.

PR #530, opened 2020-01-02
Fixes for yielding and semaphore posting on unix

KTRosenberg:
- replaced pthread_yield with sched_yield
- fixed semaphore post procedures
Note to Tetralux, please double-test this

### DIFF
--- a/core/sync/sync_darwin.odin
+++ b/core/sync/sync_darwin.odin
@@ -28,9 +28,11 @@ semaphore_destroy :: proc(s: ^Semaphore) {
 }
 
 semaphore_post :: proc(s: ^Semaphore, count := 1) {
-	assert(count == 1);
-	res := darwin.semaphore_signal(s.handle);
-	assert(res == 0);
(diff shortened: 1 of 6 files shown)


Tetralux:
Not sure how to really test this.
The demo on Linux seems to work though - although again - not really sure how to know if it's borked or not.

I'll check on Darwin in the morning.

This indicates that `pthread_yield` is the same as `sched_yield` though, so I'm optimistic: https://linux.die.net/man/3/pthread_yield

core/sync/sync_darwin.odin
```diff
@@ -28,9 +28,11 @@ semaphore_destroy :: proc(s: ^Semaphore) {
 }
 
 semaphore_post :: proc(s: ^Semaphore, count := 1) {
-	assert(count == 1);
-	res := darwin.semaphore_signal(s.handle);
-	assert(res == 0);
+	// NOTE: SPEED: If there's one syscall to do this, we should use it instead of the loop.
+	for in 0..count-1 {
```
gingerBill: Just use `0..<count`

PR #488, opened 2019-12-01
Fix heap alignment from #458

Tetralux:

PR #537, opened 2020-01-14
Fix heap allocator alignment on all platforms

Tetralux:
Supercedes https://github.com/odin-lang/Odin/pull/488.

Implements the way @gingerBill suggested to do this that is, in his estimation, the more correct way to do this. Definitely simpler.

Did some basic testing with `append`ing, `resize`ing and `ordered_remove`ing - all seems to work as intended.

If there's any other tests you can think of to try I'll do those too before merging.

### DIFF
--- a/core/os/os.odin
+++ b/core/os/os.odin
@@ -129,85 +129,66 @@ read_ptr :: proc(fd: Handle, data: rawptr, len: int) -> (int, Errno) {
 heap_allocator_proc :: proc(allocator_data: rawptr, mode: mem.Allocator_Mode,
                             size, alignment: int,
                             old_memory: rawptr, old_size: int, flags: u64 = 0, loc := #caller_location) -> rawptr {
-/*
+
 	//
-	// NOTE(tetra, 2019-11-10): The heap doesn't respect alignment.
-	// HACK: Overallocate, align forwards, and then use the two bytes immediately before
-	// the address we return, to store the padding we inserted.
-	// This allows us to pass the original pointer we got back from the heap to `free` later.
+	// NOTE(tetra, 2020-01-14): The heap doesn't respect alignment.
+	// Instead, we overallocate by `alignment + size_of(rawptr) - 1`, and insert
+	// padding. We also store the original pointer returned by heap_alloc right before
+	// the pointer we return to the user.
 	//
 
-	align_and_store_padding :: proc(ptr: rawptr, alignment: int) -> rawptr {
-		ptr := mem.ptr_offset(cast(^u8) ptr, 2);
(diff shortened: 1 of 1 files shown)


Tetralux:
Hmmm.
Actually there's a problem here.

I just thought to try calling `make_aligned([]u8, 8, 4096)` - that seems to return the same address multiple times.

```go
package test

import "core:fmt"

main :: proc() {
	using fmt;
	
	m := make(map[uintptr]uintptr);

	for i in 0..1024 {
		p := mem.make_aligned([]u8, 8, 4096);
		x := uintptr(mem.raw_data(p));

		assert(x % 4096 == 0);
		println(x);

		if _, ok := m[x]; ok {
			panic("duplicate page");
		} else {
			m[x] = x;
		}
       }
}
```
```
1951763767296
1951763767296
D:/Software/odin/test.odin(30:4) panic: duplicate page
```

Tetralux:
After a night's sleep, I've fixed it. Turned out, @gingerBill's original code was one line wrong. 😁

gingerBill:
@Tetralux I did say that my code was untested and I literally just wrote it as an example.

Tetralux:
@gingerBill Only one line, mind 😛

core/os/os.odin
```diff
@@ -129,85 +129,67 @@ read_ptr :: proc(fd: Handle, data: rawptr, len: int) -> (int, Errno) {
...
 
-	recover_original_pointer :: proc(ptr: rawptr) -> rawptr {
-		ptr := cast(^u8) ptr;
-		offset := transmute(u16) (^[2]u8)(mem.ptr_offset(ptr, -2))^;
-		ptr = mem.ptr_offset(ptr, -int(offset));
-		return ptr;
-	}
+		allocated_mem: rawptr;
+		if old_ptr != nil {
+			allocated_mem = heap_resize(old_ptr, space+size_of(rawptr));
```
gingerBill: This is a bug. The old_ptr is not the original pointer allocated.

Tetralux: It is actually the original pointer because it's retreived in `aligned_resize` - but this could be clearer, so I'll move that next to the `heap_resize` call.

MoustaphaSaad:
tested it on my code and it works, though my code is not an "interesting codebases that use memory allocation" just normal arrays, strings, and maps

Tetralux:
The commit said it was only for Windows - that was a typo; was actually for all platforms.

gingerBill:
It only need to be implemented for Windows since \*nix has `aligned_alloc` support

Tetralux:
`aligned_alloc` requires libc, right.

Tetralux:
What else needs to be done here?

PR #541, opened 2020-01-17
Add os.get_current_directory / os.set_current_directory

Tetralux:
Also changes the `os_linux` foreign function declarations to use `core:c` types for arguments and returns, finally. They actually match the C declarations now.


Tetralux:
Build failure is a false positive. `core:strings` is imported and IS used in `dlsym`.

Tetralux:
What's the status of this?

PR #543, opened 2020-01-18
Add DLL loading procs on Windows

Tetralux:
- `dlopen`
- `dlclose`
- `dlsym`

### DIFF
--- a/core/os/os_windows.odin
+++ b/core/os/os_windows.odin
@@ -1,6 +1,7 @@
 // +build windows
 package os
 
+import "core:strings"
 import "core:sys/win32"
 
 OS :: "windows";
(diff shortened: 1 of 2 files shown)


Tetralux:
I tried providing `dlerror` too but I was unable to figure out `FormatMessage` - if anyone knows exactly how to do that and what flags I need to pass I'll add it too.

Tetralux:
I definitely did not accidentally base on PR off of another PR without noticing and had to remove them. *cough* Fixed now. 🤣

gingerBill:
This is not needed as `package dynlib` is meant to do that.

Tetralux:
Does that mean that these functions should also be removed from all platforms in favour of `package dynlib` ? @gingerBill

PR #567, opened 2020-02-20
Refactor strconv `parse_*` procedures: arbitrary base and return (value, rest_of_string)

Tetralux:
- `parse_uint :: proc(s: string, base: int) -> (value: uint, rest: string)`
- `parse_int :: proc(s: string, base: int) -> (value: int, rest: string)`
- `parse_u64 :: proc(s: string, base: int) -> (value: u64, rest: string)`
- `parse_i64 :: proc(s: string, base: int) -> (value: i64, rest: string)`
- `parse_f64 :: proc(s: string) -> (value: f64, rest: string)`

```odin
n, rest_i := strconv.parse_int("1234");
assert(n == 1234 && rest_i == "");

n, rest_i = strconv.parse_int("1234blah");
assert(n == 1234 && rest_i == "blah");

f, rest_f := strconv.parse_f64("1234.56");
assert(f == 1234.56 && rest_f == "");

n, rest_i = strconv.parse_int("efef", 16); // arbitrary bases can now be provided to `parse_u64`, `parse_i64`, `parse_uint` and `parse_int`
assert(n == 0xefef && rest_i == "");
```
This is achieved by splitting `parse_u64` and `parse_i64` into two procedures: `parse_u64_of_base` and `parse_u64_maybe_prefixed`. (Likewise for `i64` variants.)

The original behavior of these functions (e.g: `parse_int(s)` would parse a string with an optional prefix) _is maintained_ in this PR.
The behavior is only changed if you pass a value for the base.

Now with nice and shiny documentation comments!

### DIFF
--- a/core/strconv/strconv.odin
+++ b/core/strconv/strconv.odin
@@ -23,7 +23,53 @@ _digit_value :: proc(r: rune) -> int {
 	return v;
 }
 
-parse_i64 :: proc(str: string) -> i64 {
+// Parses an integer value from a string, in the given base, without a prefix.
+// ```
+// n, rest := strconv.parse_i64_of_base("-1234eeee", 10);
+// assert(n == -1234 && rest == "eeee");
+// ```
+parse_i64_of_base :: proc(str: string, base: int) -> (value: i64, rest: string) {
+	assert(base <= 16, "base must be 1-16");
+	s := str;
+	neg := false;
+	if len(s) > 1 {
+		switch s[0] {
+		case '-':
+			neg = true;
+			s = s[1:];
+		case '+':
+			s = s[1:];
+		}
+	}
+
+	i := 0;
+	reached_end := true;
+	for r in s {
+		i += 1;
+		if r == '_' do continue;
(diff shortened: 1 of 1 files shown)


Tetralux:
I should note, the way it was before meant that you could not detect failure easily; parse in a specific base easily; know how much of the string was parsed and should be skipped over; or parse integers without a prefix to denote their base - like `ffff` -> `65535` - you needed `0xffff` before.

In this PR, you can check for failure by checking that `rest == ""`; `rest` is the part of the string that wasn't parsed; and you can specify a base to disable the prefix stuff.

Tetralux:
Bump?

Tetralux:
Is there a better approach you'd rather I did, @gingerBill ?

Would it be better to have the procedures remain as single-returns?

Should the procedures in this PR be added alongside the current ones, instead of replacing them?

Should they be reworked to be more the "safe" versions of the procedures that return
 `ok: bool` and fail if the string contains extra junk, instead of the rest of the string, or is returning the rest of the string valuable?

PR #575, opened 2020-02-28
Fix mem.align_backward when pointer is already aligned

Tetralux:
Previously, it would always go back - now it only goes back if it's not aligned already.

```odin
assert(mem.align_backward_uintptr(34, 32) == 32);
assert(mem.align_backward_uintptr(31, 32) == 0);
assert(mem.align_backward_uintptr(32, 32) == 32); // failed before, passes now
```


Tetralux:
What's the status on this?

PR #579, opened 2020-03-04
Fix issues with stat struct.

powerc9000:
The stat struct was the format for the 64 bit version of stat. So we need to call stat64 to get the proper data.
Also we need to use _File_Time instead of File_Time because it is a compound value. 

These changes were tested and work on my computer, MacOS 64 bit.


Tetralux:
Note: Distinct from #581 - this PR addresses the same problem as that one, but only for Darwin. The other is only for Linux.

PR #582, opened 2020-03-07
Fix strings.clone and strings.clone_to_cstring if allocation fails

Tetralux:
They currently fail a bounds check, now they return ~~an empty value~~ `(result: string, ok: bool)`.

I also fixed an oversight in `core:os` where the strings on Linux and Darwin were being allocated and deleted, rather than using the temporary allocator.

In the course of changing the return types, I made `grow_builder` return `bool` so that---like `reserve`---you can check if it succeeded or not.


Tetralux:
Added an `ok: bool` return value as well on @oskarnp's suggestion, since without it, you'd incorrectly think there was an error in some cases.

Tetralux:
Should it return the `ok` or not? I'm kinda leaning towards not honestly.

PR #583, opened 2020-03-09
LLVM: speed improvement changes.

machinamentum:
This primarily sets the data layout for the module so that a memcpyopt pass can properly run. I have observed on average a 50% or better improvement in compilation speed. I have also turned off dead-store-elim since it seemed to be taking a lot of CPU time and not giving much in return in terms of object lowering speed.

Unfortunately, I have had trouble trying to figure out how to compile community member projects so I have only tested this on a small surface area that is the included example programs.


gingerBill:
Thank you very much for everything. This branch is still not brilliant and I need to test things a lot. When its ready to be tested by the public, I will say so.

PR #592, opened 2020-03-16
Fix os.read, os.write, os.read_entire_file, os.write_entire_file on Windows

Tetralux:
The second attempt of https://github.com/odin-lang/Odin/pull/586.

I also tested `write`/`write_entire_file` and discovered they have the same problem.

Currently, you cannot read/write more than a certain amount of data with `os.read`/`os.write`, which causes `read_entire_file` and `write_entire_file` to fail if the data length is very long. Reading a large file results in an empty slice, for example.

This is due to the foreign signature for `read_file`/`write_file` being incorrect. (`i32` instead of `u32`.)
I also rewrote `os.write` and `os.read`, since, even with that fixed, they do not fill the provided buffer, even if they can, which they are supposed to.

I'll be interested in the results of your tests @gingerBill 😛

### DIFF
--- a/core/os/os.odin
+++ b/core/os/os.odin
@@ -85,6 +85,7 @@ read_entire_file :: proc(name: string) -> (data: []byte, success: bool) {
 	if data == nil {
 		return nil, false;
 	}
+	// defer if !success do delete(data);
 
 	bytes_read, read_err := read(fd, data);
 	if read_err != ERROR_NONE {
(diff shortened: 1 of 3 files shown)


Tetralux:
With the introduction of `make` performing zeroing on newly allocated memory, I discovered that `RtlFillMemory` crashes if you pass it a large region of memory.
```odin
import "core:mem"

main :: proc() {
    _ = make([]byte, mem.gigabytes(7)); // dies
}
```

When @gingerBill writes his own `memset` instead of using this, with my changes these procedures should just work.

Tetralux:
`memset` has been fixed on Windows now, which allows this to work again. 🙂

Tetralux:
What's the status of this?

PR #596, opened 2020-03-25
Add mem.Dynamic_Arena

Tetralux:
Same as `mem.Arena`, but rather than being a single buffer, expands to use as many as needed.
Can be freed with one call to `*_destroy`, or reused with `*_reset`, and has a customizable buffer size.

Made this PR to get some eyes on it since I've been meaning to for a while.
My testing on it seems to indicate that it all works, but could use some other people testing it.


Tetralux:
Assuming that `virtual.Arena` would supercede this.

PR #610, opened 2020-04-18
Add bool flag to AST data in the parser in the core library

zhibog:
I'm creating an AST printer and this info was missing until now.
It now tells you if "do" was used vs "{...}" in a block statement.


kevinw:
@zhibog as an aside can you share your ast printer?

PR #616, opened 2020-04-21
Fix potential bad optimization bug in sync.Ticket_Mutex

Tetralux:
When locking, we were not loading `m.serving` atomically and so the optimizer
could have hoisted the check out of the loop, thus resulting in an infinite loop.

Thanks to @MasterQ32 and @SpexGuy for noticing and explaining this.


Tetralux:
Updated in lieu of #617 being merged.

Tetralux:
Bump

PR #630, opened 2020-04-29
Refactor parsing procedures: Return (value, ok) instead

Tetralux:
Supercedes #567.

Gives all the number-parsing procedures an extra return value: `ok: bool`, and adds the ability to parse an integer in any base up to 16, either signed or unsigned.

```odin
v, ok := strconv.parse_int("0x1234");
// 0x1234, true
```
```odin
v, ok = strconv.parse_int("1234");
// 1234, true
```
```odin
v, ok = strconv.parse_int("-1234");
// -1234, true
```
```odin
v, ok = strconv.parse_int("asd");
// 0, false
```
```odin
v, ok = strconv.parse_int("asd", 16);
// 10, true
```
```odin
v, ok = strconv.parse_int("xxx");
// 0, false
```
.. and the same for `parse_uint`, `parse_u64`, `parse_i64`, `parse_f32` and `parse_f64`, except that the unsigned variants return `0, false` for negative values, as well.

.. along with _even shinier_ documentation comments!

If you omit the base, then they will assume base 10, but if the string has a prefix (e.g `0x`) then it'll use that to determine what base it should be.

# NOTE

One thing to be aware of right now is that if you `parse_uint("0x1234", 16)` you get `(0, true)`.
This might be a bit error prone, since you might expect an error or `0x1234` (imagine it was read from a config file somewhere.)

To fix this, I can make it return `(0, false)` if you pass a base, but the string has a prefix. Thoughts?

### DIFF
--- a/core/strconv/strconv.odin
+++ b/core/strconv/strconv.odin
@@ -23,7 +23,53 @@ _digit_value :: proc(r: rune) -> int {
 	return v;
 }
 
-parse_i64 :: proc(str: string) -> i64 {
+// Parses an integer value from a string, in the given base, without a prefix.
+// ```
+// n, rest := strconv.parse_i64_of_base("-1234eeee", 10);
+// assert(n == -1234 && rest == "eeee");
+// ```
+parse_i64_of_base :: proc(str: string, base: int) -> (value: i64, rest: string) {
+	assert(base <= 16, "base must be 1-16");
+	s := str;
+	neg := false;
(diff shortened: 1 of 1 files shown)


gingerBill:
I think string prefixes can be handled as an edge case. Do what feels correct, if you understand what I mean.

gingerBill:
But overall, this looks good.

PR #657, opened 2020-05-24
Add pop_rune to core/strings/builder.odin

krixano:
Thought this could be useful to have in the builder - a function to pop off the last rune of the builder and return that.

Also, not sure if it should be a transmute or cast on line 51.

### DIFF
--- a/core/strings/builder.odin
+++ b/core/strings/builder.odin
@@ -46,7 +46,7 @@ builder_cap :: proc(b: Builder) -> int {
 	return cap(b.buf);
 }
 
-builder_pop_rune :: proc(b: ^strings.Builder) -> (rune, int) {
+builder_pop_rune :: proc(b: ^Builder) -> (rune, int) {
 	r, size := utf8.decode_last_rune_in_string(string(b.buf[:]));
 	raw_dyn := transmute(^mem.Raw_Dynamic_Array)&b.buf;
 	raw_dyn.len -= size;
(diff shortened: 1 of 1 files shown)


krixano:
Hm.... I'm getting this error now, but not sure what it means...
```

Run ./odin check examples/demo/demo.odin -vet
  ./odin check examples/demo/demo.odin -vet
  shell: /bin/bash -e {0}
/home/runner/work/Odin/Odin/core/strings/strings.odin(4:8) 'unicode' declared but not used
##[error]Process completed with exit code 1.
```

krixano:
Ok, the error is actually coming from the latest odin-lang:master commit.

core/strings/builder.odin
```diff
@@ -46,6 +46,14 @@ builder_cap :: proc(b: Builder) -> int {
 	return cap(b.buf);
 }
 
+builder_pop_rune :: proc(b: ^strings.Builder) -> (rune, int) {
+	r, size := utf8.decode_last_rune_in_string(string(b.buf[:]));
+	raw_dyn := transmute(^mem.Raw_Dynamic_Array)&b.buf;
```
Tetralux, line 51: This can indeed just be a `cast(^mem.Raw_Dynamic_Array)` since it's just a pointer.

Tetralux:
I'd suggest calling the procedure `pop_rune`, since we already have `write_string`, `write_rune`, etc.

krixano:
Thanks for suggestions!

gingerBill:
I have just implement pop_byte and pop_rune myself now. Sorry for not merging.

PR #660, opened 2020-05-25
Fix style of core:sys/win32 foreign procedures

Tetralux:
Change the style back to that of the Windows API to aid with porting Win32 code.
I've also added aliases so that code using the Odin-style procedure names will continue to work, and can be used in the future if the person prefers to.

It was changed to snake case originally to better fit the Odin style.
Turns out, that was a bad idea.

Note: I think all the aliases are correct - I checked pretty meticulously - but would appreciate it if others verified I did them all right. 😄

### DIFF
--- a/core/dynlib/lib_windows.odin
+++ b/core/dynlib/lib_windows.odin
@@ -8,18 +8,18 @@ load_library :: proc(path: string, global_symbols := false) -> (Library, bool) {
 	// NOTE(bill): 'global_symbols' is here only for consistency with POSIX which has RTLD_GLOBAL
 
 	wide_path := win32.utf8_to_wstring(path, context.temp_allocator);
-	handle := cast(Library)win32.load_library_w(wide_path);
+	handle := cast(Library)win32.LoadLibraryW(wide_path);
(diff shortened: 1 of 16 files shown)


gingerBill:
Don't forget the data types too.

Tetralux:
> 
> 
> Don't forget the data types too.

Done.

I think I got them all... 😄 

I also went through and changed all the corelib code use the actual names, rather than the aliases.

Tetralux:
If it's too much of a pain to validate I did them all correctly, I can revert that part of this if that'd help. @gingerBill

dotbmp:
HYPE

Tetralux:
I believe this is superceded by `core:sys/windows`.

gingerBill:
I was going to merge this PR but then I realise that it would break a lot of people's code so I thought I'd just write a completely different code path which can be written from scratch.

Tetralux:
Indeed. Although, I believe I added aliases for everything so that old code would actually have still worked. 😁 

Having it all redone from scratch does at least allow for a new start though.

PR #710, opened 2020-08-13
WIP: Add More Linux syscalls

krixano:
Add More Linux Syscalls `os/linux_os.odin`: getuid, getpwnam, getpwuid, readlink, opendir, readdir, closedir, setenv, and unsetenv. Also added types dirent, passwd, uid_t, gid_t, DIR, ino_t, and off_t

The new types might need to be checked to make sure they are correct, especially uid_t, gid_t, ino_t, off_t, and DIR (which seems to be an opaque type).

We might also want to make dirent nicer to work with. Dirent has a field that tells the size of the returned record, and this size can exceed the size of the defined struct if the d_name field exceeds 256 chars/bytes (this is described in the notes section of the readdir man page).
I was thinking of having a raw dirent type, and then the wrapper functions that wrap around the _unix_opendir, _unix_readdir, and _unix_closedir syscalls can pass a more usable struct that has a string that points to the d_name of the returned dirent as well as storing the correct length, which can be figured out by the wrapper function.

### DIFF
--- a/core/os/os_linux.odin
+++ b/core/os/os_linux.odin
@@ -449,47 +449,47 @@ getpwnam :: proc(name: string) -> ^passwd {
 	return _unix_getpwnam(cstr);
 }
 
-readlink :: proc(pathname: string, buf: []u8) -> (int, os.Errno) {
+readlink :: proc(pathname: string, buf: []u8) -> (int, Errno) {
 	cstr_pathname := strings.clone_to_cstring(pathname);
 	defer delete(cstr_pathname);
 	bytes_written := _unix_readlink(cstr_pathname, cstring(#no_bounds_check &buf[0]), c.size_t(len(buf)));
 	if bytes_written == -1 {
-		return -1, os.Errno(os.get_last_error());
+		return -1, Errno(get_last_error());
 	}
-	return int(bytes_written), os.ERROR_NONE;
+	return int(bytes_written), ERROR_NONE;
 }
 
-opendir :: proc(name: string) -> (^DIR, os.Errno) {
+opendir :: proc(name: string) -> (^DIR, Errno) {
 	cstr := strings.clone_to_cstring(name);
 	defer delete(cstr);
 	result := _unix_opendir(cstr);
 	if result == nil {
-		return nil, os.Errno(os.get_last_error());
+		return nil, Errno(get_last_error());
 	}
-	return result, os.ERROR_NONE;
+	return result, ERROR_NONE;
 }
 
-readdir :: proc(dirp: ^DIR) -> (^dirent, os.Errno) {
-	previous := os.Errno(os.get_last_error());
+readdir :: proc(dirp: ^DIR) -> (^dirent, Errno) {
+	previous := Errno(get_last_error());
 
 	result := _unix_readdir(dirp);
-	err := os.Errno(os.get_last_error());
+	err := Errno(get_last_error());
 
 	if result == nil && previous != err { // If nil and errno changed, err occured
 		return nil, err;
 	} else if result == nil { // If errno not changed, end of directory stream
-		return nil, os.ERROR_NONE;
+		return nil, ERROR_NONE;
 	}
 
-	return result, os.ERROR_NONE;
+	return result, ERROR_NONE;
 }
 
-closedir :: proc(dirp: ^DIR) -> os.Errno {
+closedir :: proc(dirp: ^DIR) -> Errno {
 	result := _unix_closedir(dirp);
 	if result == 0 {
-		return os.ERROR_NONE;
+		return ERROR_NONE;
 	} else {
-		return os.Errno(os.get_last_error());
+		return Errno(get_last_error());
 	}
 }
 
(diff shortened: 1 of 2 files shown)


core/os/os_linux.odin
```diff
@@ -406,6 +443,56 @@ access :: inline proc(path: string, mask: int) -> (bool, Errno) {
 	return true, ERROR_NONE;
 }
 
+getpwnam :: proc(name: string) -> ^passwd {
+	cstr := strings.clone_to_cstring(name);
```
Tetralux, line 447: Might be good to use the temp allocator for these things.
We do that elsewhere, for instance.

You can still defer-delete it though which will prevent it from leaking if it's too big.
https://github.com/odin-lang/Odin/blob/674aeffee4483aff5030ec40082ecd77da47f7a0/core/runtime/default_allocators.odin#L106-L114

core/os/os_linux.odin
```diff
@@ -253,6 +256,29 @@ X_OK :: 1; // Test for execute permission
...
+	d_off: off_t,
+	d_reclen: c.ushort, // Size in bytes of the returned record
+	d_type: c.uchar, // File type
+
+	d_name: [256]c.char, // Name of file, null-terminated, can exceede 256 chars (in which case d_reclen exceedes the size of this struct)
+}
+
 SYS_GETTID: Syscall : 186;
 
 foreign libc {
```
Tetralux, line 284: Odin wants to be able to use `-no-crt` on Linux, and to do that usefully we need to remove all usage of libc from the stdlib so that the stdlib may still be used.
Maybe we might want to still provide like a `core:libc` package or something, but. 🤔 
Or maybe we just want to do the syscall _ourselves_ rather than relying on libc - similar to how we do `gettid`, except using inline asm, when we have it.

For now, it's probably useful to still add the new libc things here, but we want to remove them eventually - maybe when we get inline asm, or maybe put them in their own package, like I say.

Anyway, maybe we should add a comment to this foreign block along the lines of:
```
// TODO: Remove dependance on libc. Waiting on inline asm for syscalls?
```

core/os/os_linux.odin
```diff
@@ -265,16 +291,27 @@ foreign libc {
 	@(link_name="write")            _unix_write         :: proc(fd: Handle, buf: rawptr, size: c.size_t) -> c.ssize_t ---;
 	@(link_name="lseek64")          _unix_seek          :: proc(fd: Handle, offset: i64, whence: c.int) -> i64 ---;
 	@(link_name="gettid")           _unix_gettid        :: proc() -> u64 ---;
+	@(link_name="getuid")           getuid              :: proc() -> uid_t ---;
```
Tetralux, line 294: `_unix_getuid` ?

core/os/os_linux.odin
```diff
@@ -265,16 +291,27 @@ foreign libc {
...
 	@(link_name="getpagesize")      _unix_getpagesize   :: proc() -> c.int ---;
 	@(link_name="stat64")           _unix_stat          :: proc(path: cstring, stat: ^Stat) -> c.int ---;
 	@(link_name="fstat")            _unix_fstat         :: proc(fd: Handle, stat: ^Stat) -> c.int ---;
 	@(link_name="access")           _unix_access        :: proc(path: cstring, mask: c.int) -> c.int ---;
 
+	@(link_name="getpwnam")         _unix_getpwnam      :: proc(name: cstring) -> ^passwd ---;
+	@(link_name="getpwuid")         getpwuid            :: proc(uid: uid_t) -> ^passwd ---;
```
Tetralux, line 301: Similar here.

core/os/os_linux.odin
```diff
@@ -253,6 +256,29 @@ X_OK :: 1; // Test for execute permission
...
+	pw_name: cstring,
+	pw_passwd: cstring,
+	pw_uid: uid_t,
+	pw_gid: gid_t,
+	pw_gecos: cstring,
+	pw_dir: cstring,
+	pw_shell: cstring,
+}
+
+DIR :: opaque [0]byte;
```
Tetralux, line 269: Probably better to either declare this here, or to use the _actual_ size for the array, so it's at least the right size.

Or is the struct _supposed_ to never be used directly?
You could make it `DIR :: distinct rawptr;` in that case, which would be the same style as `os.Handle`.

core/os/os_linux.odin
krixano: Well, yeah, I was considering it, but some of the other functions in this file that I didn't add didn't do this. The only one that did was like set_current_directory or something (I'll have to check on this).

core/os/os_linux.odin
krixano: Ok, thanks.

core/os/os_linux.odin
krixano: There didn't need to be a wrapper around getuid, which is why I didn't have both _unix_getuid and getuid like I did with the others.

core/os/os_linux.odin
krixano: Same reason as getuid. There didn't need to be a wrapper for getpwuid. I was mostly following the style of the pre-existing stuff in here, which is also why I didn't use temp allocator, since most of the functions didn't use that.

core/os/os_linux.odin
krixano: I think that DIR is supposed to be an opaque type. I'm not sure though. I couldn't find any definition for __dirstream (which is what DIR is defined as).
I actually originally used `DIRP :: distinct rawptr`, but them someone on discord, either you or Oskar, mentioned about the `opaque` thing, so I thought I'd use that to be more consistent.

core/os/os_linux.odin
krixano: However, I'm confused... the C package isn't libc. It's just a file defining the C types to their corresponding Odin types, isn't it?

core/os/os_linux.odin
krixano: Also, are you sure about the delete with the temp allocator, because last time somebody told me this, it crashed my program, lol

core/os/os_linux.odin
krixano: Btw, libc is definitely still useful atm until it can **fully** be replaced, because I need it for my new project (I'm rewriting my shell, Paled, in Odin). These functions happened to be additional syscalls that I needed for that.

PR #731, opened 2020-09-13
Add ProcOverload_ParamNames; overload on param names

bttelle:
Allows to resolve a specific procedure of a procedure group when
the types are identical and the parameter names are unique.  The
change introduces a new ProcOverload kind that relaxes the check
of the procedure group only when types are identical.  Omitting
the parameter names at the call site will continue to complain
about an ambiguitiy with a reasonable message.  As before, mixed
positional and named parameters are not allowed.

### DIFF
--- a/examples/demo/demo.odin
+++ b/examples/demo/demo.odin
@@ -480,6 +480,36 @@ explicit_procedure_overloading :: proc() {
 	// Ambiguous answers
 	// add(1.0, 2);
 	// add(1, 2.0);
+
+	fmt.println("\n# explicit procedure overloading using unique parameter names");
+
+	resize_frame :: proc{resize_frame_xyxy, resize_frame_xywh, resize_frame_ltrb};
+
+	resize_frame_xyxy :: proc(x0, y0, x1, y1 : int) {
+		fmt.printf(" frame resized to p0=(%d,%d) p1=(%d,%d)\n", x0, y0, x1, y1);
+	}
+
+	resize_frame_xywh :: proc(x, y, width, height : int) {
+		fmt.print("(xywh)");
+		// Allowed to use procedure group name inside specific members
+		resize_frame(x0=x, y0=y, x1=x+width, y1=y+height);
+	}
+
+	resize_frame_ltrb :: proc(left, top, right, bottom : int) {
+		fmt.print("(ltrb)");
+		// Allowed to reorder named arguments
+		resize_frame(x0=left, y1=top, x1=right, y0=bottom);
+	}
+
+	// These are equivalent
+	resize_frame(x=1, y=2, width=5, height=6);
+	resize_frame(left=1, top=8, bottom=2, right=6);
+
+	// Not allowed to mix named and unnamed arguments
+	// resize_frame(1, 1, x1=5, y1=10);
+
+	// Ambiguous
+	// resize_frame(1, 1, 5, 10);
 }
 
 struct_type :: proc() {
--- a/src/check_decl.cpp
+++ b/src/check_decl.cpp
@@ -1079,6 +1079,7 @@ void check_proc_group_decl(CheckerContext *ctx, Entity *pg_entity, DeclInfo *d) 
 				break;
 			case ProcOverload_ParamCount:
 			case ProcOverload_ParamTypes:
+			case ProcOverload_ParamNames:
 				// This is okay :)
 				break;
 
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -2229,6 +2229,8 @@ enum ProcTypeOverloadKind {
 	ProcOverload_ParamCount,
 	ProcOverload_ParamVariadic,
 	ProcOverload_ParamTypes,
+	ProcOverload_ParamNames,
+
 	ProcOverload_ResultCount,
 	ProcOverload_ResultTypes,
 	ProcOverload_Polymorphic,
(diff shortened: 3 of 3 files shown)


Tetralux:
This would make this compile correctly?:
```odin
a :: proc(x: int) -> int do return 1;
b :: proc(x: i32) -> i32 do return 2;
P :: proc{a, b};

// ...

v := P(i32(0));
assert(v == 2);

**Comment by Tetralux** [comment]
I'm guessing here, but this would make this work?:
```odin
a :: proc(x: int) -> int do return 1;
b :: proc(y: int) -> int do return 2;
g :: proc{a, b};

// ...

x := g(x=0);
assert(x == 1);

y := g(y=0);
assert(y == 2);

**Comment by bttelle** [comment]
> I'm guessing here, but this would make this work?:
> 
> ```odin
> a :: proc(x: int) -> int do return 1;
> b :: proc(y: int) -> int do return 2;
> g :: proc{a, b};
> 
> // ...
> 
> x := g(x=0);
> assert(x == 1);
> 
> y := g(y=0);
> assert(y == 2);
> ```
Yes that works as long as the asserts are not at file scope.  See the examples in demo.odin.

**Comment by gingerBill** [comment]
This is a very interesting PR! 

I am not sure I would allow it because I don't really see a practical use case for it, compared to the other procedure group rules. Could you show a realistic example of where this would be useful, @bttelle?

**Comment by bttelle** [comment]
It really comes down to matters of style if you think code like this is ugly or clarifying:
```odin
sleep :: proc{sleep_us, sleep_ms, sleep_s}
sleep_us :: (microseconds : int) { ... platform native code ... |
sleep_ms :: (milliseconds : int) { sleep(microseconds = milliseconds * 1000); }
sleep_s :: (seconds : int) { sleep(milliseconds = seconds * 1000); }

sleep(seconds=5);
```
In this case, it allows you to build descriptive interfaces that are extensible without having to resort to user defined literals like C++ or polluting a public namespace with tons of action_* functions.  I think it's maybe preferable to writing `sleep(5, seconds)` where seconds is a polymorphic disambiguator and the disambiguator is positional instead of named.

In the proposal I showed how to describe a bounding box in two different ways without having to resort to an extra builder abstraction; you can just keep the normalized coordinates internally as a couple of points but still give the caller options:

```odin
resize_frame(x=1, y=2, width=5, height=6);
resize_frame(left=1, bottom=2, right=6, top=8);
```

You have a great implementation where the named parameters can be in any order, so I thought it would be interesting to have this as a tool.  I'll keep it in my tree and as I write more Odin if I use it for something more realistic I can ping you back.  I wouldn't litter the Odin standard library with this idiom but the PR is so small I thought I'd try and get it in.  I'm trying to upstream everything I rely on.

gingerBill:
As for the `sleep` example, that is better handled with the distinct typing approach we already have:
```odin
time.sleep(3 * time.Second);
time.sleep(7 * time.Millisecond);
```

As for the `resize_frame` thing, that is an interesting point. I do wonder if that is a good idea or not in practice. I'll have to give a ponder.

bttelle:
Great point about reaching for distinct types first. They make for composable interfaces.

bttelle:
@gingerBill here's an "improvement" of `container.ring_link` using this technique.  The original has two positional parameters of same type where swapping them is disastrous and where the before/after insert behavior is implicit and fixed.  This version is impossible to get wrong because you can't call it with unnamed parameters.
```odin
package main;
import "core:fmt"
import "core:container"

ring_link :: proc{ring_link_before, ring_link_after};
ring_link_before :: proc(before, elem: ^$R/container.Ring) -> ^R {
    return container.ring_link(before, elem);
}
ring_link_after :: proc(after, elem: ^$R/container.Ring) -> ^R {
    return container.ring_link(after.next, elem);
}

main :: proc() {
    IntElem :: container.Ring(int);
    a := IntElem{value=1};
    b := IntElem{value=2};
    c := IntElem{value=3};
    node := container.ring_init(&a);    
    node = ring_link(before=node, elem=&b);
    node = ring_link(elem=&c, before=node);  /* field order independent */
    node = ring_link(elem=&d, after=node);
    // Ambigous now
    // node = ring_link(node, &c);
    fmt.println(node.value, node.next.value, node.next.next.value, node.next.next.next.value);
}

> 3 2 1 4
```

krixano:
```
sleep :: proc{sleep_us, sleep_ms, sleep_s}
sleep_us :: proc(microseconds : int) { ... platform native code ... |
sleep_ms :: proc(milliseconds : int) { sleep(microseconds = milliseconds * 1000); }
sleep_s :: proc(seconds : int) { sleep(milliseconds = seconds * 1000); }

sleep(seconds=5);
```

The problem I have with this example is, what happens when someone just does `sleep(5)`? I guess it just errors out?

bttelle:
> The problem I have with this example is, what happens when someone just does `sleep(5)`? I guess it just errors out?
Yes it is then ambiguous, which is a good thing because then it is a unitless call.

bttelle:
@gingerBill I was thinking of updating this PR and taking out user facing changes to `odin.demo` and just having the important bits from this put behind a experimental feature-flag command-line option with a message like:
```
WARNING: `experimental:paramnamepoly` is an experimental language feature.
It has not been accepted into the language.  It may never be released.  Use at your own risk!
```

Let me know if that would make it easier to merge.

PR #738, opened 2020-09-16
Aarch64 (Arm64) Support

krixano:
Added Arm64 support for both the compiler, and as a target. Still a WIP as there are some problems.
However, we have the compiler successfully running on Arm64 on Linux, and it successfully compiling the demo.odin program.
The demo program, however, segfaults in the "map types" section.

*A lot* of help was given by @Tetralux , who co-authored the first commit with me.


krixano:
Ok, so... I learned there are 3/4 classifications to ARM CPUs:
* Profile
* Execution Mode
* Instruction Set - dependent on Execution Mode
* And of course the version, Armv1-Armv8 (v1-v3 are obsolete now, according to documentation)

There are 3 Profiles:
* A - Application - Supports A64, A32, and T32 ISAs
* R - Real-time - Supports only A32 and T32 ISAs
* M - Microcontroller - Variation of T32 ISA

The 2 Execution Modes are:
* Aarch64 - *only* allows A64 ISA. New in Armv8, with Armv8 Exception model
* Aarch32 - allows A32 and T32 ISAs. Maps Armv7-A Exception model, based on *PE modes*, onto Armv8 Exception model.

And the 3 ISAs:
* A64 - 64-bit registers, fixed-length 32-bit instruction encodings
* A32 - 32-bit registers, fixed-length 32-bit instruction encodings. Also called *Arm* ISA in Armv7
* T32 - variable-length instructions that use both 16-bit and 32-bit instruction encodings. Also called *Thumb* ISA in Armv7

krixano:
https://developer.arm.com/documentation/dui0774/b/compiler-command-line-options/--target

According to the above documentation on target triples, Arm recommends these triples (for no Operating System):
* **aarch64-arm-none-eabi** - Armv8-A Aarch64 Execution State
* **armv8a-arm-none-eabi** - Armv8-A *Aarch32* Execution State
* **armv7a-arm-none-eabi** - Armv7-A

In the form of *architecture-vendor-OS-ABI*

gingerBill:
This has now been officially supported, but only tested on the Apple M1 platform.

PR #749, opened 2020-10-01
Allow .a or .o files to be linked on macos.

powerc9000:
In my projects I usually make a lot of .o files from C libs. This should allow to link them without having to change file extensions to get around the compiler.


Platin21:
Maybe we should also move the dylib stuff in the same if as it doesn’t do anything different.

Platin21:
This patch doesn't work for the old backend only the new one which makes it worthless on macOS.

PR #753, opened 2020-10-09
Fix error message when importing package that does not exist

Tetralux:
Previously on Linux, if a file in your program tried to import a package that did not actually exist, `read_directory()` assumed that
the `errno` after calling `opendir()` was `ENOTDIR`.

This was incorrect.

Instead, we now switch on `errno` and check for `ENOENT`, which it is if the directory does not exist.


gingerBill (approved):

Tetralux:
Cheers.

PR #759, opened 2020-10-24
Added Perf Package

colbyhall:
The idea of this package is for measuring performance. To start off the package I've added the high res query counter functionality with procedures for time measuring.


oskarnp:
Unix version is not correct.  tv_nsec only contains the nanosecond part of the seconds. So need both tv_sec and tv_nsec.

core/perf/perf_windows.odin
```diff
@@ -0,0 +1,30 @@
+// +build windows
+package perf
+
+import "core:sys/win32"
+
+@(private="file")
+qpc_freq : i64;
```
Tetralux, line 7: Might want to consider making this `@thread_local`.

colbyhall:
So whats the next step to get this merged?

gingerBill:
I have now integrated this code into commit 9e8c46b8, and reorganized it by placing it in `package time`

PR #760, opened 2020-10-25
Add container.queue_slice_front_to_back(^Queue(T)) -> []T

Tetralux:
Provides a way to get a slice of the elements.


Tetralux:
Since, unlike `push_back`, there is no `queue_pop` that defaults to `pop_front`, the order of the elements in the slice should probably be clear.

Have renamed the procedure added by this PR to `queue_slice_front_to_back`.

gingerBill:
This doesn't make any sense queues wrap back on themselves.

PR #767, opened 2020-10-31
Update thread_windows.odin

SrMordred:
`n` was left over being always `0`. But you want `win32.WaitForMultipleObjects` to be the number of threads to wait for which u already have with `j`.


SrMordred (commented):
Right types, i believe.

PR #778, opened 2020-11-08
allow absolute filepaths on windows

Lord-Remington-the-3rd:
this removes the colon from "illegal_import_runes" so absolute filepaths are allowed on windows (*nix stuff doesn't have C: etc).


gingerBill:
This is not allow on purpose. `:` is used for collections. If you want a collection defined one 

`-collection:name=C:\absolute\path\to\collection`

Lord-Remington-the-3rd:
this fix in particular wasn't for imports tho. someone was trying to -out to an absolute file path. which doesnt work on windows because you can use the : in path strings

PR #793, opened 2020-11-25
Fix map key_ptr when internal arrays are resized

jharler:
Fix map key_ptr when internal arrays are resized


gingerBill:
No need any more since `key_ptr` was removed entirely. That was only a transition phase.

PR #797, opened 2020-12-04
Improve number parsing procedures

Tetralux:
- No longer returns true for partial string parsing:
  All characters in the string must be part of the number. ~~(Ignoring leading or trailing whitespace.)~~
  i.e: `parse_int("2.345")` no longer returns `(2, true)` - it now returns ~~`(0, false)`~~ `(2, false)` [Edit: typo]

- Return `(0, false)` on empty input strings. ~~(After trimming whitespace.)~~

### DIFF
--- a/core/strconv/strconv.odin
+++ b/core/strconv/strconv.odin
@@ -25,7 +25,8 @@ _digit_value :: proc(r: rune) -> int {
 
 // Parses an integer value from a string, in the given base, without a prefix.
 //
-// Returns ok=false if no numeric value of the appropriate base could be found.
+// Returns ok=false if no numeric value of the appropriate base could be found,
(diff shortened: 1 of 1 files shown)


Tetralux:
Worth noting: as currently implemented, the procedures will still return the value that they parsed, even if they return false.
So `parse_int("1.234")` will still return the value `1`, if you wanted that for some reason.

If this is not desired, I can twiggle them to return `0` if they fail, instead.

gingerBill:
I would remove the space trimming, but otherwise, it looks good.

Tetralux:
Done; whitespace is no longer ignored.

Tetralux:
Outstanding!

PR #804, opened 2020-12-09
Fix _queue_increase_capacity to copy correct slice

iansimonson:
This function was previously copying a large, overlapping
range of data when the queue resized. The function caused
program to fail on out of bounds slices if the queue was
pop_front-ed from (offset > 0) and resized on full.

New code copies the range [0:offset] to [end:end+offset] where
end is the previous size of the queue buffer

Additional note: this fix made a failing Day 7 Advent of Code program produce the correct result 😄


iansimonson:
On second thought this is also not quite correct in the case where you want to arbitrarily resize the queue (rather than when full) will update soon with the full fix

iansimonson:
I can close this out and re-open as an issue / post it in the discord #bugs channel if that is preferred

iansimonson:
Opened #845

PR #865, opened 2021-03-10
Add LLVM boolean compatibility to result types

matias-eduardo:
This converts `bool` result types to `llvm_bool` when using the "none" calling convention, in a similar fashion to how we [convert param types](https://github.com/odin-lang/Odin/blob/84bb3499005e89184a46dc34dc575fd879cb6f26/src/check_type.cpp#L2181-L2187). This allows us to bind and use LLVM intrinsics that return an `i1` "one bit" boolean, such as these [overflow intrinsics](https://github.com/odin-lang/Odin/blob/84bb3499005e89184a46dc34dc575fd879cb6f26/core/math/bits/bits.odin#L179-L189).

### DIFF
--- a/src/check_type.cpp
+++ b/src/check_type.cpp
@@ -2388,6 +2388,18 @@ Type *type_to_abi_compat_result_type(gbAllocator a, Type *original_type, ProcCal
 		new_type = tuple;
 	}
 
+	if (cc == ProcCC_None) {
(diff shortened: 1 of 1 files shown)


gingerBill:
Minor criticism: please use tabs for indentation. I will correct this after the merge.

PR #877, opened 2021-03-23
Add core:datetime_to_time

Kelimion:
datetime_to_time takes separate parameters for date and time values and returns a time.Time and an `ok` bool.

If the values are out of range, they're considered modulo and ok will be set to false.


Kelimion:
To be used to convert a PNG's tIME chunk to time.Time, as well as an Odin equivalent of sscanf to parse dates, what have you.

PR #878, opened 2021-03-24
Add package declaration specific error message

bynect:
If the first token is not `package` a specific error message is emitted, instead of the default `expect_token` message. This should be a little more user-friendly. 

For example:
```
//test.odin
//package main

main :: proc() {
}
```

Now outputs the error message: `test.odin(3:1) Syntax Error: Expected package declaration on first line.`
Instead of the previous generic message: `test.odin(3:1) Syntax Error: Expected 'package', got 'identifier'`

Side note: is package declaration a correct definition or should this be called something else?

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -5219,11 +5219,15 @@ bool parse_file(Parser *p, AstFile *f) {
 	}
 
 	CommentGroup *docs = f->lead_comment;
+	
+	f->package_token = f->curr_token;
+	advance_token(f);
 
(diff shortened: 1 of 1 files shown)


gingerBill:
I think you are correct that a better error message is required here. I am not sure if your approach is the correct one but something better than the current one is in need. Thank you for highlighting it.

bynect:
Thank for your feedback. I don't know if this is the best approach, but it is straight forward, just replacing the `expect_token` function and doing a manual check.

PR #879, opened 2021-03-25
Allow seeding CRC32, CRC64 & Adler32 with previous partial hash.

Kelimion:
When writing chunked formats like PNG, sometimes you just want to calculate a hash on a length field's bytes, then extend the hash with the actual contents, without first having to concatenate the two into a new buffer before passing them to a hash function.

It may also at times be cheaper to compute a rolling hash as you output. Computing a hash over the entirety of contents is not always feasible either, think large files you don't necessarily want to load completely.

```odin
Foo     := []u8{'F', 'o','o', '3', 'F', 'o', 'o', '4'};
crc     := hash.crc32(Foo[0:4]);
crc      = hash.crc32(Foo[4:], crc);
crc_all := hash.crc32(Foo);

fmt.printf("%8x %8x\n", crc, crc_all);
```
d6285ff7 d6285ff7

```odin
a32     := hash.adler32(Foo[0:4]);
a32      = hash.adler32(Foo[4:], a32);
a32_all := hash.adler32(Foo);
fmt.printf("%8x %8x\n", a32, a32_all);
```
0c5102b0 0c5102b0

### DIFF
--- a/core/hash/crc.odin
+++ b/core/hash/crc.odin
@@ -1,14 +1,14 @@
 package hash
 
-crc32 :: proc(data: []byte) -> u32 #no_bounds_check {
-	result := ~u32(0);
+crc32 :: proc(data: []byte, seed := u32(0)) -> u32 #no_bounds_check {
+	result := ~u32(seed);
 	for b in data {
 		result = result>>8 ~ _crc32_table[(result ~ u32(b)) & 0xff];
 	}
 	return ~result;
 }
-crc64 :: proc(data: []byte) -> u64 #no_bounds_check {
-	result := ~u64(0);
+crc64 :: proc(data: []byte, seed := u32(0)) -> u64 #no_bounds_check {
+	result := ~u64(seed);
 	for b in data {
 		result = result>>8 ~ _crc64_table[(result ~ u64(b)) & 0xff];
 	}
--- a/core/hash/hash.odin
+++ b/core/hash/hash.odin
@@ -2,9 +2,9 @@ package hash
 
 import "core:mem"
 
-adler32 :: proc(data: []byte) -> u32 {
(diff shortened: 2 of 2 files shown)


core/hash/crc.odin
```diff
@@ -1,14 +1,14 @@
 package hash
 
-crc32 :: proc(data: []byte) -> u32 #no_bounds_check {
-	result := ~u32(0);
+crc32 :: proc(data: []byte, seed := u32(0)) -> u32 #no_bounds_check {
```
gingerBill, line 3: Should the default seed be `~u32(0)` instead?

Kelimion: No, the seed that is passed in, 0 by default, gets inverted to become the ~0 if run as before.
This way, to extend the hash you don't have to invert the previous hash first.

Kelimion: I'm using the the CRC in the PNG reader without passing the optional parameter, and the Adler32 with the optional parameter as a rolling hash as I output bytes. They're correct, just more convenient.

Old CRC32 init:
`result := u32(~0);`

New CRC32 init:
```
seed := 0;
result := u32(~seed);
```

They're equivalent. When the optional seed for a preceding CRC run is not passed, the function will sub in the default 0 and it'll be inverted. The optimizer will elide that ~ and load result with ~0 directly when no seed is passed, I'm sure.

Doing it this way means that chained hashing doesn't require inverting the previous hash when passing it:

 ```
 crc := crc32(data);
 crc  = crc32(more_data, crc);
 
 versus:
 
 crc := crc32(data);
 crc  = crc32(more_data, ~crc);
 ```

gingerBill (approved):

Issue #895, opened 2021-04-05
Can't build demo.odin


Platin21:
## Context

* LLVM API
This is enabled in my case and uses llvm11.1 (as for the other are no binaries available)

* Operating System:
MacOS 11.2.3 on x86_AMD64 

* Please paste `odin version` output:
Commit => d3665331c76cd10658d921ec141759c4ec833693
odin version 0.13.1-d3665331

## Expected Behavior

The demo program builds and links without any problems. 

## Current Behavior

The compilation of the demo program fails.

## Failure Information (for bugs)

It seems like there is a function used in the codegen which is not existent for this architecture.
This seems to be `___extendhfsf2` 

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. type => odin build examples/demo/demo.odin

### Failure Logs

Undefined symbols for architecture x86_64:
  "___extendhfsf2", referenced from:
      _fmt.fmt_arg in demo.o
ld: symbol(s) not found for architecture x86_64

#### Notes: 
@ThisDrunkDane it seems like the build system doesn't check wether demo.odin did even run correctly or somehow didn't catch this as it can be clearly seen for odin run that it get's the same output as I do.

Platin21:
Hmm interesting it seems like this something inside the compiler_rt library is this suppose to be used?

Kelimion:
`extendhfsf2` is used by the new f16 functionality.

Platin21:
@Kelimion Yes something I expected too from the name of it. What I don't get though is from where does it get normally delivered and is this something llvm generates or is this added explicitly through adding that function?

Platin21:
I suspect it uses this => https://code.woboq.org/llvm/compiler-rt/lib/builtins/extendhfsf2.c.html on other platforms but I don't think we do link on macOS with the compiler-rt of clang/llvm.

Kelimion:
> I suspect it uses this => https://code.woboq.org/llvm/compiler-rt/lib/builtins/extendhfsf2.c.html on other platforms but I don't think we do link on macOS with the compiler-rt of clang/llvm.

A bit surprising it's not in the Mac release given that I see references to `extendhfsf2.c` dating back to 2015.

Kelimion:
It's also in the latest clang on Apple's site: https://opensource.apple.com/source/clang/clang-800.0.42.1/src/projects/compiler-rt/lib/builtins/extendhfsf2.c.auto.html

Platin21:
[...]

PR #896, opened 2021-04-05
Add's Handling for Return Values for Linker-Stage

Platin21:
I added return values so that if the linker fails it also gets reflected in the return value of the program. 
Tried to use gb_exit and exit but zsh behaved weird with both of them. 

This is related to #895


gingerBill (approved):

Platin21:
This will show the current macOS error that happens because of some linking bug.

Platin21:
And it did exactly what I expected 👍

gingerBill:
Perfect!

Platin21:
Doesn't look like it changed can we re-run this.

Platin21:
Ok looks good now with that change. Do we still want this? @ThisDrunkDane though it was already in the compiler that when the linker fails it returns the codes.

PR #904, opened 2021-04-14
Suggestions for core:os2.

Kelimion:


Kelimion:
An additional suggestion not included in this PR would be to have a core:os/locale helper that retrieves fmt format strings for date, time, etc, in a cross-platform way, as well as an ISO language code. Using this fmt could then display core:time variables in a locale-specific way.

Companion code for internationalization could be added to core:i18n, the simplest implementation of which would comprise a .po file loader and a string->string map that returns the key if the key is not found, so that i18n.lookup("Untranslated") would return "Untranslated" or the translation thereof if present.

gingerBill:
The locale for time will be handled purely by `package time` and completely separate from the OS, especially since it is not OS specific data.

As for ISO language code stuff, that may be better in a completely different package too, unrelated to `package os`.

PR #909, opened 2021-04-23
Add install to Makefile, build.sh and build-m1.sh

alichraghi:


krixano:
The install in the makefile should depend on release rather than debug, I would think?

AliChraghi:
> The install in the makefile should depend on release rather than debug, I would think?

Because normally (all) was on "debug, demo" I did this

krixano:
Currently, if someone were to use `make install`, it would build a debug binary and *then* move that. Imo, that should not be the expected behavior. Debugging does not require installing, hence why the default is debugging, but with installing, you would usually want a *release* build.

AliChraghi:
> Currently, if someone were to use `make install`, it would build a debug binary and _then_ move that. Imo, that is not the expected behavior. Debugging does not require installing, hence why the default (all) is debug, but with installing, you would usually want a _release_ build.

I would tell gingerBill

krixano:
Sure, we can let gingerbill decide. I just wanted to mention this.

An alternative way that I think is more common is to not have dependencies on install at all, and require the user to run make themselves *before* doing `make install`. I believe I usually see this more often.

gingerBill:
What needs to happen before this is to make the path to `core:` configurable through an environment variable such as `ODIN_CORE_PATH`.

AliChraghi:
> What needs to happen before this is to make the path to `core:` configurable through an environment variable such as `ODIN_CORE_PATH`.

I do not think this PR have problem
We can do it later

gingerBill:
> > What needs to happen before this is to make the path to `core:` configurable through an environment variable such as `ODIN_CORE_PATH`.
> 
> I do not think this PR have discrepancy
> We can do it later

PR #910, opened 2021-04-24
add return to _readlink

atkurtul:


gingerBill:
Two things wrong with this PR:

* This is clearly a patch over a compiler bug, so please make an issue for the bug, not a patch
* You added extra things to the Makefile which is not allowed yet

PR #914, opened 2021-04-25
Removal of the Old Backend

gingerBill:
Total removal of the old backend in favour of the backend using the integrated LLVM C API.


.github/workflows/ci.yml
```diff
@@ -29,11 +29,11 @@ jobs:
       - name: build odin
         run: make release
       - name: Odin run
-        run: |
-          ./odin run examples/demo/demo.odin
-          ./odin run examples/demo/demo.odin -llvm-api
+        run: ./odin run examples/demo/demo.odin
       - name: Odin check
         run: ./odin check examples/demo/demo.odin -vet
+      - name: Odin version
```
ThisDrunkDane, line 35: Why do you want this extra step to run? It's just extra time spend in CI unless you plan on going and checking it?

build.bat
```diff
@@ -59,7 +66,10 @@ set linker_settings=%libs% %linker_flags%
 del *.pdb > NUL 2> NUL
 del *.ilk > NUL 2> NUL
 
-cl %compiler_settings% "src\main.cpp" /link %linker_settings% -OUT:%exe_name%
+rem cl %compiler_settings% "src\main.cpp" /link %linker_settings% -OUT:%exe_name%
+
+cl %compiler_settings% "src\main.cpp" /link %linker_settings% -OUT:%exe_name% ^
+	&& odin run examples/demo
```
ThisDrunkDane, line 72: Could you remove the `&& odin run examples/demo`? there is the check on line 75 that if we're not building release mode we run the demo. For the CI the demo is run explicitly in it's own step

src/build_settings.cpp
```diff
@@ -71,9 +71,11 @@ TargetEndianKind target_endians[TargetArch_COUNT] = {
 	TargetEndian_Little,
 };
 
+#ifndef ODIN_VERSION_RAW
+#define ODIN_VERSION_RAW "dev-2021-unknown"
```
ThisDrunkDane, line 75: Personally I think it should just be `dev-unknown-unknown`

src/main.cpp
```diff
@@ -1194,7 +1184,8 @@ bool parse_build_flags(Array<String> args) {
 							break;
 
 						case BuildFlag_UseLLVMApi:
-							build_context.use_llvm_api = true;
+							gb_printf_err("-llvm-api flag is not required any more\n");
```
ThisDrunkDane, line 1187: Why error out on the flag specifically? I don't remember Odin yelling at you about unknown flags?

src/main.cpp
```diff
@@ -2064,7 +2049,7 @@ int main(int arg_count, char const **arg_ptr) {
 		#endif
 
 		#ifdef GIT_SHA
-		gb_printf("-%s", GIT_SHA);
+		gb_printf(":%s", GIT_SHA);
```
ThisDrunkDane, line 2052: Why the change?

src/main.cpp
```diff
@@ -2187,450 +2158,69 @@ int main(int arg_count, char const **arg_ptr) {
...
+	}
 
-		if (build_context.build_mode == BuildMode_Object) {
-			// Ignore the linker
-			if (build_context.show_timings) {
-				show_timings(&checker, timings);
+	remove_temp_files(gen.output_base);
+
+#if defined(GB_COMPILER_MSVC)
+	if (false) {
```
ThisDrunkDane, line 2187: `if(false)` Why not #ifdef it out completely?

src/main.cpp
```diff
@@ -2187,450 +2158,69 @@ int main(int arg_count, char const **arg_ptr) {
...
-						LIT(output_base),
-						LIT(build_context.resource_filepath)
-					);
-
-          if(result != 0) {
-            return 1;
-          }
-
-          result = system_exec_command_line_app("msvc-link",
-						"\"%.*slink.exe\" \"%.*s.obj\" \"%.*s.res\" -OUT:\"%.*s.%s\" %s "
```
ThisDrunkDane, line 2412: Where did the linker call go? Don't we still need that with the LLVM C API?

ThisDrunkDane (commented):
I mainly only looked at the CI-related stuff

build.bat
gingerBill: I will do, this was an accidental commit

src/main.cpp
gingerBill: It just looked aesthetically wrong. Making it `:` make it not look like a date if the hash began with two numbers

src/main.cpp
gingerBill: Oversight

src/main.cpp
gingerBill: The linker call is in a separate procedure. The old backend did everything inline.

src/main.cpp
ThisDrunkDane: Hmm, not something I would see but a fair reason

src/main.cpp
ThisDrunkDane: Ah fair enough

gingerBill: See `linker_stage`

PR #925, opened 2021-04-30
Fix `core:sys/win32` tests to use `core:testing`.

Kelimion:
Updated the `core:sys/win32` tests to use `core:testing`.


ThisDrunkDane (approved):
LGTM

ThisDrunkDane:
@gingerBill

gingerBill:
I'll merge it, I was just hoping to have the win32 generation done by now so that most of this would not be needed.

PR #929, opened 2021-05-01
Put #no_bounds_check to the loops it is needed at for Base64. I had to do this, …

zhibog:
…because it seems like it no longer applies it to the whole procedure if you put it at the procedure level. I noticed it when making the crypto lib compliant with strict style. I was getting out of bounds errors.

Feel free to close the PR if this behavior of #no_bounds_check is not intentional.


Kelimion:
Maybe first find out why it triggers OOB and fix that problem before hiding it.
There's obviously a correctness issue if it goes OOB in the first place?

PR #983, opened 2021-05-14
WIP: Linux os v2

rytc:
Work on removing libc and moving over to the new os v2 structure.

Current Status:
- Implemented syscalls: `open, close, lseek, read, write, fsync, fstat, lstat, readlink, mkdir, getcwd, rmdir, chmod, umask, chdir, truncate, rename, link, symlink, chown, lchown, utime, setxattr, pipe, pipe2, mmap, munmap, mrepamp`
- Syscalls are located in `core/sys/unix/syscalls_linux.odin` so if anything changes with the inline assembly, it only needs to be changed in one place.

Sub-modules:
- **Stat:** Done
- **Path:** Done (permissions are kinda wonky)
- **User:** Done? (Looks like it needs API work still. Currently hard-coded for each platform)
- **File:** Done (Need guidance on how `_read_from` should work. chmod needs permission passed in)
- **Pipe:** Done
- **Env:** Odin needs to expose the env variables to the program from the main function. See notes in `env_linux.odin`
- **Heap:** Linux does not have a "HeapAlloc" like Windows, so this will require writing a heap allocator ontop of mmap
- **Temp_File:** Need to define how these procedures will work.
- **Process:** Not done (needs API work still?)


atkurtul:
You are moving stuff into wrong registers. You can find the correct signatures [here](https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md). Also it could be a better idea to just use constants in the asm string for syscall numbers

rytc:
Submitted a change to use the r* registers, the issue with read still persists.

atkurtul:
```c
read::proc"contextless"(fd: int, buf: ^[]u8) -> int {
  return asm(int, ^u8, int) -> int #sideeffects {
        ` xorl %eax,%eax
          syscall`,
        "={rax},{rdi},{rsi},{rdx}",
    }(fd, &buf[0], len(buf));
};
```
This works for me

rytc:
Alright, it's all working now. I fleshed out stat/fstat so that it now fills out the whole File_Info struct, through I need to decide on how to handle the path allocation for that.

PR #1003, opened 2021-05-27
Fix deleting substring in filepath.rel

jockus:


gingerBill:
Has this been tested on Windows?

jockus:
Only tested on windows

jockus:
Platform should make no difference however - even if the volume length is 0 the changes are correct.

PR #1014, opened 2021-06-10
fix string defines through command line arguments

atkurtul:


gingerBill:
The problem here is purely with regards to how `argv` is handled and how it strips quotes. The solution is to see if you can get the unstripped string rather than trying to make a bodge and guess.

A good example is that `"true"` and `true` are currently parsed the same by the `argv`.

atkurtul:
Only accepted values through `-define` seem to be booleans, strings and integers. Assuming delivering a string literal with proper quotes through the arguments is not uniform across shells and is very hard., i think this solves the current problem.

PR #1016, opened 2021-06-12
 straightforward align functions

bumbread:
These SWAR algorithms (a) remove branch, (b) have less operations performed, (c) also assume that the argument to the function is a power of two, effectively not changing the outcome.


krux02:
I am not responsible for anything here, so my approval doesn't weight anything yet. But you have my approval. This is the exact same version of align that I used in the nim compiler here: 


https://github.com/nim-lang/Nim/blob/5d1608c9764416e74b36f6a772f16a66d58ddc28/compiler/sizealignoffsetimpl.nim#L16

PR #1062, opened 2021-07-25
Minimal passes for opt:0 and -debug to preserve debug information reporting

JesseRMeyer:
Early out as soon as possible while maintaining the minimum LLVM optimization passes for opt:0 to preserve debug variable reporting.

This was tested with the official demo.odin and my own odin program on opt:0, 1, and 2, with and without -debug.  But I would consider this PR only minimally tested.

### DIFF
--- a/src/llvm_backend_opt.cpp
+++ b/src/llvm_backend_opt.cpp
@@ -39,15 +39,14 @@ void lb_populate_function_pass_manager(lbModule *m, LLVMPassManagerRef fpm, bool
 	optimization_level = gb_clamp(optimization_level, 0, 2);
 
 	lb_add_must_preserve_predicate_pass(m, fpm, optimization_level);
-
-	if (ignore_memcpy_pass) {
-		lb_basic_populate_function_pass_manager(fpm);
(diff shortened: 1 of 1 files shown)


src/llvm_backend_opt.cpp
```diff
@@ -39,15 +39,14 @@ void lb_populate_function_pass_manager(lbModule *m, LLVMPassManagerRef fpm, bool
...
-	if (ignore_memcpy_pass) {
-		lb_basic_populate_function_pass_manager(fpm);
-		return;
-	} else if (optimization_level == 0) {
+	if (optimization_level == 0) {
 		LLVMAddMemCpyOptPass(fpm);
-		lb_basic_populate_function_pass_manager(fpm);
 		return;
 	}
+	else if (ignore_memcpy_pass) {
```
gingerBill, line 46: You cannot do the `AddMemCopyOptPass` if `ignore_memcpy_pass` is set

JesseRMeyer:
Closing to re-open an updated PR alternative.

PR #1065, opened 2021-08-01
Add `strings.cut`, which returns a substring.

Kelimion:


Kelimion:
Fixed, and also return a clone of the input string if  `rune_length <= 0` instead of `== 0`, as per comments above the proc.

PR #1073, opened 2021-08-09
Potential way to detect C main entry point and output the main entry point when needed

thePHTest:
I suspect this might not be handling the -no-entry-point flag correctly


thePHTest:
Fixed by today's commit: https://github.com/odin-lang/Odin/commit/fb0a3ab7c14d4bc3b821cef723ec6ea3e956c956

PR #1095, opened 2021-08-22
Rewrite thread_pool.cpp

nakst:
Comments are appreciated. I'm new to the Odin code base.

Waiting threads spin on a `yield()` call. Usually I'd use an auto-reset Win32 event object, but I don't know how this translates to POSIX threads. It doesn't seem to cause a performance penalty though.

I have not tested this on Windows, but it appears to fix the threading issues I and other people were having on Linux.


nakst:
I've run `./odin build examples/demo/demo.odin` (modified to early out before calling LLVM) in a while loop for 15 minutes on my processor on Linux without any hangs or crashes. This makes me reasonably confident the new code is working.

gingerBill (approved):

PR #1098, opened 2021-08-23
SDL2: Add basic image loading example.

Kelimion:
`SDL2.dll` needs to be available.
If `USE_SDL2_IMAGE` is enabled, `SDL2_image.dll` and `libpng16-16.dll` also need to be present.
If `USE_SDL2_IMAGE` is not enabled (the default), the example uses `core:image/png` to load the image.

On Windows this means placing them next to the executable.


Kelimion:
Replaced by https://github.com/odin-lang/examples.

PR #1112, opened 2021-08-31
Optional Semicolons

gingerBill:
Make semicolons truly optional within the language Odin. This effectively makes `-insert-semicolon` on by default now (and not opt-out-able). This is a NON-BREAKING CHANGE, which means all valid code before the PR is valid code with the PR.

(This has been planned for over a year)

### Features

Two new features to the compiler:

* `odin strip-semicolon` utility function which will strip unneeded tokens (i.e. statement terminator semicolons). Takes the same parameters as `odin check`
* `-strict-style` will print a syntax error on an unneeded token (i.e. statement terminator semicolons).

The entirety of `core` and vendor` have had their semicolons stripped.

### Rationale

Rationale behind making statement terminator semicolons fully optional:

* Previous rules for semicolons were inconsistent
* Previous rules for semicolons were a little random
* Better to go full one way or the other than the half-arsed approach that existed
* Enforcing semicolons everywhere would have made a lot of people mad
* Aids with tooling such as code generation
* Minor point regarding advertisement: many people (mostly idiots) will dismiss a language in [the current year] purely for having semicolons still

### Rules

Full rules for semicolon insertion (directly from the compiler):
```c
switch (token->kind) {
case Token_Invalid:
case Token_Comment:
	// Preserve insert_semicolon info (from previous token)
	break;
case Token_Ident:
case Token_context:
case Token_typeid:
case Token_break:
case Token_continue:
case Token_fallthrough:
case Token_return:
case Token_or_return:
	/*fallthrough*/
case Token_Integer:
case Token_Float:
case Token_Imag:
case Token_Rune:
case Token_String:
case Token_Undef:
	/*fallthrough*/
case Token_Question:
case Token_Pointer:
case Token_CloseParen:
case Token_CloseBracket:
case Token_CloseBrace:
	/*fallthrough*/
case Token_Increment:
case Token_Decrement:
	/*fallthrough*/
	// Insert semicolon after this token if newline is found
	t->insert_semicolon = true;
	break;
default:
	// DO NOT insert semicolon after this token if newline is found
	t->insert_semicolon = false;
	break;
}
```


gingerBill:
These are the current optional semicolon rules (__before__ this PR).
```
Current optional semicolon rules
--------------------------------

A semicolon is optional if the statement is followed by one of these tokens on the same line:
    } 
    ) 
    else
    "EOF" 

After any block ("{}") (any control flow block)

If a constant value declaration's last expression is one of the following expressions, then a semicolon is not required:

    '#type'    of any type on the list (recursively)
    'distinct' of any type on the list (recursively)
    pointer    of any type on the list (recursively)

    procedure literal

    // Only optional at file scope, required in a procedure body
    struct type
    union type
    enum type

A semicolon is optional after the following declarations:
    package declaration        (package name)
    import declarations        (import "foo")
    foreign import declaration ("foreign import "bar")
    foreign block declaration  (foreign lib { })
```

My person opinions about these rules is that they are a complete [bodge](https://www.urbandictionary.com/define.php?term=bodge). And I don't want permanent bodges in the language's rules. And the bodges still have issues to them, which would require adding even more rules to make them "feel good".

gingerBill:
So it looks like the main arguments **against** making optional semicolons are:
1) I'm used to semicolons
2) I don't mind typing semicolons
3) Semicolons look good to read (partially due to being used to it)
4) Semicolons are good to type because they act like a full-stop/period on a sentence in many languages
5) I see little benefit over making them optional
6) Consistent with other languages such as C or GLSL, when swapping between them

Counter-points:
1) is just habit
2) is fine IF you knew what the rules were and if you had to type them everywhere, then you'd realize how much you have come to rely on optional semicolons. My bet is that the vast majority would hate it and want some form of optional semicolons in place regardless
3) I agree this is true, but so are newlines
4) I agree this is true but it's mostly due to (1)
5) That's because of (1) (2) and not understanding why people dislike semicolons
6) Odin is already different enough from those, especially with regards to declaration syntax, that trying to unify them is not really a good idea

gingerBill:
These are the arguments **for** making semicolons optional:

1) semicolons are visual clutter and don't aid reading practice
2) semicolons are not needed in other "modern" languages in [the current year]
3) the parsing rules are a lot more consistent, simplified, and not a bodge
4) easier for code generation tools
5) productivity is slightly increased when you don't have to type semicolons, many typos arising from forgetting to type a semicolon

Counter-pointers:

1) they do aid in reading in that they show the end of the statement, relying on a newline is not necessarily that easy to read
2) why do other language TRENDS matter that much? I don't care about fads.
3) But the previous ones felt fine, why change?
4) make better code generation tools
5) In a statically typed and compiled language, those typos are quickly caught. And the slight increase to productivity is highly debatable

PR #1148, opened 2021-09-09
Add vendor wren

nico-ec:


vendor/Wren/wren.odin
```diff
@@ -0,0 +1,203 @@
...
+    bindForeignClassFn : BindForeignClassFn,
+    writeFn : WriteFn,
+    errorFn : ErrorFn,
+    initialHeapSize : c.size_t,
+    minHeapSize : c.size_t,
+    heapGrowthPercent : c.int,
+    userData : rawptr,
+}
+
+@(default_calling_convention="c")
```
Kelimion, line 77: You could use `@(default_calling_convention="c", link_prefix="wren")`

That way you don't have to provide the link name for each procedure.

```odin
 @(link_name="wrenInitConfiguration")
    InitConfiguration :: proc(configuration : ^Configuration) ---;
```
Would turn into:
```odin
InitConfiguration :: proc(configuration : ^Configuration) ---;
```

Kelimion (changes requested):
Please use `link_prefix`, like so:
```odin
@(default_calling_convention="c", link_prefix="wren")
foreign wren {
    InitConfiguration :: proc(configuration : ^Configuration) ---;
}
```
That way you don't have to provide the link name for each procedure.

Skytrias:
nice work :+1: 
maybe strip-semicolon too :smile:

Kelimion (approved):

PR #1161, opened 2021-09-13
fix custom alignment clamping warning

krux02:
I was following some conversation on discord about this message, and that it was incorrect. here a fix.


gingerBill (approved):

krux02:
Ok, this error message has been removed. So this PR is outdated.

PR #1194, opened 2021-09-25
Allow `-define:name=string`.

Kelimion:
Slightly rewrote `build_param_to_exact_value` parsing logic.

First, try to parse as a bool, then as an integer or float.
If both fail, treat the value as a string.

To force a string define for e.g. `true`, use `-define:name='true'`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -688,35 +688,55 @@ void add_flag(Array<BuildFlag> *build_flags, BuildFlagKind kind, String name, Bu
 
 ExactValue build_param_to_exact_value(String name, String param) {
 	ExactValue value = {};
-	if (str_eq_ignore_case(param, str_lit("t")) ||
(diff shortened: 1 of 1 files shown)


Kelimion:
It works like this:
`-define:name=value`, where if the value is `t`, `true`, `f`, `false` or can be parsed as a number, it will. It's parsed as a string otherwise.
If you want to force it to be a string, wrap the value in `'`.

String with spaces:
    Windows:
    `odin run . "-define:FOO=Foozle Foo"`
    Linux:
    `odin run . "-define:FOO=Foozle Foo"`

Number coerced to string:
    Windows:
    `odin run . -define:FOO='1234a'` <- without the quotes it's parsed as an untyped int 1234.
    Linux:
    `odin run . -define:FOO=\'1234a\'` or `odin run . "-define:FOO='1234a'"`

PR #1203, opened 2021-09-30
Add a stopwatch to time.odin

thePHTest:


gingerBill:
As this is a stopwatch, `tick_now()` should probably be used since it is high precision and a relative clock.

thePHTest:
Updated to use `tick_now()`
Should I move this over to perf.odin instead with all of the other Tick related code?

maxtrem271991:
tick_now

thePHTest:
Added overloads of time.clock(...) for Duration and Stopwatch

maxtrem271991:
@thePHTest

PR #1256, opened 2021-10-31
Add a way to securely scrub memory

Yawning:
Per https://github.com/odin-lang/Odin/discussions/1253

This adds `core:intrinsics.mem_zero_volatile` and `core:mem.zero_explicit`.  The latter routine is the more widely useful of the two and is intended to provide equivalent functionality to `explicit_bzero`/`SecureZeroMemory`/`OPENSSL_cleanse`.


gingerBill:
Overall, this is a good PR!

I'm not sure on the name for `mem.zero_explicit` but the concept is good.

PR #1271, opened 2021-11-05
don't try to use __cpuid() on arm64

hdooley:
__cpuid() isn't available on arm architecture. we might be able to recreate this functionality using sysctl, but it looks like we're already bailing on calling cpuid on arm anyway.

### DIFF
--- a/src/bug_report.cpp
+++ b/src/bug_report.cpp
@@ -146,10 +146,17 @@ void report_windows_product_type(DWORD ProductType) {
 #endif
 
 void odin_cpuid(int leaf, int result[]) {
-	#if defined(GB_COMPILER_MSVC)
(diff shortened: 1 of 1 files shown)


Kelimion:
I have been thinking about recreating `cpuid` on ARM, as you say. On Linux we can parse `/proc/cpuinfo`, but I hope there's a better way. For MacOS on M1 and iOS, sysctl might do the trick? I don't have an M1 to work on, though.

Anyway, approved.

hdooley:
I can look into this for you. sysctl is probably the ticket.

Kelimion:
> I can look into this for you. sysctl is probably the ticket.

Appreciated.

PR #1276, opened 2021-11-06
Add macOS Homebrew support to the Makefile.

NotKyon:
This change detects if Homebrew is installed, and if so asks it for the LLVM 11 installation directory via `brew --prefix llvm@11`. This enables Odin to be built easily on macOS for users of [Homebrew](https://brew.sh/). The only prerequisite is the installation of LLVM 11 via `brew install llvm@11`. After that, no changing of toolchains or environment variable fiddling is needed to build Odin straight away.

Previously, if newer versions of LLVM were installed, Odin would not build or would fail to find LLVM 11 on macOS. This change enables targeting the specific version of LLVM needed by Odin.


gingerBill:
Minor issue with this approach is that it assumes that the user will be using Homebrew on macOS which may not be the case. This needs to support both cases.

NotKyon:
Made some changes. Latest commit resolves the conflict between the updated Makefile and the one I'm providing.

The change now checks to see if the user specified a LLVM_CONFIG and lets that take precedence over checking whether Homebrew is available or not. If none is specified, then we check to see if Homebrew is installed at all (as we previously did), and then to see if the required version of LLVM (via `$(LLVM_VERSION)`) is installed in Homebrew. If it is we retrieve its path and use that to determine the exact `llvm-config` to use.

This change allows the user to remain in complete control over which LLVM version and `llvm-config` they wish to use, while allowing Homebrew users to benefit from automatic installation detection, playing more nicely with the myriad LLVM installations some users will have.

Makefile
```diff
@@ -9,19 +9,25 @@ OS=$(shell uname)
...
     ARCH=$(shell uname -m)
-    LLVM_CONFIG=llvm-config
 
     # LLVM Version Setting  
     LLVM_VERSION_PATTERN="^11\."
     LLVM_VERSION="11"
     ifeq ($(ARCH), arm64)
-	LLVM_VERSION="13"
+        LLVM_VERSION="13"
         LLVM_VERSION_PATTERN="^13"
```
Platin21, line 18: Can we add the fix for the pattern too? 
```suggestion
        LLVM_VERSION_PATTERN="^13\."
```

NotKyon: Sure thing. Just committed the suggestion (and made a small aesthetic change, swapping two of the prior lines to match the later lines).

NotKyon:
The recent improvements to the Makefile in the main branch have improved the general workflow for me. I'll close this pull request for now and possibly revisit in the future if LLVM installation conflicts (which can be automatically resolved with Homebrew) should become an issue once more.

PR #1277, opened 2021-11-06
core/crypto: Add some "modern" primitives

Yawning:
See https://github.com/odin-lang/Odin/discussions/1275 for some context.

 * [x] Constant-time memory compare
 * [x] X25519
 * [x] poly1305
 * [x] ChaCha20 (IETF, XChaCha20)
 * [x] chacha20poly1305
 * [x] Cryptographic entropy source
   * [x] Address Linux systemcall portability issues. 
 * [x] Investigate performance some more (poly1305 is lower than what it should be, probably should look at it with vTune)
 * [x] Figure out why the rand_bytes test breaks horribly when built with `-o:speed` -> #1307 

This is intended:
 * To be enough (in combination with the existing hash function support) to implement a working subset of the Noise Protocol Framework (the full protocol requires AES-GCM and X448).
 * To serve as a playground/discussion point for the associated discussion regarding library/API design (The API is not final and subject to change).


Yawning:
Ok, this should be ready for (initial) review now.

Major software implementation references used:
 * fiat-crypto (1-BSD)
 * monocypher (public domain)
 * poly1305-donna (public domain)
 * Various bits of my work over the years.

The quirks in how I implemented things are as follows:
 * poly1305 uses a faster (but not fiat-derived) deserialization routine for performance reasons when applicable.
 * The X25519 RFC has an optional "check and reject if all-zero output" for contributory behavior.  This is intentionally omitted to match historical curve25519 behavior, and because not doing provides some additional flexibility. 
 * The curve25519 field inversion does a bit more work than strictly required for X25519 in anticipation of supporting Ed25519/ristretto255.

TODO (at a later date probably?):
 * More tests would be a good idea (eg: Wycheproof)
 * Someone that's "Not Me(TM)" needs to add crypto.rand_bytes support for other operating systems.
   * BSD/Darwin - `getentropy` (Warning: 256-bytes per request)
   *  Windows - `RtlGenRandom`/`CryptGenRandom`/`BCryptGenRandom`
 * It might be interesting to test the standard library on Linux 386/arm/aarch64, since it has a non-zero probability of working now.

Yawning:
Hm, the `rand_bytes` test fails when the binary is built with `-o:speed` and the compiler is using LLVM 11.  When it is built with optimization and LLVM 13, the tests pass.  I suspect that I am somehow hitting an optimization issue, but the giant block of LLVM IR isn't the easiest thing to read.

zhibog:
Looks very nice. Good for merge I'd say.

gingerBill:
Is this ready to merge?

Yawning:
> Is this ready to merge?

I think so.  There is always more to do, but my intention is to do it incrementally.

PR #1281, opened 2021-11-07
Preserve debug info opt 0

JesseRMeyer:
This PR differs from the last one in two ways:

The logic is simplified (including removing stale, dead, code). 
The MemCpyOpt pass has become optional due to recent LLVM work on master, which is now reflected in this PR.

Tested on the demo.odin for all combinations with -debug and -opt:[0-2, small, speed]

Lastly, this PR is up to date with master.


JesseRMeyer:
A community member has reported the following code causes this PR to crash on Windows 10, which I confirmed.  LLVM dereferences a null pointer in `LLVMRunFunctionPassManager(fpm, p->value);` from `lb_run_function_pass_manager`.

```package main
import "core:hash"
main :: proc() {
  a := hash.fnv32({})
}
```

JesseRMeyer:
The PR interacts poorly (for some yet to be determined reason) with @optimization= procedure tags.  A momentary work around is to comment them out.

PR #1289, opened 2021-11-09
Add functionality to export build timings.

Kelimion:
Usage:
`odin build examples/demo -show-extra-timings -export-timings:json -export-timings-file:timings.json`
`odin build examples/demo -show-timings -export-timings:csv -export-timings-file:timings.csv`

Example output:
```json
{
        "totals": [
                {"name": "total_packages",  "count": 23},
                {"name": "total_files",     "count": 114},
                {"name": "total_lines",     "count": 44413},
                {"name": "total_tokens",    "count": 257884},
                {"name": "total_file_size", "count": 1134058},
        ],
        "timings": [
                {"name": "Total Time", "millis": 1720.383},
                {"name": "initialization", "millis": 2.019},
                {"name": "parse files", "millis": 57.618},
                {"name": "type check", "millis": 316.631},
                {"name": "LLVM API Code Gen", "millis": 1239.287},
                {"name": "msvc-link", "millis": 104.822},
        ],
}
```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1256,87 +1256,88 @@ bool parse_build_flags(Array<String> args) {
 							break;
 						}
 
-						case BuildFlag_Debug:
+						case BuildFlag_Debug: {
 							build_context.ODIN_DEBUG = true;
 							break;
-
-						case BuildFlag_DisableAssert:
+						}
+						case BuildFlag_DisableAssert: {
 							build_context.ODIN_DISABLE_ASSERT = true;
 							break;
-
-						case BuildFlag_NoBoundsCheck:
+						}
+						case BuildFlag_NoBoundsCheck: {
 							build_context.no_bounds_check = true;
 							break;
-
-						case BuildFlag_NoDynamicLiterals:
+						}
+						case BuildFlag_NoDynamicLiterals: {
 							build_context.no_dynamic_literals = true;
 							break;
-
-						case BuildFlag_NoCRT:
+						}
+						case BuildFlag_NoCRT: {
 							build_context.no_crt = true;
 							break;
-
-						case BuildFlag_NoEntryPoint:
+						}
+						case BuildFlag_NoEntryPoint: {
 							build_context.no_entry_point = true;
 							break;
-
-						case BuildFlag_UseLLD:
+						}
+						case BuildFlag_UseLLD: {
 							build_context.use_lld = true;
 							break;
-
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -929,10 +933,10 @@ bool parse_build_flags(Array<String> args) {
 						}
 
 						if (ok) switch (bf.kind) {
-						case BuildFlag_Help:
+						case BuildFlag_Help: {
```
gingerBill, line 936: Please don't add blocks around cases when they were not needed in the first place.

src/main.cpp
```diff
@@ -1508,6 +1567,107 @@ bool parse_build_flags(Array<String> args) {
...
+
+		gb_fprintf(&f, "\t],\n");
+
+		gb_fprintf(&f, "}\n");
+	}
+
+	/*
+		CSV export
+	*/
+	else if (build_context.export_timings_format == TimingsExportCSV) {
```
gingerBill, line 1651: Minor style issue: keep to the `} else if` style if possible and add the comment either after the `) {` or on the next line

src/timings.cpp
```diff
@@ -169,14 +169,19 @@ f64 time_stamp(TimeStamp const &ts, u64 freq, TimingUnit unit) {
 	}
 }
 
-void timings_print_all(Timings *t, TimingUnit unit = TimingUnit_Millisecond) {
+void timings_print_all(Timings *t, TimingUnit unit = TimingUnit_Millisecond, bool timings_are_finalized = false) {
```
gingerBill, line 172: Is there anything that sets `timings_are_finalized` to `true` at any point?

gingerBill (commented):
Overall logic seems absolutely fine, just some minor style issues.

src/timings.cpp
Kelimion: Yes, `main.cpp:show_timings():1693` originally called this procedure and then finalized the timings. As I'm iterating over the timings again for the export, I don't want to stop the clock twice. So now it's called first for the stdout output as before, only with the implicit `false`, and then the export calling it sets it to `true`.

src/main.cpp
Kelimion: Addressed in 8c5c45a04c2a525e048538d764cc35f44239230b.

src/timings.cpp
gingerBill: Okay. Makes sense!

PR #1292, opened 2021-11-10
Add test command to CLI usage output

kevinsjoberg:
It was not obvious to me that there existed a `test` command since it was not documented. I reused the same documentation as you get when you run `odin test -help`. If you want me to adjust the documentation, I'd be happy to do so. 🙂

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -540,6 +540,7 @@ void usage(String argv0) {
 	print_usage_line(1, "          one must contain the program's entry point, all must be in the same package.");
 	print_usage_line(1, "run       same as 'build', but also then runs the newly compiled executable.");
 	print_usage_line(1, "check     parse and type check .odin file");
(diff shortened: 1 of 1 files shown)


gingerBill:
One of the reasons why it is not there is because it's not supported well on most platforms. So I didn't want to advertise it until it is.

kevinsjoberg:
Oh, I see. That makes sense. Should I close this for now then?

Is there anything I can assist with in terms of extending platform support?

gingerBill:
We need to improve support for both Linux and Darwin. What needs to be done is effectively catch a panic on any test and recover from it. On Windows this is achieved with using a separate thread per test (still ran serially) and a custom kernel exception handler. The same approach (if necessary) can be taken by those other platforms too.

We'll probably also need to make a minor check for `odin test` which checks to see that platform supports `odin test` at all or not.

I'll close this PR for the time being.

PR #1296, opened 2021-11-10
Do not filter test procedures when filter is empty

kevinsjoberg:
When #1293 got merged, a bug surfaced where it would unconditionally filter test cases no matter what. This have the side effect of filtering out all test cases unless you explicitly name them.

If `build_context.test_names` is empty, we do not need to perform any filtering.


DanielGavin:
Would be nice to get this merged soon.

Issue #1307, opened 2021-11-13
amd64 syscall intrinsic does not specify the correct clobber constraints


Yawning:
Note: Not using the template because I root-caused the issue, and it affects all current versions.

Some of my test cases fail with optimization enabled.  I root caused the issue to some of my values getting obliterated by the systemcall intrinsic.  The current amd64 systemcall intrinsic does not specify any clobber constraints, which is universally incorrect (with some extra platform specific annoyances).

Universal issues:
 * (Fixed: #1308) ~The `SYSCALL` instruction uses `RCX` to save the return address, and `R11` to save RFLAGS.~
 * (Fixed: #1308) ~Some but not all system calls also clobber `memory`.~

Platform specific issues:
 * FreeBSD will also clobber `R8`, `R9`, and `R10`, and uses `CF`(RFLAGS gets clobbered) to indicate an error.
 * I have no idea what Darwin does.

Yawning:
Hm, the BSD variants still needs fixing here...

PR #1308, opened 2021-11-13
src: Fix the syscall intrinsic code generation for Linux and Windows

Yawning:
The old assembly generated for the syscall intrinsic did not specify
clobber constraints.  This adds RCX and R11 (that are clobbered by
the instruction itself), and memory (that is clobbered by some
system calls) to the assembly constraints.

Note: This is still incorrect on FreeBSD, which clobbers more registers
and uses the carry flag instead of -errno in rax to indicate an error.

Partial fix for #1307

### DIFF
--- a/src/llvm_backend_proc.cpp
+++ b/src/llvm_backend_proc.cpp
@@ -2002,7 +2002,22 @@ lbValue lb_build_builtin_proc(lbProcedure *p, Ast *expr, TypeAndValue const &tv,
 						constraints = gb_string_appendc(constraints, regs[i]);
 						constraints = gb_string_appendc(constraints, "}");
 					}
-					
+
+					// The SYSCALL instruction stores the address of the
+					// following instruction into RCX, and RFLAGS in R11.
+					//
+					// RSP is not saved, but at least on Linux it appears
+					// that the kernel system-call handler does the right
+					// thing.
+					//
+					// Some but not all system calls will additionally
+					// clobber memory.
+					//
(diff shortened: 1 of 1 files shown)


src/llvm_backend_proc.cpp
```diff
@@ -2002,7 +2002,22 @@ lbValue lb_build_builtin_proc(lbProcedure *p, Ast *expr, TypeAndValue const &tv,
...
+					// following instruction into RCX, and RFLAGS in R11.
+					//
+					// RSP is not saved, but at least on Linux it appears
+					// that the kernel system-call handler does the right
+					// thing.
+					//
+					// Some but not all system calls will additionally
+					// clobber memory.
+					//
+					// TODO: FreeBSD is different and will also clobber
```
gingerBill, line 2016: I'd recommend just placing this in an if statement then. e.g.
```c
if (build_context.metric.os != "freebsd") {
   ...
}
```

src/llvm_backend_proc.cpp
```diff
@@ -2002,7 +2002,22 @@ lbValue lb_build_builtin_proc(lbProcedure *p, Ast *expr, TypeAndValue const &tv,
...
+					// R8, R9, and R10.  Additionally CF is used to
+					// indicate an error instead of -errno.
+					constraints = gb_string_appendc(constraints, ",~{rcx},~{r11},~{memory}");
```
gingerBill, line 2019: Does having these always added make a difference on both Linux and Darwin for lengths of syscalls?

PR #1310, opened 2021-11-14
Speed up Odin raw version generation for build.bat

Kelimion:
Getting the raw version string the old way, then jumping to `end_of_build`:
```
 -DODIN_VERSION_RAW=\""dev-2021-11"\"
[Finished in 128ms]
```
Getting the raw version string the old way, then jumping to `end_of_build`:
```
 -DODIN_VERSION_RAW=\""dev-2021-11"\"
[Finished in 77ms]
```


gingerBill:
Bit of a hack, not sure I like that much that you have to build a c executable to do this.

Kelimion:
> Bit of a hack, not sure I like that much that you have to build a c executable to do this.

Compiling it on the fly would add half a second, so the executable  is provided.

PR #1318, opened 2021-11-19
vendor/sdl2: add static libraries for Linux

ftphikari:
SDL bindings assume the latest version of libraries, however most Linux
distributions only supply tagged versions, the last of which for ttf is more
than 2 years old.

This commit adds compiled Linux/glibc libraries for all SDL modules.


vendor/sdl2/image/sdl_image.odin
```diff
@@ -3,8 +3,12 @@ package sdl2_image
 import "core:c"
 import SDL ".."
 
-when ODIN_OS == "windows" { foreign import lib "SDL2_image.lib"    }
-when ODIN_OS == "linux"   { foreign import lib "system:SDL2_image" }
+when ODIN_OS == "windows" { foreign import lib "SDL2_image.lib" }
+when ODIN_OS == "linux" {
+	foreign import lib "linux-SDL2_image.a" // commit @168ceb5
+	@require foreign import "system:dl"
+	@require foreign import "system:pthread"
```
gingerBill, line 10: If you are wanting to do this, you need to write it like this:
```odin
foreign import lib {
	"linux-SDL2_image.a",
	"system:dl",
	"system:pthread",
}
```

gingerBill: Same with every other instance of these imports.

gingerBill (changes requested):
SDL2 should not use static libraries for Linux for numerous reasons, and the SDL team recommend people dynamically linking to SDL.

ftphikari:
> SDL2 should not use static libraries for Linux for numerous reasons, and the SDL team recommend people dynamically linking to SDL.

That is what they say. I don't necessarily agree, since they broke the behavior of programs in the past. I don't want my program to work one way, and then not work differently when they release a new version. Still, they themselves provide the build script to compile a static library, which is what i did. And this static library has an option to be overwritten by a dynamic one at run-time by using a special environment variable.

So should we really care about what they say in light of those reasons? Or does portability of an executable matters more?

ftphikari:
Since we agreed to have shared libs on Linux and to use workarounds for now, i'm closing this.

PR #1320, opened 2021-11-19
Allow custom LLVM_CONFIG

mohd-akram:
This patch was required to package odin for MacPorts (now available via `sudo port install odin`).


mohd-akram:
@gingerBill Is this good to merge?

PR #1331, opened 2021-11-24
Recover from closing brace not found in field list

DanielGavin:
The current parser breaks completely after incorrect field lists:
```
My_Struct :: struct {
	type: int,
	wrong
	indent: int,
	width: int,
}
```
The fix is to keep looking for a closing brace or EOF. 

This change is really necessary to prevent OLS from giving wrong information or crashing.

### DIFF
--- a/core/odin/parser/parser.odin
+++ b/core/odin/parser/parser.odin
@@ -416,7 +416,16 @@ expect_closing_brace_of_field_list :: proc(p: ^Parser) -> tokenizer.Token {
 		str := tokenizer.token_to_string(token)
 		error(p, end_of_line_pos(p, p.prev_tok), "expected a comma, got %s", str)
 	}
-	return expect_token(p, .Close_Brace)
+	expect_brace := expect_token(p, .Close_Brace)
+
+	if expect_brace.kind != .Close_Brace {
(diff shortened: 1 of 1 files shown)


core/odin/parser/parser.odin
```diff
@@ -416,7 +416,16 @@ expect_closing_brace_of_field_list :: proc(p: ^Parser) -> tokenizer.Token {
 		str := tokenizer.token_to_string(token)
 		error(p, end_of_line_pos(p, p.prev_tok), "expected a comma, got %s", str)
 	}
-	return expect_token(p, .Close_Brace)
+	expect_brace := expect_token(p, .Close_Brace)
+
+	if expect_brace.kind != .Close_Brace {
+		for p.curr_tok.kind != .Close_Brace && p.curr_tok.kind != .EOF {
```
gingerBill, line 422: I'd probably also check for a semicolon (and make sure it is not a newline too).

DanielGavin: Alright added that. This should hopefully be good enough for now. I might later down the line try to recover the other fields, but the most important thing is to recover all the other declarations outside of the struct.

gingerBill:
@DanielGavin Is this still required?

PR #1342, opened 2021-11-30
Initial version of `core:encoding/xml`.

Kelimion:
```
	An XML 1.0 / 1.1 parser

	A from-scratch XML implementation, loosely modelled on the [spec](https://www.w3.org/TR/2006/REC-xml11-20060816).

	Features:
		- Supports enough of the XML 1.0/1.1 spec to handle the 99.9% of XML documents in common current usage.
		- Simple to understand and use. Small.

	Caveats:
		- We do NOT support HTML in this package, as that may or may not be valid XML.
		  If it works, great. If it doesn't, that's not considered a bug.

		- We do NOT support UTF-16. If you have a UTF-16 XML file, please convert it to UTF-8 first. Also, our condolences.
		- <[!ELEMENT and <[!ATTLIST are not supported, and will be either ignored or return an error depending on the parser options.

	TODO:
	- Optional CDATA unboxing.
	- Optional `&gt;`, `&#32;`, `&#x20;` and other escape substitution in tag bodies.

	MAYBE LATER:
	- XML writer?
	- Serialize/deserialize Odin types?
```


Kelimion:
Removed another unnecessary bit. Down to 634 lines of Odin, excluding blanks and comments.
Around 1500 lines in total, including the example, test suite and test documents.

It'll soon grow a bit to add the optional CDATA unboxing and SGML escape decoding, but it's a testament to Odin that you can have this kind of parser in so little lines of code.

gingerBill:
I won't merge this until after tomorrow's release.

core/encoding/xml/debug_print.odin
```diff
@@ -0,0 +1,73 @@
...
+
+	List of contributors:
+		Jeroen van Rijn: Initial implementation.
+*/
+import "core:fmt"
+
+/*
+	Just for debug purposes.
+*/
+print :: proc(doc: ^Document) {
```
gingerBill, line 18: I'd probably make this take an `io.Writer` just so that it is more "agnostic"

core/encoding/xml/debug_print.odin
```diff
@@ -0,0 +1,73 @@
...
+		println(" --- ")		
+	}
+}
+
+print_element :: proc(indent: int, element: ^Element) {
+	if element == nil { return }
+	using fmt
+
+	tab :: proc(indent: int) {
+		tabs := "\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t"
```
gingerBill, line 47: Is this enough tabs? /s

core/encoding/xml/example/xml_example.odin
```diff
@@ -0,0 +1,55 @@
...
+
+import "core:encoding/xml"
+import "core:mem"
+import "core:fmt"
+
+Error_Handler :: proc(pos: xml.Pos, fmt: string, args: ..any) {
+
+}
+
+FILENAME :: "../../../../tests/core/assets/xml/nl_NL-xliff-1.0.xliff"
```
gingerBill, line 11: Wouldn't `"core:tests/tests/core/assets/xml/nl_NL-xliff-1.0.xliff"` work fine here?

core/encoding/xml/xml_reader.odin
```diff
@@ -0,0 +1,651 @@
...
+	val: string,
+}
+
+Attributes :: [dynamic]Attr
+
+Options :: struct {
+	flags:            Option_Flags,
+	expected_doctype: string,
+}
+Option_Flags :: bit_set[Option_Flag]
```
gingerBill, line 125: I'd probably place this right next to `Option_Flag`, and I would also specify the underlying integer size too

core/encoding/xml/example/xml_example.odin
Kelimion: I don't know. Does `os.open` resolve those?

gingerBill: `#load` does

Kelimion: I just tried, but `tests` isn't a collection like `core` and `vendor` are, and `tests` doesn't live under core.

gingerBill: You are correct, I am wrong.

core/encoding/xml/debug_print.odin
Kelimion: Done.

core/encoding/xml/xml_reader.odin
Kelimion: Done.

gingerBill:
Is this read enough to merge?

Kelimion:
> Is this read enough to merge?

It works, but the optimization refactor is going to change things enough that I'd like to hold off. I'll turn it into a draft PR for now.

gingerBill:
Should we begin merging this even though it is not yet ready? (after the conflicts have been resolve, obviously)

PR #1351, opened 2021-12-06
Linux: Fix missing file error reported by #load

corruptmemory:
Under Linux (and other *nix) `realpath` will return NULL when resolving
a given path if the file or path does not exist.  When this happens the
error message returned by #load during compile time will display an
error with a missing file name or path.

Given the following program:

```odin
package main

import "core:fmt"

main :: proc() {
  stuffs := #load("test.file")
  fmt.printf("stuffs %v\n", stuffs)
}
```

If the file `test.file` is missing or the path is invalid you would get
the following error:

```
Failed to `#load` file: ; invalid file or cannot be found
```

However, the correct behavior should be:

```
Failed to `#load` file: test.file; invalid file or cannot be found
```

This PR corrects this

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -782,15 +782,17 @@ String path_to_fullpath(gbAllocator a, String s) {
 
 
 String get_fullpath_relative(gbAllocator a, String base_dir, String path) {
-	u8 *str = gb_alloc_array(heap_allocator(), u8, base_dir.len+1+path.len+1);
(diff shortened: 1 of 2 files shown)


corruptmemory:
I see that there is a MacOS failure.  I'm presuming this is related to my change.  However, I do not have access to a Mac to see what I can do to fix it.

corruptmemory:
The failed test seems to have succeeded on my fork: https://github.com/corruptmemory/Odin/runs/4425775225?check_suite_focus=true

PR #1354, opened 2021-12-07
[core:container/bit_array] Create new package.

Kelimion:
```odin
package dynamic_bit_array

/*
The Bit Array can be used in several ways:

-- By default you don't need to instantiate a Bit Array. It'll grow as-needed.

package test

import "core:fmt"
import ba "core:container/bit_array"

main :: proc() {
	bits: ba.Bit_Array
	defer ba.destroy(&bits)

	// returns `true`
	fmt.println(ba.set(&bits, 42))

	// returns `false`, `false`, because this Bit Array wasn't created to allow negative indices.
	was_set, was_retrieved := ba.get(&bits, -1)
	fmt.println(was_set, was_retrieved) 
}


-- A Bit Array can optionally allow for non-zero indices, if the mininum value was given during creation:

package test

import "core:fmt"
import ba "core:container/bit_array"

main :: proc() {
	Foo :: enum int {
		Negative_Test = -42,
		Bar           = 420,
		Leaves        = 69105,
	}

	bits := ba.create(int(max(Foo)), int(min(Foo)))
	defer ba.destroy(&bits)

	fmt.printf("Set(Bar):           %v\n", ba.set(&bits, Foo.Bar))
	fmt.printf("Get(Bar):           %v\n", ba.get(&bits, Foo.Bar))
	fmt.printf("Set(Negative_Test): %v\n", ba.set(&bits, Foo.Negative_Test))
	fmt.printf("Get(Leaves):        %v\n", ba.get(&bits, Foo.Leaves))
	fmt.printf("Get(Negative_Test): %v\n", ba.get(&bits, Foo.Negative_Test))
	fmt.printf("Freed.\n")
}
*/
```

### DIFF
--- a/core/container/bit_array/bit_array.odin
+++ b/core/container/bit_array/bit_array.odin
@@ -1,0 +1,124 @@
+package dynamic_bit_array
+
+import "core:intrinsics"
+
+/*
+	Note that these constants are dependent on the backing being a u64.
+*/
+@(private="file")
+INDEX_SHIFT :: 6
+
+@(private="file")
+INDEX_MASK  :: 63
+
+Bit_Array :: struct {
+	bits: [dynamic]u64,
+	bias: int,
+}
+
+/*
+	In:
+		- ba:    ^Bit_Array - a pointer to the Bit Array
+		- index: The bit index. Can be an enum member.
+
+	Out:
+		- res:   The bit you're interested in.
+		- ok:    Whether the index was valid. Returns `false` if the index is smaller than the bias.
+
+	The `ok` return value may be ignored.
+*/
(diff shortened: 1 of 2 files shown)


gingerBill:
Should it be called bit vector or array?

Kelimion:
Renamed.

gingerBill:
What is the difference between `Bit_Vector` and `Bit_Array`?

Kelimion:
> What is the difference between `Bit_Vector` and `Bit_Array`?
The name. Bit Array is more in line with other types in Odin.

gingerBill:
Then I'd remove `Bit_Vector` entirely

Kelimion:
> Then I'd remove `Bit_Vector` entirely

They weren't supposed to have been left in. Removed.

PR #1361, opened 2021-12-10
fix bug with fields_proc discarding last byte

Naboris:
```odin
package main

import "core:bytes"
import "core:fmt"
import "core:strings"
import "core:unicode"

main :: proc() {
	str := "Odin Programming Language"
	s := transmute([]u8)str

	fmt.println(strings.fields_proc(str, unicode.is_space))
	// [Odin, Programming, Languag]

	fmt.println(bytes.fields_proc(s, unicode.is_space))
	// [[79, 100, 105, 110], [80, 114, 111, 103, 114, 97, 109, 109, 105, 110, 103], [76, 97, 110, 103, 117, 97, 103]]
}
```

### DIFF
--- a/core/bytes/bytes.odin
+++ b/core/bytes/bytes.odin
@@ -1143,7 +1143,7 @@ fields_proc :: proc(s: []byte, f: proc(rune) -> bool, allocator := context.alloc
 	}
 
 	if start >= 0 {
-		append(&subslices, s[start : end])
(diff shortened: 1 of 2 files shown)


gingerBill:
This is not a fix, this is a bodge.

gingerBill:
I've fixed it now in 85f8c8df

PR #1374, opened 2021-12-16
fix math.prod

indiscible:
accumulator was not initialized to one


Kelimion:
Good catch. The result would always be zero otherwise. Will merge as soon as the CI clears.

gingerBill (approved):

indiscible:
was doing the latest advent of code...

PR #1375, opened 2021-12-17
add zeroing to new region from realloc

jasonKercher:
On Linux, `calloc` (`heap_resize` in os_linux.odin) is used for the default allocator's `aligned_alloc` procedure which has the benefit of providing zero'd memory.  However, `aligned_resize`, utilizes `realloc` which does not provide zero'd memory.  As far as I know, there really isn't a built in way to do this.  This change manually takes care of it for non-windows environments.

### DIFF
--- a/core/os/os.odin
+++ b/core/os/os.odin
@@ -206,11 +206,20 @@ heap_allocator_proc :: proc(allocator_data: rawptr, mode: mem.Allocator_Mode,
 		}
 	}
 
-	aligned_resize :: proc(p: rawptr, old_size: int, new_size: int, new_alignment: int) -> ([]byte, mem.Allocator_Error) {
+	aligned_resize :: proc(p: rawptr, old_size: int, new_size: int, new_alignment: int) -> (new_memory: []byte, err: mem.Allocator_Error) {
 		if p == nil {
 			return nil, nil
 		}
-		return aligned_alloc(new_size, new_alignment, p)
+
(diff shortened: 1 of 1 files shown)


core/os/os.odin
```diff
@@ -206,11 +206,20 @@ heap_allocator_proc :: proc(allocator_data: rawptr, mode: mem.Allocator_Mode,
...
 	}
 
-	aligned_resize :: proc(p: rawptr, old_size: int, new_size: int, new_alignment: int) -> ([]byte, mem.Allocator_Error) {
+	aligned_resize :: proc(p: rawptr, old_size: int, new_size: int, new_alignment: int) -> (new_memory: []byte, err: mem.Allocator_Error) {
 		if p == nil {
 			return nil, nil
 		}
-		return aligned_alloc(new_size, new_alignment, p)
+
+		new_memory = aligned_alloc(new_size, new_alignment, p) or_return
```
gingerBill, line 214: Wouldn't it be better to correct `heap_resize`?

Kelimion: That's actually a good point.

Kelimion: > Wouldn't it be better to correct `heap_resize`?
Actually, it can't be fixed there because `heap_resize` has no clue what the old size is, so it doesn't know what to zero.

This patch does appear to be the best compromise.

PR #1379, opened 2021-12-18
___$startup_runtime for MacOS

weshardee:
Unlike Linux, MacOS requires 3 underscores for '___$startup_runtime'


weshardee:
This was fixed a while back but it appears the issue was re-introduced.

PR #1383, opened 2021-12-21
Fix slice into_dynamic zeroing len of input slice

DevinR528:
I'm assuming this is the intended behavior? When converting a slice to a dynamic array the length was zeroed out returning something with an equal capacity but no elements.

### DIFF
--- a/core/slice/slice.odin
+++ b/core/slice/slice.odin
@@ -23,7 +23,7 @@ swap_between :: proc(a, b: $T/[]$E) {
 	n := builtin.min(len(a), len(b))
 	if n >= 0 {
 		ptr_swap_overlapping(&a[0], &b[0], size_of(E)*n)
-	}	
+	}
 }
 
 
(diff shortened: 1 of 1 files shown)


gingerBill:
That is not a bug since it's creating a buffer backed dynamic array which cannot resize.

A better name or better docs is required.

DevinR528:
Oh duh, it can't move it from the stack without copying the elements hmm. Would this work for docs and name

```odin
// Converts slice into the backing buffer for the returned dynamic array without cloning or allocating.
// Since this is backed by a non-dynamic slice it cannot be resized. 
// The returned array has a length of zero but a capacity equal to the input slice's length.
into_empty_dynamic :: proc(a: $T/[]$E) -> [dynamic]E {...}
```

gingerBill:
In `package mem` it's called `mem.buffer_from_slice`
https://github.com/odin-lang/Odin/blob/master/core/mem/mem.odin#L172

But `slice.buffer_from_slice` is a little verbose.

Tetralux:
`into_dynamic_buffer` perhaps?

PR #1386, opened 2021-12-23
Removes unneeded lookups / Adds sret to call site which fixes the mac…

Platin21:
Fixes the arm64 Abi bug that happened on macOS mainly because someone didn't want to add the right attribute to the call site which makes problems when you call through a pointer. And also removes the needless lookups that we had for the ABI types on arm side.


CarwynNelson:
Thanks so much for fixing this. I just thought to try Odin on my new M1 Pro that I got from work and encountered this problem, and I was just about to go and watch the issue that had been reported expecting it to take a while to be fixed.

PR #1389, opened 2021-12-23
Add slice/scanner proc

ap29600:
Similar to `slice/reduce`, produces a new slice containing the partial results from the reduction:

```odin
s := []int{1, 2, 3, 4, 5}
slice.scanner(s, 0, proc(n,m:int)->int{return n+m}) // []int{1, 3, 6, 10, 15}
```

### DIFF
--- a/core/slice/slice.odin
+++ b/core/slice/slice.odin
@@ -304,6 +304,27 @@ filter :: proc(s: $S/[]$U, f: proc(U) -> bool, allocator := context.allocator) -
 	return r[:]
 }
 
+scanner :: proc (s: $S/[]$U, initializer: $V, f: proc(V, U)->V, allocator := context.allocator) -> []V {
+  if len(s) == 0 { return {} }
+  p := as_ptr(s)
+
+  res := make([]V, len(s), allocator)
+
+  q := as_ptr(res)
+  l := len(res)
+
+  r := initializer 
+
+  for l > 0 {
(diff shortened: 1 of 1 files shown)


core/slice/slice.odin
```diff
@@ -304,6 +304,27 @@ filter :: proc(s: $S/[]$U, f: proc(U) -> bool, allocator := context.allocator) -
...
+
+  for l > 0 {
+    r = f(r, p^)
+    q^ = r
+    p = intrinsics.ptr_offset(p, 1)
+    q = intrinsics.ptr_offset(q, 1)
+    l -= 1
+  }
+
+  return res
```
gingerBill, line 326: Why are you using pointer arithmetic here when normal slices can be used?

ap29600: I measured performance and this was about twice as fast at O3 on a cumulative sum. I can change this Bach to the slice syntax if you'd rather have the clean version. (^_^)

gingerBill: No need to change. I was just querying. And with the latest commit on master, you can do this:
```
p = p[1:]
q = q[1:]
```

ap29600: I see, and this would produce similar code because it's just incrementing the pointer behind the scenes, correct?

gingerBill: Yes.

PR #1395, opened 2021-12-25
factor out alloca generation into a helper

hdooley:


src/llvm_backend_general.cpp
```diff
@@ -216,6 +216,14 @@ LLVMValueRef llvm_one(lbModule *m) {
 	return LLVMConstInt(lb_type(m, t_i32), 1, false);
 }
 
+LLVMValueRef llvm_alloca(lbProcedure *p, LLVMTypeRef llvm_type, isize alignment, char const* name) {
+	LLVMValueRef val = LLVMBuildAlloca(p->builder, llvm_type, name);
```
gingerBill, line 220: You are missing the `LLVMPositionBuilderAtEnd(p->builder, p->decl_block->block);` here

PR #1400, opened 2021-12-30
Add ARM64 support on Windows

Sharundaar:
This PR adds arm64 as a possible target for the compiler on windows. To do this it adds the LLVM-C lib and dll built for ARM64 and modify the build script to shuffle those around instead of the x86/x64 .dll depending on the vc target architecture env var (setup by vcvarsall normally)

The main changes were to remove usage of x64/x86 intrinsics in favor of the macos ARM path. I also had to remove the libtommath/mp_set_double.c compilation since it was generating a blocking warning, since the function didn't seem in use in the project I think it's fine ?

I couldn't test it against the examples files since it uses some win32 calls that have the stdcall calling convention which are unavailable on arm. But I did manage to get the SDL and STB running to output png files and do some basic rendering/window management which were good enough for my use case.


.gitignore
```diff
@@ -276,3 +276,4 @@ shared/
 *.ll
 
 *.sublime-workspace
+/LLVM-C.dll
```
gingerBill, line 279: Why has this been added the .gitignore?!

gingerBill:
I'm extremely surprised that it was this simple to port to ARM64 on windows.

Sharundaar:
Me too, I was actually happy to see that LLVM already supports windows ARM pretty well... There's probably a bunch of edge case I haven't tested as I'm not doing too much low level stuff but it seems to be a good starting point.

gingerBill:
Which versions of the DLLs are you using? 12.0.1 still?

Sharundaar:
Oh that's a good point... I tested a bunch of stuff an it might be the 13.0 actually, lemme fix that

Sharundaar:
Eh boi that's annoying, LLVM has a Windows on Arm version for 12.0.0 and 13.0.0 but not for 12.0.1, so I guess to get those .lib I'd have to build that part of the project on my machine which I'm not sure is powerful enough to get it in a reasonable amount of time. I'd have to check a bit more

gingerBill:
12.0.0 is bugged and cannot work 
We are using 12.0.1 on Windows for the time beginning.

I might have even built the DLL myself.

Sharundaar:
Sounds good 👍 I have what I need here to build the lib, it's just gonna take a while from my experience with LLVM. I'll update the PR with the right lib tomorrow (or when it's done compiling... which might be next year at this point x) )

Platin21:
For arm to correctly work we right now use llvm13 on MacOS the other one generates invalid il code. It works seemingly flawlessly.

gingerBill:
We'll probably not support Windows ARM64, at least not officially anytime soon.

PR #1424, opened 2022-01-12
os_linux additions + libc to syscalls

jasonKercher:
I added `rename`, `remove`, `make_directory`, `remove_directory`, `is_dir`, `is_file` to match `core/os/file_windows.odin`, so now `odinfmt` can will compile on Linux!

I moved a handful of libc calls to straight syscalls.  So far, I was only able to test on amd64, but it should be able to handle arm64, arm and i386 as well. If someone wants to test that on a different arch, here is a quick test program:

```odin
package main

import "core:testing"
import "core:os"

@test
test_os_linux :: proc(t: ^testing.T) {
        err: os.Errno
        fd: os.Handle
        n: int

        mybytes : [4]u8 = "fgkc"

        err = os.make_directory("temp")
        testing.expect_value(t, err, 0)
        testing.expect(t, os.is_dir("temp"))
        testing.expect(t, !os.is_file("temp"))

        /* try to open bad path */
        fd, err = os.open("dir/that/doesnt/exist.txt", os.O_WRONLY, 0o666)
        testing.expect_value(t, fd, -1)
        testing.expect(t, err != 0)

        /* open new file for writing */
        fd, err = os.open("temp/shnt.txt", os.O_CREATE | os.O_TRUNC | os.O_WRONLY, 0o666)
        testing.expect(t, fd != -1)
        testing.expect_value(t, err, 0)

        /* write some shit */
        n, err = os.write(fd, mybytes[:])
        testing.expect_value(t, n, 4)
        testing.expect_value(t, err, 0)

        /* close new file */
        err = os.close(fd)
        testing.expect_value(t, err, 0)

        /* directory not empty */
        err = os.remove_directory("temp")
        testing.expect(t, err != 0)

        /* make directory empty */
        err = os.rename("temp/shnt.txt", "shnt.txt")
        testing.expect_value(t, err, 0)
        testing.expect(t, !os.is_file("temp/shnt.txt"))

        /* NOW remove it... */
        err = os.remove_directory("temp")
        testing.expect_value(t, err, 0)

        /* check file now... */
        testing.expect(t, os.is_file("shnt.txt"))

        /* remove the file */
        err = os.remove("shnt.txt")
        testing.expect_value(t, err, 0)

        /* check for file one more time */
        testing.expect(t, !os.is_file("shnt.txt"))

}
```

### DIFF
--- a/core/os/os_linux.odin
+++ b/core/os/os_linux.odin
@@ -666,9 +666,9 @@ _readlink :: proc(path: string) -> (string, Errno) {
 	buf := make([]byte, bufsz)
 	for {
 		rc := _unix_readlink(path_cstr, &(buf[0]), bufsz)
-		if rc == -1 {
+		if rc < 0 {
 			delete(buf)
-			return "", Errno(get_last_error())
+			return "", _get_errno(rc)
 		} else if rc == int(bufsz) {
 			// NOTE(laleksic, 2021-01-21): Any cleaner way to resize the slice?
 			bufsz *= 2
--- a/core/os/os_linux.odin
+++ b/core/os/os_linux.odin
@@ -266,33 +266,142 @@ X_OK :: 1 // Test for execute permission
 W_OK :: 2 // Test for write permission
 R_OK :: 4 // Test for read permission
 
+AT_FDCWD            :: -100
+AT_REMOVEDIR        :: uintptr(0x200)
+AT_SYMLINK_NOFOLLOW :: uintptr(0x100)
+
+_unix_open :: proc(path: cstring, flags: int, mode: int = 0o000) -> Handle {
+	when ODIN_ARCH != "arm64" {
+		res := int(intrinsics.syscall(unix.SYS_open, uintptr(rawptr(path)), uintptr(flags), uintptr(mode)))
+	} else { // NOTE: arm64 does not have open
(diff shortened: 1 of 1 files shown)


Platin21:
Hmm would have put the syscalls into sys/linux and called them from os. But yeah

jasonKercher:
> Hmm would have put the syscalls into sys/linux and called them from os. But yeah

If that is preferred, I can change it.  I didn't want to break any existing code that is calling `_unix_*` from `os` which are links into libc.  It is a bit ambiguous currently and there is a bit of goofiness with `errno` in libc vs calculating it from the syscall return.

Platin21:
Ohh i meant because we want to move to a new os library anyways and normal os will be deprecated so i though it would be a good idea to move it over. 

Yeah i wouldn’t touch the unix stuff just put it in sys/linux makes more sense. Especially as the unix stuff isn’t necessarily correct anyways. As different platforms seem to have different error returns and values even sometimes.

So overall maybe better to have it working in os2 and leaving os just for compatibility reasons. But just my thinking..

jasonKercher:
Ah okay.  I see that.  os2 is pretty incomplete, but I can see the general idea is to merge the different platforms under a one umbrella.  I will move the syscall code to sys, and move new os stuff to os2.  I just want to keep the *new* functions so `odinfmt` can compile for Linux.  Currently, `odinfmt` uses `rename`, `remove`, `is_dir`, `is_file` and they only exist in `*_windows.odin` files.

Platin21:
Ahh makes sense

PR #1441, opened 2022-01-22
Makes it possible to use llvm 12 13 and 11 for building on macOS

Platin21:
I thin this should make it possible to build with all 3 versions on macOS on the x86 side. For the arm side I don't see that as useful as only llvm13 generates the correct debug info files and some other stuff is also broken in older llvm versions.


gingerBill:
This might be a good change to. Just trying different versions until it works. You will need to check if the versions are 11.1.0, 12.0.1, and 13.

Platin21:
Sure let me see the problem is more then testing if it works verifiably :

Platin21:
Fixed it so that we can now say which llvm versions we want I suspect this was how it makes more sense?

Platin21:
Removed the maybe not supported llvm versions for x86 for arm I still would prefer to keep the requirement up to be at least llvm13 else we get some bug reports for llvm11 not being able to create debug info correctly for arm64 macOS which is a llvm bug and will never be fixed especially not in older versions.

PR #1469, opened 2022-02-04
add `iterator` to `core:container/bit_array`

ap29600:


gingerBill:
Wouldn't an interface like this be simpler?
```odin
it: uint
for ok, index in bit_array.iterator(the_bit_array, &it) {
}
// or
it := bit_array.iterator_make(the_bit_array)
for ok, index in bit_array.iterator(&it) {
}
```
be a bit simpler?

ap29600:
my iterator stores both word and sub-word offset in order to skip the whole word, admittedly it could just be the global offset but that would require some modulo arithmetic and it's simpler to do it this way.

I'll add the second option since that can wrap the interface I already have.

gingerBill:
Why are you skipping elements? What if you want to iterate over those just to know if they are empty or not?

ap29600:
Well, the idea was that this would be an iterator over the *set* bits of the array. This makes more sense to me than to just iterate over all the bits because that can be achieved pretty much optimally with 
```odin
for i in min_bit..=max_bit {
    val, ok := bits.get(&array, i)
   ...
}
```
which is pretty much the same amount of code as constructing an iterator and using it, 
and should be pretty much optimal if bits.get is inlined.

On the other hand, iterating only the set bits is a little harder to do efficiently if, say, the words are somewhat sparse (say 10% filled on average) so I thought an iterator that does the skipping for you would be nice to have.

Maybe both should be provided?

Kelimion:
```cpp
Iterator_Mode :: enum u8 {
   Set,
   Unset,
   All,
}

it := bit_array.iterator_make(the_bit_array, .Set) // return indices of set bits only.
```

ap29600:
I'll think about it, ideally I would differentiate the procs and not the struct.

ap29600:
I'm not convinced it makes sense to implement an iterator over all values with the way the interface is right now.
The `Bit_Array` struct currently does not remember what the max value of the enum is, meaning that if a bit_array is
constructed with values of some enum type `T` that has 10 values, an array of one `u64` is used to store the value.
when the values are iterated over, the iterator will also provide values `10..=63` set to `false`, which make no sense in the context of the array containing values of type `T`.
to make ths make sense it would be necessary to either store the maximum represented value as part of the type, or provide a range in the iterator construction.

Kelimion:
That enum's 10 values don't have to be 0..9.
But you're right that it makes sense to store min/max values on the `Bit_Array` if we're going to be iterating over it. It wouldn't make setting a bit that much more expensive.

Two extra branches.
```cpp
ba.min_index = min(ba.min_index, biased_index)
ba.max_index = max(ba.max_index, biased_index)
```

Of course you're also paying that cost for each bit set if you don't use the iterator, so perhaps `make_iterator` could keep track of this information itself and do a quick scan of the bit array given to it (any leg that's 0 can be skipped) to compute the min/max bits set at the time of iterator construction.

I wonder what @gingerBill thinks is the way to go.

gingerBill:
Would it be this?
```odin
for set, idx in iterate_by_all(&it) {}
for idx in iterate_by_set(&it) {}
for idx in iterate_by_unset(&it) {}
```

Kelimion:
> Would it be this?
> 
> ```odin
> for set, idx in iterate_by_all(&it) {}
> for idx in iterate_by_set(&it) {}
> for idx in iterate_by_unset(&it) {}
> ```

I agree that should be the iterator API.

On a related note, what do you think of keeping track of minimum and maximum bounds so it's clear what "all" is? If done in `bit_array.set`, it's two extra branches you're paying even if you're not using the iterator(s).

ap29600:
> Would it be this?
> 
> ```odin
> for set, idx in iterate_by_all(&it) {}
> for idx in iterate_by_set(&it) {}
> for idx in iterate_by_unset(&it) {}
> ```

this seems good, I'll implement it.

PR #1471, opened 2022-02-05
Feature/fix odin fmt

Platin21:
This PR adds missing os calls for odin fmt on macOS

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
 
+# For macOS
+.DS_Store
+
 # Build results
 [Dd]ebug/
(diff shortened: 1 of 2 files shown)


Kelimion:
Oops:
```
Run ./odin check examples/demo -vet
/Users/runner/work/Odin/Odin/core/os/os_darwin.odin(4[7](https://github.com/odin-lang/Odin/runs/5079671462?check_suite_focus=true#step:7:7)5:22) Cannot assign value 'path' of type 'string' to 'cstring' in argument
/Users/runner/work/Odin/Odin/core/os/os_darwin.odin(474:2) 'path_cstr' declared but not used
```

Platin21:
@Kelimion yep typo fix incoming 👍

PR #1476, opened 2022-02-06
Replace `any` with `union` for subtyping in `core:odin/ast`

gingerBill:
Replace `any` with `union` for subtyping in `core:odin/ast` to provide much better type safety checks

Ping: @DanielGavin


DanielGavin:
@gingerBill Good to merge now.

PR #1495, opened 2022-02-12
Updated vendor/sdl2/ttf LIB and DLL.

FancyKillerPanda:
Updated the LIB and DLL for SDL_ttf to match the version that has been ported in code (2.0.18).

Also incremented the version number in SDL2_ttf.odin as features from newer versions already exist (such as `RenderUTF8_Solid_Wrapped`), but it seems the version number had not been updated when that was implemented.

LIB and DLL were downloaded from [SDL2_ttf-devel-2.0.18-VC.zip](https://github.com/libsdl-org/SDL_ttf/releases/download/release-2.0.18/SDL2_ttf-devel-2.0.18-VC.zip).


FancyKillerPanda:
I was looking at the release notes from https://discourse.libsdl.org/t/sdl-ttf-2-0-18-released/34422

SDL2_ttf.odin has the functions from there, but the PATCHLEVEL constant is still set to 15.

PR #1519, opened 2022-02-19
Maintain a running count printed, so fprintf returns correct sizes for non-files

colrdavidson:
(and also remove the need for file_size/double-fstat calls per printf)


colrdavidson:
It should probably handle the error cases from io.write_string, but we weren't handling them before, so I wasn't super concerned. io returns 0 sizes on failure, so things shouldn't get too screwy.

The more interesting debate is "what happens when write_byte fails?" Arguably for consistency's sake, write_byte should return a 1/0 too, just to make running tallies less weird

gingerBill:
There's actually an extra parameter to most of io write calls which does this increment, use that rather than the return value itself.

PR #1526, opened 2022-02-20
Freestanding target for amd64

gingerBill:


bumbread:
Upon compilation with this flag, the linker generates references to `os.write`, which is undesireable

image

Also the compiled file still calls `runtime.bounds_check_error` on every assignment to a slice member.

gingerBill:
On a freestanding target, `runtime._os_write` does nothing.
```
_os_write :: proc "contextless" (data: []byte) -> (int, _OS_Errno) {
	return 0, -1
}
```

gingerBill:
Also, whereabouts is this `runtime.os_write` being called? It shouldn't be called if you have `-foreign-error-procedures` set

bumbread:
I don't have this flag set, why is this a separate option?

gingerBill:
Maybe it should be set if you are targeting a freestanding target.

bumbread:
Based on our yesterday's discord discussion I added a few comments to the [freestanding target proposal](https://github.com/odin-lang/Odin/discussions/1525#discussioncomment-2214842).

PR #1528, opened 2022-02-21
Add doc comments to strings.split() and strings.split_n()

Tetralux:


gingerBill:
I don't recommend the leading asterisks. It's inconsistent with the other comments are a pain to copy and paste.

PR #1539, opened 2022-02-23
Some documentation improvements

FancyKillerPanda:
I've been working on some of the formatting and content of the documentation. It's far from done but it's a start.

I also changed `//+ignore` to `//+private` as ignore doesn't do anything, which caused some of the procedures from example/internal parts of packages to appear in the docs.

### DIFF
--- a/core/container/bit_array/bit_array.odin
+++ b/core/container/bit_array/bit_array.odin
@@ -3,7 +3,7 @@ package dynamic_bit_array
 import "core:intrinsics"
 
 /*
-	Note that these constants are dependent on the backing being a u64.
+Note that these constants are dependent on the backing being a u64.
 */
 @(private="file")
 INDEX_SHIFT :: 6
@@ -27,25 +27,24 @@ Bit_Array_Iterator :: struct {
 }
 
 /*
-	In:
-		- ba:   ^Bit_Array - the array to iterate over
+In: <br>
+- `ba`: the array to iterate over <br>
 
-	Out:
-		- it:   ^Bit_Array_Iterator - the iterator that holds iteration state
+Out: <br>
+- `it`: the iterator that holds iteration state <br>
 */
 make_iterator :: proc (ba: ^Bit_Array) -> (it: Bit_Array_Iterator) {
 	return Bit_Array_Iterator { array = ba }
 }
 
 /*
-	In:
-		- it:    ^Bit_Array_Iterator - the iterator struct that holds the state.
+In: <br>
+- `it`: the iterator struct that holds the state. <br>
 
-	Out:
-		- set:    bool - the state of the bit at `index`
-		- index:  int - the next bit of the Bit_Array referenced by `it`.
-		- ok:	  bool - `true` if the iterator returned a valid index,
-			  `false` if there were no more bits
+Out: <br>
+- `set`: the state of the bit at `index` <br>
+- `index`: the next bit of the `Bit_Array` referenced by `it`. <br>
+- `ok`:	`true` if the iterator returned a valid index, `false` if there were no more bits <br>
 */
 iterate_by_all :: proc (it: ^Bit_Array_Iterator) -> (set: bool, index: int, ok: bool) {
 	index = it.word_idx * NUM_BITS + int(it.bit_idx) + it.array.bias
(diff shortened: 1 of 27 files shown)


Kelimion:
You may want to revert [Changed all occurences of //+ignore to //+private.](https://github.com/odin-lang/Odin/pull/1539/commits/d71da1ea05984fcd78624049178b35db0a2e8827)

They're not the same thing and introduced breaking changes as a result.

core/container/bit_array/bit_array.odin
```diff
@@ -157,14 +154,14 @@ get :: proc(ba: ^Bit_Array, #any_int index: uint, allocator := context.allocator
 }
 
 /*
-	In:
-		- ba:    ^Bit_Array - a pointer to the Bit Array
-		- index: The bit index. Can be an enum member.
+In: <br>
```
Kelimion, line 157: Please don't add HTML tags to documentation. Gross.
If they're needed on the output side, that ought to be handled by the documentation generator.

core/container/bit_array/bit_array.odin
```diff
@@ -27,25 +27,24 @@ Bit_Array_Iterator :: struct {
 }
 
 /*
-	In:
-		- ba:   ^Bit_Array - the array to iterate over
+In: <br>
```
gingerBill, line 30: Please don't use `<br>` within the documentation. The docs need to be readable from the test and from `odin doc` which is a terminal viewer.

We should use a better markup to make it clear that it is a list below

core/encoding/json/doc.odin
```diff
@@ -0,0 +1,57 @@
+/*
+The `encoding/json` package can be used to parse and validate JSON files. A simple example is shown below.
+
+`test.json`
+<pre><code class="hljs" data-lang="json">
```
gingerBill, line 5: No HTML tags. 

This is better solved with the documentation generation tool rather than just hacking it in.

core/path/filepath/path.odin
```diff
@@ -18,6 +20,16 @@ is_slash :: proc(c: byte) -> bool {
 	return c == '\\' || c == '/'
 }
 
+// `split` splits the path immediately following the last separator,
+// separating it into a directory and file name component.
+//
+// The returned values have the property that `path = dir+file`.
+//
+// Examples: <br>
```
gingerBill, line 28: No HTML tags

core/path/filepath/path_unix.odin
```diff
@@ -13,14 +13,21 @@ SEPARATOR :: '/'
...
 LIST_SEPARATOR :: ':'
 
+// `is_reserved_name` will return `true` if `path` is reserved on the current operating system.
 is_reserved_name :: proc(path: string) -> bool {
 	return false
 }
 
+// `is_abs` checks whether the path is absolute.
+//
+// Examples: <br>
```
gingerBill, line 23: No HTML tags.

core/path/filepath/path_windows.odin
```diff
@@ -31,7 +32,11 @@ is_UNC :: proc(path: string) -> bool {
 	return volume_name_len(path) > 2
 }
 
-
+// `is_abs` checks whether the path is absolute.
+//
+// Examples: <br>
```
gingerBill, line 37: No HTML tags.

core/path/filepath/walk.odin
```diff
@@ -3,25 +3,29 @@ package filepath
...
-// In the case of an error, the info argument will be 0
-// If an error is returned, processing stops
-// The sole exception is if 'skip_dir' is returned as true:
-// 	when 'skip_dir' is invoked on a directory. 'walk' skips directory contents
-// 	when 'skip_dir' is invoked on a non-directory. 'walk' skips the remaining files in the containing directory
+// and the procedure can decide how to handle that error (and `walk` will not descend into that directory).
+// In the case of an error, the `info` argument will be `0`.
+// If an error is returned, processing stops.
+//
+// The sole exception is if `skip_dir` is returned as `true`: <br>
```
gingerBill, line 16: No HTML tags. Need better list semantics for the generator.

gingerBill (changes requested):
No HTML tags, the generator's markup language needs improving instead.

FancyKillerPanda:
I'm closing this as discussed with Jeroen. I'll try working on improving the doc generator so that we can use better formatting here without HTML tags. I'll keep this code around though so I can later put it back (with modifications to remove the bad bits :) ).

PR #1543, opened 2022-02-23
Add basic static HTTP server impl

colrdavidson:


core/net/http/server.odin
```diff
@@ -14,4 +14,203 @@
...
-package http
\ No newline at end of file
+package http
+
+import "core:fmt"
+import "core:strings"
+import "core:net"
+import "core:os"
+
+guess_mime :: proc(filename: string) -> (mime: string) {
```
Kelimion, line 24: Or no? Easier to expand in the future.
```cpp
guess_mime :: proc(filename: string) -> (mime: string) {
	suffix := ... grab suffix
	switch suffix {
	}
	return "application/octet-stream"
}
```

PR #1544, opened 2022-02-24
Changed `//+ignore` to `//+build ignore` and emit a warning for unknown tags

FancyKillerPanda:
I changed the files that used `//+ignore` to use `//+build ignore` as they should.

Also removed the ignore tag from `core:math/big/internal.odin` since those procedures can be called from elsewhere, like the demo.

Also made the parser emit a warning when encountering an unknown tag, so that tags like `//+ignore` will emit `Warning: Ignoring unknown tag '+ignore'`

### DIFF
--- a/core/compress/gzip/example.odin
+++ b/core/compress/gzip/example.odin
@@ -1,4 +1,4 @@
-//+ignore
+//+build ignore
 package gzip
 
 /*
--- a/core/compress/zlib/example.odin
+++ b/core/compress/zlib/example.odin
@@ -1,4 +1,4 @@
-//+ignore
+//+build ignore
 package zlib
 
 /*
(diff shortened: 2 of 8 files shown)


Kelimion:
I'll let @gingerBill decide whether he wants the parser to emit a warning on ignoring a tag.

gingerBill:
"as they should."
What does this mean?

I wanted `//+ignore` to be different from `//+build` because `ignore` is not an OS nor Architecture.

FancyKillerPanda:
From what I can tell, `//+ignore` is not a tag that actually exists (the parser never checks for it). `//+build ignore` is what the parser checks for, and so I changed the few `//+ignore` directives used to `//+build ignore` (most in the source were already `//+build ignore`, there were just a few that weren’t).

PR #1547, opened 2022-02-24
[net] Cleanup some errors

Tetralux:
- Add DNS_Error to Network_Error
- Move Resolve_Error in common.odin
- Ensure no error enums have named zero values
- Added documenting comment to Network_Error which explains why

### DIFF
--- a/core/net/common.odin
+++ b/core/net/common.odin
@@ -57,7 +57,7 @@ ODIN_NET_TCP_NODELAY_DEFAULT :: #config(ODIN_NET_TCP_NODELAY_DEFAULT, true)
 Maybe :: runtime.Maybe
 
 General_Error :: enum {
-	Unable_To_Enumerate_Network_Interfaces,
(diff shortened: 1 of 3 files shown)


core/net/common.odin
```diff
@@ -82,10 +90,16 @@ Network_Error :: union {
...
+}
+
+
+Resolve_Error :: enum {
+	Unable_To_Resolve = 1,
 }
 
 DNS_Error :: enum {
-	Invalid_Hostname_Error,
+	Invalid_Hostname_Error = 1,
```
Tetralux, line 102: Ideally, I'd add an `#assert` to `Network_Error` to ensure none of the enums have a named zero value, but I don't know quite how I'd approach that.
Maybe we want `intrinsics.type_union_variants_have_value(Network_Error, nil)` 😄 

I could add such an assert to every enum individually, but that kinda gross honestly...

PR #1549, opened 2022-02-24
delete allocated memory with clone_to_cstring

semarie:
if I correctly understand `clone_to_cstring` code, it is allocating memory, and this memory should be released using `delete()` when not longer needed.

the following PR adds `defer delete(x)` when `clone_to_cstring` is used for foreign code, avoiding leaking memory.

please note that I am not able to properly test the PR, as I don't have any of the OS impacted, and I am fairly new with odin code. so the PR might contains errors.

### DIFF
--- a/core/os/os_darwin.odin
+++ b/core/os/os_darwin.odin
@@ -344,9 +344,8 @@ get_last_error_string :: proc() -> string {
 }
 
 open :: proc(path: string, flags: int = O_RDWR, mode: int = 0) -> (Handle, Errno) {
-	cstr := strings.clone_to_cstring(path)
+	cstr := strings.clone_to_cstring(path, context.temp_allocator)
 	handle := _unix_open(cstr, i32(flags), u16(mode))
-	delete(cstr)
 	if handle == -1 {
 		return INVALID_HANDLE, 1
 	}
--- a/core/os/os_freebsd.odin
+++ b/core/os/os_freebsd.odin
@@ -7,149 +7,149 @@ import "core:runtime"
 import "core:strings"
 import "core:c"
 
-Handle :: distinct i32;
-File_Time :: distinct u64;
-Errno :: distinct i32;
-Syscall :: distinct i32;
+Handle :: distinct i32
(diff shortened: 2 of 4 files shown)


core/dynlib/lib_windows.odin
```diff
@@ -19,6 +19,7 @@ unload_library :: proc(library: Library) -> bool {
 
 symbol_address :: proc(library: Library, symbol: string) -> (ptr: rawptr, found: bool) {
 	c_str := strings.clone_to_cstring(symbol, context.temp_allocator)
```
Kelimion, line 21: You'll notice here that `clone_to_cstring` is called with the explicit `context.temp_allocator` parameter, meaning there is no memory that needs freeing. Doing so would result in a bad free.

Kelimion:
I think the PR should be reworked so that where `clone_to_cstring` is used with foreign code, you add `, context.temp_allocator` where it doesn't already have it, and get rid of the `defer delete` in the places you added them.

Kelimion:
That and the `;` is optional as a line ending. For the core library we decide to treat their presence as an error.

semarie:
thanks for the review. I will rework it.

Kelimion:
This is incidentally also why many of us develop using the `Tracking_Allocator`.
It'll tell you if you leaked something and where it was allocated, and if a pointer was doubly freed.
The way I'm printing those here has the same location format as source code errors, so it works nicely with editors allowing you to jump to them.

```odin
main :: proc() {
	track: mem.Tracking_Allocator

	call_your_real_code()

	fmt.println("Cleaning up.")
	for _, leak in track.allocation_map {
		fmt.printf("%v leaked %v bytes\n", leak.location, leak.size)
	}
	for bf in track.bad_free_array {
		fmt.printf("%v allocation %p was freed badly\n", bf.location, bf.memory)
	}
	fmt.println("Done.")
}
```

e.g. when I accidentally on purpose forget to destroy allocations:
image

Kelimion:
Excellent work. Thank you.

PR #1550, opened 2022-02-24
[core:net] Fix http.Client and resolve()

Tetralux:
- Remove erroneous use of `or_return` in `resolve()`.
- Add some basic useful timeouts on HTTP sockets in `send_request()`.
- Move logic for ensuring `'Connection'` header is sent, into `send_request()` so that `http.get()` and `http.Client` both benefit from it.

### DIFF
--- a/core/net/http/http.odin
+++ b/core/net/http/http.odin
@@ -20,6 +20,7 @@ import "core:net"
 import "core:strings"
 import "core:strconv"
 import "core:fmt" // for panicf
+import "core:time"
 
 /*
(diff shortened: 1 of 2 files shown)


core/net/http/http.odin
```diff
@@ -144,11 +142,13 @@ send_request :: proc(r: Request, allocator := context.allocator) -> (socket: net
 		port = 80
 	}
 
-
 	// TODO(tetra): SSL/TLS.
 	skt, err := net.dial_tcp(host, port)
 	if err != nil do return
 
+	// TODO(tetra): Make this configurable?
```
Kelimion, line 149: Yes.

PR #1553, opened 2022-02-24
Calling convention is only written explicitly in doc if it's not 'odin'.

FancyKillerPanda:


gingerBill:
Should we ignore this the generator rather itself?

FancyKillerPanda:
Do you think it's something the website generator should do? Because then when reading `odin doc` from the terminal wouldn't the `proc "odin"` still show up, even though it's assumed to be the default?

gingerBill:
Both `odin doc` and the website pkg generator shouldn't show `proc "odin"`, but we should store that data into the file format because it is useful.

PR #1557, opened 2022-02-25
initial OpenBSD support

semarie:
__missing code path__
- [x] in [gb_file_copy()](https://github.com/odin-lang/Odin/blob/master/src/gb/gb.h#L6025), `sendfile()` is unsupported on OpenBSD, but it could be implemented in userland
- [ ] `absolute_path_from_handle()` : unsupported on OpenBSD
- [ ] `read_dir()` : use `absolute_path_from_handle()` for getting dir name from fd
        but all callers are opening fd just before calling it...
- [x] in [src/build_settings.cpp](https://github.com/odin-lang/Odin/blob/master/src/build_settings.cpp#L695) `internal_odin_root_dir()` : it need to get the path of the running process - it is an unsupported operation on OpenBSD
        as workaround, using `ODIN_ROOT` env works well
- [x] amount of installed RAM in report - (see [report_ram_info()](https://github.com/odin-lang/Odin/blob/master/src/bug_report.cpp#L214))

__odin build__
- [x] `Makefile` changes
  - no dl library (functions are inside libc)
  - stdc++ isn't the default standard C++ library (it is c++, but no in all archs) - it seems that clang choose the right C++ library even if `-lstdc++` is specified
  - need iconv library (external third-party)

__runtime problems__
- segfault at `__$cleanup_runtime` (it only occurs when compiled with `-o:speed`)
  - crypto_hash test
  - test_hash test
  - botan_hash test

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,9 @@ jobs:
           cd tests/vendor
           make
         timeout-minutes: 10
+      - name: Odin check examples/all for OpenBSD amd64
+        run: ./odin check examples/all -vet -strict-style -target:openbsd_amd64
+        timeout-minutes: 10
   build_macOS:
     runs-on: macos-latest
     steps:
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 GIT_SHA=$(shell git rev-parse --short HEAD)
 DISABLED_WARNINGS=-Wno-switch -Wno-macro-redefined -Wno-unused-value
-LDFLAGS=-pthread -ldl -lm -lstdc++
+LDFLAGS=-pthread -lm -lstdc++
 CFLAGS=-std=c++14 -DGIT_SHA=\"$(GIT_SHA)\"
 CFLAGS:=$(CFLAGS) -DODIN_VERSION_RAW=\"dev-$(shell date +"%Y-%m")\"
 CC=clang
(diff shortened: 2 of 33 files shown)


semarie:
outside the mentioned segfault problem, running (manually) the CI on OpenBSD works fine.

```
$ odin report

        Odin: dev-2022-02:3a469dc1
        OS:   OpenBSD 7.0 GENERIC.MP#350 amd64
        CPU:  AMD Athlon 3000G with Radeon Vega Graphics     
        RAM:  Unknown.
```

semarie:
I think it is good to review now.

some notes:
- `absolute_path_from_handle()` is the bigger problem I have for now. some usage in `core` could be avoided, but not all. so I will keep it unresolved for now. it affects `stat()` functions family in `core:os`
- `__$cleanup_runtime` segfault is still present when compiling with `-o:speed`

semarie:
It should be also possible to add a specific entry in CI for doing (minimal) testing for OpenBSD (from a supported target by github):

`odin check examples/all -vet -strict-style -target:openbsd_amd64`

gingerBill:
If you add that to the CI.yml on your branch, it should work.

gingerBill:
How far off done is this PR?

semarie:
for me, it is in a mergeable state. it provides a usable odin compiler on OpenBSD. it doesn't mean all is done and perfect, but futher development could be done with others PR.

gingerBill:
I will merge this in then and await the new PRs!

PR #1565, opened 2022-02-28
openbsd: defaults to PIE executable

semarie:
OpenBSD uses PIE code by default to allow the system to load the binary at a random location.

don't pass -no-pie to preserve this behaviour, and build objects with -fPIC (LLVMRelocPIC) (-fPIC is required for PIE).


semarie:
thanks

Issue #1580, opened 2022-03-03
core:compress errors on 32 bits architectures


semarie:
## Context

When `core:compress` is used on 32bits architectures (any OS), the following errors shown:

```
$ odin check examples/all -target:windows_386
/home/semarie/repos/odin/odin/core/compress/common.odin(142:1) Compile time assertion: size_of(Context_Memory_Input) == 64
/home/semarie/repos/odin/odin/core/compress/gzip/gzip.odin(108:31) Cannot cast 'max(u32le)' as 'int' from 'u32le'
/home/semarie/repos/odin/odin/core/compress/gzip/gzip.odin(108:2) Invalid declaration type 'int(max(u32le))'
```

There are two differents issues:
- `Context_Memory_Input` struct contains a pointer, so the size of the struct depend of `size_of(builtin.rawptr)`
- `GZIP_MAX_PAYLOAD_SIZE :: int(max(u32le))`, on 32 bits, a `int` is too small to hold `max(u32le) = 4294967295`

### Steps to Reproduce

Using `odin check` with 32bit target is enough to reproduce, even on 64bit arch.

```
$ odin check examples/all -target:windows_386
```
or
```
$ odin check examples/all -target:linux_386
```

semarie:
cc @Kelimion, as it seems you wrote the initial code of `core:compress`. if you have pointers on the proper fix for this issue.

Kelimion:
I'll fix it today. Thanks for bringing it to my attention, @semarie.

Kelimion:
Fixed by #1581.

semarie:
Thanks !

PR #1581, opened 2022-03-03
32 bit fixes

Kelimion:
Fixes 32-bit cleanliness of `core:compress`.

Also, make the tests return error level 1 if a test fails so the CI notices.


Kelimion:
Fixes #1580.

PR #1582, opened 2022-03-03
CI: add linux i386, Darwin arm64 and Windows 386

semarie:
this PR adds more checks in CI, particulary for `linux_386` `darwin_arm64` and `windows_386`.

I kept `linux_arm64`, `freebsd_amd64` and `freebsd_386` outside this PR as there are currently failing.

I added them in differents jobs in order to spread the load.


gingerBill:
The architecture should be renamed to `i386` from `386`

PR #1583, opened 2022-03-03
Semi colons and EOF

semarie:
As found in #1582, the parser doesn't check for semi-colons when EOF is the next token.

The following PR makes the check to be done for EOF too, and remove semi-colons from some files.


semarie:
arg, it will conflicts with #1582. I corrected the parser so it find the semi-colons in `core/compress/common.odin` too.

Issue #1587, opened 2022-03-06
`core:container/bit_array` leaks memory on `clear`ing


ap29600:
## Context

```
        Odin: dev-2022-03:07062324
        OS:   Arch Linux, Linux 5.16.10-arch1-1
```

## Expected Behavior

calling `bit_array.clear` sets all bits in the `Bit_Array` to `false`

## Current Behavior

calling `bit_array.clear` sets the `bits` field of the `Bit_Array` to `nil`, leaking the previous allocated array. 

### Steps to Reproduce

run the following code with valgrind (or check leaks with the tracking allocator)

```odin
package test
import "core:container/bit_array"
main :: proc () {
	b := bit_array.create(100)
	for i in 0..100 {
		bit_array.set(&b, 100)
		bit_array.clear(&b)
	}
	bit_array.destroy(&b)
}
```

PR #1588, opened 2022-03-06
Fix leak in `core:container/bit_array`

ap29600:
calling `clear` on a `bit_array` no longer leaks the previous
allocation, instead it sets all bits to `false` preserving the same
backing dynamic array.


ap29600:
this fixes #1587

PR #1590, opened 2022-03-06
Add missing `intrinsics.unaligned_load`.

Kelimion:
`intrinsics.unaligned_load` existed but wasn't listed in the intrinsics file, going overlooked as a result.


gingerBill:
There is also unaligned_store

PR #1595, opened 2022-03-07
Add support for left/right pad

colrdavidson:


gingerBill:
Why are you even bothering with the Builder when a simple bytes buffer is all that is required?

colrdavidson:
Following the pattern used in left_justify/right_justify. Happy to switch things over, one sec.

gingerBill:
Well, you can use the builder if you just reserve the precise size required.

qgcarver:
many such cases

PR #1598, opened 2022-03-08
Add `core:encoding/varint` with LEB128 encoding, decoding and tests.

Kelimion:
LEB128 variable length integer encoding is used in DWARF, DEX and others.

Also make tests in general less spammy and don't print `[PASS]`, only report failures and progress.


gingerBill:
What if the individual just wants to decode the value but not as an integer? i.e. it could be bigger than `u128`?

Kelimion:
> What if the individual just wants to decode the value but not as an integer? i.e. it could be bigger than `u128`?

I had a chat about this with @colrdavidson and while theoretically it should use a bigint implementation, in practice any such unbounded leb128 indicates a corrupted DWARF file. I commented as such in the package as well.

It would be easy to return a union of (u128, big.Int) where if this path runs out of room, it makes a big.Int with the currently accumulated value and then continues as a big.Int, or even just do things "big" to begin with. But practically speaking is there a use case we know where these values even take up more than 8 bytes?

I'd not be surprised if libDwarf nopes out when it doesn't fit in a u64.

gingerBill (approved):

gitlost:
Note this accidentally added the binary "tests/core/crypto_hash".

PR #1612, opened 2022-03-11
fix thread data races

graphitemaster:
The following PR reworks the Linux thread implementation to avoid data races on construction and join which only worked on systems with full-sequential consistency (e.g x86). Helgrind reported tons of errors which are now corrected for.

The three states that a thread can be in (started, joined, done) are handled with a simple bit_set now, the start gate was switched over to sync2 (no longer need to construct or destruct mutexes and condition variables) and the start gate serves dual-purpose as a mechanism to block the thread until start and also synchronize fields of the thread structure rather than using atomic instructions which were failing. Some of them were producing false-negatives in Helgrind and were fine, but this is just greatly simplified now and keeps it silent which makes debugging thread problems far easier.

The is_done method is still safely implemented with a full-sequential consistent atomic_load to avoid acquiring a lock each time it's called.

Some of the concerns about calling `pthread_join` when the thread has already returned as mentioned in the comments is just not true, but joining a thread multiple times is of course illegal and this handles that legitimate case the same way.

### DIFF
--- a/core/thread/thread_unix.odin
+++ b/core/thread/thread_unix.odin
@@ -35,7 +35,7 @@ _create :: proc(procedure: Thread_Proc, priority := Thread_Priority.Normal) -> ^
 
 		t.id = sync.current_thread_id()
 
-		if .Started not_in t.flags {
+		for (.Started not_in t.flags) {
 			sync.wait(&t.cond, &t.mutex)
 		}
 
--- a/core/thread/thread_unix.odin
+++ b/core/thread/thread_unix.odin
@@ -4,33 +4,22 @@ package thread
 
 import "core:runtime"
 import "core:intrinsics"
-import "core:sync"
+import sync "core:sync/sync2"
 import "core:sys/unix"
+
+Thread_State :: enum u8 {
+	Started,
+	Joined,
+	Done,
+}
 
 // NOTE(tetra): Aligned here because of core/unix/pthread_linux.odin/pthread_t.
 // Also see core/sys/darwin/mach_darwin.odin/semaphore_t.
 Thread_Os_Specific :: struct #align 16 {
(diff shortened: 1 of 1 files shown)


graphitemaster:
The mac build is failing because there is no implementation of a mutex in the sync2 package so it falls back to a spin-lock implementation which is misusing atomic instructions

gingerBill:
@graphitemaster What needs to be done to fix the spin-lock?

graphitemaster:
@gingerBill The `atomic_compare_exchange_weak_acquire` in `lock_slow` is the problem, it appears to be using `acquire` semantics for the success order and `seq_cst` for the failure order, when atomic instructions require that failure order never be greater of a requirement than success order. Since full sequential consistency is far greater of a requirement than acquire semantics it fails. Conceptually, almost none of these intrinsics (for read-modify-write operations) actually make sense since `intrinsics.atomic_cxchgweak*` always presumes `seq_cst` for failure order, it means success order can only ever also be `seq_cst` and nothing else, as none of the other orders are a higher requirement. So these intrinsics don't make any sense nor do anything meaningful (and might actually be miscompiling code) and should be removed:
* `intrinsics.atomic_cxchgweak_acq(object, expected^, desired)`
* `intrinsics.atomic_cxchgweak_rel(object, expected^, desired)`
* `intrinsics.atomic_cxchgweak_acqrel(object, expected^, desired)`

I remember recognizing this when I was porting the C standard library, I had mapped the `atomic_compare_exchange_*` functions 1:1 with Odin's intrinsics but when reading the standard it made no sense that some combinations were valid, everything in the `seq_cst` failure case except `seq_cst` for success case should also be removed in `core:c/libc` with some asserts detecting a higher ordering for failure compared to success which is what libc and libc++ do as well to help detect misuses of the API.

PR #1616, opened 2022-03-12
add easing + flux package

Skytrias:
easing + flux animation package

would be cool if float arrays would work, but they clash with `math.sin`, `math.cos`, ... procedures that arent overloaded for the possible array types

```odin
package main

import "core:math/ease"

main :: proc() {
  flux := ease.flux_init(f32)
  defer ease.flux_destroy(flux)  

  x: f32
  tween := ease.flux_to(&flux, &x, 100, .Quadratic_Out, time.Second, time.Second) // last param is optional delay
  // tween.data is a rawptr, set to &x by default
  tween.on_complete = proc(flux: ^ease.Flux_Map, data: rawptr) { // tween.data pushed to call in ease.flux_update()
    // ...
  }
  
  // main loop
  for {
    ease.flux_update(&flux)
  }
}
```

### DIFF
--- a/core/math/ease/ease.odin
+++ b/core/math/ease/ease.odin
@@ -1,0 +1,466 @@
+package ease
+
+import "core:math"
+import "core:intrinsics"
+import "core:time"
+
+@(private) PI_2 :: math.PI / 2
+
+// converted to odin from https://github.com/warrenm/AHEasing
+// with additional enum based call
+
+// Modeled after the parabola y = x^2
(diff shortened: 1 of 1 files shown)


Skytrias:
can do cool animations like rxi's previous showcase of his flux lua library

https://user-images.githubusercontent.com/22886746/158018314-bfd5cf99-d124-4450-9f1a-761273dec6d2.mp4

PR #1621, opened 2022-03-14
Core out Makefile, add freebsd building

colrdavidson:


gingerBill:
This just needs testing macOS ARM but it should be ready to merge.

colrdavidson:
Tested on M1 OSX Monterey, works fine!

PR #1640, opened 2022-03-21
Add split_iterator_byte

Lperlind:
Add split_iterator_byte. It functions exactly like split_iterator but takes in a byte seperator rather than a string seperator.
The intention is to provide a faster split parsing if the seperator is known to be byte size.

Splitting by byte rather than string in my obj parser I saw 20% decrease in parsing time, I thought it naturally fit in the core.
If I need to extend to match all the other split functions (which only take string right now) I can in the near future.

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -339,6 +339,25 @@ _split_iterator :: proc(s: ^string, sep: string, sep_save: int) -> (res: string,
 	return
 }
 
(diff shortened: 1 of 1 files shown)


gingerBill:
To keep it consistent with the rest of the, please call it either `split_byte_iterator` or `split_by_byte_iterator`.

PR #1642, opened 2022-03-21
Add user32.odin with most basic procedures to core:sys/windows

themole:

PR #1643, opened 2022-03-21
sys/windows: added rudimentary User32.lib bindings

ftphikari:


themole:
I think we were aiming for the same thing. we might want to join efforts on this topic and maybe talk to @gingerBill about how to do this properly

gingerBill:
@ftphikari @themole Please decide amongst yourself as to the branch to be merged.

ftphikari:
> I think we were aiming for the same thing. we might want to join efforts on this topic and maybe talk to @gingerBill about how to do this properly

Absolutely, I just don't know where?

ftphikari:
@themole ok, from what I can see our type definitions are different, notice that UINT_PTR is unitptr in my bindings, and uint in yours. Is that meaningfull difference? It may be. I took my type definitions from sys/win32. Anyhow, I would be happy if you looked through my bindings since they are smaller, and added what is not in yours.

themole:
@ftphikari I merged your master into mine and used your definition for UINT_PTR. see https://github.com/odin-lang/Odin/pull/1642

ftphikari:
@themole seems good to me, closing this pull request

PR #1662, opened 2022-03-27
sys/windows: add GetSystemMetrics

ftphikari:


core/sys/windows/types.odin
```diff
@@ -483,7 +483,116 @@ HWND_BOTTOM    :: HWND( uintptr(1))     //  1
...
+SM_CXDOUBLECLK       :: 36
+SM_CYDOUBLECLK       :: 37
+SM_CXICONSPACING     :: 38
+SM_CYICONSPACING     :: 39
+SM_MENUDROPALIGNMENT :: 40
+SM_PENWINDOWS        :: 41
+SM_DBCSENABLED       :: 42
+SM_CMOUSEBUTTONS     :: 43
+
+SM_CXFIXEDFRAME :: SM_CXDLGFRAME  // ;win40 name change
```
Kelimion, line 532: Nothing older than Win 8.1 is supported, so I wouldn't bother with win40 aliases.

ftphikari: This is just a comment from the original WinUser.h, I didn't make it. I imported all constants that are Windows 8.1 or older.

PR #1663, opened 2022-03-27
add math easing package

Skytrias:
*had to recreate the pull request, since my fork didnt push this on a different branch*
easing + flux animation package

would be cool if float arrays would work, but they clash with `math.sin`, `math.cos`, ... procedures that arent overloaded for the possible array types

ease example, ease functions can be used individually if wanted e.g. `ease.sine_in(p: f32)` 

```odin
package main

import "core:math/ease"

main :: proc() {
	COUNT :: 5

	for type in 0..<len(ease.Ease) {
		type := ease.Ease(type)
		fmt.eprintf("%20s ", type)
		
		for i in 0..<COUNT {
			value := ease.ease(type, f64(i) / (COUNT - 1))

			if value < 0 {
				fmt.eprintf("%.3f ", value)
			} else {
				fmt.eprintf(" %.3f ", value)
			}
		}
		
		fmt.eprintln()
	}
}
```

```
              Linear  0.000  0.250  0.500  0.750  1.000 
        Quadratic_In  0.000  0.062  0.250  0.562  1.000 
       Quadratic_Out  0.000  0.438  0.750  0.938  1.000 
    Quadratic_In_Out  0.000  0.125  0.500  0.875  1.000 
            Cubic_In  0.000  0.016  0.125  0.422  1.000 
           Cubic_Out  0.000  0.578  0.875  0.984  1.000 
        Cubic_In_Out  0.000  0.062  0.500  0.938  1.000 
          Quartic_In  0.000  0.004  0.062  0.316  1.000 
         Quartic_Out  0.000  0.684  0.938  0.996  1.000 
      Quartic_In_Out  0.000  0.031  0.500  0.969  1.000 
          Quintic_In  0.000  0.001  0.031  0.237  1.000 
         Quintic_Out  0.000  0.763  0.969  0.999  1.000 
      Quintic_In_Out  0.000  0.016  0.500  0.984  1.000 
             Sine_In  0.000  0.076  0.293  0.617  1.000 
            Sine_Out  0.000  0.383  0.707  0.924  1.000 
         Sine_In_Out  0.000  0.146  0.500  0.854  1.000 
         Circular_In  0.000  0.032  0.134  0.339  1.000 
        Circular_Out  0.000  0.661  0.866  0.968  1.000 
     Circular_In_Out  0.000  0.067  0.500  0.933  1.000 
      Exponential_In  0.000  0.006  0.031  0.177  1.000 
     Exponential_Out  0.000  0.823  0.969  0.994  1.000 
  Exponential_In_Out  0.000  0.016  0.500  0.984  1.000 
          Elastic_In  0.000 -0.005 -0.022  0.068  1.000 
         Elastic_Out  0.000  0.932  1.022  1.005  1.000 
      Elastic_In_Out  0.000 -0.011  0.500  1.011  1.000 
             Back_In  0.000 -0.161 -0.375 -0.108  1.000 
            Back_Out  0.000  1.108  1.375  1.161  1.000 
         Back_In_Out  0.000 -0.188  0.500  1.188  1.000 
           Bounce_In -0.000  0.041  0.281  0.527  1.000 
          Bounce_Out  0.000  0.473  0.719  0.959  1.000 
       Bounce_In_Out -0.000  0.141  0.500  0.859  1.000
```

flux example: 

```odin
package main

import "core:math/ease"

main :: proc() {
  flux := ease.flux_init(f32)
  defer ease.flux_destroy(flux)  

  x: f32
  tween := ease.flux_to(&flux, &x, 100, .Quadratic_Out, time.Second, time.Second) // last param is optional delay
  // tween.data is a rawptr, set to &x by default
  tween.on_complete = proc(flux: ^ease.Flux_Map, data: rawptr) { // tween.data pushed to call in ease.flux_update()
    // ...
  }
  
  // main loop
  for {
    ease.flux_update(&flux)
  }
}
```

### DIFF
--- a/core/math/ease/ease.odin
+++ b/core/math/ease/ease.odin
@@ -1,0 +1,466 @@
+package ease
+
+import "core:math"
+import "core:intrinsics"
+import "core:time"
+
+@(private) PI_2 :: math.PI / 2
+
+// converted to odin from https://github.com/warrenm/AHEasing
+// with additional enum based call
+
+// Modeled after the parabola y = x^2
+quadratic_in :: proc "contextless" (p: $T) -> T where intrinsics.type_is_float(T) {
+	return p * p
+}
+
+// Modeled after the parabola y = -x^2 + 2x
+quadratic_out :: proc "contextless" (p: $T) -> T where intrinsics.type_is_float(T) {
+	return -(p * (p - 2))
+}
+
+// Modeled after the piecewise quadratic
+// y = (1/2)((2x)^2)             ; [0, 0.5)
+// y = -(1/2)((2x-1)*(2x-3) - 1) ; [0.5, 1]
+quadratic_in_out :: proc "contextless" (p: $T) -> T where intrinsics.type_is_float(T) {
+	if p < 0.5 {
+		return 2 * p * p;
+	}	else {
+		return (-2 * p * p) + (4 * p) - 1
+	}
+}
+
+// Modeled after the cubic y = x^3
+cubic_in :: proc "contextless" (p: $T) -> T where intrinsics.type_is_float(T) {
+	return p * p * p
+}
+
+// Modeled after the cubic y = (x - 1)^3 + 1
+cubic_out :: proc "contextless" (p: $T) -> T where intrinsics.type_is_float(T) {
+	f := p - 1
+	return f * f * f + 1
+}
+
+// Modeled after the piecewise cubic
+// y = (1/2)((2x)^3)       ; [0, 0.5)
+// y = (1/2)((2x-2)^3 + 2) ; [0.5, 1]
+cubic_in_out :: proc "contextless" (p: $T) -> T where intrinsics.type_is_float(T) {
+	if p < 0.5 {
+		return 4 * p * p * p
+	} else {
+		f := (2 * p) - 2
+		return 0.5 * f * f * f + 1
+	}
+}
+
+// Modeled after the quartic x^4
+quartic_in :: proc "contextless" (p: $T) -> T where intrinsics.type_is_float(T) {
+	return p * p * p * p
+}
+
+// Modeled after the quartic y = 1 - (x - 1)^4
+quartic_out :: proc "contextless" (p: $T) -> T where intrinsics.type_is_float(T) {
+	f := p - 1
+	return f * f * f * (1 - p) + 1
+}
+
+// Modeled after the piecewise quartic
+// y = (1/2)((2x)^4)        ; [0, 0.5)
+// y = -(1/2)((2x-2)^4 - 2) ; [0.5, 1]
+quartic_in_out :: proc "contextless" (p: $T) -> T where intrinsics.type_is_float(T) {
+	if p < 0.5 {
+		return 8 * p * p * p * p
+	}	else {
+		f := p - 1
+		return -8 * f * f * f * f + 1
+	}
+}
+
+// Modeled after the quintic y = x^5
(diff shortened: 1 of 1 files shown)


core/math/ease/ease.odin
```diff
@@ -0,0 +1,466 @@
...
+	values: map[^T]Flux_Tween(T),
+}
+
+Flux_Tween :: struct($T: typeid) {
+	value: ^T,
+	start: T,
+	diff: T,
+	goal: T,
+
+	// using ticks for timing instead
```
Kelimion, line 338: `using ticks for timing instead` of what, seconds, and why? This comment without context of what is replaced by ticks, and why is confusing. More so because it's the first iteration of the package and there's nothing to "instead" against that could be revealed by a diff.

So maybe either delete the comment, or change it to explain why ticks are used there?

core/math/ease/ease.odin
```diff
@@ -0,0 +1,466 @@
...
+
+// update all tweens, wait for their delay if one exists
+// calls callbacks in all stages, when they're filled
+// deletes tween from the map after completion
+flux_update :: proc(flux: ^Flux_Map($T), dt: f64) where intrinsics.type_is_float(T) {
+	size := len(flux.values)
+	now := time.tick_now()
+
+	for key, tween in &flux.values {
+		if tween.delay != 0 {
```
Kelimion, line 422: Is this correct? The way I read it, if you make a tween with a delay of 1 second and update it each frame, then any remainder of the delay is lost.

i.e.: a tween with a delay of 1 second, updated at a `dt` of 0.45:
0.45
0.9
1.35 ... the 0.35 remainder gets ignored instead of forwarding the tween by that much?

Kelimion: Also, why `now` and `diff` to figure out how much the delay has advanced by instead of subtracting `dt` from `delay` until `delay <= 0` and using any remainder as the rest of `dt` to then advance the tween by?

Skytrias: > 

thats how it works in the original flux repo https://github.com/rxi/flux/blob/master/flux.lua#L150
cant remember why i changed it but i think the using core time was more correct? gonna try it out again

Kelimion: I see. I just think it's going to be surprising if you have frame stutters and your delay ends up longer than specified.

At 60fps any `dt` remainder in the `delay` being ignored won't be obvious, but if `dt` is big then a (sub)second delay before the animation starts will effectively be stretched until the next update, and that'll be noticeable.

Skytrias: > Is this correct? The way I read it, if you make a tween with a delay of 1 second and update it each frame, then any remainder of the delay is lost.
> 
> i.e.: a tween with a delay of 1 second, updated at a `dt` of 0.45: 0.45 0.9 1.35 ... the 0.35 remainder gets ignored instead of forwarding the tween by that much?

using the core:time tick diffs arent based on the dt at all, so i dont get what you mean :sweat_smile:

Kelimion: That's exactly what I mean. It keeps track of the delay separately to `dt`, which is just weird. It means that if you don't update it for a second and only then start pumping the update again, the delay will have already passed.

Or if you update it once every 450ms for some reason (doing a lot of processing), then instead of 350ms being applied to your tween, they get dropped on the floor because they were left over from the delay.

Why is `delay` decoupled from `dt` instead of being advanced by it like the tween itself is? It's strange and leads to unexpected situations when you have long frames.

Skytrias: that makes sense yeah, i'll try out the `dt` based approach again :+1:

PR #1664, opened 2022-03-27
add string documentation & examples, fix & cleanup string_multi

Skytrias:
also includes a bug fix on `split_iterator` which would end up looping endlessly

any feedback on the comment style? i heard the documentation format is being worked on at the moment, so code comments will need an update at some point

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -1,16 +1,21 @@
+// simple procedures to manipulate UTF-8 encoded strings
 package strings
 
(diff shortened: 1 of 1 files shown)


gitlost:
I think the comments look very good, would suggest capitalizing the first letter and adding a full stop to each sentence.

(also typo "same to" -> "Same as")
(also consistent use of third person singular, e.g. "split" -> "Splits" in some places)

PR #1671, opened 2022-03-30
Make no crt work on Linux

colrdavidson:


gingerBill:
oooooooo

colrdavidson:
This doesn't do the hard work, and get odin working entirely standalone, you still need to link it against something that has _start and our remaining libc hooks. It is helpful for the libc linux burndown as-is though.

gingerBill:
I'll merge it any way.

PR #1673, opened 2022-03-30
Brand New `package sync` and Atomics Intrinsics

gingerBill:


core/sync/extended.odin
```diff
@@ -146,10 +146,10 @@ Auto_Reset_Event :: struct {
 }
 
 auto_reset_event_signal :: proc(e: ^Auto_Reset_Event) {
-	old_status := atomic_load_relaxed(&e.status)
+	old_status := atomic_load_explicit(&e.status, .seq_cst)
```
graphitemaster, line 149: Probably wanted `.relaxed` here

core/sync/extended.odin
```diff
@@ -146,10 +146,10 @@ Auto_Reset_Event :: struct {
...
 	for {
 		new_status := old_status + 1 if old_status < 1 else 1
-		if _, ok := atomic_compare_exchange_weak_release(&e.status, old_status, new_status); ok {
+		if _, ok := atomic_compare_exchange_weak_explicit(&e.status, old_status, new_status, .seq_cst, .seq_cst); ok {
```
graphitemaster, line 152: This is not the same but also not dangerous either so should be fine

core/c/libc/stdatomic.odin
```diff
@@ -189,102 +177,104 @@ atomic_exchange_explicit :: #force_inline proc(object: ^$T, desired: T, order: m
...
 	value: T; ok: bool
-	#partial switch (failure) {
+	#partial switch failure {
 	case .seq_cst:
 		assert(success != .relaxed)
-		#partial switch (success) {
+		#partial switch success {
 		case .seq_cst:
-			value, ok := intrinsics.atomic_cxchg(object, expected^, desired)
+			value, ok = intrinsics.atomic_compare_exchange_strong_explicit(object, expected^, desired, .seq_cst, .seq_cst)
```
Kelimion, line 196: Are these cases here to be explicit?

You could have:
```odin
#partial switch failure {
	case .seq_cst:
		assert(success != .relaxed)
		value, ok = intrinsics.atomic_compare_exchange_strong_explicit(object, expected^, desired, success, .seq_cst)
```

Kelimion (commented):
Looks good to me. I really like that the type checker checks pairings.

Only thing I'd maybe change is to get rid of `#partial switch success` and just use `success` as a param directly after the `assert` to weed out the unwanted case(s).

core/c/libc/stdatomic.odin
gingerBill: The orders must be constant, and those enums don't match

Kelimion (approved):

PR #1679, opened 2022-03-31
Have get_std_handle() no longer make the handles uninheritable

Tetralux:
This caused all handles returned by GetStdHandle() to also not be inheritable,
which prevents you from handing them to child processes that you might create.

This fixes that.


Kelimion:
This is probably related to the issue I ran into with my chess frontend. Created 3 new handles (stdin, stdout, stderr) and handed them to Stockfish, but it just wouldn't work. Going to need to review `win32.SetHandleInformation` use for that specific kind of use case, which you'd also encounter if you wanted to write a terminal emulator.

Issue #1067, opened 2021-08-03
Add optional message to `#assert`, like `assert`.


Kelimion:
```odin
FORCE_64_BIT :: false;
FORCE_32_BIT :: true;
#assert(!(FORCE_32_BIT && FORCE_64_BIT), "Cannot force 32-bit and 64-bit big backend simultaneously.");
```
Unlike `assert`, `#assert` takes only 1 parameter, the condition. It would be lovely if the compile time assert also allowed for an optional message, so we don't have to do this:

```odin
FORCE_64_BIT :: true;
FORCE_32_BIT :: true;
when (FORCE_32_BIT && FORCE_64_BIT) { #panic("Cannot force 32-bit and 64-bit big backend simultaneously."); };
```

PR #1686, opened 2022-04-02
Allow optional message for `#assert`.

Kelimion:
Currently `#assert` can only take a boolean condition.
This PR adds a second (optional) message string parameter.

It helps avoid this pattern: `when !cond do #panic("message")`.

```odin
_main :: proc() {
	#assert(int == u64)
	#assert(int == u64, "Foozle")
}
```
```
W:/Odin-other/Odin-test/test.odin(7:2) Compile time assertion: int == u64
	Called within '_main' :: proc()
W:/Odin-other/Odin-test/test.odin(8:2) Compile time assertion: int == u64 ("Foozle")
	Called within '_main' :: proc()
```

### DIFF
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -830,8 +830,8 @@ bool check_builtin_procedure(CheckerContext *c, Operand *operand, Ast *call, i32
 			}
 
 		} else if (name == "assert") {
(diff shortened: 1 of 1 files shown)


gitlost:
Would close https://github.com/odin-lang/Odin/issues/1067

PR #1698, opened 2022-04-05
handle cleanup if dns record parsing fails a little better

colrdavidson:
Give this an eyeball before you merge it. Looks like core_net is currently not building due to an issue in mem, probably needs a master -> core_net merge before this can be properly tested.


Kelimion:
Incorporated the dns.odin changes.

PR #1702, opened 2022-04-07
Compiler: Add early error for output path being a directory.

Kelimion:
- Introduce new `Path` type and an array of build paths on the build context.
- Resolve input and output paths/files early (before parsing).
- Error early if inputs are missing or outputs are directories.
- Plumb new file path generation into linker stage instead of its adhoc method.

TODO:
- Remove more adhoc file path generation in parser and linker stage.
- Make intermediate object file generation use new path system.
- Round out and robustify Path helper functions.


Kelimion:
Updated to current master branch.

PR #1703, opened 2022-04-08
add pretty json output in marshal

Skytrias:
adds `pretty_print` option to the `json.marshal` procedure and will properly indent generated json by indentations - struct or array scopes

could add more options like `use_spaces` into the `Pretty` struct but this is enough for my taste :+1:

### DIFF
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -17,13 +17,19 @@ Marshal_Error :: union #shared_nil {
 	io.Error,
 }
 
-marshal :: proc(v: any, allocator := context.allocator) -> (data: []byte, err: Marshal_Error) {
+Pretty :: struct {
+	apply: bool,
+	indentation: int,
+}
+
+marshal :: proc(v: any, pretty: Pretty = {}, allocator := context.allocator) -> (data: []byte, err: Marshal_Error) {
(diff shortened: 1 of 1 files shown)


core/encoding/json/marshal.odin
```diff
@@ -17,25 +17,31 @@ Marshal_Error :: union #shared_nil {
...
 }
 
-marshal_to_builder :: proc(b: ^strings.Builder, v: any) -> Marshal_Error {
-	return marshal_to_writer(strings.to_writer(b), v)
+marshal_to_builder :: proc(b: ^strings.Builder, v: any, pretty: ^Pretty) -> Marshal_Error {
+	return marshal_to_writer(strings.to_writer(b), v, pretty)
 }
 
-marshal_to_writer :: proc(w: io.Writer, v: any) -> (err: Marshal_Error) {
+marshal_to_writer :: proc(w: io.Writer, v: any, pretty: ^Pretty) -> (err: Marshal_Error) {
```
gingerBill, line 44: Please don't make it require this parameter. Either make a wrapper procedure or make it default to `nil` and work without it.

Skytrias: ah sorry for the late reply, i'll just use `pretty: ^Pretty = nil` and not use a bool flag inside the struct

PR #1704, opened 2022-04-09
Add convenience functions to print formatted cstrings to temp memory.

themole:
Add procedures ctprint, ctrpintln and ctprintf that uses contexs temp_allocator and prints formatted cstrings. This reduces the amount of copies needed when interacting with c libraries.

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -116,6 +116,30 @@ tprintf :: proc(fmt: string, args: ..any) -> string {
 	return strings.to_string(str)
 }
 
+// ctprint procedure returns a cstring that was allocated with the current context's temporary allocator
+ctprint :: proc(args: ..any, sep := " ") -> cstring {
+	str: strings.Builder
+	strings.init_builder(&str, context.temp_allocator)
+	sbprint(buf=&str, args=args, sep=sep)
+	strings.write_byte(&str, 0)
+	return strings.to_cstring(str)
+}
+// ctprintln procedure returns a cstring that was allocated with the current context's temporary allocator
+ctprintln :: proc(args: ..any, sep := " ") -> cstring {
+	str: strings.Builder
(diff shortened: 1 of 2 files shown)


core/fmt/fmt.odin
```diff
@@ -116,6 +116,30 @@ tprintf :: proc(fmt: string, args: ..any) -> string {
...
+	return strings.to_cstring(str)
+}
+// ctprintf procedure returns a cstring that was allocated with the current context's temporary allocator
+ctprintf :: proc(fmt: string, args: ..any) -> cstring {
+	str: strings.Builder
+	strings.init_builder(&str, context.temp_allocator)
+	sbprintf(&str, fmt, ..args)
+	strings.write_byte(&str, 0)
+	return strings.to_cstring(str)
+}
```
gingerBill, line 142: You would need to add the `context.allocator` forms to make this consistent.

core/strings/builder.odin
```diff
@@ -146,6 +146,12 @@ to_string :: proc(b: Builder) -> string {
 	return string(b.buf[:])
 }
 
+// cast the builder byte buffer to a cstring and return it. assert that the buffer ends with a terminating nul byte.
+to_cstring :: proc(b: Builder) -> cstring {
+	assert(b.buf[len(b.buf)-1] == 0)
+	return cstring(&b.buf[0])
+}
```
gingerBill, line 153: This is a very bad assumption to make. Don't do this.

themole: I asserted to let users know this is a requirement for the conversion to work and to avoid creating a string of undefined length. Wouldn't it be even more dangerous to not assert here? How would this be done the odin way?

gingerBill: To do this safely, it would have been better to check if the last value was `0`, and if not, append `0`. And if the length of the buffer is zero or could not append, return `nil`.

PR #1705, opened 2022-04-10
Add convenience functions to print formatted cstrings

themole:


Skytrias:
would be really nice to have for beginners

ftphikari:
I like this

gingerBill:
I'd rather have this as a utility procedure outside of the `package fmt`. I don't really want to have the `cstring` stuff in there if I can help it.

PR #1726, opened 2022-04-18
Add Netpbm image format support

WalterPlinge:
Initial commit of my Netpbm image format support.
Includes a helper file and the start of a doc file.
I left a couple question comments with `//?`, namely about whether specific validation is needed on this library's end.
Minimally touched `core:image/common.odin` (I just put it in alphabet order).
Maybe some of the error values could be moved to `General_Image_Error`, as a couple are now shared between specific image formats.


WalterPlinge:
Apparently what I mentioned about error enums makes it fail, because they cannot be used implicitly when there are two of the same value. Should it be used explicitly, or would it be better to pull out the common errors?

Kelimion:
Looks solid from a first glance. I do see you've moved `Invalid_Number_Of_Channels` from QOI to General, but I see no corresponding patch to the QOI reader. Did I not use that return value?

I should probably move `Invalid_*` from QOI to General.

The PBM reader and writer could also use some tests, like reading the PNGs, writing them out as PBM, reading them in again, and comparing pixel data. It's what I did for QOI. I can write those tests.

Kelimion:
Refactored API slightly, added tests, fixed bugs. The entire PNG suite can now roundtrip through Netpbm and result in the same image data out the other side.

PR #1746, opened 2022-04-26
`core:sync` Improvements

gingerBill:
Improvements/Simplifications to `Cond`/`Atomic_Cond` and `Sema/``Atomic_Sema`.

Add `Parker`


Kelimion:
How does it fare on this code, with and without the `-define:ODIN_SYNC_SEMA_USE_FUTEX=false`?
```cpp
package main

import "core:fmt"
import "core:thread"

task_proc :: proc(task: thread.Task) {
    data := cast(^i64)task.data
    data^ += 500+13*64
}

main :: proc() {
    pool: thread.Pool
    i := 1
    for true {
        fmt.printf("Round %v\n", i)
        i += 1
        fmt.println("before init")
        thread.pool_init(&pool, 25, context.allocator)
        fmt.println("before start")
        thread.pool_start(&pool)
        fmt.println("before add task")
        data: i64
        thread.pool_add_task(&pool, task_proc, &data, 0)
        fmt.println("before join")
        thread.pool_join(&pool)
        fmt.println("before destroy")
        thread.pool_destroy(&pool)
    }
}
```

Kelimion:
I like it. The `Parker` is lovely and simple.

PR #1755, opened 2022-04-29
Update `vendor/miniaudio` to v0.11.9

bkrypt:
This updates the `vendor/miniaudio` package to the latest version of miniaudio, v0.11.9 at the time of writing.

I won't lie, when I read the miniaudio breaking API changes from `v0.10.x` to `v0.11.x` and said to myself "eh, this doesn't seem like that big of a task", I totally underestimated the size of the task. 😅 

Please let me know if this needs to go through a more formal process, or the if the PR requires more. This was really just a major side-quest sparked by wanting to start with audio in my project (instead of graphics as I normally do), really liking the sound of some of the features in `v0.11.x` of miniaudio, and wanting to use them in Odin.


vendor/miniaudio/common.odin
```diff
@@ -13,13 +13,15 @@ when ODIN_OS == .Windows {
...
-LOG_LEVEL_ERROR   :: 1
+/* SIMD alignment in bytes. Currently set to 32 bytes in preparation for future AVX optimizations. */
+SIMD_ALIGNMENT :: 32
+
+log_level :: enum c.int {
+	LOG_LEVEL_DEBUG   = 4,
+	LOG_LEVEL_INFO    = 3,
+	LOG_LEVEL_WARNING = 2,
+	LOG_LEVEL_ERROR   = 1,	
+}
```
gingerBill, line 24: Why has this become an enum rather than the global constants?

vendor/miniaudio/common.odin
```diff
@@ -191,6 +193,7 @@ format :: enum c.int {
 	s24     = 3,     /* Tightly packed. 3 bytes per sample. */
 	s32     = 4,
 	f32     = 5,
+	count,
```
gingerBill, line 196: This is not required because `len(format)` would have been equivalent (iff `count` doesn't exist)

vendor/miniaudio/device_io_types.odin
```diff
@@ -312,39 +357,34 @@ device_info :: struct {
...
+	deviceType:                 device_type,
+	sampleRate:                 u32,
+	periodSizeInFrames:         u32,
+	periodSizeInMilliseconds:   u32,
+	periods:                    u32,
+	performanceProfile:         performance_profile,
+	noPreSilencedOutputBuffer:  b8,   /* When set to true, the contents of the output buffer passed into the data callback will be left undefined rather than initialized to zero. */
+	noClip:                     b8,   /* When set to true, the contents of the output buffer passed into the data callback will be clipped after returning. Only applies when the playback sample format is f32. */
+	noDisableDenormals:         b8,   /* Do not disable denormals when firing the data callback. */
+  noFixedSizedCallback:       b8,   /* Disables strict fixed-sized data callbacks. Setting this to true will result in the period size being treated only as a hint to the backend. This is an optimization for those who don't need fixed sized callbacks. */
```
gingerBill, line 369: Indentation issue

vendor/miniaudio/common.odin
bkrypt: Just to keep in line with the change from `#define` constants (in `v0.10.x`) to `enum` (in `v0.11.x`) for the same in the actual miniaudio source header.

PR #1756, opened 2022-04-30
sys/windows: add a couple of procedures and types

ftphikari:


core/sys/windows/gdi32.odin
```diff
@@ -62,5 +62,11 @@ foreign gdi32 {
 	ChoosePixelFormat :: proc(hdc: HDC, ppfd: ^PIXELFORMATDESCRIPTOR) -> c_int ---
 	SwapBuffers :: proc(HDC) -> BOOL ---
 
+	SetDCBrushColor :: proc(hdc: HDC, color: COLORREF) -> COLORREF ---
+	GetDCBrushColor :: proc(hdc: HDC) -> COLORREF ---
 	PatBlt :: proc(hdc: HDC, x, y, w, h: c_int, rop: DWORD) -> BOOL ---
 }
+
+RGB :: proc(r, g, b: u8) -> COLORREF {
```
Kelimion, line 70: I'd perhaps gone with this:
```cpp
RGB :: proc(rgb: [3]u8) -> COLORREF {
    res: [4]u8 = {0, rgb.b, rgb.g, rgb.r} // ABGR
    return transmute(COLORREF)res
}
```
And also put a comment on COLORREF that it's packed as ABGR.

core/image/common.odin
```diff
@@ -1131,4 +1131,4 @@ write_bytes :: proc(buf: ^bytes.Buffer, data: []u8) -> (err: compress.General_Er
 		return compress.General_Error.Resize_Failed
 	}
 	return nil
-}
\ No newline at end of file
+}
```
Kelimion, line 1134: Not sure what this meaningless change to this file is doing in the patch?

ftphikari: Sublime automatically inserted newline :/

core/sys/windows/gdi32.odin
```diff
@@ -62,5 +62,13 @@ foreign gdi32 {
 	ChoosePixelFormat :: proc(hdc: HDC, ppfd: ^PIXELFORMATDESCRIPTOR) -> c_int ---
 	SwapBuffers :: proc(HDC) -> BOOL ---
 
+	SetDCBrushColor :: proc(hdc: HDC, color: COLORREF) -> COLORREF ---
+	GetDCBrushColor :: proc(hdc: HDC) -> COLORREF ---
 	PatBlt :: proc(hdc: HDC, x, y, w, h: c_int, rop: DWORD) -> BOOL ---
 }
+
+// Windows colors are packed as ABGR
+RGB :: #force_inline proc "contextless" (r, g, b: u8) -> COLORREF {
```
Kelimion, line 71: Test failed because params are still `r, g, b: u8` instead of `rgb: [3]u8`

core/sys/windows/types.odin
```diff
@@ -55,6 +55,8 @@ UINT_PTR :: uintptr
 ULONG :: c_ulong
 UCHAR :: BYTE
 NTSTATUS :: c.long
+COLORREF :: DWORD
```
Kelimion, line 58: Maybe add this comment here as well. `// Windows colors are packed as ABGR`

ftphikari:
I use `r, g, b: u8` instead of `rgb: [3]u8` because that's how the original macro does it

Kelimion:
> I use `r, g, b: u8` instead of `rgb: [3]u8` because that's how the original macro does it, is there a performance reason or something why I should do it with an array instead?

Not a performance reason, but it's more compatible with how people use color types in Odin, as a `[3]u8` or `[4]u8`.

PR #1758, opened 2022-04-30
sys/windows: add a couple procedures and tons of constants

ftphikari:


core/sys/windows/user32.odin
```diff
@@ -114,10 +114,14 @@ foreign user32 {
 		cy: c_int,
 		uFlags: UINT,
 	) -> BOOL ---
+	MoveWindow :: proc(hWnd: HWND, X, Y, hWidth, hHeight: c_int, bRepaint: BOOL) -> BOOL ---
 	GetSystemMetrics :: proc(nIndex: c_int) -> c_int ---
 	AdjustWindowRect :: proc(lpRect: LPRECT, dwStyle: DWORD, bMenu: BOOL) -> BOOL ---
 	AdjustWindowRectEx :: proc(lpRect: LPRECT, dwStyle: DWORD, bMenu: BOOL, dwExStyle: DWORD) -> BOOL ---
 
+	SystemParametersInfoA :: proc(uiAction, uiParam: UINT, pvParam: PVOID, fWinIni: UINT) -> BOOL ---
```
gingerBill, line 122: I'd recommend not having any of the `A` functions if there is an `W` equivalent.

ftphikari: Do you mean all `A` procedures from all libs should be removed from `sys/windows`?

gingerBill: For this specific function, I don't see any reason why the `A` one needs to exist if the `W` does. 

I know Windows will in the future (or maybe even already) make all the `A` functions UTF-8 and do the conversion to UTF-16 for use already, but currently we are relying on our own conversions for this to the `W` calls.

Now I'm not sure if we should remove all the `A` stuff or not. It's a complicated API design issue.

Kelimion: > For this specific function, I don't see any reason why the `A` one needs to exist if the `W` does.
> 
> I know Windows will in the future (or maybe even already) make all the `A` functions UTF-8 and do the conversion to UTF-16 for use already, but currently we are relying on our own conversions for this to the `W` calls.
> 
> Now I'm not sure if we should remove all the `A` stuff or not. It's a complicated API design issue.

Let's tackle that question as part of the Great OS/SYS Restructuring of 2022 and prefer W for now.

ftphikari: I'm sure if Microsoft defined 2 functions differently, they are implemented differently, because they only did that for select number of them.

ftphikari: Either way, this is how it is defined in the WinUser.h, so I'll pass on the responsibility of removing A bindings to you guys.

ftphikari:
Is there anything else holding this back?

Kelimion:
> Is there anything else holding this back?
Removal of the A variants, as Bill requested.

ftphikari:
Done, but don't forget that these 2 procedures do, in fact, act differently, depending if it is the W or A variant, since pvParam could be a string and depending if it's W or A variant, string formats are different.

gingerBill:
> Done, but don't forget that these 2 procedures do, in fact, act differently, depending if it is the W or A variant, since pvParam could be a string and depending if it's W or A variant, string formats are different.

We haven't forgot. We want to deprecate the A stuff.

ftphikari:
Is there anything else holding this back?

PR #1768, opened 2022-05-04
Make the link order of foreign imports deterministic

gingerBill:
Foreign libraries will now be imported in a deterministic order based on a few factors:

* `@(priority_index=<int>)` (new attribute)
* Package ordering (determined based on import usage (priority sort))
* File ordering (alphabetical within the package)
* Order of import in source

### DIFF
--- a/src/llvm_backend.cpp
+++ b/src/llvm_backend.cpp
@@ -43,6 +43,13 @@ GB_COMPARE_PROC(foreign_library_cmp) {
 	if (x == y) {
 		return 0;
 	}
+	GB_ASSERT(x->kind == Entity_LibraryName);
+	GB_ASSERT(y->kind == Entity_LibraryName);
+
+	cmp = i64_cmp(x->LibraryName.priority_index, y->LibraryName.priority_index);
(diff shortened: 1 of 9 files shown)


src/llvm_backend.cpp
```diff
@@ -29,29 +29,46 @@ void lb_add_foreign_library_path(lbModule *m, Entity *e) {
...
+		isize order_y = y->pkg ? y->pkg->order : 0;
+		cmp = isize_cmp(order_x, order_y);
+		if (cmp) {
+			return cmp;
 		}
+	}
+	if (x->file != y->file) {
+		String fullpath_x = x->file ? x->file->fullpath : (String{});
+		String fullpath_y = y->file ? y->file->fullpath : (String{});
+		String file_x = filename_from_path(fullpath_x);
```
Kelimion, line 65: I'll do a `String` -> `Path` pass on the rest of the compiler later. For now this looks fine to me.

Kelimion (approved):
Followed it as you made the changes. Looks good to me.

PR #1770, opened 2022-05-06
sys/windows: fix RGB macro

ftphikari:


ftphikari:
This looks better to you?

PR #1775, opened 2022-05-08
Ctrl-C handler on Windows.

Kelimion:
Allows a console app to listen to `Ctrl-C` and handle it themselves by having the OS ignore it, or install a trap handler for keys like these.

See: https://docs.microsoft.com/en-us/windows/console/setconsolectrlhandler

```cpp
when ODIN_OS == .Windows {

	set_ignore_ctrl_c :: proc(ignore: bool) -> (res: bool) {
		return bool(win32.set_console_ctrl_handler(nil, win32.BOOL(ignore)))
	}
}
```


ftphikari:
Why does this not follow the Windows naming scheme like other `sys/windows` procedures?

Kelimion:
> Why does this not follow the Windows naming scheme like other `sys/windows` procedures?
1) As you are well aware, that package is going to see serious refactoring along with `core:os`, and isn't as homogeneous in its naming as you make it out to be.

2) What's even the point of commenting on closed pull requests of working and tested code? Our time is more precious than that.

PR #1792, opened 2022-05-16
Os2 linux

jasonKercher:
This PR fills in most of os2 for Linux without libc.


core/os/os2/heap_linux.odin
```diff
@@ -0,0 +1,722 @@
...
+		diff := int(aligned_ptr - ptr)
+		if (size + diff) > space {
+			return nil, .Out_Of_Memory
+		}
+
+		aligned_mem = rawptr(aligned_ptr)
+		mem.ptr_offset((^rawptr)(aligned_mem), -1)^ = allocated_mem
+
+		return mem.byte_slice(aligned_mem, size), nil
+	}
```
gingerBill, line 177: If you have a custom memory allocator, then you have no need for doing this trick, you can do the alignment directly.

jasonKercher: That's a good point.  I think that went over my head because I didn't want to change the signature of `heap_alloc`.  I could do either
```odin
heap_alloc :: proc(size: int, alignment: int = 8) -> rawptr
```
or
```odin
heap_alloc :: proc(size: int) -> rawptr {
        return heap_alloc_aligned(size, 8)   // < This called directly by the allocator
}
```

gingerBill: Don't worry about changing the signature. It might be a much better idea to be the following:
```odin
heap_alloc :: proc(size, alignment: int) -> (rawptr, runtime.Allocation_Error) {...}
```

gingerBill:
I'll merge this even if it needs some work.

Issue #1787, opened 2022-05-13
Bug in microsoft_crazyness.h


khvorov45:
## Context

```
Odin: dev-2022-05:cc81057d
OS:   Windows 10 Professional (version: 21H1), build 19043.1645
CPU:  Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
RAM:  16338 MiB
```

## Bug

Sometimes (about 1 in a 100) compilation failes with
```
Output path G:/Projects/wirede must have an appropriate extension
```
The actual path is
```
G:/Projects/wiredeck/build/wiredeck.exe
```
I put some print statements into the compiler and found that the correct path turned into
```
G:/Projects/wirede@@/build/wiredeck.exe
```
where `@` is how I marked `\0`

I tracked it down to here (microsoft_crazyness.h `find_windows_kit_root` line 249)
```c
DWORD required_length;
...
DWORD length = required_length + 2;  // The +2 is for the maybe optional zero later on. Probably we are over-allocating.
wchar_t *value = (wchar_t *)calloc(1, length);
...
if (value[length]) {
    value[length+1] = 0;
}
```
Note that `length` is already the length with the null terminator included (so `length + 1` is wrong?).
Also note that `length` is length in bytes, not in `wchar_t`. So `value[length + 1]` is actually setting something to 0 that's way outside the actual allocated region. 

I printed the pointer values for one of the failures I got
```
value: 1db34c96dd0 | length: 40 | build_context.out_filepath.text: 1db34c96e10
```
So `value` happened to be allocated just before `build_context.out_filepath.text` and so `value[length+1]` pointed 18 bytes into `build_context.out_filepath.text` setting the two bytes (`whar_t`) to 0.

The correct version should be something like
```c
wchar_t* last_wchar = (wchar_t*)((uint8_t*)value + length - 1)
if (last_wchar[0]) {
    last_wchar[0] = 0;
}
```
But even this is probably unnecessary. I read the documentation of `RegQueryValueExW` and it says
>If the data has the REG_SZ, REG_MULTI_SZ or REG_EXPAND_SZ type, this size includes any terminating null character or characters unless the data was stored without them. For more information, see Remarks.

Since the `value` buffer is always 2 bytes larger than `required_length` and it's always zeroed (`calloc`) there should always be a (double) null-terminator on it presumably? So it shouldn't matter if the data was stored with or without the null terminators, our `value` buffer will have at least 2 regardless, so this 
```
if (value[length]) {
[...]

PR #1809, opened 2022-05-27
Refactor ms_craziness.h

Kelimion:
Factor out or wrap a lot of the original libc wide string handling, using the `String` type instead.
We can also free any memory allocated by this code wholesale by calling `mc_free_all`.


Kelimion:
Addresses #1787.

PR #1819, opened 2022-06-02
Allocator: tidying up Query_Info return values when not part of Query_Features

IanLilleyT:
Some allocators were returning `nil, nil` instead of `nil, .Mode_Not_Implemented` when `Query_Info` was not part of the `Query_Features`

### DIFF
--- a/core/mem/allocators.odin
+++ b/core/mem/allocators.odin
@@ -886,6 +886,10 @@ tracking_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
 		}
 	case .Free:
 		delete_key(&data.allocation_map, old_memory)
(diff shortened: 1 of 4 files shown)


core/mem/allocators.odin
```diff
@@ -886,6 +886,10 @@ tracking_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
 		}
 	case .Free:
 		delete_key(&data.allocation_map, old_memory)
+	case .Free_All:
+		if data.clear_on_free_all {
+			clear_map(&data.allocation_map)
+		}
```
IanLilleyT, line 892: I'm just making the ordering of the switch case consistent here

PR #1820, opened 2022-06-02
[runtime] Add builtin `shrink` for dynamic arrays and maps

Tetralux:
Asks the allocator to shrink the backing allocation to the current __length__, or a capacity of the user's choosing.

```odin
shrink(&array) // shrinks to len(array)
shrink(&array, N) // shrink to N capacity

shrink(&map) // shrinks down to len(map)
shrink(&map, N) // shrink to N capacity
```

If `new_cap > cap()`, then `shrink()` is a no-op.

If `new_cap > len()`, then `len()` is left unchanged.

### DIFF
--- a/core/runtime/core_builtin.odin
+++ b/core/runtime/core_builtin.odin
@@ -129,6 +129,9 @@ reserve :: proc{reserve_dynamic_array, reserve_map}
 @builtin
 resize :: proc{resize_dynamic_array}
 
+// Shrinks the capacity of a dynamic array or map down to the current length, or the given capacity.
+@builtin
+shrink :: proc{shrink_dynamic_array, shrink_map}
 
 @builtin
 free :: proc{mem_free}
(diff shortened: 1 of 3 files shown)


core/runtime/core_builtin.odin
```diff
@@ -536,6 +557,54 @@ resize_dynamic_array :: proc(array: ^$T/[dynamic]$E, length: int, loc := #caller
...
+shrink_dynamic_array :: proc(array: ^$T/[dynamic]$E, new_cap := -1, loc := #caller_location) -> bool {
+	if array == nil {
+		return false
+	}
+	a := (^Raw_Dynamic_Array)(array)
+
+	new_cap := new_cap if new_cap != -1 else a.len
+
+	if new_cap > a.cap {
+		return true
```
gingerBill, line 578: Should this be `true`?

core/runtime/dynamic_array_internal.odin
```diff
@@ -41,6 +41,37 @@ __dynamic_array_reserve :: proc(array_: rawptr, elem_size, elem_align: int, cap:
...
+
+	// NOTE(tetra, 2020-01-26): We set the allocator before earlying-out below, because user code is usually written
+	// assuming that appending/reserving will set the allocator, if it is not already set.
+	if array.allocator.procedure == nil {
+		array.allocator = context.allocator
+	}
+	assert(array.allocator.procedure != nil)
+
+	if new_cap > array.cap {
+		return true
```
gingerBill, line 55: Should this be `true`?

core/runtime/core_builtin.odin
Tetralux: I figured that it should report that it was successful if there was no shrinking required, since you'd generally only report an error if it was unable to perform its job.

core/runtime/dynamic_array_internal.odin
Tetralux: Same here.

Tetralux:
Out of interest, I'll also document here that I've set it up to return a `bool`, like the other reservation procedures.
This does make it a bit more annoying to use in real code, if the real code is set up to handle `mem.Allocator_Error`s instead, probably warranting the need to make your own wrapping procedure.

A pass should probably be done at some point to make more use of `mem.Allocator_Error` return values to enable more convenient use of `or_return` for memory allocation, but I've done done that in this PR.

core/runtime/core_builtin.odin
Tetralux: I'm not sure if it's useful to distinquish the third one or not. 🤔 
If we want to do that though, maybe we could do `-> (did_shrink: bool, err: Allocator_Error)` ?

Tetralux: Could always return `(did_shrink: bool, err: Allocator_Error)` instead I suppose. That actually seems easy enough to do.

PR #1830, opened 2022-06-04
add max heap implementation to slice package

graphitemaster:
This adds an implementation of a max heap to the slice package. Heaps are a type of data structure which aren't actually a separate type, but rather a property or invariant which can be maintained on an existing array of data in place.

This is the most efficient implementation of such a data structure, inspired by [the following paper on binary heaps](https://web.archive.org/web/20160205023201/http://www.stats.ox.ac.uk/__data/assets/pdf_file/0015/4173/heapbuildjalg.pdf) (specifically the Floyd algorithm) and a simple API inspired by the C++ standard library (they actually got something right)

The following procedures are provided by this package.
 * `heap.make(slice, compare)` Which makes a max heap inside the slice with given comparator.
 * `heap.push(slice, compare)` Which inserts the element at the end into the max heap.
 * `heap.pop(slice, compare)` Which swaps the value at 0 with the end, making a subrange into a heap.
 * `heap.sort(slice, compare)` Which sorts the heap and removes the max heap property.
 * `heap.is_heap_until(slice, compare)` Which examines the slice to find the largest range that is a max heap. This returns an index.
 * `heap.is_heap(slice, compare)` Which tells you if a given slice is a max heap.

The comparator passed to all of these is of the form `proc(lhs, ths: T) -> bool`, the package is written to take `compare: $T` when it should take `$compare: $T`, however some bugs in Odin's handling of constant parapoly procedures is a bit broken, once fixed these should be switched over.

Since the comparator is user supplied, a min heap can be created by using the opposite comparison, i.e `>` instead of `<`. Likewise, other types of heaps can be constructed (e.g heap score) simply by setting the comparison.

Everything is commented and complexity analysis of how expensive each procedure is, is also provided.

### DIFF
--- a/core/slice/heap/heap.odin
+++ b/core/slice/heap/heap.odin
@@ -30,8 +30,7 @@ make :: proc(data: []$T, compare: $C) {
 
 	// start from data parent, no need to consider children
 	for start := (length - 2) / 2; start >= 0; start -= 1 {
-		sift_down(data, compare, start);
-		start -= 1;
+		sift_down(data, compare, start)
 	}
 }
 
--- a/core/slice/heap/heap.odin
+++ b/core/slice/heap/heap.odin
@@ -1,0 +1,232 @@
+/*
+	Copyright 2022 Dale Weiler <weilercdale@gmail.com>.
+	Made available under Odin's BSD-3 license.
+
+	List of contributors:
+		Dale Weiler: Initial implementation
+*/
+
+// Package implements a generic max heap in-place on a slice for any type.
+package heap
+
+/*
+	Constructs a max heap in slice given by data with comparator. A max heap is
+	a range of elements which has the following properties:
+
+	1. With N = len(data), for all 0 < i < N, data[(i - 1) / 2] does not compare
+	   less than data[i].
+
+	2. A new element can be added using push in O(log n) time.
+
+	3. The first element can be removed using pop in O(log n) time.
+
+	The comparator compares elements of type T and can be used to construct a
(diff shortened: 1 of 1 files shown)


core/slice/heap/heap.odin
```diff
@@ -0,0 +1,231 @@
...
+	   less than data[i].
+
+	2. A new element can be added using push in O(log n) time.
+
+	3. The first element can be removed using pop in O(log n) time.
+
+	The comparator compares elements of type T and can be used to construct a
+	max heap (less than) or min heap (greater than) for T.
+*/
+make :: proc(data: []$T, compare: $C) {
```
gingerBill, line 26: Why is this called `make`? And not something like `construct` or `init` or whatever?

graphitemaster: Because all the other implementations call it make and I think it's called that in the algorithm? It's not a one time construction of something, you're expected to use it multiple times on the same array to make it into a heap because the heap property is easily invalidated under mutation so it's often used multiple time when you need the heap property.

gingerBill: The reason `make` feels off is because we have `make` in Odin and it has a specific meaning. So this having `make` and not returning anything is a bit off.

gingerBill:
I'll merge it and then decide on a better name later.

core/slice/heap/heap.odin
graphitemaster: Other implementations call it heapify, but `heap.heapify` seems silly to me.

PR #1844, opened 2022-06-14
Added missing #includes on OpenBSD

Ronaldr1985:
Resolves issues with compiling Odin on OpenBSD, sys/resource.h and sys/wait.h are required for WIFEXITED and WEXITSTATUS.

### DIFF
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -eu
 
 GIT_SHA=$(git rev-parse --short HEAD)
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -25,11 +25,6 @@
 #include <psapi.h>
 #endif
 
-#if defined(GB_SYSTEM_OPENBSD)
-#include <sys/resource.h>
-#include <sys/wait.h>
-#endif
-
 #include <math.h>
 #include <string.h>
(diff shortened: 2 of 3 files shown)


src/common.cpp
```diff
@@ -25,6 +25,11 @@
 #include <psapi.h>
 #endif
 
+#if defined(GB_SYSTEM_OPENBSD)
+#include <sys/resource.h>
+#include <sys/wait.h>
```
semarie, line 30: only `sys/wait.h` should be required for `WIFEXITED` and `WEXITSTATUS` macro.

also, I would put the missing includes inside `src/gb/gb.h` under the `#if defined(GB_SYSTEM_OPENBSD)`

build_odin.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
semarie, line 1: I agree that it is the more portable way to invoke `bash`

semarie (changes requested):

semarie:
@Ronaldr1985 ping ?

Ronaldr1985:
Apologies for the large delay, I was busy and then I got sick unfortunately.

However, I have moved the include to gb/gb.h and removed it from common.cpp.

build_odin.sh
Ronaldr1985: Yes, this means that the script can still run on OpenBSD without requiring changes to the SHABANG.

PR #1852, opened 2022-06-20
Improve build_odin.sh

sus-impost0r:
- Fix a bug in `config_darwin` where a llvm-config version less than 13 would be accepted because a parameter expansion was forgotten.
- Port the script to POSIX sh (replace `[[ ]]` with `[ ]`, remove set -eu)
- Quote Every Fucking Substitution consistently.
- Style changes: Braces around variable names. Single quotes when no substitution is required.

### DIFF
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -10,142 +10,293 @@ CC='clang'
 OS="$(uname)"
 
 panic() {
-        printf '%s\n' "$1"
-        exit 1
+	printf '%s\n' "$1"
+	exit 1
 }
 
 # TODO replace awk with something portable
 version() { echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }'; }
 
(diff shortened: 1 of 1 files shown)


gingerBill:
First is not a bug. We do support versions 11, 12, and 13.

build_odin.sh
```diff
@@ -1,111 +1,263 @@
...
 version() { echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }'; }
 
 config_darwin() {
-	ARCH=$(uname -m)
-	LLVM_CONFIG=llvm-config
+	ARCH="$(uname -m)"
+	LLVM_CONFIG='llvm-config'
 
 	# allow for arm only llvm's with version 13
-	if [ ARCH == arm64 ]; then
```
sus-impost0r, line 24: Sorry, I was unclear. I'm referring to this.

PR #1854, opened 2022-06-21
Add runtime.dll_forward_reason for Windows DLLs.

Kelimion:
Some plugin architectures require you to know when your DLL (written in Odin) is about to be unloaded, or another thread's about to make use of it.

`fwdReason` as passed to the DLL's entry point will now populate `dll_forward_reason` on Windows to allow you to handle this.

### DIFF
--- a/core/runtime/core.odin
+++ b/core/runtime/core.odin
@@ -267,6 +267,19 @@ type_table: []Type_Info
 
 args__: []cstring
 
+when ODIN_OS == .Windows {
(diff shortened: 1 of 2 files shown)


Kelimion:
I'd put the global in `entry_windows.odin` or `os_specific_window.odin`, but those are marked as `//+private`, so we're obliged to add it to `core.odin` in an `when ODIN_OS == .Windows` block.

Issue #1859, opened 2022-06-25
Support `&&` with `bool` and `b32` operands


hstormo:
## Context
odin version dev-2022-06:eec9be71

## Expected Behavior
`bool` and `b32` represent the same thing, so it should be possible to `&&` them together.

## Current Behavior
Doing this produces a "mismatched types" compilation error.

PR #1860, opened 2022-06-25
Fix issue 1859

Pablo96:
This PR allow any boolean type to be `&&` or `||` with any other boolean type and thus fixing the [issue 1859](https://github.com/odin-lang/Odin/issues/1859)


gingerBill:
This is a bodge and not the correct logic to handle this.

PR #1876, opened 2022-07-08
Allow passing config options to build script

mohd-akram:
This allows specifying:

- CXX
- CPPFLAGS
- CXXFLAGS
- LDFLAGS
- ODIN_VERSION
- LLVM_CONFIG

It also allows building when the directory is not a git repository (eg. when downloading a .zip archive from GitHub).

### DIFF
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -1,12 +1,20 @@
 #!/bin/bash
 set -eu
 
-GIT_SHA=$(git rev-parse --short HEAD)
+: ${CXX=clang++}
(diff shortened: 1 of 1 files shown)


mohd-akram:
Is this good to merge? Need this to update Odin in MacPorts. There are no functional changes in this, we simply allow specifying some options as the previous Makefile allowed.

gingerBill:
This just requires some validation on our end to make sure it works correctly, that is all.

Issue #1858, opened 2022-06-25
Debug info tracking


JesseRMeyer:
```go
main :: proc() {

    bob := 0
    for i := 0; i < 10; i += 1 {
        bob += i
    }

    i := 0 //comment this out to 'fix' debug info
```

`i` will always report 0 unless the second `i` is commented out, and then correctly reports its value.

If this logic is hoisted into its own function, `i` simply reports nonsense values, even after the second `i` is overwritten, implying the possibility of reporting an incorrect stack location.

`odin build debug_example -o:minimal -debug`

```
        Odin: dev-2022-06
        OS:   Windows 10 Home Basic (version: 21H2), build 19044.1766
        CPU:  Intel(R) Core(TM) i5-3570K CPU @ 3.40GHz
        RAM:  32717 MiB
```

gingerBill:
This should be fixed now.

JesseRMeyer:
As mentioned in the other PR, variable state tracking has improved but the problem is not fixed in its entirety.
![bob](https://user-images.githubusercontent.com/5846309/179520179-c5d0a329-17b1-4a49-bbf4-8d449768187d.JPG)

Try playing with the provided code to get a feel for the variable tracking weirdness.  If `i := 0 //comment this out to 'fix' debug info` is commented out, then `i` in the loop goes out of tracking scope on the condition check (is that intended?).  If it not commented out, `i`i is not out of scope, but reports uninitialized values.  

Visual Studio's debug view makes this a little clearer.
![ms_bob](https://user-images.githubusercontent.com/5846309/179520800-9598c688-4fb6-4976-a6b7-2b6ec691ead7.JPG)

fabiansperber:
That is (kind of) expected. The initialization will only happen once you step over that line, so when stepping the first time on the `for` line, it should change from uninitialized to 0 (next breakpoint on that line is before checking the condition).

We could have an(the) extra zero-initialization step (A) happen before the debug breakpoint, but in the long term, imo, there should not be an initialization step except the one that assigns the value being used (`0` here).
For example `for i:=slow_function(); i< 10; i += 1 {` we would want the function call to not yet have happened when a breakpoint is triggered on the `for`, so there would be no value to assign to `i`.

But that is just defining expectations, can make it work either way.

(A) Right now, there are two stores writing 0 because zero-initialization is always done for variables, regardless if there is a value assigned in the declaration. My guess is that this was only a workaround for some other issue and not the intended way for variable initialization to work.

[...]

PR #1877, opened 2022-07-08
Improve debug info

fabiansperber:
Improve generation of llvm debug source locations:
- begin/end of procedure (e.g. previously stepping in the debugger would jump from end of proc to the beginning)
- fix scopes of some instruction in the for loop #1858

Embed a natvis file when building the compiler on windows as a debug build to improve debugging experience (at least in msvc) by providing views for String and Array<>

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -58,7 +58,7 @@ set libs= ^
 set linker_flags= -incremental:no -opt:ref -subsystem:console
 
 if %release_mode% EQU 0 ( rem Debug
-	set linker_flags=%linker_flags% -debug
+	set linker_flags=%linker_flags% -debug /NATVIS:src\odin_compiler.natvis
 ) else ( rem Release
 	set linker_flags=%linker_flags% -debug
 )
--- a/src/llvm_backend_debug.cpp
+++ b/src/llvm_backend_debug.cpp
@@ -43,6 +43,10 @@ LLVMMetadataRef lb_debug_location_from_ast(lbProcedure *p, Ast *node) {
 	GB_ASSERT(node != nullptr);
 	return lb_debug_location_from_token_pos(p, ast_token(node).pos);
 }
+LLVMMetadataRef lb_debug_end_location_from_ast(lbProcedure *p, Ast *node) {
+	GB_ASSERT(node != nullptr);
+	return lb_debug_location_from_token_pos(p, ast_end_token(node).pos);
+}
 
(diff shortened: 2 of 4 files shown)


gingerBill:
The work looks really good so far, if this fixes a lot of the issues, A LOT OF PEOPLE will be very very happy.

JesseRMeyer:
This is a marked improvement for #1858.  However, the behavior is not perfect, as shown below.



`i`'s value tracking gets busted during the test condition check, but is restored on the assignment.

graphitemaster:
First of all, thank you @fabiansperber I'm very happy to see some progress being made on debug support. I think this is a very good start. I've pulled these changes and tested locally and can confirm this fixes some issues.

I'd like to extend a hand here. If there's anything you need help with or would like to combine forces and work on this together please don't hesitate to ask. I'd like to improve upon the debug information as well. I have some time off work to work on personal projects and improving debug support would greatly benefit me and it seems you have some LLVM experience.

Have you at all run into miscompilation issues with `-debug`. Those are primarily my focus right now, not exactly getting debug support to be 100% but just having it not break things would be huge itself. Unfortunately these changes don't help with that right now but that's okay, as these are already quite helpful on some smaller codebases I have written in Odin.

gingerBill:
I've implemented a lot of what you've done! (I did it since this was still a draft)

fabiansperber:
I've rebased my local branch with some more changes. The tracking of the loop variable should work now, @JesseRMeyer

Also fixed having all different variable-variants of a type-switch visible at the same time. Note that the typed variable is not available on the case statement itself, only on the first actual statement in that block - somehow llvm does still not like the debug info enough to do that
```
switch actual in a {
  case u64:   // actual is not available here :(
    fmt.println("u64") // actual is available here
  case string:
    fmt.println("string")
}
```

fabiansperber:
Also sorry for the late response, @graphitemaster, I did not have any miscompilations per se due to `-debug`, only the debug info not working correctly, e.g. being at stepping through wrong locations or "variables optimized out".

My little LLVM experience is only from looking at other compilers use LLVM and looking at the LLVM IR / comparing to clang's output. It's not too difficult to look at small examples in both odin and C and compare the IR output to see what's different and possibly wrong. Or as in some cases of this MR to just check if what is being generated, does make sense.

As for helping out, I see that there is a list of debug problems in #1877 which I will take a look at, when I have some free time. Also you can just ping me on discord as well ;)

PR #1885, opened 2022-07-13
Fix using defer in the for loop's init, condition, or post

fabiansperber:
Test file to show when which defer call happens:

```
package issue

import "core:fmt"

init_counter := 0
check_counter := 0
inc_counter := 0
loop_counter := 0

increment_counter :: proc (_: int, c: ^int, m: string) {
  c^ += 1
  fmt.println("inc:", m, c^)
}

@(deferred_in=increment_counter)
wrap :: proc (v: int, _: ^int, m: string) -> int {
  fmt.println("wrap called with", m)
  return v
}

main :: proc() {
  //for i := 0; i < 3; i +=1 do wrap(0, &loop_counter, "D")
  for i := wrap(0, &init_counter, "A"); i < wrap(3, &check_counter, "B"); i += wrap(1, &inc_counter, "C") {
    wrap(0, &loop_counter, "D")
    fmt.println("loop ------")
  }
  
  fmt.println("init_counter: ", init_counter)
  fmt.println("check_counter: ", check_counter)
  fmt.println("inc_counter: ", inc_counter)
  fmt.println("loop_counter: ", loop_counter)
}
```

Wrong output on master (bfb082cda4):
```
wrap called with A
wrap called with B
wrap called with D
loop ------
inc: D 1
inc: B 1
inc: A 1            <- should not be here at all
wrap called with C  <- deferred statement from condition was already executed at this point, IMO should be after the increment's deferred statement
                <- also missing C's deferred statement
wrap called with B
wrap called with D
loop ------
inc: D 2
inc: B 2
inc: A 2
wrap called with C
wrap called with B
wrap called with D
loop ------
inc: D 3
inc: B 3
inc: A 3
wrap called with C
wrap called with B <- missing deferred statement for B when the loop ends
init_counter:  3
check_counter:  3
inc_counter:  0
loop_counter:  3
inc: C 1    <- one deferred increment happening at the end of main but shouldn't escape for's scope
```
Expected output/output with this change:
```
wrap called with A
wrap called with B
wrap called with D
loop ------
inc: D 1
wrap called with C
inc: C 1
inc: B 1
wrap called with B
wrap called with D
loop ------
inc: D 2
wrap called with C
inc: C 2
inc: B 2
wrap called with B
wrap called with D
loop ------
inc: D 3
wrap called with C
inc: C 3
inc: B 3
wrap called with B
inc: B 4
inc: A 1
init_counter:  1
check_counter:  4
inc_counter:  3
loop_counter:  3
```

### DIFF
--- a/src/llvm_backend_stmt.cpp
+++ b/src/llvm_backend_stmt.cpp
@@ -216,10 +216,7 @@ void lb_open_scope(lbProcedure *p, Scope *s) {
 			if (ast_file != nullptr) {
 				file = lb_get_llvm_metadata(m, ast_file);
 			}
-			LLVMMetadataRef scope = nullptr;
-			if (p->scope_stack.count > 0) {
-				scope = lb_get_llvm_metadata(m, p->scope_stack[p->scope_stack.count-1]);
-			}
+			LLVMMetadataRef scope = lb_get_current_debug_scope(p);
 			if (scope == nullptr) {
 				scope = lb_get_llvm_metadata(m, p);
 			}
@@ -1722,6 +1719,8 @@ void lb_build_if_stmt(lbProcedure *p, Ast *node) {
 	lb_start_block(p, done);
 }
 
+bool contains_deferred_call(Ast *node);
+
 void lb_build_for_stmt(lbProcedure *p, Ast *node) {
 	ast_node(fs, ForStmt, node);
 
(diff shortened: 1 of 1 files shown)


gingerBill:
I'd rather disallow deferred attributes within for loops than try and "fix" them.

PR #1896, opened 2022-07-22
Support for Quaternion marshal unmarshal json

sumofat:
At this time Quaternions are not supported in marshalling and unmarshalling json.

This PR hopes to alleviate that issue as it is  a very simple addition that is helpful.

Serializing any kind of 3d or 2d in quaternions is quite common therefor I propose this PR to help with that and any other use case that this may be beneficial.

### DIFF
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -116,7 +116,23 @@ marshal_to_writer :: proc(w: io.Writer, v: any) -> (err: Marshal_Error) {
 		io.write_byte(w, ']')    or_return
 
 	case runtime.Type_Info_Quaternion:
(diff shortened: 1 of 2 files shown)


gingerBill:
I will not accept this purely due to the issues of memory layout of quaternions. There are two common possibilities which could be written:

* xyzw (imaginary parts followed by real part)
* wxyz (real part followed by imaginary parts)

Because of the two possibilities without any specific annotation, I will not allow this.

sumofat:
Which I understand why if there are two memory layouts but if in umarshal I converted the data to a quat than assigned it via .x .y notation than 
this would still preserve the memory order correct?

Issue #1882, opened 2022-07-11
SIMD extraction printing strange error vs similar code success


CircArgs:
## Context
```odin
package test

import "core:fmt"
import "core:simd"

WECTOR_WIDTH :: 8
Vector3 :: #simd[4]f64

Wector3_Lane :: #simd[WECTOR_WIDTH]f64

Wector3 :: struct {
    x, y, z: Wector3_Lane,
}

main :: proc() {
    x := simd.extract(Wector3{x = {1}}.y, 0)
    fmt.print(x)
}
```
## `odin report` output:
        Odin: dev-2022-07-nightly:1676c643
        OS:   Ubuntu 20.04.4 LTS, Linux 5.4.0-121-generic
        CPU:  Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz
        RAM:  15935 MiB

## Expected Behavior

Expected to print `0`

## Current Behavior
```src/llvm_backend_const.cpp(874): Assertion Failure: `index_tav.mode == Addressing_Constant` 
Illegal instruction (core dumped)```

## Failure Information (for bugs)

Changing to `x := simd.extract(Wector3{x = {1}}.x, 0)` works

```odin
main :: proc() {
    wec := Wector3{x = {1}}
    x := simd.extract(wec.y, 0)
    fmt.print(x)
}
``` 
also works

Issue #1834, opened 2022-06-07
simd.from_array and .to_array causes compiler panic (Linux/macOS)


kururen:
## Context

Odin: dev-2022-06-nightly:ba5f7c4e
OS:   Fedora Linux 36 (Workstation Edition), Linux 5.17.9-300.fc36.x86_64
CPU:  Intel(R) Core(TM) i5-4300U CPU @ 1.90GHz
RAM:  7850 MiB

Using from_array or to_array with integer and boolean types causes the compiler to panic with the following error message:
`src/llvm_abi.cpp(813): Panic: Unhandled integer width for vector type`

For float types, the message is different:
`src/array.cpp(13): Assertion Failure: '0 <= index && index < count' Index 2 is out of bounds ranges 0..<2`

## Expected Behavior

Both methods successfully convert from a #simd type to a regular array.

## Current Behavior

Compiler panic / asserts

## Failure Information (for bugs)
Minimal example:
```
package bug

import "core:simd"

main :: proc() {
    data: #simd[64]u8
    arr := simd.to_array(data)
}
```
or
```
package bug

import "core:simd"

main :: proc() {
    arr: [64]u8
    data := simd.from_array(data)
}
```
respectively.

The problem seems to be related to the polymetric parameters used by both procedures, since using transmute(...) directly does not result in a panic. It is independent from the array size used. This also happened while testing on macOS (Catalina).

CircArgs:
Similar error with 

Code:
```Odin
add_vectors :: proc(a, b: simd.f64x4) -> simd.f64x4 {
	return a + b
}

main :: proc() {
	x := add_vectors(simd.f64x4{1, 2, 3, 4}, simd.f64x4{1, 2, 3, 4})
	fmt.print(x)
}
```

Error:
```
src/array.cpp(13): Assertion Failure: `0 <= index && index < count` Index 2 is out of bounds ranges 0..<2
Illegal instruction (core dumped)
```

CircArgs:
Not a real clue what the implications of this are, but my little example above works fine without the lines in `src/array.cpp` e.g. commenting out:

```cpp
  T &operator[](isize index) {
    // #if !defined(NO_ARRAY_BOUNDS_CHECK)
    // 		GB_ASSERT_MSG(0 <= index && index < count, "Index %td is out of bounds
    // ranges 0..<%td", index, count); 	#endif
    return data[index];
  }
```

Tetralux:
[...]

PR #1899, opened 2022-07-23
Remove clamp from vector sizing: 1834, 1882

CircArgs:
I believe this is a potential fix for https://github.com/odin-lang/Odin/issues/1834, https://github.com/odin-lang/Odin/issues/1882

### DIFF
--- a/src/llvm_abi.cpp
+++ b/src/llvm_abi.cpp
@@ -233,7 +233,7 @@ i64 lb_sizeof(LLVMTypeRef type) {
 			i64 elem_size = lb_sizeof(elem);
 			i64 count = LLVMGetVectorSize(type);
 			i64 size = count * elem_size;
-			return gb_clamp(next_pow2(size), 1, build_context.max_align);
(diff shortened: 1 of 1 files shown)


gingerBill:
The issue is that I believe vector types don't necessarily have the same alignment as their size (rounded up to the nearest power of two). Restricting it to the maximum (likely) alignment on that platform makes the most sense.

CircArgs:
@CircArgs thanks for taking a look! I opened this based on the latest Master https://github.com/odin-lang/Odin/issues/1906.

Some notes for hat motivated the change of this PR in case it could help at all:
It seemed like the `data` field of `Array` should be the same length as the vector and [here](https://github.com/odin-lang/Odin/blob/f174c805a9bc9ff984c89789776f3f2b2fd831d5/src/llvm_abi.cpp#L572) because of that `lb_sizeof` clamp you end up with less (2 in my case when I guess I expected 4 for `f64x4` since my max alignment is 16)

PR #1447, opened 2022-01-24
Added vendor glfw support for Mac OS

phillvancejr:
Updated bindings and as per Ginger Bill's request added the mac specific libraries to a subdirectory in glfw/lib

PR #1916, opened 2022-07-28
GLFW 3.3.8 Darwin universal and Windows static libs

wjlroe:
This swaps out the intel-only static libs for GLFW on Darwin (which breaks builds on Apple Silicon) for the universal static lib (containing both intel and arm64 libraries).

This also updates the Windows static libs and DLLs to the same version, the latest version: GLFW 3.3.8

This fixes builds using GLFW on M1 and M2 Macs after #1447


Kelimion:
I don't think this will work on Windows. Looking at the files changed, I don't see `bindings.odin`.

The PR appears to delete `vendor/glfw/lib/glfw3_mt.lib`, `vendor/glfw/lib/glfw3.lib`, `vendor/glfw/lib/glfw3.dll` and replace them with a single file `vendor/glfw/lib/glfw3dll.lib`, however, the imports are not updated to point to this new library.

wjlroe:
> I don't think this will work on Windows. Looking at the files changed, I don't see `bindings.odin`.
> 
> The PR appears to delete `vendor/glfw/lib/glfw3_mt.lib`, `vendor/glfw/lib/glfw3.lib`, `vendor/glfw/lib/glfw3.dll` and replace them with a single file `vendor/glfw/lib/glfw3dll.lib`, however, the imports are not updated to point to this new library.

That's not what's happened - I replaced one releases files with another. Github is rendering the difference in file sizes in the most misleading way possible. Each of those files has been reduced by about 2% and it shows that in this UI as 98% deleted! 

Here's another way to look at that commit:

Screenshot 2022-07-28 at 15 52 47

Kelimion:
Oh, wow. Now you mention it. Misleading UI, indeed!

PR #1919, opened 2022-08-01
Update dxgi.odin

Hyp-X:
Fixed the last parameters of `CreateDXGIFactory*` procedures to be `^rawptr` instead of `rawptr`
The original functions have `void **` parameter
Note: 
Also fixed [DXGIGetDebugInterface1](https://docs.microsoft.com/en-us/windows/win32/api/dxgi1_3/nf-dxgi1_3-dxgigetdebuginterface1)
The original last parameter is `void **pDebug` . The naming confusion is MS's fault.


Hyp-X:
Ooops, this my first time doing this and it looks I screwed up. This PR should be just on file

PR #1927, opened 2022-08-03
Debug symbols for range-based for loop over number range

IanLilleyT:
Before this PR the debugger showed `0` for `val` and `i` at every step of the loop. Now it shows `0` `1` `2`.

```
sum := 0
for val, i in 0 ..< 3 {
	sum += val
}
```

This PR doesn't fix debug symbols for the other kinds of range-based for loops. Those looked more difficult to fix but I think it's the same underlying problem.

I tested with `cppvsdbg` in vscode.

### DIFF
--- a/src/llvm_backend_stmt.cpp
+++ b/src/llvm_backend_stmt.cpp
@@ -481,7 +481,7 @@ void lb_build_range_interval(lbProcedure *p, AstBinaryExpr *node,
 	lbAddr index;
 	if (val1_type != nullptr) {
 		Entity *e = entity_of_node(rs->vals[1]);
-		index = lb_add_local(p, t_int, e, false);
+		index = lb_add_local(p, val1_type, e, false);
 	} else {
 		index = lb_add_local_generated(p, t_int, false);
 	}
--- a/src/llvm_backend_stmt.cpp
+++ b/src/llvm_backend_stmt.cpp
@@ -472,9 +472,9 @@ void lb_build_range_interval(lbProcedure *p, AstBinaryExpr *node,
 	lbAddr value;
 	if (val0_type != nullptr) {
 		Entity *e = entity_of_node(rs->vals[0]);
-		value = lb_add_local(p, val0_type ? val0_type : lower.type, e, false);
+		value = lb_add_local(p, val0_type, e, false);
 	} else {
-		value = lb_add_local_generated(p, val0_type ? val0_type : lower.type, false);
+		value = lb_add_local_generated(p, lower.type, false);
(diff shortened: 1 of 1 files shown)


src/llvm_backend_stmt.cpp
```diff
@@ -478,10 +469,22 @@ void lb_build_range_interval(lbProcedure *p, AstBinaryExpr *node,
 	lbValue lower = lb_build_expr(p, node->left);
 	lbValue upper = {}; // initialized each time in the loop
 
-	lbAddr value = lb_add_local_generated(p, val0_type ? val0_type : lower.type, false);
+	lbAddr value;
+	if (val0_type != nullptr) {
+		Entity *e = entity_of_node(rs->vals[0]);
+		value = lb_add_local(p, val0_type ? val0_type : lower.type, e, false);
```
gingerBill, line 475: Why is `val0_type` in a ternary when the above if statement checks for it?

IanLilleyT: oops! fixed

IanLilleyT:
one thing I've been meaning to ask is when would `val0_type` or `val1_type` be defined? Is it possible to initialize the value some other way?

src/llvm_backend_stmt.cpp
```diff
@@ -478,10 +469,22 @@ void lb_build_range_interval(lbProcedure *p, AstBinaryExpr *node,
...
+	} else {
+		value = lb_add_local_generated(p, lower.type, false);
+	}
 	lb_addr_store(p, value, lower);
 
-	lbAddr index = lb_add_local_generated(p, t_int, false);
+	lbAddr index;
+	if (val1_type != nullptr) {
+		Entity *e = entity_of_node(rs->vals[1]);
+		index = lb_add_local(p, t_int, e, false);
```
gingerBill, line 484: This should be `e->type` not `t_int`

IanLilleyT: all set

gingerBill: Still wrong because you still have `lb_add_local_generated(p, t_int, false)`

IanLilleyT: I'm confused... Line 486 from the diff? Wouldn't it need a fallback type when there's no entity for the iteration value?

gingerBill: Sorry, I read it wrong 😆

IanLilleyT: all good 👍

Issue #1924, opened 2022-08-03
math.log2 returns truncated result


slembcke:
## Context
	Odin: dev-2022-08-nightly:73beed04
	OS:   Pop!_OS 22.04 LTS, Linux 5.18.10-76051810-generic
	CPU:  AMD Ryzen 7 3700X 8-Core Processor             
	RAM:  15927 MiB

## Expected Behavior

`math.log2()` should compute the base 2 logarithm of a number. Ex: `math.log2(f32(10))` should return approximately 3.32.

## Current Behavior

Calling `math.log2()` always returns a truncated integral result. Ex: `math.log2(f32(10))` returns 3.00.
[In the source](https://github.com/odin-lang/Odin/blob/master/core/math/math.odin#L188) it's an alias for `logb()` which is very unexpected since that's a distinct function in C. I believe this was made in error, since in every other language I know of (including C), `log2()` returns the full result.

blakeb211:
request to close issue

PR #1932, opened 2022-08-06
fixed math.log2

blakeb211:
### What
- Change math:log2 proc, which called logb() and returned truncated result, to call log, log_f16, log_f32, etc instead. Now it returns the correct full floating-point result.

 ### Why
 - Issue opened: math.log2 returns truncated result #1924 

### Testing
- I have tested this on the Issue's case, log2(f16(10)), log2(f32(10)), log2(f64(10)) and it returns 3.32 ...

### DIFF
--- a/core/math/math.odin
+++ b/core/math/math.odin
@@ -185,16 +185,23 @@ log       :: proc{
 	log_f64, log_f64le, log_f64be,
 }
 
-log2_f16   :: logb_f16
-log2_f16le :: logb_f16le
-log2_f16be :: logb_f16be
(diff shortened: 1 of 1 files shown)


slembcke:
Looks like it does ln(x)/ln(base) now. I kinda assume there aren't perf or precision issues nowadays?
https://github.com/odin-lang/Odin/blob/7b539e3025907ab5b40762f90ca4f613cee7eab5/core/math/math.odin#L179

Kelimion:
ln(2) could always be made into a constant for the log2 case, if need be.

PR #1933, opened 2022-08-06
Removed use of deprecated functions. Cleaned up most deprecated use o…

lerno:
…f LLVMGetElementType.

Please do a solid regression testing on this. I've tested it with the sample program and tried to verify my changes, but there are a lot of different places I had to update, so please check carefully before merging this.

### DIFF
--- a/src/llvm_backend_type.cpp
+++ b/src/llvm_backend_type.cpp
@@ -104,7 +104,8 @@ lbValue lb_type_info(lbModule *m, Type *type) {
 	};
 
 	lbValue value = {};
-	value.value = LLVMConstGEP(lb_global_type_info_data_ptr(m).value, indices, gb_count_of(indices));
+	lbValue data = lb_global_type_info_data_ptr(m);
+	value.value = LLVMConstGEP2(lb_type(m, type_deref(data.type)), data.value, indices, gb_count_of(indices));
 	value.type = t_type_info_ptr;
 	return value;
 }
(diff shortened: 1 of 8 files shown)


src/llvm_backend_type.cpp
```diff
@@ -123,10 +124,16 @@ lbValue lb_get_type_info_ptr(lbModule *m, Type *type) {
...
 	lbValue res = {};
 	res.type = t_type_info_ptr;
-	res.value = LLVMConstGEP(lb_global_type_info_data_ptr(m).value, indices, cast(unsigned)gb_count_of(indices));
+	lbValue data = lb_global_type_info_data_ptr(m);
+	res.value = LLVMConstGEP2(lb_type(m, type_deref(data.type)), data.value, indices, cast(unsigned)gb_count_of(indices));
 	return res;
 }
 
+// The use of this method needs to be eliminated.
+LLVMTypeRef lb_llvm_get_pointer_type(LLVMTypeRef type)
```
gingerBill, line 133: Why isn't this just called `llvm_get_element_type` rather than `LLVMGetElementType`?

lerno: Because I want to separate the valid use of LLVMGetElementType (using it on arrays, vectors and matrices) vs the deprecated use (using it on pointers). So this one is the one I have to remove all uses of.

gingerBill: I've already removed most use cases of it already!

PR #1935, opened 2022-08-07
Removed windows 8 from microsoft_craziness.h

IanLilleyT:
Ever since https://github.com/odin-lang/Odin/pull/478 the Windows 8 and 8.1 development kits have not worked because it looks for the path `C:\Program Files (x86)\Windows Kits\8.1\Lib\winv6.3\ucrt` which does not exist, and errors with `LINK : fatal error LNK1104: cannot open file 'libucrt.lib'`.

This can be repro'd by going to https://developer.microsoft.com/en-us/windows/downloads/sdk-archive/ and downloading the `Windows 8.1 SDK` and commenting out the code block under `// Look for a Windows 10 entry`, forcing it to get paths from the Windows 8 SDK.

Microsoft is [dropping Windows 8.1 support ](https://support.microsoft.com/en-us/office/windows-8-and-windows-8-1-end-of-support-and-office-34e28be4-1e4f-4928-b210-3f45d8215595) on January 10, 2023. The Visual Studio Installer doesn't even list the Window 8.1 SDK. Also, since no one has mentioned Windows 8 build problems in the two years since the original PR, I think it's safe to assume that no one who's building the compiler is using Windows 8.

### DIFF
--- a/src/microsoft_craziness.h
+++ b/src/microsoft_craziness.h
@@ -220,7 +220,7 @@ struct DECLSPEC_UUID("42843719-DB4C-46C2-8E7C-64F1816EFD5B") DECLSPEC_NOVTABLE I
 
 // The beginning of the actual code that does things.
 struct Version_Data_Utf8 {
-	i32 best_version[4];  // For Windows 8 versions, only two of these numbers are used.
+	i32 best_version[4];
 	String best_name;
 };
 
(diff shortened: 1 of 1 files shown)


IanLilleyT:
Obsoleted by https://github.com/odin-lang/Odin/pull/1937

PR #1937, opened 2022-08-07
Find rc.exe in Windows SDK

IanLilleyT:
This PR finds the location of `rc.exe` in the Windows SDK. Before it would error when passing `-resource:icon.rc`

It includes changes from https://github.com/odin-lang/Odin/pull/1935 so that should be merged first. Or close that one and only review this. Whatever is easier.

Along the way I found several bugs in the environment variable parsing and ended up refactoring a good portion of `microsoft_craziness.h`

To test:
- download [icon.zip](https://github.com/odin-lang/Odin/files/9277669/icon.zip)
- run `odin build main.odin -file -resource:icon.rc`

Before: 
```
Failed to execute command: "rc.exe" ...
```
After:

image

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -204,7 +204,7 @@ enum BuildPath : u8 {
 	BuildPath_Main_Package,     // Input  Path to the package directory (or file) we're building.
 	BuildPath_RC,               // Input  Path for .rc  file, can be set with `-resource:`.
 	BuildPath_RES,              // Output Path for .res file, generated from previous.
-	BuildPath_Win_SDK_Root,     // windows_sdk_root
+	BuildPath_Win_SDK_Bin_Path, // windows_sdk_bin_path
 	BuildPath_Win_SDK_UM_Lib,   // windows_sdk_um_library_path
 	BuildPath_Win_SDK_UCRT_Lib, // windows_sdk_ucrt_library_path
 	BuildPath_VS_EXE,           // vs_exe_path
@@ -1336,7 +1336,7 @@ bool init_build_paths(String init_filename) {
 
 		if ((bc->command_kind & Command__does_build) && (!bc->ignore_microsoft_magic)) {
 			// NOTE(ic): It would be nice to extend this so that we could specify the Visual Studio version that we want instead of defaulting to the latest.
-			Find_Result_Utf8 find_result = find_visual_studio_and_windows_sdk_utf8();
+			Find_Result find_result = find_visual_studio_and_windows_sdk();
 			defer (mc_free_all());
 
 			if (find_result.windows_sdk_version == 0) {
@@ -1357,8 +1357,8 @@ bool init_build_paths(String init_filename) {
 			if (find_result.windows_sdk_um_library_path.len > 0) {
 				GB_ASSERT(find_result.windows_sdk_ucrt_library_path.len > 0);
 
-				if (find_result.windows_sdk_root.len > 0) {
-					bc->build_paths[BuildPath_Win_SDK_Root]     = path_from_string(ha, find_result.windows_sdk_root);
+				if (find_result.windows_sdk_bin_path.len > 0) {
+					bc->build_paths[BuildPath_Win_SDK_Bin_Path]  = path_from_string(ha, find_result.windows_sdk_bin_path);
 				}
 
 				if (find_result.windows_sdk_um_library_path.len > 0) {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -283,6 +283,9 @@ i32 linker_stage(lbGenerator *gen) {
 			String vs_exe_path = path_to_string(heap_allocator(), build_context.build_paths[BuildPath_VS_EXE]);
 			defer (gb_free(heap_allocator(), vs_exe_path.text));
 
+			String windows_sdk_bin_path = path_to_string(heap_allocator(), build_context.build_paths[BuildPath_Win_SDK_Bin_Path]);
+			defer (gb_free(heap_allocator(), windows_sdk_bin_path.text));
+
 			char const *subsystem_str = build_context.use_subsystem_windows ? "WINDOWS" : "CONSOLE";
 			if (!build_context.use_lld) { // msvc
 				if (build_context.has_resource) {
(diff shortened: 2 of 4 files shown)


Kelimion:
I can't reproduce your bug.

After adding an icon file and fixing `icon.rc` to point to it, your example compiles fine for me as-is, and results with an .exe file with embedded icon as expected.

Can you tell us the output of `odin report`?
```
W:\Odin-other\Odin-test>odin build main.odin -file -resource:icon.rc
W:\Odin-other\Odin-test>
```

IanLilleyT:
Are you running odin from a developer command prompt? I'm not.

```
Odin: dev-2022-08:659c3c52
OS:   Windows 11 Professional (version: 21H2), build 22000.795
CPU:  Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
RAM:  32605 MiB
```

Kelimion:
I am running it from a developer prompt. That you are not explains the problem you're running into.

IanLilleyT:
It's analogous to running `rc.exe` from a regular command prompt. It says:

```
'rc.exe' is not recognized as an internal or external command, operable program or batch file.
```

Likewise, calling `rc.exe` from the code without `windows_sdk_bin_path` in front of it can't find the program.

Kelimion:
> It's analogous to running `rc.exe` from a regular command prompt. It says:
> 
> ```
> 'rc.exe' is not recognized as an internal or external command, operable program or batch file.
> ```
> 
> Likewise, calling `rc.exe` from the code without `windows_sdk_bin_path` in front of it can't find the program.

Right, so it's not a bug at all. The only thing we could do here is have `odin` report the non-existence of `rc.exe` same as it does for other parts of the SDK.

Typing `cl` or `link` from a regular prompt also isn't going to work.

IanLilleyT:
The difference is `cl` and `link` are called during compilation where you are expected to be in a developer command prompt

`rc.exe` is called from the binary where there's no expectation of that. Users would have to download visual studio and work inside a developer command prompt to be able to use the `-resource` flag

Kelimion:
Odin calls `link` as well, and therefore expects to be run from a developer prompt. If it can't find `link`, that it can't find `rc` is inconsequential.

IanLilleyT:
unlike `rc`, `link` is already preceded with its path, so there's no problem finding `link` when running outside the developer command prompt, as long as the compiler is compiled with the same windows sdk as your system.

Kelimion:
Managed to replicate and verify the fix. Nice work.

For future reference, it would've been helpful if the refactor and fix each had their own dedicated commit instead of being squashed into one.

IanLilleyT:
Thank you, I will keep that in mind next time

krixano:
> Odin calls link as well, and therefore expects to be run from a developer prompt. If it can't find link, that it can't find rc is inconsequential.

I've never needed to run Odin from the developer prompt in years. You only need to do it for building Odin (which honestly shouldn't need to be done considering most other projects' build systems are able to find the files as well). Any decent compiler for Windows would automatically find the necessarily VS libraries and binaries. This would just be lazy programming if it couldn't.

IanLilleyT:
> but I can't say that's the weirdest thing within Odin's community

Like the way I was treated when I asked if the PR could be reopened: https://discord.com/channels/568138951836172421/568146055409958913/1006319433217753178

Why bother working on PRs when you get a reaction like that?

gingerBill:
@krixano 
> I've never needed to run Odin from the developer prompt in the years I've been using it.

Which is what all of this microsoft_craziness.h was about. And thank you @IanLilleyT for getting `rc.exe` to work. I apologize to you Ian on a lot of the reactions mainly due to all of the confusion between loads of different people. What was interesting is `rc.exe` is in the same directory as `link.exe` for me but doesn't appear to be for other people, which just adds to the insanity.


@krixano 
> Any decent compiler for Windows would automatically find the necessary VS libraries and binaries. This would just be lazy programming if it couldn't.

This is categorically false in that NONE of the visual studio compilers do this, and even languages like Rust states in their docs that you should `call vcvarsall.bat` since they don't have any of it preprogrammed. 

> I would have expected a person who is working on the programming language to have known this, but I can't say that's the weirdest thing within Odin's community.

I know how the Visual Studio toolchain works and related tools. Please don't be condescending to others.

IanLilleyT:
Thank you @gingerBill

Regarding baking paths in the compiler - it becomes a problem when you ship binaries and the user doesn't have the same exact SDK version. An option is to search for the visual studio paths at runtime (i.e. when it's compiling your odin files) when the baked-in path doesn't exist on the user's system. Of course, this has performance / ease-of-use / trickier code tradeoffs

Kelimion:
As an aside, you can use `-show-debug-messages` and Odin will (among other things) print out the paths it found/constructed.

IanLilleyT:
Oh you're right! My mistake

Kelimion:
On a separate but related note, @gingerBill:

Do you think the use case for `-resource` is likely to remain limited to injecting a `.ico` payload, and won't end up being used to inject animated cursors, binary payloads, translation strings and version numbers as also supported by `.rc`/`.res`?

If so, it wouldn't take much to write a bit of code to verify a `.ico` file is valid, turn it into a `.res` file containing it, and leave out the need for `rc.exe` entirely. Given we've established that MS likes to move it around, adding a dedicated `-icon` option and deprecating `-resource` would make things less likely to break if some future SDK decides to complicate matters further.

Ekopalypse:
For me, I can say that I currently use it for version information and wanted to embed dialogs in the future.

Kelimion:
> For me, I can say that I currently use it for version information and wanted to embed dialogs in the future.
That's good to know, thanks.

IanLilleyT:
I love the idea of an `-icon` option in addition to the `-resource` flag. To explore the idea further, how about taking a regular image file (bmp, png, etc) and doing mipmaps and converting to os-specific format under the hood?

Problems with this idea:
- Which image formats get support? Is it worth bundling image codecs in the compiler?
- If you want custom designs for different sizes the compiler flag gets more complicated. Once the compiler is self-hosting and exposed as a library this becomes more straightforward, but that's a ways off.

PR #1939, opened 2022-08-10
fix #1840 invalid LLVM code gen for arithmetics between f32 and shifted untyped integer

### LINKED ISSUE #1840: invalid LLVM code gen for arithmetics between `f32` and shifted untyped integer
## Context

Odin: dev-2022-06-nightly:ba5f7c4e
OS:   Windows 10 Unknown Edition (00000063) (version: 21H1), build 19043.1706
CPU:  Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
RAM:  16300 MiB

## Expected Behavior

The following code snippet should either compile or fail with proper error messages:
```odin
package test
main :: proc() {
    r := u32(3)
    a := f32(1.0)
    x := a * (1<<r)
}
```

## Current Behavior

Compilation fails with invalid LLVM code gen.

Changing the last line to `x := a * f32(1<<r)` also triggers invalid code gen.

But this works: `x := a * f32(u32(1)<<r)`

### Steps to Reproduce

compile and run the code snippet above

### Failure Logs

```text
>odin run test.odin -file
LLVM CODE GEN FAILED FOR PROCEDURE: test.main
define internal void @test.main(i8* noalias nocapture nonnull %__.context_ptr) {
decls:
  %0 = alloca i32, align 4
  %1 = alloca float, align 4
  %2 = alloca float, align 4
  br label %entry

  %3 = bitcast i8* %__.context_ptr to %runtime.Context*
  store i32 0, i32* %0, align 4
  store i32 3, i32* %0, align 4
  store float 0.000000e+00, float* %1, align 4
  store float 1.000000e+00, float* %1, align 4
  store float 0.000000e+00, float* %2, align 4
  %4 = load float, float* %1, align 4
  %5 = load i32, i32* %0, align 4
  %6 = uitofp i32 %5 to float
  %7 = icmp ult float %6, i0 32
  %8 = shl float 1.000000e+00, %6
  %9 = select i1 %7, float %8, float 0.000000e+00
  %10 = fmul float %4, %9
  store float %10, float* %2, align 4
  ret void
}

Both operands to ICmp instruction are not of the same type!
  %7 = icmp ult float %6, i0 32
Shifts only work with integral types!
  %8 = shl float 1.000000e+00, %6
```

jaspergeer:
Closes #1840 

Before this PR, shift expressions with untyped const LHS operands would result in invalid LLVM IR if their parent node had a non-integral type.

* converts untyped const LHS operands of shift expressions with non-integer typed parent nodes to `t_int`
* adds tests to tests/issues

This is my first PR and I'm new to the codebase so feedback is very much appreciated!

### DIFF
--- a/src/llvm_backend_expr.cpp
+++ b/src/llvm_backend_expr.cpp
@@ -1345,7 +1345,13 @@ lbValue lb_build_binary_expr(lbProcedure *p, Ast *expr) {
 	case Token_Shr: {
 		lbValue left, right;
 		Type *type = default_type(tv.type);
-		left = lb_build_expr(p, be->left);
+
+		if (lb_is_expr_untyped_const(be->left) && !is_type_integer(type)) {
+			type = t_int;
+			left = lb_expr_untyped_const_to_typed(p->module, be->left, t_int);
+		} else {
+			left = lb_build_expr(p, be->left);
+		}
 
 		if (lb_is_expr_untyped_const(be->right)) {
 			// NOTE(bill): RHS shift operands can still be untyped
--- a/tests/issues/run.bat
+++ b/tests/issues/run.bat
@@ -12,6 +12,9 @@ build\test_issue
 ..\..\odin build test_issue_1592.odin %COMMON% -file
 build\test_issue
 
+..\..\odin build test_issue_1840.odin %COMMON% -file
+build\test_issue
+
 @echo off
 
 rmdir /S /Q build
--- a/tests/issues/run.sh
+++ b/tests/issues/run.sh
@@ -13,6 +13,9 @@ $ODIN build test_issue_829.odin $COMMON -file
 $ODIN build test_issue_1592.odin $COMMON -file
 ./build/test_issue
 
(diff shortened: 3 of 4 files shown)


gingerBill:
This is more a front-end problem than a backend one. I'm not sure if this should be allowed to begin with or not.

jaspergeer:
I did experiment with addressing this case in the type checker but it resulted in an explicit cast being needed for arithmetic using shifted integers and floats. Since untyped RHS operands are converted to typed in the backend as well so I figured I'd put it there. What exactly would be the intended behavior in this case? Should the compiler throw an error or perform an implicit cast?

gingerBill:
An explicit cast might have been the correct solution here. Because what integer type should `1 << n` be before you cast? Depending on what that it is make a lot of difference as to what the conversion to the float type does.

jaspergeer:
Right, that's a good point. Correct me if I'm wrong but it looks like integer literals are assigned t_int in the absence of other type information so a shifted literal should also have type t_int?

PR #1944, opened 2022-08-11
Improvements to `#load`

gingerBill:
Numerous improvements to `#load` et al.

* Allow for `#load(path) or_else default`
* Deprecate `#load_or`
* Cache `#load` data
* Cache `#load_hash` hashes
* Allow for giving the data type: `#load(path, type)`, where `type` must be one of the following:
    * `string`
    * `[]T` where `T` must be a simple data type containing not (internal) pointers

### DIFF
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -1075,6 +1075,459 @@ bool check_builtin_simd_operation(CheckerContext *c, Operand *operand, Ast *call
 }
 
 
+bool check_builtin_procedure_directive(CheckerContext *c, Operand *operand, Ast *call, i32 id, Type *type_hint) {
+	ast_node(ce, CallExpr, call);
+	ast_node(bd, BasicDirective, ce->proc);
+	String name = bd->name.string;
+	if (name == "location") {
+		if (ce->args.count > 1) {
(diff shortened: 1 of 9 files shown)


ThisDrunkDane:
Any reason the load type can't just be `T`? where T is a basic type like `int64` or something? If you for some reason just wanna load in a number?

gingerBill:
I want to reduce the confusion of what the return type is. And if it really just a single number `#load(path, []i64)[0]` will work just fine

ThisDrunkDane:
It would, but a feels a bit nonsense you have to do that? Like What clarity do we really gain?

gingerBill:
I can already guess people will do `#load(path, i64)` and expect it to be a slice of `i64` rather than a single element since the default type will be `[]byte`

ThisDrunkDane:
Really? I would explicitly not think if it as a slice unless I put a slice in there. It would be very weird for me if it was a slice

ThisDrunkDane:
And if the default is note `byte` but `[]byte` then I would not think so?

gingerBill:
@ThisDrunkDane Because I've explicitly seen people write generic procedures which take the element type and return the slice time because that makes more sense to them in their mental framework. There was a very good reason I made `make` take the return type.

Issue #1968, opened 2022-08-21
BSD: unable to build due to WIFEXITED and WEXITSTATUS


ashemedai:
## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System & Odin Version: FreeBSD 12.3-STABLE, but also applies to NetBSD and OpenBSD
Tree at b3e3b6c6

## Expected Behavior

Successful Odin build.

## Current Behavior

Failed Odin build.

### Failure Logs

```
./build_odin.sh debug
+ clang++ src/main.cpp src/libtommath.cpp -Wno-switch -Wno-macro-redefined -Wno-unused-value '-DODIN_VERSION_RAW="dev-2022
-08"' '-DGIT_SHA="b3e3b6c6"' -std=c++14 -I/usr/local/llvm11/include -std=c++14 -fno-exceptions -D__STDC_CONSTANT_MACROS -D
__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -L/usr/local/llvm11/lib -g -pthread -lm -lstdc++ -lLLVM-11 -o odin
src/main.cpp:111:6: error: use of undeclared identifier 'WIFEXITED'
        if (WIFEXITED(exit_code)) {
            ^
src/main.cpp:112:15: error: use of undeclared identifier 'WEXITSTATUS'
                exit_code = WEXITSTATUS(exit_code);
                            ^
2 errors generated.
*** Error code 1
```

FreeBSD, NetBSD or OpenBSD do not know either `WIFEXITED` or `WEXITSTATUS` for `system()` (`waitpid()`) if `<sys/wait.h>` is not included.

This allows me to build without error on FreeBSD again, unsure if this is how it should be fixed though. Also unsure at the moment if it could be collapsed with the preceding section, I don't think it can hurt since `sys/types.h` and `sys/sysctl.h` are quite common includes on all three.

```patch
diff --git a/src/build_settings.cpp b/src/build_settings.cpp
index 9d5c3e55..bed8e7c9 100644
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -3,6 +3,10 @@
 #include <sys/sysctl.h>
 #endif

+#if defined(GB_SYSTEM_FREEBSD) || defined(GB_SYSTEM_NETBSD) || defined(GB_SYSTEM_OPENBSD)
+#include <sys/wait.h>
+#endif
+
[...]

PR #1969, opened 2022-08-22
fix: Expand OpenBSD include for wait.h to all Unix

ashemedai:
closes: #1968


semarie:
build (and run) fine on OpenBSD.

PR #1979, opened 2022-08-26
Fixed marshal_to_writer not supporting i128

Despacito696969:
Previously json.marshal(i128(696969)) would print 0 as there was no `u = u128(i)` initialization.


Despacito696969:
This is incomplete because there is also i128le and i128be which I forgot about. Please somebody just type that in and close this PR.

Issue #1985, opened 2022-08-29
-build-mode:obj overwrites -out: extension


nico-abram:
## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System & Odin Version: Windows 10 21H2 and dev-2022-08:3d4698d
* Please paste `odin report` output:

```
Where to find more information and get into contact when you encounter a bug:

        Website: https://odin-lang.org
        GitHub:  https://github.com/odin-lang/Odin/issues


Useful information to add to a bug report:

        Odin: dev-2022-08:3d4698d
        OS:   Windows 10 Home Basic (version: 21H2), build 19044.1889
        CPU:  AMD Ryzen 7 4800H with Radeon Graphics
        RAM:  15789 MiB

```

## Expected Behavior

Explicitly setting an output filename when generating an object file should use the given extension

## Current Behavior

Running `.\odin.exe build a.odin -file -build-mode:obj -out:a.s` generates `a.obj` instead of `a.s`

PR #1986, opened 2022-08-29
Don't overwrite extension when explicitly generating object files

nico-abram:
Fixes https://github.com/odin-lang/Odin/issues/1985


Kelimion:
Didn't compile. Issue fixed.

PR #1989, opened 2022-08-29
Add CMark bindings

Kelimion:


vendor/cmark/cmark.odin
```diff
@@ -0,0 +1,464 @@
...
+*/
+package cmark
+
+import "core:c"
+import "core:c/libc"
+
+BINDING_VERSION :: Version_Info{major = 0, minor = 30, patch = 2}
+
+when ODIN_OS == .Windows {
+	// To use `import "vendor:cmark"` and copy the `.dll` next to your program.
```
gingerBill, line 15: Any reason it has to be a DLL and cannot be statically linked?

Kelimion: I'll have a look at the nmake file to see it it can be built statically.

Kelimion:
You can now do this:
```odin
context.allocator = cm.get_default_mem_allocator_as_odin()

// Allocated on CMark's allocator
foo := make([dynamic]int, 42)
defer delete(foo)

for i in 0..<42 {
	foo[i] = i
}
fmt.println("Before resize:", foo)
resize(&foo, 13)
fmt.println("After resize:", foo)
```

PR #2005, opened 2022-09-01
have sort_with_indices allocate. Add a couple convenience procs for using the indices result to sort other slices.

thePHTest:
I need to add some tests for the sort_by_indices procs. Also, would love any suggestions for better naming.

### DIFF
--- a/tests/core/slice/test_core_slice.odin
+++ b/tests/core/slice/test_core_slice.odin
@@ -127,11 +127,13 @@ test_sort_by_indices :: proc(t: ^testing.T) {
 
 		{
 			indices := make([]int, test_size)
+			defer delete(indices)
 			for _, i in indices {
 				indices[i] = i
 			}
 
 			sorted_indices := slice.sort_by_indices(indices, f_idx)
+			defer delete(sorted_indices)
 			for v, i in sorted_indices {
 				idx_pass := v == f_idx[i]
 				expect(t, idx_pass, "Expected the sorted index to be the same as the result from sort_with_indices")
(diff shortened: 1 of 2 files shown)


Kelimion:
I just added tests for `slice.sort_with_indices`, Phil.

thePHTest:
@Kelimion I should add tests for the new sort_by_indices procs that can sort a slice given the indices that were the result of sort_with_indices.

Kelimion:
> @Kelimion I should add tests for the new sort_by_indices procs that can sort a slice given the indices that were the result of sort_with_indices.

Ah, I got you. Feel free to add them to `tests/core/slice`.

Kelimion:
Let's keep this PR open until those new procedures are tested as well, and then merge it?

thePHTest:
I added the tests. I could also add some more procs like `sort_by_cmp_with_indices` etc to make it exhaustive. Also, a sort_soa could be implemented fairly easily now. Or those could come later.

thePHTest:
and if you have a better name for the procs than `sort_by_indices` and `sort_By_indices_overwrite` let me know and I'm happy to change them,

thePHTest:
I suppose we could also add a `sort_by_indices_swap` that doesn't allocate or do the overwrite so that the caller can provide the slice to swap with.

tests/core/slice/test_core_slice.odin
```diff
@@ -94,4 +91,69 @@ test_sort_with_indices :: proc(t: ^testing.T) {
...
+
+		// Sort
+		f_idx := slice.sort_with_indices(vals)
+		defer delete(f_idx)
+
+		// Verify sorted test values
+		rand.init(&r, seed)
+
+		{
+			indices := make([]int, test_size)
```
Kelimion, line 129: Looks like you're no longer cleaning up `indices := make([]int, test_size)` here or in the next block?

PR #2025, opened 2022-09-05
Add windows.ShowCursor

englerj:


englerj:
I'm unsure what course of action to take here. It seems that the foreign declaration of ShowCursor from windows (User32) is colliding with ShowCursor from Raylib.

Lperlind:
I need ShowCursor too and would like to see some resolution to this PR.

Kelimion:
It is under discussion. What we'll likely have to do is one of the following:
- For Windows, dynamically import `ShowCursor` and wrap it to avoid the link conflict.
- For RayLib don't link with `ShowCursor` on Windows, and instead call the Windows version there.

PR #2036, opened 2022-09-09
add debug info for named procedure return values

nico-bb:
This is an attempt at adding debug info for named procedure return values.
It seems to work properly when I try it, but I would very much like someone to triple check that I didn't break anything.


JesseRMeyer:
This assumes all named procedure return values are returned as pointers, but register sized types are not.  As is, this PR will crash.  Prevent that with an `if`, at the cost of not generating debug info for some return types. 

```
c++
if (return_ptr_value.value != nullptr) {
	lb_add_debug_param_variable(p, return_ptr_value.value, e->type, e->token, result_index+1, p->decl_block);
}
```

PR #2040, opened 2022-09-10
Add windows.timeGetTime()

ryancsh:
Here is the windows documentation for it: https://docs.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timegettime


Kelimion:
What is motivation to use `timeGetTime` compared to `core:time`'s `now()`, which also gets the system time in larger precision?

ryancsh:
The motivation for this pull request is API completeness mainly.

Here's what's written on the [Odin homepage](https://odin-lang.org/): 
> Batteries Included. Odin comes with high quality packages out of the box in its [core](https://github.com/odin-lang/Odin/tree/master/core) library.

From the way it's written, I was assuming that providing as much of the Windows API as possible, which is under ```core:sys```, was something that aligned with the goals of the project.

At the moment, I don't have any specific use for ```timeGetTime``` in particular. I was merely testing different Windows-specific timing functions and found that one missing and so added it.

PR #2048, opened 2022-09-12
Use pow of two capacity for hash maps to allow for `& (n-1)` instead of `% n`

gingerBill:


core/runtime/dynamic_map_internal.odin
```diff
@@ -300,17 +302,36 @@ __dynamic_map_set :: proc(h: Map_Header, hash: Map_Hash, value: rawptr, loc := #
...
-		// index = __dynamic_map_find(h, hash).entry_index
-		// assert(index >= 0)
 	}
 	
 	return __dynamic_map_get_entry(h, index)
 }
 
 
+@(private="file")
+next_pow2 :: proc "contextless" (n: int) -> int {
```
Kelimion, line 312: This is a bit poorly named. If you pass it a power of two, it returns that same power of two, not the next one.

e.g.
```
next_pow2(001): 1
next_pow2(002): 2
next_pow2(003): 4
next_pow2(004): 4
next_pow2(005): 8
next_pow2(006): 8
next_pow2(007): 8
next_pow2(008): 8
next_pow2(009): 16
next_pow2(010): 16
next_pow2(011): 16
next_pow2(012): 16
next_pow2(013): 16
next_pow2(014): 16
next_pow2(015): 16
next_pow2(016): 16
next_pow2(017): 32
```

core/runtime/dynamic_map_internal.odin
```diff
@@ -300,17 +302,36 @@ __dynamic_map_set :: proc(h: Map_Header, hash: Map_Hash, value: rawptr, loc := #
...
+	n := n
+	if n <= 0 {
+		return 0
+	} else if n <= 2 {
+		return n
+	}
+	n -= 1
```
Kelimion, line 319: If replaced thusly, it does produce 16 for 8, 32 for 16, and so on:
```cpp
if intrinsics.count_ones(n) != 1 {
    n -= 1
}
```

Kelimion (commented):
Looks good to me in principle, but does it doesn't seem to actually grow if the capacity is already a power of two, because `next_pow2` returns the same power of two you input. I suggested a fix.

core/runtime/dynamic_map_internal.odin
Kelimion: Oh, I see. The `next_pow2` is only for the initial reservation, and it just doubles after that. I withdraw my objections.

core/runtime/dynamic_map_internal.odin
gingerBill: Good test!

gingerBill:
I forgot to say how much I love that this PR number was 2048, a power of two 😆

PR #2055, opened 2022-09-17
Map Internals Improvements

gingerBill:
Improve the internals of a map with the following:

* Change internal  map indices to use a distinct `uint` rather than just `int`
    * Uses explicit sentinel (`~Map_Index(0)` rather than use any negative number
* `__dynamic_map_get`  and `__dynamic_map_set` use only simple parameters (non-aggregate types) to simplify compiler internals
* Cache map header per value where possible to minimize stack usage
* Add `"contextless"` where possible to dynamic_map_internal.odin
* Update json/unmarshal.odin where needed
* Wrap `__dynamic_map_find` for certain cases

### DIFF
--- a/core/runtime/dynamic_map_internal.odin
+++ b/core/runtime/dynamic_map_internal.odin
@@ -205,7 +205,7 @@ __slice_resize :: proc(array_: ^$T/[]$E, new_count: int, allocator: Allocator, l
 	return false
 }
 
-__dynamic_map_reset_entries :: proc(using header: Map_Header, loc := #caller_location) {
+__dynamic_map_reset_entries :: proc "contextless" (using header: Map_Header, loc := #caller_location) {
 	for i in 0..<len(m.hashes) {
 		m.hashes[i] = MAP_SENTINEL
 	}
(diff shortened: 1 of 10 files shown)


core/runtime/dynamic_map_internal.odin
```diff
@@ -364,52 +373,53 @@ __dynamic_map_find :: proc(using h: Map_Header, hash: Map_Hash) -> Map_Find_Resu
 	return fr
 }
 
-__dynamic_map_add_entry :: proc(using h: Map_Header, hash: Map_Hash, loc := #caller_location) -> int {
-	prev := m.entries.len
-	c := __dynamic_array_append_nothing(&m.entries, entry_size, entry_align, loc)
+__dynamic_map_add_entry :: proc(using h: Map_Header, hash: Map_Hash, loc := #caller_location) -> Map_Index {
```
Kelimion, line 376: Should this also be `contextless`?

gingerBill: I am going to inline this procedure because it is used nowhere else

PR #2058, opened 2022-09-18
Fix darwin lib dir stbi

sumofat:
The darwin build libs were pointing to the wrong directory.


gingerBill:
None of those calls are part of the official C standard library

sumofat:
ok that was an accident

Issue #2073, opened 2022-09-22
Improve `odin test`.


Kelimion:
`odin test` presently prints `Tests failed: 1` (without a newline) on non-Windows platforms, instead of the more elaborate:
```
[Package: test]
[Test: trivial_test]
W:/Odin-other/Odin-test/test.odin(246:13): 1 does not equal 2.
[trivial_test : FAILURE]
```
In addition, on these platforms it also fails to print that assertion that 1 does not equal 2.

Lastly, let's allow `-out` to work for `odin test`.

PR #2078, opened 2022-09-23
Refactor tests to use default expect and log from core:testing

codename-irvin:
Refactored tests/core in the following ways:

- Remove that entire when ODIN_TEST { ... } else { ... } wrapper.
- Annotate any of the tests run by main previously with @test if they're not already annotated thusly.
- Update build.bat and Makefile under tests/core to use odin test.
- Remove superfluous code from tests/common


Kelimion:
There are a few problems to iron out with `odin test` that we're now running into by switching to it instead of `odin run`.

I'll try to address those together with the work on #2073.

Kelimion:
We'll have to park this PR for a bit until `odin test` has seen some maintenance rework. There are a few interactions with the CI that are hard to replicate locally, but it means that for the moment we'll have to stay with `odin run`.

PR #2093, opened 2022-09-28
Change llvm-copy-to-zip path to use real one, instead of link

flplv:


Kelimion:
image

This fails. Going to revert.

PR #2094, opened 2022-09-28
Revert "Change llvm-copy-to-zip path to use real one, instead of link"

Kelimion:
Reverts odin-lang/Odin#2093


flplv:
will fix.. thank

flplv:
this revert is not needed, this PR (https://github.com/odin-lang/Odin/pull/2096) fixes the issue... sorry about that

PR #2101, opened 2022-10-01
sys/windows: add a bunch of stuff

ftphikari:


core/sys/windows/advapi32.odin
```diff
@@ -128,4 +128,34 @@ foreign advapi32 {
 		lpData: LPCVOID,
 		cbData: DWORD,
 	) -> LSTATUS ---
+
+	GetFileSecurityA :: proc(
```
gingerBill, line 132: Please don't add the `A` versions, only the `W` versions please

PR #2117, opened 2022-10-08
Add glfw.GetCocoaWindow

janivanecky:


Kelimion:
To avoid this error, you'd have to break `native.odin` into OS-specific files.
e.g. `native_windows.odin` with `//+build windows` at the top,
`native_darwin.odin` + `//+build darwin`,
`native_linux.odin` + `//+build linux`.

janivanecky:
Thanks! Updated.

Kelimion:
You no longer need the `when ODIN_OS == .Darwin`, et al, because the `_darwin.odin` and `//+build darwin` already preclude those files from being processed on other operating systems. Looks good otherwise.

janivanecky:
Gotcha, updated.

PR #2120, opened 2022-10-09
Fix signature for `shouldTerminateAfterLastWindowClosed` delegate proc

jceipek:
The existing `shouldTerminateAfterLastWindowClosed` proc used in `ApplicationDelegate` declares no return value, which prevents programmers using Odin from choosing whether closing the last window in an NSApplication should terminate the application.

Per [Apple's documentation](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428381-applicationshouldterminateafterl?language=objc), `NSApplicationDelegate`'s `applicationShouldTerminateAfterLastWindowClosed` method should return a `BOOL` value:

```
- (BOOL)applicationShouldTerminateAfterLastWindowClosed:(NSApplication *)sender;
```

This PR fixes the signature discrepancy by using ~a `bool` signature.
NOTE: strictly speaking, BOOL is not equivalent to a C bool per https://developer.apple.com/documentation/objectivec/bool/, but I suspect that doesn't matter in this case.~
an `NS.BOOL` signature.

```
Odin: dev-2022-10-nightly:775c9648
OS:   macOS Monterey 12.5 (build: 21G115, kernel: 21.6.0)
CPU:  Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
RAM:  32768 MiB
```

### DIFF
--- a/vendor/darwin/Foundation/NSApplication.odin
+++ b/vendor/darwin/Foundation/NSApplication.odin
@@ -11,7 +11,7 @@ ActivationPolicy :: enum UInteger {
 ApplicationDelegate :: struct {
 	willFinishLaunching:                  proc "c" (self: ^ApplicationDelegate, notification: ^Notification),
 	didFinishLaunching:                   proc "c" (self: ^ApplicationDelegate, notification: ^Notification),
-	shouldTerminateAfterLastWindowClosed: proc "c" (self: ^ApplicationDelegate, sender: ^Application),
+	shouldTerminateAfterLastWindowClosed: proc "c" (self: ^ApplicationDelegate, sender: ^Application) -> bool,
 
 	user_data: rawptr,
 }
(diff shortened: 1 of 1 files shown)


Lperlind:
Why not use NS.BOOL?

https://github.com/odin-lang/Odin/blob/master/vendor/darwin/Foundation/NSTypes.odin#L19

jceipek:
I didn't know about `NS.BOOL`. It seems appropriate but might become annoying if https://github.com/odin-lang/Odin/blob/b6a65fac36ddcc2656e393be0e654679c2c4744f/vendor/darwin/Foundation/NSTypes.odin#L19 becomes `distinct` as hinted in the TODO.

Lperlind:
I wouldn't worry about the cast. It's very typical when interfacing with foreign code.
For example sys windows everything is marked 1:1 with the windows types.

jceipek:
I made the change in [63086c7](https://github.com/odin-lang/Odin/pull/2120/commits/63086c7eaf3c0e97b1e8991fc94e146fb4efac38)

PR #2124, opened 2022-10-12
`core:mem/virtual` - Unify `Static_Arena` and `Growing_Arena` into `Arena`

gingerBill:
Unify `Static_Arena` and `Growing_Arena` into `Arena`

`Arena` defaults to `.Growing`, with minimum block size defaults.

Can be initialized to be `.Growing` or `.Static` explicitly.


jrfondren:
Cobra noticed that this PR causes this program to segfault:

```odin
package main

import "core:fmt"
import "core:mem/virtual"


Tree :: struct {
	left, right: ^Tree,
}

main :: proc() {
	arena: virtual.Arena
	allocator := virtual.arena_allocator(&arena)
	defer virtual.arena_free_all(&arena)

	tree := new(Tree, allocator)
	fmt.println(tree == tree.left)
}
```

It segfaults as the allocation fails with memory out of error, despite (on examination of the arena) the allocator successfully allocating plenty of memory in its first block. `arena_init_static` gets the error from `alloc_from_memory_block`, which gets it from the first of the lines changed by this patch:

```
diff --git a/core/mem/virtual/virtual.odin b/core/mem/virtual/virtual.odin
index b209d06e..03690124 100644
--- a/core/mem/virtual/virtual.odin
+++ b/core/mem/virtual/virtual.odin
@@ -136,13 +136,13 @@ alloc_from_memory_block :: proc(block: ^Memory_Block, min_size, alignment: uint)
 	}
 
 	alignment_offset := calc_alignment_offset(block, uintptr(alignment))
-	size, size_ok := intrinsics.overflow_add(min_size, alignment_offset)
-	if !size_ok {
+	size, overflow := intrinsics.overflow_add(min_size, alignment_offset)
+	if overflow {
 		err = .Out_Of_Memory
 		return
 	}
 
-	if to_be_used, ok := intrinsics.overflow_add(block.used, size); !ok || to_be_used > block.reserved {
+	if to_be_used, overflow := intrinsics.overflow_add(block.used, size); overflow || to_be_used > block.reserved {
 		err = .Out_Of_Memory
 		return
 	}
```

As intrinics.overflow_add returns true <em>on overflow</em>, rather than returning an 'ok' boolean:

```odin
package main

import "core:fmt"
import "core:intrinsics"

main :: proc() {
	n : uint = 100
	sum, overflow := intrinsics.overflow_add(n, n)
	fmt.println(sum, overflow)
}
```
output:
```
200 false
```

yay:
So what the view on unit testing in standard librarry in Odin?

gingerBill:
@jrfondren this has been fixed now. I added that original change (before the fix) as someone mentioned it and I honestly didn't properly test it.

gingerBill:
> So what's the view on unit testing in standard librarry in Odin?

May I ask why are you asking it in this particular PR?

yay:
@gingerBill Just seems like having some would be helpful to verify correctness while developing and to prevent regressions in the future. I took a look at the code and the only area that's covered by tests is the crypto stuff.

Kelimion:
> @gingerBill Just seems like having some would be helpful to verify correctness while developing and to prevent regressions in the future. I took a look at the code and the only area that's covered by tests is the crypto stuff.

And compression stuff and image stuff and other things, not just crypto.

yay:
@Kelimion Ah, you're right, I searched for `@(test` only, not `when ODIN_TEST`.

PR #2141, opened 2022-10-20
Add caprintf and ctprintf to fmt

ChuuniMage:
Formatted cstring procs to work with ubiquitous cstring APIs


core/fmt/fmt.odin
```diff
@@ -162,7 +162,25 @@ panicf :: proc(fmt: string, args: ..any, loc := #caller_location) -> ! {
 	p("Panic", message, loc)
 }
 
+// formatted printing for cstrings
+caprintf :: proc(format: string, args: ..any) -> cstring {
+	str: strings.Builder
+	strings.builder_init(&str)
+	fmt.sbprintf(&str, format, ..args)
```
Kelimion, line 169: You're referring to package `fmt` from within itself.
There is no `core:fmt/fmt` which would lead to this `fmt.` namespacing here.

PR #2176, opened 2022-11-04
fix #2129 Segfault in compiler when void function used for its return value

### LINKED ISSUE #2129: Segfault in compiler when void function used for its return value
## Context

With latest odin from master

```
        Odin: dev-2022-10:412ca362
        OS:   Manjaro Linux, Linux 5.10.147-1-MANJARO
        CPU:  Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
        RAM:  15953 MiB
```

and this program:

```odin
package main

f :: proc() { }

g :: proc() -> int {
	return f()
}
```

## Expected Behavior

A compiler error and then normal exit.

## Current Behavior

A compiler error and then a segfault in the compiler:

```
$ odin build noret4.odin -file
/home/jfondren/odin/bug/noret4.odin(6:9) 'f()' call does not return a value and cannot be used as a value
Segmentation fault (core dumped)
```

jaspergeer:
Before this PR, `is_type_untyped` would not check if the type passed to it was null, resulting in a segfault in situations such as #2129. The function now returns `false` when passed a null type.

Below is the compiler message displayed when the code from #2129 is run:
```
/git/Odin% ./odin build test.odin -file                       
/home/jasper/git/Odin/test.odin(6:9) 'f()' call does not return a value and cannot be used as a value
/home/jasper/git/Odin/test.odin(1:2) Undefined entry point procedure 'main'
```
Note the lack of a segfault.

### DIFF
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -1257,6 +1257,9 @@ bool is_type_typed(Type *t) {
 }
 bool is_type_untyped(Type *t) {
 	t = base_type(t);
+	if (t == nullptr) {
+		return false;
+	}
 	if (t->kind == Type_Basic) {
 		return (t->Basic.flags & BasicFlag_Untyped) != 0;
 	}


jaspergeer:
Just realizing that I'm not sure if a null type should be considered typed or untyped. Is this correct behavior?

PR #2185, opened 2022-11-12
Update strings.odin

xzores:
The name index, is not the best name in a core library, I would suggest renaming the function to get_index_string or just index_string.


Kelimion:
`strings.index` already tells the reader that you're looking for the index in a string. Calling it `get_index_string` is redundant.

PR #2189, opened 2022-11-14
Update raylib to version 4.2

pJotoro:
I unfortunately couldn't test this beyond removing compile errors because of CRT linking hell. That said, the API itself has barely changed, so I doubt I made any mistakes here. I just followed raylib's change log word for word.


guest256:
raylib has already been updated to 4.5 in odin

PR #2200, opened 2022-11-15
Update macOS release map.

Kelimion:


Kelimion:
This detection will also be used to set the default for `-minimum-os-version` to the host's macOS version.

PR #2216, opened 2022-11-26
Automatically initialize x64 environment if CL.exe is missing

oskarnp:
also make sure found CL.exe is for x64

credit to mmozeiko


Kelimion:
Clever.

PR #2222, opened 2022-11-29
add thread.pool_clear

thePHTest:


thePHTest:
The new commit ow safely exhausts the semaphore.

Kelimion:
I'd love to see GraphiteMaster run this through the valgrind thread wobblificator. I recall he had a few ideas about rewriting the thread pool from scratch. His input would be most welcome.

gingerBill:
It might be better to just completely rewrite it from scratch 🤣

thePHTest:
I've been thinking about this more. It might be best to provide a few different thread pool implementations in core. Multiple producer + multiple consumer & single producer + multiple consumer for example.

PR #2231, opened 2022-12-05
sys/windows: fix wgl function loading in accordance with OpenGL wiki

ftphikari:


core/sys/windows/wgl.odin
```diff
@@ -87,6 +87,12 @@ foreign Opengl32 {
 }
 
 // Used by vendor:OpenGL
+// https://www.khronos.org/opengl/wiki/Load_OpenGL_Functions#Windows
 gl_set_proc_address :: proc(p: rawptr, name: cstring) {
-	(^rawptr)(p)^ = wglGetProcAddress(name)
+	func := wglGetProcAddress(name)
+	if uintptr(func) == 0 || uintptr(func) == 1 || uintptr(func) == 2 || uintptr(func) == 3 || uintptr(func) == ~uintptr(0) {
```
gingerBill, line 93: This could be simplified with:
```odin
switch uintptr(func) {
case 0, 1, 2, 3, ~uintptr(0):
```

Kelimion: Looks good otherwise. Follows Khronos' best practices and observed MSDN guidance.

PR #2236, opened 2022-12-09
Added support for quaternions in marshal/unmarshal procs

oelnenimmat:


gingerBill:
So this is not an oversight as to why these are not added. The biggest reason is because the memory layout of quaternions vary from place to place. They could be xyzw internally or wxyz. Because of this, it's not a good idea to rely on this too much if possible.

A comment should have explained this.

graphitemaster:
Suggestion, maybe we allow it but only of the quaternion in question has a tag on it like `xyzw` or `wxyz` which the reflection tools can use to read or write in the correct order?

gingerBill:
That might be a solution to the problem e.g. `json:"name,xyzw"` or something

ThisDevDane:
I think that would be a good compromise, I think it would be a shame to exclude one of the core built-in types like this from the official json implementation

PR #2255, opened 2022-12-17
Add more bindings to sys/windows

Lperlind:
More bindings to kernel32 and user32.


Lperlind:
Looks like it conflicts with raylib :/

Hyrtwol:
fyi had the same issue the work a-round was ```
	when !intrinsics.is_package_imported("raylib") {
		ShowCursor :: proc(bShow: BOOL) -> INT ---
	}
```
not ideal but usually you don't use raylib and win32 at the same time :p
I think if you rebase up this pr should work.

PR #2256, opened 2022-12-17
Add more utility procedures to small array

Lperlind:
Just some more procedures that I'm using in my project which I thought I should just put out there.

### DIFF
--- a/core/container/small_array/small_array.odin
+++ b/core/container/small_array/small_array.odin
@@ -1,6 +1,8 @@
 package container_small_array
 
 import "core:builtin"
+import "core:runtime"
+_ :: runtime
 
 Small_Array :: struct($N: int, $T: typeid) where N >= 0 {
 	data: [N]T,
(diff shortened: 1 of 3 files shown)


core/container/small_array/small_array.odin
```diff
@@ -32,6 +32,20 @@ get_ptr :: proc "contextless" (a: ^$A/Small_Array($N, $T), index: int) -> ^T {
 	return &a.data[index]
 }
 
+get_safe :: proc(a: $A/Small_Array($N, $T), index: int) -> (T, bool) {
```
Skytrias, line 35: looks good to me

core/container/small_array/small_array.odin
```diff
@@ -102,6 +116,19 @@ consume :: proc "odin" (a: ^$A/Small_Array($N, $T), count: int, loc := #caller_l
 	a.len -= count
 }
 
+ordered_remove :: proc "odin" (a: ^$A/Small_Array($N, $T), index: int, loc := #caller_location) {
```
Skytrias, line 119: could maybe do what the builtin `order_remove` does -> https://github.com/odin-lang/Odin/blob/master/core/runtime/core_builtin.odin#L61

maybe also use runtime.bounds_check_error_loc with `#no_bounds_check` flag

core/container/small_array/small_array.odin
```diff
@@ -111,6 +138,18 @@ push_back_elems :: proc "contextless" (a: ^$A/Small_Array($N, $T), items: ..T) {
 	a.len += n
 }
 
+insert :: proc "contextless" (a: ^$A/Small_Array($N, $T), item: T, index: int) -> bool {
```
Skytrias, line 141: could be renamed to `inject_at` to keep the naming scheme consistent with core

Lperlind:
Added tests and addressed comments

PR #2262, opened 2022-12-18
Make all functions static

creikey:


creikey:
Too many errors, regex was a bad idea...

PR #2272, opened 2022-12-21
Fixed memory leak in dir_darwin.odin.

sir-w7:
This is a minor fix to core:os/dir_darwin.odin where there is an allocation to context.allocator that is followed by neither a delete nor a return, thereby causing a memory leak. 
```
    dirpath: string
    dirpath, err = absolute_path_from_handle(fd)

    if err != ERROR_NONE {
        return
    }
```

### DIFF
--- a/core/os/dir_darwin.odin
+++ b/core/os/dir_darwin.odin
@@ -14,10 +14,11 @@ read_dir :: proc(fd: Handle, n: int, allocator := context.allocator) -> (fi: []F
 
 	dirpath: string
 	dirpath, err = absolute_path_from_handle(fd)
-
 	if err != ERROR_NONE {
 		return
 	}
+
+	defer delete(dirpath)
 
 	n := n
 	size := n


core/os/dir_darwin.odin
```diff
@@ -14,11 +14,12 @@ read_dir :: proc(fd: Handle, n: int, allocator := context.allocator) -> (fi: []F
 
 	dirpath: string
 	dirpath, err = absolute_path_from_handle(fd)
-
 	if err != ERROR_NONE {
 		return
 	}
 
+	defer delete(dirpath)
+
```
gingerBill, line 22: I'd move the `defer delete(dirpath)` to before the `err` check.

thePHTest: I took a look at this and `absolute_path_from_handle` only allocates when there isn't an error. In the error case it returns ` "" `

sir-w7: Yeah, looking at the Darwin implementation of `absolute_path_from_handle`, it appears to only allocate when there isn't an error (as thePHTest says).

PR #2273, opened 2022-12-21
Fix `core:slice.rotate_left`

ap29600:
This commit includes two fixes:
- a temporary cast to make the function compile
- a fix to a logic error that caused the function to hang or return incorrect results

### DIFF
--- a/core/slice/ptr.odin
+++ b/core/slice/ptr.odin
@@ -73,25 +73,17 @@ ptr_rotate :: proc(left: int, mid: ^$T, right: int) {
 		left, mid, right := left, mid, right
 
 		// TODO(bill): Optimization with a buffer for smaller ranges
-		if left >= right {
-			for {
-				ptr_swap_non_overlapping(ptr_sub(mid, right), mid, right)
+		for left > 0 && right > 0 {
+			if left >= right {
+				ptr_swap_non_overlapping(ptr_sub(mid, right), mid, right * size_of(T))
 				mid = ptr_sub(mid, right)
 
 				left -= right
-				if left < right {
-					break
-				}
-			}
-		} else {
-			for {
-				ptr_swap_non_overlapping(ptr_sub(mid, left), mid, left)
+			} else {
+				ptr_swap_non_overlapping(ptr_sub(mid, left), mid, left * size_of(T))
 				mid = ptr_add(mid, left)
 
 				right -= left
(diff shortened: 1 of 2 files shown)


core/slice/slice.odin
```diff
@@ -218,8 +218,10 @@ rotate_left :: proc(array: $T/[]$E, mid: int) {
 	n := len(array)
 	m := mid %% n
 	k := n - m
-	p := raw_data(array)
-	ptr_rotate(mid, ptr_add(p, mid), k)
+	// FIXME: (ap29600) this cast is a temporary fix for the compiler not matching
+	// [^T] with $P/^$T
+	p := cast(^int)raw_data(array)
```
gingerBill, line 223: Why are you casing to `(^int)`?

core/slice/ptr.odin
```diff
@@ -73,25 +73,17 @@ ptr_rotate :: proc(left: int, mid: ^$T, right: int) {
...
 				left -= right
-				if left < right {
-					break
-				}
-			}
-		} else {
-			for {
-				ptr_swap_non_overlapping(ptr_sub(mid, left), mid, left)
+			} else {
+				ptr_swap_non_overlapping(ptr_sub(mid, left), mid, left * size_of(T))
```
gingerBill, line 83: Why have you changed this logic?

gingerBill (changes requested):
Why have you changed the logic and done false casts?

core/slice/slice.odin
ap29600: that's a mistake, my bad.

core/slice/ptr.odin
ap29600: The logic needed to be changed because it's possible for the ordering of the left and right sizes to change multiple times before the algorithm terminates.
With the privious logic:
```
[1 2 3 4 5 6 7 8]
|      ^        |
iteration 1: left = 3, right = 5
swaps the first 3 with the following 3
[4 5 6 1 2 3 7 8]
       |     ^  |
iteration 2: left = 3, right = 2
range is not fully rotated, we still need to swap [1 2 3] with [7 8],
however the previous logic terminated here due to right becoming smaller than left.

```

Skytrias:
tested the `rotate_left` and it seems to work now :+1:

PR #2276, opened 2022-12-22
Show error message on macOS if llvm-config cannot be found

milend:
This adds a friendly error message on how exactly to make the build work if `llvm-config` cannot be found. The key parts are being able to set `LLVM_CONFIG` env variable and making sure the system `clang++` comes first in `$PATH` (the vanilla `clang++` compiler in the LLVM install needs additional flags to be able to find the macOS SDKs etc, so compilation doesn't work out of the box).

### DIFF
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -28,6 +28,16 @@ config_darwin() {
 	ARCH=$(uname -m)
 	: ${LLVM_CONFIG=llvm-config}
 
+	if ! command -v "${LLVM_CONFIG}" &> /dev/null; then
+		echo "You can provide path to llvm-config by either:"
+		echo "  1) Prepending 'LLVM_CONFIG=/path/to/bin/llvm-config' to your command"
+		echo "  2) Appending the directory containing llvm-config to your PATH"
+		echo "    - For example, execute 'export PATH=\"\$PATH:/path/to/llvm/bin\"' before the build command"
+		echo "    - Appending ensures your system's clang++ compiler comes first in your PATH"
+		echo "To get the path to your LLVM installation, execute 'brew info llvm@version'"
+		panic "Unable to find lvvm-config"
+	fi
+
 	# allow for arm only llvm's with version 13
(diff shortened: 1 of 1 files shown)


milend:
@gingerBill In the [docs](https://odin-lang.org/docs/install/#for-macos) for macOS, it says:

> On newer versions of macOS, some headers are not installed by default. Open macOS_SDK_headers_for_macOS_*.pkg in /Library/Developer/CommandLineTools/Packages/.

That shouldn't be necessary as in the [requirements section](https://odin-lang.org/docs/install/#requirements), they have already installed Xcode, so users can just activate the Xcode installation.

What are your thoughts on updating the docs, so that we remove all the instructions to deal with the command line tools package and just rely on Xcode?

By the way, I'm presuming there's no requirement for Odin to be compiled with the *same* compiler that comes as part of the LLVM installation from Homebrew.

Platin21:
Hmm would have made it just so it searches for 11.x.x and 12.x.x and 14.x.x in the path as that covers the entire base of compiler versions of clang we support. Technically we don’t require the brew clang e.g you can use anyone that fits the requirements. (I think apple is already using a newer version.. )

There isn’t really a requirement for having it build your odin programs with the same clang version as you use for compiling the compiler itself but you’d still need the right llvm config because of llvm-c being a hard requirement.

Btw. it will ask you to install the cmd line tools only if there isn’t any export for it yet it seems. The sdk and headers are taken from the dev package folder so..

gingerBill:
@milend The point of the command line tools is that you can use them WITHOUT installing Xcode. Xcode itself is a huge dependency on top of just the command line tools.

milend:
@gingerBill:

> The point of the command line tools is that you can use them WITHOUT installing Xcode

Yup, the point I was trying to make is that the [requirements section](https://odin-lang.org/docs/install/#requirements) already requires installing *both* Xcode and Command Line Tools but having both is unnecessary.

PR #2280, opened 2022-12-26
Fixed empty output_path.name when building a folder with no subfolders

DragosPopse:
`odin build folder` would result in an output with an empty filename `./folder/.exe`, while `odin build folder/folder2` would result in `./folder2.exe`. Now the first exemple results in `./folder.exe` as expected.


DragosPopse:
Note: paths are being resolved a bit finicky across the entire compiler. Issues with them might resurface.

PR #2284, opened 2022-12-30
fixup are_types_identical for comparing procs and checking if parameter names differ

thePHTest:
This fixes bugs with `odin doc examples/all -all-packages -doc-format` incorrectly de-duping proc types and resulting in incorrect documentation on the website.


thePHTest:
One note-- there are more cases in are_types_identical_internal for Type_Map and Type_SimdVector where this same change could be made to call into are_types_identical_internal(.., check_tuple_names) but I don't think it makes sense for those cases? Regardless might not hurt to make this same change to those, lmk what you think

PR #2289, opened 2023-01-02
Fix out of range error with _alloc_command_line_arguments in darwin

Lperlind:
If defaulting to nil allocator `res` will be set to a zero sized slice.
Thus we get an out of range error on the line
`res[i] = string(arg)`

Instead base the for loop iteration on the size of res (like on other platforms).


Lperlind:
I feel like macos actions have been failing a lot lately...

PR #2291, opened 2023-01-04
Add a lot of bindings to NSFoundation

Lperlind:
Just a ton of bindings I created for when I was porting my project to macos.
All of them are pretty straightforward, with the exception of the WindowDelegate.

I implemented the full window delegate protocol. To make life a little better than how the application delegate functions I register a class at runtime for the delegate.
Usage code looks like:
```odin
my_delegate := NS.WindowDelegateTemplate {
    // Implement delegate here. All functions are optional
    windowWillResizeToSize = proc(sender: ^NS.Window, size: NS.Size) -> NS.Size {
        fmt.println("Resized to:", size)
        return size
    },
}

// register and allocate the window delegate
// context is optional, with context set to runtime.default_context() if not explicitly provided
// this context is the odin context for all delegate calls
delegate := NS.window_delegate_register_and_alloc(my_delegate, "My_Delegate", context)
window->setDelegate(delegate)
```
The delegate template and odin context is copied into the created object. It returns nil if an allocation fails or the provided name has already been registered with.

Currently I haven't released a helper release mechanism. You can call `delegate->release()` but the class will still have its name registered.

I haven't extensively tested all the callbacks on the delegate. I just decided to implement it all in one swoop after I had what I needed done.
`NSUndoManager`, `NSScreen` and `NSPasteboard` were added simply for the WindowDelegate bindings

### DIFF
--- a/vendor/darwin/Foundation/NSApplication.odin
+++ b/vendor/darwin/Foundation/NSApplication.odin
@@ -8,10 +8,49 @@ ActivationPolicy :: enum UInteger {
 	Prohibited = 2,
 }
 
+ApplicationTerminateReply :: enum UInteger {
+	TerminateCancel = 0,
+	TerminateNow = 1,
+	TerminateLater = 2,
+}
+
+ApplicationPresentationOptionFlag :: enum UInteger {
+	AutoHideDock = 0,
+	HideDock = 1,
+	AutoHideMenuBar = 2,
+	HideMenuBar = 3,
+	DisableAppleMenu = 4,
+	DisableProcessSwitching = 5,
+	DisableForceQuit = 6,
+	DisableSessionTermination = 7,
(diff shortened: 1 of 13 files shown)


Platin21:
Hmm why does NSValue get the delegate methods?

Lperlind:
For application delegate?

Platin21:
Yeah for the app delegate my suggestion would be make a new class that just takes these. Instead of mangling NSValue..

Lperlind:
Yes that's what I do for the window delegate. I didn't intend to change application delegate that much. That's just how it is currently in the repo.

I can probably fully implement application delegate later this week. Following how I did window delegate (if approved)

Lperlind:
Gonna cut the review up a bit and rebase on master since the other PR got merged.

PR #2292, opened 2023-01-05
add get core count

colrdavidson:


core/os/os_freebsd.odin
```diff
@@ -702,6 +703,18 @@ get_page_size :: proc() -> int {
 	return page_size
 }
 
+get_processor_thread_count :: proc() -> int {
```
gingerBill, line 706: I'd just name this `processor_core_count` and on each platform specific file, prefix call it with `_` and mark it `@(private)`. And then have have a public interface that calls the package private procedure.

PR #2293, opened 2023-01-05
fix futex error handling

colrdavidson:


src/threading.cpp
```diff
@@ -562,6 +562,7 @@ gb_internal void tpool_wake_addr(Futex *addr) {
 		if (ret >= 0) {
```
gingerBill, line 562: This needs to be a `ret == 0`, right?

colrdavidson: I'm not 100% sure. Zig's futex does something similar to my change (>=0 and then flip ret), but this is one of those weird not-actually-documented functions, because it's not *really* exposed to users. Between the Apple docs and the Zig docs, I don't really know what to believe, but I hope the Zig guys got it right, and the interface changes a bit between the libc wrapper and the syscall.

https://github.com/ziglang/zig/blob/a27bfae0365e8d6d7da695a114497f97b4b20a43/lib/std/Thread/Futex.zig#L164

Issue #2277, opened 2022-12-23
json unmarshal fails on some float inputs


ftphikari:
## Context

        Odin: dev-2022-12:56e050fb
        OS:   Windows 10 Unknown Edition (0000007d) (version: 21H2), build 19044.2364
        CPU:  Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz
        RAM:  16263 MiB

## Code
```odin
package bug

import "core:fmt"
import "core:encoding/json"

Data :: struct {
	a: f32,
}

main :: proc() {
	data: Data
	str := `{"a":0.011111111111111111111}`
	bytes := transmute([]u8)str
	json.unmarshal(bytes, &data)
	fmt.println(data)
}
```

## Expected behavior
The program parses the number truncated.

## Ouput
The program hangs.

DragosPopse:
tracked it down to the parsing of floats. Looking into it further. It's not exactly a json marshalling error, it's `strconv`

DragosPopse:
The bug goes further, because  `str := {"a":0.81111111111111111}` will output `data = 0.0`

nico-bb:
I also had to do some digging recently and it indeed is `strconv.float64()` that doesn't handle high number of digit places (around 15-17+). It is very unexpected because the json parser silently ignores the error value from `float64()` and as a user you don't even realize that it may have failed.

ftphikari:
Seems this bug was fixed by commit ea9fe397e572b33f4db70a878ef7c5fc6b8c5d43, can people here confirm?

MarenFayre:
So, I wrote a script to brute force all the f32,
I wanted to test the "float-to-string" part against my [bindings to fastfloat](https://github.com/MarenFayre/odin-fastfloat), then validate every value for `parse_f32`.

But the first test was actually tripped during ftoa, this value:
```odin
import "core:fmt"
main :: proc() {
    buf: [4096]u8
    fmt.bprintf(buf[:], "%.170f", f32(16777216.0))

    fmt.println(16777216.0)   // 16777216.000
    fmt.printf("%s\n", buf)   // 16777210.0000000000000000000000000000000.......
}                             //        ^ Error
```

And here is the complete test I was doing:
```odin
package test_all

import "core:strconv"
import "core:fmt"
import "../fastfloat"

is_nan :: proc(f: f32) -> bool {  // I think this is correct?
    bits := transmute(u32)(f)
[...]

PR #2312, opened 2023-01-20
Replace fallback f64 parsing path with libc.strtod

MarenFayre:
Temporary fix for #2277

1. Disabled fallback path in favour of `libc.strtod`.
    While I dislike libc, the fallback path `decimal_to_float_bits / decimal.Decimal` simply does not seem to work on my side.  
    After some research this seemed like the most commonly used fallback.
 
 2. Added **break** from **trunc_block**
 
The `parse_f64` code seems like it was originally a copy of Go's `atof64`, but parts of it have evolved where it is hard to compare them for correctness.

Go separates the parsing logic like this:
```go
f, ok = atofHex() 
f, ok = atof64exact()
f, ok = eiselLemire64()
bits, overflow = decimal.set(string).floatBits()
```
But what is effectively our `atof64exact` is not a separate proc. Returning to indicate "keep parsing with a different algorithm" returns from the main function, skipping the fallback path. Although this part of the logic is slightly different from Go so I am not 100% sure of what was being attempted here with the block.  
**Adding this break prevents the fallback from being skipped in the repro example case.**

The `decimal.Decimal` fallback path foes not seem to work for me and the nested proc "decimal_to_float_bits::end" gets optimized out preventing me from debugging it, besides not being able to printf without import cycles.

In any case, `strtod` does not exhibit the following issues:
```odin
fmt.println(strconv.parse_f64("0.81111111111111111"))  // 0.000 true
```
```odin
// Loops forever as `decimal_to_float_bits` is stuck in for loops computing something

import "core:fmt"
import "core:encoding/json"
Data :: struct {
	a: f64,
}
main :: proc () {
	data: Data
	str := `{"a":0.011111111111111111111}`
	bytes := transmute([]u8)str
	json.unmarshal(bytes, &data)
	fmt.println(data)
	fmt.printf("%.60f\n", data.a)
}
```

NOTES:
- Code appears to be missing Go's fast `eiselLemire64` "Eisel-Lemire" path when naive extraction fails -- this is the most commonly used and fastest algorithm (with 99% float coverage).
- Decimal fallback path: 
  https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/strconv/decimal.go seems like a partial copy with some optimisations missing (`leftCheat`) -- could this explain the insanely slow parsing in some cases?  
- `strtod` Safety?
    1. Set errno to 0 and check it afterwards.
    2. Pass a &ptr:^u8 as second argument and check  ptr=&cstr => no number was found. (though I assume this is already handled.)
 
**Warning:** This has only tested against those two cases!

### DIFF
--- a/core/strconv/generic_float.odin
+++ b/core/strconv/generic_float.odin
@@ -287,8 +287,8 @@ round_shortest :: proc(d: ^decimal.Decimal, mant: u64, exp: int, flt: ^Float_Inf
 
 @(private)
 decimal_to_float_bits :: proc(d: ^decimal.Decimal, info: ^Float_Info) -> (b: u64, overflow: bool) {
-	end :: proc "contextless" (d: ^decimal.Decimal, mant: u64, exp: int, info: ^Float_Info) -> (b: u64) {
-		bits := mant & (u64(1)<<info.mantbits - 1)
+	end :: proc "contextless" (d: ^decimal.Decimal, mant: u64, exp: int, info: ^Float_Info) -> (bits: u64) {
+		bits = mant & (u64(1)<<info.mantbits - 1)
 		bits |= u64((exp-info.bias) & (1<<info.expbits - 1)) << info.mantbits
 		if d.neg {
 			bits |= 1<< info.mantbits << info.expbits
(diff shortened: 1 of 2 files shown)


gingerBill:
Please never fallback to libc for anything.

PR #2327, opened 2023-02-10
New default `context.temp_allocator`

gingerBill:
Replace the current default temporary allocator which is based on a ring buffer with a growing arena.

This new approach means the user need to explicitly call `free_all(context.temp_allocator)` to free the data.

### The Rational

* The current approach is "unsafe" by default due to the ring buffer design. This can be confusing to some newcomers to Odin.
* If an allocation was too large to fit within the ring buffer, it would be allocated to the leaking array. These were "hidden leaks" to many as people didn't realize you had to call `free_all(context.temp_allocator)` still.
* The arena approach allows for the ability to place temporary "watermarks" (similar to `virtual.Arena_Temp`)

### How it Works

The arena works by storing a linked list of memory blocks which track the current memory allocations. When a memory block cannot contain the requested allocation size, it will allocate a new memory block that can contain the new requested allocation size with the backing allocator (defaults to `runtime.default_allocator()`).


Platin21:
I do wonder would it maybe make sense to give a stack allocated block as first block?

One thing that i maybe over saw is also when do we create the temp arena? At the program start?

gingerBill:
@Platin21 
> I do wonder would it maybe make sense to give a stack allocated block as first block?

No because this would easily go out of bounds when used in weird ways. 
 
> One thing that i maybe over saw is also when do we create the temp arena? At the program start?

The temp arena is actually created on first use or at the start of the thread and destroyed on program/thread end.

PR #2336, opened 2023-02-20
Add TSC frequency getter

colrdavidson:


colrdavidson:
This needs some batting back and forth on API design.
As-is, this is probably the "safest" API choice, although it will potentially introduce a 2 second sleep in *some* systems, like Windows that want to call it.

We could make it take a "return immediately on fail" flag or something, or try to work in the hacky solution to get the data from Windows directly, although my confidence in the hacky solution failing appropriately if it's on the wrong version of Windows, is not high.

A user on the HMN forums worked out how to get these without approximation on Windows, although you have to grab the data from NtQuerySystemInformation.
https://hero.handmade.network/forums/code-discussion/t/7485-queryperformancefrequency_returning_10mhz_bug/2

core/time/perf.odin
```diff
@@ -40,6 +40,55 @@ _tick_duration_end :: proc "contextless" (d: ^Duration, t: Tick) {
 	d^ = tick_since(t)
 }
 
+when ODIN_ARCH == .amd64 {
+	x86_has_invariant_tsc :: proc "contextless" () -> bool {
```
gingerBill, line 44: Should this be private to the package?

core/time/perf.odin
```diff
@@ -40,6 +40,55 @@ _tick_duration_end :: proc "contextless" (d: ^Duration, t: Tick) {
...
+
+has_invariant_tsc :: proc "contextless" () -> bool {
+	when ODIN_ARCH == .amd64 {
+		return x86_has_invariant_tsc()
+	}
+
+	return false
+}
+
+_get_tsc_frequency_fallback :: proc "contextless" () -> u64 {
```
gingerBill, line 66: I would just nest this procedure in the one below.

core/time/perf.odin
```diff
@@ -40,6 +40,55 @@ _tick_duration_end :: proc "contextless" (d: ^Duration, t: Tick) {
...
+	sleep(2 * Second)
+
+	tsc_end := intrinsics.read_cycle_counter()
+	tick_end := tick_now()
+
+	time_diff := u128(duration_nanoseconds(tick_diff(tick_begin, tick_end)))
+	return u64((u128(tsc_end - tsc_begin) * 1_000_000_000) / time_diff)
+}
+
+get_tsc_frequency :: proc "contextless" () -> (u64, bool) {
```
gingerBill, line 79: Does this need the `get` prefix when there is never going to be a `set`?

gingerBill (approved):

PR #2356, opened 2023-02-28
Add option to link to glfw3 dynamically

flysand7:
When implementing multi-module applications (for example applications that use hot reloading), static linking with GLFW3 creates problems related to duplication of global state in the different Odin modules.

This patch implements a configurable option to tell GLFW3 package to link to .DLL instead of .LIB. In order to use the option, supply `-define:GLFW_DYNAMIC=true` argument to odin compiler option and copy .dll file from `vendor:glfw/lib` folder to the same folder as your executable, or other folder where windows would search for the folder.

On linux dynamic linking (as of now) should be on by default. I didn't do any work for other OSes.

### DIFF
--- a/vendor/glfw/bindings/bindings.odin
+++ b/vendor/glfw/bindings/bindings.odin
@@ -3,13 +3,24 @@ package glfw_bindings
 import "core:c"
 import vk "vendor:vulkan"
 
+GLFW_DYNAMIC :: #config(GLFW_DYNAMIC, false)
+
 when ODIN_OS == .Windows {
-	foreign import glfw { 
-		"../lib/glfw3_mt.lib",
-		"system:user32.lib", 
-		"system:gdi32.lib", 
-		"system:shell32.lib",
-	} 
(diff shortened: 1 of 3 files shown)


vendor/glfw/bindings/bindings.odin
```diff
@@ -3,13 +3,24 @@ package glfw_bindings
...
+	when GLFW_DYNAMIC {
+		foreign import glfw {
+			"../lib/glfw3dll.lib",
+			"system:user32.lib", 
+			"system:gdi32.lib", 
+			"system:shell32.lib",
+		}
+	} else {
+		foreign import glfw {
+			"../lib/glfw3mt.lib",
```
Kelimion, line 18: It looks like you forgot the underscore here, `../lib/glfw3_mt.lib`, and that's maybe why the test now fails.

flysand7: Added the missing underscore, should be working now.
Great how one of these lib files is underscored and the other isn't

PR #2380, opened 2023-03-13
Add -no-tls flag

flysand7:
Freestanding applications that do not initialize TLS may crash upon using `@thread_local` variables. This patch adds a `-no-tls` build flag, so that these applications can use `@thread_local` variables and treat them as normal variables (if there's only one thread, `@thread_local` should be the same as simply a global variable).

The need for the flag is due to some functionality in `core:` package declaring `@thread_local` variables that can not be avoided, in particular `default_temp_allocator_*` in `runtime` package.

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -275,6 +275,7 @@ struct BuildContext {
 	bool   no_output_files;
 	bool   no_crt;
 	bool   no_entry_point;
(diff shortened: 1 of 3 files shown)


gingerBill:
Maybe rename it to `-no-thread-local`?

flysand7:
That might be a better name

Issue #2388, opened 2023-03-16
imports enclosed in when are type checked


DragosPopse:
## Context

When trying to separate an import unavailable on a certain platform through a `when` statement, the compiler does type checking on the import, and obviously fails. I feel this is a big issue, because this makes a lot of cross platform setups nearly impossible to achieve (example below), especially desktop <-> wasm
![image](https://user-images.githubusercontent.com/31741257/225557492-95b8669a-2601-48b4-928c-c29b0ed952e7.png)
![image](https://user-images.githubusercontent.com/31741257/225557575-d95faf38-b8eb-4513-921b-f595cc5306c3.png)

* Operating System & Odin Version:
Odin: dev-2023-03:ca15eb26
        OS:   Windows 10 Professional (version: 21H2), build 19044.2728
        CPU:  AMD Ryzen 9 3900X 12-Core Processor
        RAM:  32716 MiB

## Expected Behavior

The import should be completely ignored if inclosed in a `when` that doesn't evaluate to true. 

## Current Behavior

`#panic` is being invoked from `os_js.odin`, which shows that the import is type checked. It also follows up with a bunch of undefined type errors.

## Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. `when ODIN_OS != .JS do import "core:os`
2. Compiler with `-target:js_wasm32`
3. Pain 

### Failure Logs

![image](https://user-images.githubusercontent.com/31741257/225558523-6734577d-1248-4dab-b3b8-11d406bb30c3.png)

Feoramund:
It is not possible anymore to import packages from within when statements, making this issue no longer applicable. The following error will arise:
```
Cannot use 'import' within a 'when' statement.
Prefer using the file suffixes (e.g. foo_windows.odin) or '//+build' tags
```

DragosPopse:
this is quite a stale and "false" issue at this point, bill clarified the situation many times. i'll close it.

PR #2397, opened 2023-03-20
Made most libraries panic on js targets instead of not compiling

DragosPopse:
In reference to #2388 . Writing cross platform code for wasm + desktop gets quite annoying when you can't enclose your library usage in `when` statements. This PR fixes the issue by implementing all non supported procedures as a simple `panic` call. This will allow usage of those libraries inside `when ODIN_OS == ...` statements, and also give better error messages when you are using a procedure that is not supported on your current platform.


core/dynlib/lib_js.odin
```diff
@@ -0,0 +1,15 @@
...
+	panic("core:dynlib not supported by JS target")
+}
+
+_unload_library :: proc(library: Library) -> bool {
+	panic("core:dynlib not supported by JS target")
+}
+
+_symbol_address :: proc(library: Library, symbol: string) -> (ptr: rawptr, found: bool) {
+	panic("core:dynlib not supported by JS target")
+}
```
gingerBill, line 15: Would this be better to just return `false` for all of these?

Kelimion: Maybe, but if not, then replace `panic("...")` with `unimplemented("...")`.

DragosPopse: I didn't know `unimplemented` exists. What is the difference between that and `panic`? But yeah maybe for the dynlib stuff a false could be nicer to use than straight busting your program. I'll update it

DragosPopse:
Also, coming back to my discord post. Would it be sensible to apply the same treatment to `intrinsics.wasm_grow_memory` ? I'm asking because I am not particularly sure how would I do that in a builtin proc.

core/dynlib/lib_js.odin
Kelimion: It's still a `panic`, only it makes it easier to search for things that are `unimplemented`.

DragosPopse: Good call, i'll update my pr with your suggestion in a few hours

DragosPopse:
Updated. Let me know if anything else is needed

Kelimion:
> Also, coming back to my discord post. Would it be sensible to apply the same treatment to `intrinsics.wasm_grow_memory` ? I'm asking because I am not particularly sure how would I do that in a builtin proc.

That's a @gingerBill question.

gingerBill:
No because it is a built-in intrinsic and something at the wasm level itself.

PR #2406, opened 2023-03-24
Fix typo in warning message in parser

aloussase:


krixano:
Not to be a grammar police or anything, but there's yet another mistake that was missed here :D 
`has` should also be changed to `have`, since `ranges` is plural.

aloussase:
Are you sure? I see it as saying that the '..' syntax has been deprecated, not the ranges themselves.

krixano:
Ah! You know what... I was seeing "for ranges" as part of the noun phrase, actually. But the intended meaning might have rather been something more like `'..' in ranges has now been deprecated`, as in "for" being a preposition rather than talking about for loops.

PR #2411, opened 2023-03-26
Fmt docs

jon-lipstate:
Added Code Documentation to fmt.

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -44,6 +44,34 @@ Info :: struct {
 // Custom formatter signature. It returns true if the formatting was successful and false when it could not be done
 User_Formatter :: #type proc(fi: ^Info, arg: any, verb: rune) -> bool
 
+// Example User Formatter:
+// SomeType :: struct {
+// 	value: int,
+// }
+// // Custom Formatter for SomeType
+// User_Formatter :: proc(fi: ^fmt.Info, arg: any, verb: rune) -> bool {
+// 	m := cast(^SomeType)arg.data
+// 	switch verb {
+// 	case 'v', 'd':
+// 		fmt.fmt_int(fi, u64(m.value), true, 32, verb)
+// 	case:
+// 		return false
+// 	}
+// 	return true
+// }
+// //
+// main :: proc() {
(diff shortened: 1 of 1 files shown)


core/fmt/fmt.odin
```diff
@@ -44,6 +44,34 @@ Info :: struct {
...
+// Example User Formatter:
+// SomeType :: struct {
+// 	value: int,
+// }
+// // Custom Formatter for SomeType
+// User_Formatter :: proc(fi: ^fmt.Info, arg: any, verb: rune) -> bool {
+// 	m := cast(^SomeType)arg.data
+// 	switch verb {
+// 	case 'v', 'd':
+// 		fmt.fmt_int(fi, u64(m.value), true, 32, verb)
```
Kelimion, line 56: Separate but related observation: The `bit_size` argument is a bit confusing. Why is the `bit_size` of a `u64` 32 and not 64?

core/fmt/fmt.odin
```diff
@@ -44,6 +44,34 @@ Info :: struct {
...
+// 	}
+// 	return true
+// }
+// //
+// main :: proc() {
+// 	// Ensure the fmt._user_formatters map is initialized
+// 	if fmt._user_formatters == nil {
+// 		fmt._user_formatters = new(map[typeid]fmt.User_Formatter)
+// 	}
+// 	// Register the custom formatter directly to the _user_formatters map
```
Kelimion, line 68: I agree that you can do that directly, but perhaps the example should promote using `set_user_formatters` & `register_user_formatter` instead.

Users who want to set it up directly can easily figure out how to do so by reading those procedures. What if we were to make that global private in the future?

Kelimion:
Wow, that's an awesome PR. Very good effort.

core/fmt/fmt.odin
jon-lipstate: fixed

core/fmt/fmt.odin
jon-lipstate: fixed, i also added an if guard around set_user_formatters, do you agree with that?

PR #2412, opened 2023-03-26
text/table: Initial implementation

oskarnp:


core/text/table/table.odin
```diff
@@ -0,0 +1,399 @@
...
+import "core:runtime"
+import "core:strings"
+
+Cell :: struct {
+	text: string,
+	alignment: Cell_Alignment,
+}
+
+Cell_Alignment :: enum {
+	Left = 1,
```
gingerBill, line 25: Why is this set to `1` and not just `0`?

core/text/table/table.odin
```diff
@@ -0,0 +1,399 @@
...
+	tbl.colw = make([dynamic]int, tbl.table_allocator)
+	tbl.format_allocator = format_allocator
+	return tbl
+}
+_init_with_virtual_arena :: proc(tbl: ^Table, format_arena: ^virtual.Arena, table_allocator := context.allocator) -> ^Table {
+	return _init(tbl, virtual.arena_allocator(format_arena), table_allocator)
+}
+_init_with_mem_arena :: proc(tbl: ^Table, format_arena: ^mem.Arena, table_allocator := context.allocator) -> ^Table {
+	return _init(tbl, mem.arena_allocator(format_arena), table_allocator)
+}
```
gingerBill, line 60: I'd recommend just making this `init_with_allocator` `init_with_virtual_arena` `init_with_mem_arena` instead of the `_init` prefix.

core/text/table/table.odin
```diff
@@ -0,0 +1,399 @@
...
+	}
+	switch val in value {
+	case nil:
+		cell.text = ""
+	case string:
+		cell.text = string(val)
+	case cstring:
+		cell.text = string(val)
+	case:
+		cell.text = format(tbl, "%v", val)
```
gingerBill, line 106: This case allocates but the others don't. Is this as expected or is it just relying on the arena-like memory allocation?

core/text/table/table.odin
gingerBill: Or better yet, why not have `Default = 0,`?

core/text/table/table.odin
```diff
@@ -0,0 +1,399 @@
...
+row :: proc(tbl: ^Table, values: ..any, loc := #caller_location) {
+	if tbl.nr_cols == 0 {
+		if len(values) == 0 {
+			panic("Cannot create row without values unless knowing amount of columns in advance")
+		} else {
+			tbl.nr_cols = len(values)
+		}
+	}
+	tbl.nr_rows += 1
+	for col := 0; col < tbl.nr_cols; col += 1 {
```
gingerBill, line 154:
```odin
for col in 0..<tbl.nr_cols {
```

core/text/table/table.odin
```diff
@@ -0,0 +1,399 @@
...
+
+	io.write_string(w, "<table>\n")
+	if tbl.caption != "" {
+		io.write_string(w, "<caption>")
+		io.write_string(w, tbl.caption)
+		io.write_string(w, "</caption>\n")
+	}
+
+	align_attribute :: proc(cell: ^Cell) -> string {
+		#partial switch cell.alignment {
```
gingerBill, line 218: Why is this `#partial`?

core/text/table/table.odin
```diff
@@ -0,0 +1,399 @@
...
+					write_byte_repeat(w, max(1, tbl.colw[col]-1, tbl.lpad, tbl.rpad), '-')
+				case .Center:
+					io.write_byte(w, ':')
+					write_byte_repeat(w, max(1, tbl.colw[col]-2, tbl.lpad, tbl.rpad), '-')
+					io.write_byte(w, ':')
+				case .Right:
+					write_byte_repeat(w, max(1, tbl.colw[col]-1, tbl.lpad, tbl.rpad), '-')
+					io.write_byte(w, ':')
+				case:
+					panic("unhandled cell alignment")
```
gingerBill, line 349: Does this need to panic or can it be treated like `.Left`?

core/text/table/table.odin
```diff
@@ -0,0 +1,399 @@
...
+		pad := colw - len(text)
+		odd := pad & 1 != 0
+		write_byte_repeat(w, pad/2, ' ')
+		io.write_string(w, text)
+		write_byte_repeat(w, pad/2 + 1 if odd else pad/2, ' ')
+	case .Right:
+		write_byte_repeat(w, colw - len(text), ' ')
+		io.write_string(w, text)
+	case:
+		panic("unhandled alignment")
```
gingerBill, line 388: Does this need to panic or can it be treated like `.Left`?

core/text/table/table.odin
```diff
@@ -0,0 +1,399 @@
...
+	write_byte_repeat(w, rpad, ' ')
+}
+
+stdio_writer :: proc() -> io.Writer {
+	return io.to_writer(os.stream_from_handle(os.stdout))
+}
+
+strings_builder_writer :: proc(b: ^strings.Builder) -> io.Writer {
+	return strings.to_writer(b)
+}
```
gingerBill, line 399: I'd add these to a `utility.odin` file as they are not really part of this API, kind of.

core/text/table/table.odin
oskarnp: Yes, that is intended. That's why I need the free_all for formatted literals to avoid to free on each cell separately.

krixano:
Not sure if you want to implement this, but it could be useful for ASCII tables if there were an option for a column max length, as well as handling text-wrapping for that.

oskarnp:
> Not sure if you want to implement this, but it could be useful for ASCII tables if there were an option for a column max length, as well as handling text-wrapping for that.

I’d love to but think I’ll leave that for a future 2.0 version and not this PR.

PR #2414, opened 2023-03-27
core:strings docs

jon-lipstate:
Please verify you agree with Builder Line 147 & 182 (destroy proc modifications).

### DIFF
--- a/core/strings/ascii_set.odin
+++ b/core/strings/ascii_set.odin
@@ -2,20 +2,23 @@
 package strings
 
 import "core:unicode/utf8"
-// Ascii_Set is designed to store ASCII characters efficiently as a bit-array
-// Each bit in the array corresponds to a specific ASCII character, where the value of the bit (0 or 1) 
-// indicates if the character is present in the set or not.
-Ascii_Set :: distinct [8]u32
 
-// Creates an Ascii_Set with unique characters from the input string.
-//
-// Inputs:
-// - chars: A string containing characters to include in the Ascii_Set.
-//
-// Returns:
-// - as: An Ascii_Set with unique characters from the input string.
-// - ok: false if any character in the input string is not a valid ASCII character.
-//
+/*
+Ascii_Set is designed to store ASCII characters efficiently as a bit-array
+Each bit in the array corresponds to a specific ASCII character, where the value of the bit (0 or 1) 
+indicates if the character is present in the set or not.
+*/
+Ascii_Set :: distinct [8]u32
+/*
+Creates an Ascii_Set with unique characters from the input string.
+
+Inputs:
+- chars: A string containing characters to include in the Ascii_Set.
+
+Returns:
+- as: An Ascii_Set with unique characters from the input string.
+- ok: false if any character in the input string is not a valid ASCII character.
+*/
 ascii_set_make :: proc(chars: string) -> (as: Ascii_Set, ok: bool) #no_bounds_check {
 	for i in 0..<len(chars) {
 		c := chars[i]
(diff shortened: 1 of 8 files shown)


core/strings/conversion.odin
```diff
@@ -57,14 +69,19 @@ to_valid_utf8 :: proc(s, replacement: string, allocator := context.allocator) ->
...
-*/
+// Converts the input string `s` to all lowercase characters.
+//
+// *Allocates Using Provided Allocator*
+//
+// Inputs:
+// - s: Input string to be converted.
+// - allocator: (default: context.allocator).
+//
+// Usage:
```
Lperlind, line 80: You should use Example: rather than Usage:
This will generate a colourised code block on the website that is collapsible.

core/strings/ascii_set.odin
```diff
@@ -2,10 +2,20 @@
 package strings
 
 import "core:unicode/utf8"
-
+// Ascii_Set is designed to store ASCII characters efficiently as a bit-array
```
Lperlind, line 5: Can't multi select lines on mobile unfortunately.
Since tabbing is significant (it determines code blocks) you should do all the documentation within a block comment, i.e. /* */
Look at package dynlib for example.

Lperlind (commented):
Sorry on mobile if the comments are a bit awkward.

core/strings/ascii_set.odin
```diff
@@ -3,9 +3,22 @@ package strings
...
+Creates an Ascii_Set with unique characters from the input string.
 
-// create an ascii set of all unique characters in the string
+Inputs:
+- chars: A string containing characters to include in the Ascii_Set.
+
+Returns:
+- as: An Ascii_Set with unique characters from the input string.
+- ok: false if any character in the input string is not a valid ASCII character.
+*/
```
Lperlind, line 21: This will be formatted incorrectly on the website.
You need to have 2 trailing spaces in markdown to add a break and not have everything be on a single line.
i.e.
`Inputs:  ` rather than `Inputs:`

Furthermore to make this stand out more on the website it might be beneficial to make the `Inputs`, `Returns`, etc. to be bold.
i.e.
`**Inputs**  `

Note you don't need to do this for `Example:` since that has its own formatting rules when generating the website documentation.

jon-lipstate:
Going to reopen on a clean branch, i made something of a mess on my copy.

core/strings/conversion.odin
```diff
@@ -57,13 +71,20 @@ to_valid_utf8 :: proc(s, replacement: string, allocator := context.allocator) ->
...
+*Allocates Using Provided Allocator*
+
+Inputs:
+- s: Input string to be converted.
+- allocator: (default: context.allocator).
+
+Example:
+```odin
+	strings.to_lower("TeST") -> test
+```
```
Lperlind, line 86: The ticks and odin is not required. The website already knows that odin docs are being generated. Only a tab is required and markdown will format it also properly. You should also make all your examples fully self contained.
i.e. (note that tabs are significant here)
```
Example:

	import "core:strings"
	import "core:fmt"
	import "core:mem"

	strings_to_lower_example :: proc() -> (err: mem.Allocator_Error) {
		s :: "TeST"
		s_lower_cased := strings.to_lower(s, context.allocator) or_return
		fmt.println(s_lower_cased)
	}

Output:

	test

```

The `Output:` is not necessary but the website will properly format and test it soon so would be good to start.

Lperlind: Not a rule of thumbs but it will be easier to generate specific things if the example procedure name follows the form
`<Package_Name>_<Procedure_Name>_example`

PR #2421, opened 2023-03-28
Code Docs for strconv

jon-lipstate:


jon-lipstate:
@Lperlind @Kelimion i think its ready to look at. I left an issue in the code for `quote`, either i misunderstand it entirely or it isnt behaving as expected. the old docs align to what i expect.

`!! ISSUE !! NOT EXPECTED -- "\"hello\"" was expected  `

Lperlind:
To make the cli pass you need to make the examples in the following form
```
<proc_name>_example
```
I know I stated in the previous PR to include package, but we simplified things so it's no longer required.

Lperlind:
Take your time with getting the examples right. Your paren change just produces a lot of invalid code.

PR #2436, opened 2023-04-03
Make tests scripts error if a test fails

Lperlind:
Additionally fixes tests that were found broken because of this.
* Keep vet happy
* Fix crlf encoding on windows platforms in compression tests
* remove non existant text/lua test


sabahkarami (commented):
ddddd

PR #2438, opened 2023-04-05
Add script for removing platform-specific libs

Naught00:
Utility script for removing binary libraries from other platforms.

It is useful for doing a system install of the vendor library collection, for example.

### DIFF
--- a/misc/remove_libraries_for_other_platforms.sh
+++ b/misc/remove_libraries_for_other_platforms.sh
@@ -6,6 +6,12 @@ panic() {
 	exit 1
 }
 
+assert_vendor() {
+	if [ $(basename $(pwd)) != 'vendor' ]; then
+		panic "Not in vendor directory!"
+	fi
+}
+
 remove_windows_libraries() {
 	find . -type f -name '*.dll' | xargs rm -f 
 	find . -type f -name '*.lib' | xargs rm -f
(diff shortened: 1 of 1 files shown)


misc/remove_libraries_for_other_platforms.sh
```diff
@@ -0,0 +1,46 @@
...
+	find . -type d -name '*macos*' | xargs rm -rf
+}
+
+remove_linux_libraries() {
+	find . -type f -name '*.so' | xargs rm -f 
+	find . -type d -name 'linux' | xargs rm -rf
+}
+
+case $OS in
+	Linux)
```
Kelimion, line 26: Maybe there could be an `assert_vendor` sub routine that verifies you're calling it from `vendor`, using `basename $(pwd)`?

Just so that you're not accidentally nuking anything else.

Naught00: Just added in new commit

Kelimion: Great, but you'd probably want to run that on Darwin and *BSD as well?

Naught00: Done

Kelimion (approved):
Waiting on CI to complete to merge.

PR #2452, opened 2023-04-11
core:http

jon-lipstate:
Support basic HTTP, not to exceed HTTP/1.1

Planned Contents: Status Codes, Verbs, URIs, Headers, Request / Response


jon-lipstate:
I think the consensus is to close/abort this pr for now (?). 

@Tetralux @Kelimion 
I do think i made some improvements of uri parsing, what are your opinions for merging that aspect into `core:net`? My implementation still is missing some features that would be desired such as relative path resolution & canonicalization, but it does break up the uri into components as defined in RFC3987.

Example of decode/encode as-written:
```console
Baseline Example:
decode:   http  user:deprecated  example.com  4242  ["path", "to", "the", "..", "resource"]  map[query="value"]  fragment
core:net  http  user:deprecated@example.com:4242   /path/to/the/../resource   map[query="value#fragment"]
encode:  http://user:deprecated@example.com:4242/path/to/the/../resource?query=value#fragment
RFC Example:
decode:   ftp   cnn.example.com&story=breaking_news   10.0.0.1   ["top_story.htm"]  map[foo="3"]  bar
core:net  ftp   cnn.example.com&story=breaking_news@10.0.0.1  /top_story.htm  map[foo="3#bar"]
encode:  ftp://cnn.example.com&story=breaking_news@10.0.0.1/top_story.htm?foo=3#bar
```

on the `percent_decode` front [whatwg](https://url.spec.whatwg.org/#percent-decode) specs out that a non-hex octet is passed through without error, eg `%2A_%FQ_%9 -> *_%FQ_%9` where octet no. 2 is invalid, and no. 3 is not 2 characters.

Kelimion:
It's a slippery slope, as Bill intimated. Best done as a third-party package if someone does want to tackle it. We can link to it from the wiki and so forth.

PR #2462, opened 2023-04-16
Fix: header directories in Unix build script

lefp:


Kelimion:
What's the purpose of this "fix"? It builds the same with and without this patch.

lefp:
The unpatched version failed to build on one of my machines. Patched version works on both.

Sorry about skipping that detail.

PR #2465, opened 2023-04-16
Fix put preferable llvm-config-14 first.

shizeeg:
Fix building on systems with default LLVM v15 and higher, Arch Linux for example.


jcmdln:
This PR is stale and can likely be closed. The `build_odin.sh` script has been refactored and there's a separate PR to add checks for LLVM14  in https://github.com/odin-lang/Odin/pull/2893. The order of when LLVM14 is checked is important so that Odin always cuts a release with an officially supported version of LLVM as outlined in that PR.

PR #2469, opened 2023-04-19
Fixing user formatters

xzores:
The current way user formatters work are not great. We should all different parts of the code to add their own userformatter, only knowing about their own local code. We should be able to call register_user_formatter without calling set_user_formatters, and set_user_formatters should just copy the formatters not require the user to keep them alive.

Here is how I would do it.

I am new to this pull request thing, I don't know if I did it 100% correctly.


xzores:
It was not my intention to upload "Update strings.odin", how do it change that?

gingerBill:
The entire point of it being a pointer is not cause any implicit memory allocation and for the user to manager the `map` themselves. This is not a mistake in the design.

xzores:
But how can you expect your user do that? It would conflict with any library that uses formatters.

2 points:
- This is going to cause friction.
- You will have (implicit) memory allocations anyway (I don't know if you would classify it as implicit)

If a user sets a map with set_user_formatters and then afterwards calls register_user_formatter you will have an memory allocation anyway. 
Also what you are doing is to have a pointer in global memory to some other global memory. I assume that whatever map you pass will have to be in global memory since it need for the duration of the program. And that is just a little weird.

- library example:
Let us take the scenario where one has installed 2 libraries that both use user formatters. 
First you want to initialize library A by calling it's init function. Now lib A was made with user formatters and it requires the user to call set_user_formatters before its init is called.
So the user does that, now the user call lib B. 
Lib B also uses user formatters, but it calls set_user_formatters in its init function. 
It is important that A's init is called before B's.
What should the user do here? hack and slash something together?
Do you have a standard for how the libs should interact with each other?
What happens if that standard is violated, the libs just become incompatible.

Not only will this problems, but it is highly unlikely that a new user will just know how to setup the user formatters without asking an stack overflow.

gingerBill:
> It would conflict with any library that uses formatters.

That is by design. I understand what you are wanting.

> This is going to cause friction.

Friction is not always bad, especially when you want to reduce the chance of things like this.

> You will have (implicit) memory allocations anyway 

Not at the core library level like this where fmt should be able to work perfectly fine in a freestanding build where the context has no default allocators.

PR #2470, opened 2023-04-20
Separate int size from word/pointer size

gingerBill:
Currently, the compiler assumes that the size of `int` is equivalent to the word/pointer size. This PR will make a distinction between `size_of(int)` and `size_of(rawptr)`.

This PR will also add the (pseudo-)architectures `wasm64p32` which will make `size_of(int) == 8` and `size_of(rawptr) == 4`.

This approach would also hypothetically allow for other architectures like `amd64p32` (i.e. [Linux's `x32`](https://en.wikipedia.org/wiki/X32_ABI)).


colrdavidson (approved):
I don't see anything obviously wrong here. Hopefully I've got test-file that should fail to OOM on hand, will see if I can stretch wasm64p32 vs wasm32

colrdavidson:
Ah, nope, triggering some asserts still.

`src/llvm_backend_const.cpp(163): Assertion Failure: `value_count == elem_count` { %runtime.Type_Info*, i32, i64 } 2 3`

src/build_settings.cpp
```diff
@@ -361,131 +361,149 @@ gb_internal isize MAX_ERROR_COLLECTOR_COUNT(void) {
...
 	str_lit("e-m:e-p:32:32-i64:64-n32:64-S128"),
 };
 
 
-gb_global TargetMetrics target_js_wasm64 = {
+gb_global TargetMetrics target_freestanding_wasm64p32 = {
+	TargetOs_freestanding,
+	TargetArch_wasm64p32,
+	4, 8, 8, 16,
+	str_lit("wasm32-freestanding-js"),
```
MrSmith33, line 481: Is `wasm32` correct here? Same question about other `wasm64p32` targets

PR #2479, opened 2023-04-23
Change default executable extension for non-Windows platforms

Naught00:
Putting a program into your path on a UNIX system with a file extension means that you have to type the extension out for every invocation of the program. A better default is to have no extension at all since most people will end up removing it anyway.

This change does not affect Windows since the .exe extension is set after the default extension if compiling on Windows.

One problem that arises from this is when you are building a package from the same directory that the package's directory is located in: the program name and directory name conflict. The second commit fixes this by adding the .bin extension to fix the conflict.

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -1502,7 +1502,7 @@ gb_internal bool init_build_paths(String init_filename) {
 
 		if (build_context.metrics.os == TargetOs_windows) {
 			output_extension = STR_LIT("exe");
-		} else if (!str_eq(init_filename, str_lit(".")) && path_is_directory(init_filename)) {
+		} else if (!str_eq(init_filename, str_lit(".")) && path_is_directory(last_path_element(init_filename))) {
 			// Add .bin extension to avoid collision
 			// with package directory name
 			output_extension = STR_LIT("bin");
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -1502,16 +1502,9 @@ gb_internal bool init_build_paths(String init_filename) {
 
 		if (build_context.metrics.os == TargetOs_windows) {
 			output_extension = STR_LIT("exe");
-		} else if (str_eq(init_filename, str_lit("."))) {
-			// Avoid conflict in edge case where directory to be compiled is
-			// the same as the current directory's name
-			if (path_is_directory(last_path_element(get_current_directory()))) {
-				// Add .bin extension to avoid collision
-				// with package directory name
-				output_extension = STR_LIT("bin");
-			}
-		// Path could be absolute or relative
 		} else if (path_is_directory(last_path_element(bc->build_paths[BuildPath_Main_Package].basename))) {
+			// Add .bin extension to avoid collision
+			// with package directory name
 			output_extension = STR_LIT("bin");
 		}
 	} else if (build_context.build_mode == BuildMode_DynamicLibrary) {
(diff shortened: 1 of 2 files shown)


DaraghD:
LGTM!

Kelimion:
I'll give it a thorough review tomorrow, as this problems has proven stickier than one might imagine.

Kelimion:
More commits keep being added. Let me know when you think it works on Windows, Linux and Mac in all of the edge cases. :-)

Naught00:
I have tested it on Windows and Linux. It works in all the edge cases I could think of.

Kelimion:
I did say it was a sticky problem. ;-)

Naught00:
This is ready now.

Kelimion:
Thanks for the changes. I'll try it under all edge cases I know of under Win/Lin/Mac tomorrow.

Kelimion:
Looks really, really solid, @Naught00. 👍 

One edge-case left: `PWD: /` (or `C:\`) for Windows.
On Windows and Linux it reports "LLVM: Permission denied". On Mac it fails with "LLVM: Read-only file system".

It is allowed when you `sudo odin` the thing on Linux.

I know it's a TOCTOU issue, but I think we should try opening the file for writing without truncation and closing it again. If that succeeds we have the rights, regardless of the path or user. If it fails we can warn the user gracefully instead of failing through LLVM.

What do you think, @gingerBill?

Naught00:
Is the "LLVM: Permission denied" error related to these changes? I get the same message on the latest commit from master.

If I try to actually build the root directory as a package with:
```
odin build /
```

the last_path_element() function chops off the / and this assertion fires on line 1412 in build_settings.cpp:
```
               String build_project_name  = last_path_element(bc->build_paths[BuildPath_Main_Package].basename);
               GB_ASSERT(build_project_name.len > 0);
               bc->ODIN_BUILD_PROJECT_NAME = build_project_name;
```
However, these both seem like separate issues.

Kelimion:
> Is the "LLVM: Permission denied" error related to these changes? I get the same message on the latest commit from master.
That's exactly my point. ;-)

> However, these both seem like separate issues.
Yes and no.

Originally the output file path was generated at the linker stage, after the compiler had done all the work. The patches that you're trying to fix moved all that work to `build_settings.cpp` to give the user feedback as soon as possible and not let LLVM be the complainant at the end of the process.

Your patch addresses the extension part of it and checks whether it conflicts with a directory. It doesn't address the one remaining check: is the destination writable? You're correct that the current master branch also doesn't check this, but it's a natural extension (pun intended) of the patch set you're patching's intent to fail gracefully early.

Do you want to add the check or shall I merge this PR and add the write permission check? I don't mind either way.

src/build_settings.cpp
```diff
@@ -1624,6 +1625,22 @@ gb_internal bool init_build_paths(String init_filename) {
 		return false;
 	}
 
+	if (path_is_directory(bc->build_paths[BuildPath_Output])) {
+		String output_file = path_to_string(ha, bc->build_paths[BuildPath_Output]);
+		defer (gb_free(ha, output_file.text));
+		gb_printf_err("Output path %.*s is a directory.\n", LIT(output_file));
+		return false;
+	}
+	//nocheckin char const *pathname = (char *)bc->build_paths[BuildPath_Output].basename.text;
```
Kelimion: Oops?

Naught00: Yes, just fixed

Naught00:
I was going to make write_directory() analogous to read_directory() but used a bool instead of an enum since we are only checking for a permission error.

PR #2501, opened 2023-04-30
Fix #2489

### LINKED ISSUE #2489: Bug trying to use setsockopt with a duration (timeval) due to confusion between microseconds and nanoseconds on Linux
## Context

I am writing code using a TCP socket reading from the network on Linux, and attempting to set a read timeout on it, using `setsockopt` and `SO_RCVTIMEO`.

According to `man 7 socket`: 

> The argument is a struct timeval.

which seems to be handled correctly by `core/net/socket_linux.odin`:

```odin
	case
		.Linger,
		.Send_Timeout,
		.Receive_Timeout:
			t, ok := value.(time.Duration)
			if !ok do panic("set_option() value must be a time.Duration here", loc)

			nanos := time.duration_nanoseconds(t)
			timeval_value.nanoseconds = int(nanos % 1e9)
			timeval_value.seconds = (nanos - i64(timeval_value.nanoseconds)) / 1e9

			ptr = &timeval_value
			len = size_of(timeval_value)

```
And in my user code, I try to pass a small timeout: `1*time.Millisecond` (see at the bottom for the minial code reproducing the bug).

However, the system call returns `EDOM` when running the code: `setsockopt(3, SOL_SOCKET, SO_RCVTIMEO_NEW, "\0\0\0\0\0\0\0\0@B\17\0\0\0\0\0", 16) = -1 EDOM (Numerical argument out of domain)`

The linux code returning this `EDOM` is located in `net/core/sock.c` in the linux source tree:

```c
static int sock_set_timeout(long *timeo_p, sockptr_t optval, int optlen,
			    bool old_timeval)
{
	struct __kernel_sock_timeval tv;
	int err = sock_copy_user_timeval(&tv, optval, optlen, old_timeval);

	if (err)
		return err;

	if (tv.tv_usec < 0 || tv.tv_usec >= USEC_PER_SEC) // <----- This line most likely returns our error
		return -EDOM;

// [...]
```
 
and indeed, `man structval` shows:

```c
       struct timeval {
           time_t       tv_sec;   /* Seconds */
           suseconds_t  tv_usec;  /* Microseconds */
       };
```

And the Odin code is trying to treat the second field as nanoseconds, instead of microseconds. In `core/os/os_linux.odin`:

```odin
Timeval :: struct {
	seconds: i64,
	nanoseconds: int,
}
```

I suggest fixing this second field by making it microseconds and fixing all locations using it to have the correct value (i.e. unit), or adding a new struct like this:

```odin
TimevalMicroseconds :: struct {
	seconds: i64,
	microseconds: int,
}
```
and using it when preparing arguments for the syscall `setsockopt`.

```
        Odin: dev-2023-04:f2ec4381
        OS:   Debian GNU/Linux 12 (bookworm), Linux 6.1.0-7-amd64
        CPU:  11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
        RAM:  15673 MiB
```

## Expected Behavior

The `setsockopt` syscall works and returns 0.

### Steps to Reproduce

```odin
package main

import "core:net"
import "core:fmt"
import "core:os"
import "core:time"

main :: proc() {
  socket, err := net.dial_tcp_from_hostname_with_port_override("localhost", int(8000))
  if err != nil {
    fmt.eprintf("Err creating socket: %s", err)
    os.exit(1)
  }
  defer net.close(socket)

  err = net.set_option(socket, net.Socket_Option.Receive_Timeout, 1*time.Millisecond)
  if err != nil {
    fmt.eprintf("Err setsockopt: %s", err)
    os.exit(1)
  }
}
```

Output:

```
Err setsockopt: %!(BAD ENUM VALUE=33)%
```

Strace output:

```
[...]
socket(AF_INET, SOCK_STREAM, IPPROTO_TCP) = 3
setsockopt(3, SOL_SOCKET, SO_REUSEADDR, [1], 4) = 0
connect(3, {sa_family=AF_INET, sin_port=htons(8000), sin_addr=inet_addr("127.0.0.1")}, 128) = 0
setsockopt(3, SOL_TCP, TCP_NODELAY, [1], 4) = 0
setsockopt(3, SOL_SOCKET, SO_RCVTIMEO_NEW, "\0\0\0\0\0\0\0\0@B\17\0\0\0\0\0", 16) = -1 EDOM (Numerical argument out of domain)
write(2, "Err setsockopt: ", 16Err setsockopt: )        = 16
[...]
```

PS: Ah, and the `EDOM` errno code is missing from `core/os/os_linux.odin`, it should be added with the value `33`:

```odin
EPIPE:          Errno : 32
                                         // <--- EDOM here
ERANGE:         Errno : 34 /* Result too large */
```

JamesDSource:
I've basically just implemented a solution provided in the issue. I've also verified that it worked in a personal project of mine.


JamesDSource:
@gingerBill  Any reason why this was closed? Is the fix not good or is there something I missed?

jasonKercher:
I can't speak for Bill, but I can say that the issue does correctly point out that `struct timeval` is seconds and microseconds and not seconds and nanoseconds.  I think adding a new struct with a different name may be confusing since our `os.timeval` is indeed incorrect.

JamesDSource:
Ok, That makes sense to me. I guess one thing that I should have checked is how often that struct is used, and a quick looks shows it's only for Darwin and Linux socket timeout values. According to apple [docs](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getsockopt.2.html#//apple_ref/doc/man/2/getsockopt) it looks like they also take microseconds, so it's fine to change the existing struct completely. I'll submit another request for that.

PR #2513, opened 2023-05-05
add time.weekday proc

laytan:


core/time/time.odin
```diff
@@ -170,6 +170,12 @@ day :: proc "contextless" (t: Time) -> (day: int) {
 	return
 }
 
+weekday :: proc "contextless" (t: Time) -> (weekday: Weekday) {
+    abs := _time_abs(t)
+	sec := (abs + u64(Weekday.Monday) * SECONDS_PER_DAY) % SECONDS_PER_WEEK
+	return Weekday(int(sec) / SECONDS_PER_DAY)
```
Kelimion, line 176: Inconsistent indentation. Please use tabs for indentation, spaces for alignment.

Kelimion:
While I like the idea of a proc like this to get the weekday, an even more useful way to provide it would be to have a `to_julian` proc that provides an `f64` with a Julian date (where the fractional part is the time of day).

The day of week helper proc only has to % 7 and add an offset to bias back to Sunday then, and you have the useful Julian date measure you can use to calculate the difference in days between two dates with little hassle.

With that said, I have all that implemented in a datetime package that'll drop in the not so distant future. We can work with your version of day of week until that lands.

laytan:
@Kelimion Sounds good, looking forward to your datetime package! I have fixed the tabs/spacing.

PR #2520, opened 2023-05-09
Allow for custom sleep duration in tsc frequency fallback

matias-eduardo:
This gives the user more control over the spectrum of precision vs. load time on Windows. Even with lower sleep time, Spall's output is still very useful.

NOTE: A better API might be to allow the user to pass the `freq` as a param to `create_context` in case they already paid for it beforehand, but this seems fine for now.


colrdavidson:
Looks reasonable to me. I like the suggestion to pass the frequency as an option too, it was definitely a thorn in my side to need to use a sleep on Windows. :/

I've been thinking about adding an optional timestamp event when you begin and close the trace to allow the frontend to guess the frequency without needing the sleep, but it is definitely safer to have a fallback if the end of the trace fails to get logged on program crash.

Kelimion:
> This gives the user more control over the spectrum of precision vs. load time on Windows. Even with lower sleep time, Spall's output is still very useful.
> 
> NOTE: A better API might be to allow the user to pass the `freq` as a param to `create_context` in case they already paid for it beforehand, but this seems fine for now.

You can still do that.
`context_create_with_frequency :: proc(filename: string, frequency: f64) -> (...)`
`context_create_with_sleep :: proc(filename: string, sleep := 2 * time.Second) -> (...)`
`context_create :: proc{context_create_with_sleep, context_create_with_frequency}`

Where the latter calls the former after probing the frequency.

matias-eduardo:
Yes, that's a better approach. It would allow for what cloin mentioned about storing the value in a file and reusing it from there, which seems a whole lot more sensible than adding any amount of sleep to each run.

PR #2521, opened 2023-05-09
remove misleading defer if example

elusivePorpoise:
Using defer if cannot be used to change the return value (without an extra scope) as shown in the example, Should be removed.

### DIFF
--- a/examples/demo/demo.odin
+++ b/examples/demo/demo.odin
@@ -2145,6 +2145,10 @@ or_return_operator :: proc() {
 			return -345 * z, zerr
 		}
 
+		defer if err != nil {
+			fmt.println("Error in", #procedure, ":" , err)
+		}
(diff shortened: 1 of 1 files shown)


examples/demo/demo.odin
```diff
@@ -2145,12 +2145,6 @@ or_return_operator :: proc() {
 			return -345 * z, zerr
 		}
 
-		// If the other return values need to be set depending on what the end value is,
-		// the 'defer if' idiom is can be used
-		defer if err != nil {
-			n = -1
```
gingerBill, line 2151: I would have just changed this specific line to something like `fmt.println("an error has occured", err)` or whatever, and not completely remove the entire thing.

gingerBill (changes requested):

PR #2533, opened 2023-05-12
add encoding/hex and use it to expand the percent decoding chars

laytan:
The percent_decode proc had a bunch of encodings that were not being decoded, when looking into it and testing, decoding it using hex is enough and covers everything I tried.

Then I felt like hex encoding and decoding has a place in the encoding/ namespace.

### DIFF
--- a/core/encoding/hex/hex.odin
+++ b/core/encoding/hex/hex.odin
@@ -1,0 +1,74 @@
+package hex
+
+import "core:strings"
+
+encode :: proc(src: []byte, allocator := context.allocator) -> []byte #no_bounds_check {
+	dst := make([]byte, len(src) * 2, allocator)
+    for i := 0; i < len(src); i += 1 {
+		v := src[i]
+        dst[i]   = HEXTABLE[v>>4]
+        dst[i+1] = HEXTABLE[v&0x0f]
+        i += 2
+    }
+
+	return dst
+}
+
+
+decode :: proc(src: []byte, allocator := context.allocator) -> (dst: []byte, ok: bool) #no_bounds_check {
+	if len(src) % 2 == 1 {
+		return
+	}
+
+	dst = make([]byte, len(src) / 2, allocator)
+	for i, j := 0, 1; j < len(src); j += 2 {
+		p := src[j-1]
+		q := src[j]
+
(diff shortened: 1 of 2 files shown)


core/encoding/hex/hex.odin
```diff
@@ -0,0 +1,74 @@
...
+		dst[i] = (a << 4) | b
+		i += 1
+	}
+
+	return dst, true
+}
+
+// Decodes the given sequence into one byte.
+// Should be called with one rune worth of the source, eg: 0x23 -> '#'.
+decode_sequence :: proc(str: string) -> (byte, bool) {
```
Kelimion, line 40: This can be simplified a bit if we take advantage of the sequence having a length of 2 without the prefix. And while we're at it, let's also handle the uppercase prefix seeing as we handle uppercase hex digits.

```odin
decode_sequence :: proc(str: string) -> (res: byte, ok: bool) {
	str := str
	if strings.has_prefix(str, "0x") || strings.has_prefix(str, "0X") {
		str = str[2:]
	}

	if len(str) != 2 {
		return 0, false
	}

	upper := hex_digit(str[0]) or_return
	lower := hex_digit(str[1]) or_return

	return upper << 4 | lower, true
}
```

core/encoding/hex/hex.odin
```diff
@@ -0,0 +1,74 @@
+package hex
+
+import "core:strings"
+
+encode :: proc(src: []byte, allocator := context.allocator) -> []byte #no_bounds_check {
```
Kelimion, line 5: Please fix the indentation in this proc.

core/encoding/hex/hex.odin
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		val += u8(hd) << uint(4 * index)
+	}
+
+	return val, true
+}
+
+@(private)
+HEXTABLE := [16]byte {
```
Kelimion, line 58: Odin uses tabs for indentation, spaces for alignment. This file seems to use 4 spaces interchangeably with a tab. Kindly fix.

Kelimion (changes requested):
Good effort.

Kelimion:
Spurious error we see in CI sometimes.

PR #2535, opened 2023-05-13
Fix #2515 - Implement backward shift on `map` on insert and reseed hashes on resize

### LINKED ISSUE #2515: Map sometimes is just wrong
## Context

Happens both on windows and linux.

```odin version dev-2023-05:9f392097```

In my code I had a weird bug where sometimes tracking allocator had entries from dynamic array, that one line after declaration had `defer delete(...)`. Sometimes my program would just straight up stop working because it got into an infinite loop in `runtime.map_insert_hash_dynamic`. I was able to make a reproducible examples for both of these bugs, unfortunately they are pretty big.

### Steps to Reproduce

Use following `main.odin`:
```odin
package main

import "core:mem"
import "core:math/rand"
import "core:fmt"
import "core:os"

main :: proc() {
   data, data_ok := os.read_entire_file("logs.txt")
   if !data_ok {
      fmt.panicf("Error: couldn't read the logs.txt")
   }
   defer delete(data)


   allocation_map: map[rawptr]struct{}
   defer delete(allocation_map)

   bugged_loc :: rawptr(uintptr(0x104D078))

   data_iter := data
   for len(data_iter) > 0 {
      // Takes from a buffer size_of(T) bytes and returns them as a T
      read_value :: proc(data_iter: []u8, $T: typeid) -> (data_iter_result: []u8, value: T) {
         elem_slice := mem.ptr_to_bytes(&value)
         copy_slice(dst = elem_slice, src = data_iter[:len(elem_slice)])
         data_iter_result = data_iter[len(elem_slice):]
         return
      }

      data_type: u8
      data_iter, data_type = read_value(data_iter, u8)

      switch data_type {
         case 'S':
            memory_loc: rawptr
            memory_loc_0: u64le
            data_iter, memory_loc_0 = read_value(data_iter, u64le)
            memory_loc = cast(rawptr)cast(uintptr)memory_loc_0

            if (bugged_loc == memory_loc) {
               fmt.println("ASSIGN:", memory_loc)
            }

            allocation_map[memory_loc] = {}
         case 'D':
            memory_loc: rawptr
            data_iter, memory_loc = read_value(data_iter, rawptr)

            if (bugged_loc == memory_loc) {
               fmt.println("DELETE")
            }

            delete_key(&allocation_map, memory_loc)

         case:
            fmt.panicf("This shouldn't happen!\n")
      }
   }

   for elem, _ in allocation_map {
      fmt.println("PRESENT KEY:", elem)
   }
}
```

This program just opens a `logs.txt` and executes operations on the map stored in this file. It also prints every operation on the map that has `bugged_loc` key.

[logs.txt](https://github.com/odin-lang/Odin/files/11413208/logs.txt)

This file prints:
```
ASSIGN: 0x104D078
DELETE
ASSIGN: 0x104D078
DELETE
ASSIGN: 0x104D078
ASSIGN: 0x104D078
ASSIGN: 0x104D078
DELETE
ASSIGN: 0x104D078
DELETE
PRESENT KEY: 0x104D078
PRESENT KEY: 0x6EB258
PRESENT KEY: 0x757888
PRESENT KEY: 0x1047A38
```

Which makes no sense, since we've just deleted the `bugged_loc=0x104D078` from our map and didn't assign it again.

[logs2.txt](https://github.com/odin-lang/Odin/files/11413209/logs2.txt)

This one just stops working. GDB shows that we're stuck here:

```
...
#2  0x000000000040b26d in runtime.map_insert_hash_dynamic (m=0x7fffffffd810, info=0x443b50, h=8753646848103601834, ik=140737488345048, iv=140737488344904)
    at /home/imaginary/Programy/Odin/core/runtime/dynamic_map_internal.odin:412
#3  0x00000000004107cf in runtime.__dynamic_map_set (m=0x7fffffffd810, info=0x443b50, hash=8753646848103601834, key=0x7fffffffd7d8 "H\243\t\001", value=0x7fffffffd748 "", loc=...)
    at /home/imaginary/Programy/Odin/core/runtime/dynamic_map_internal.odin:788
#4  0x000000000040fea0 in main.main () at main.odin:43
```

And we're infinite looping in `runtime.map_insert_hash_dynamic`.

jasonKercher:
This fixes issue #2515 and changes the `map` implementation in a couple ways.

The original issue stems from [map_insert_hash_dynamic](https://github.com/odin-lang/Odin/blob/master/core/runtime/dynamic_map_internal.odin#LL402C1-L421C5):

```odin
		if map_hash_is_empty(element_hash) {
			k_dst := map_cell_index_dynamic(ks, info.ks, pos)
			v_dst := map_cell_index_dynamic(vs, info.vs, pos)
			intrinsics.mem_copy_non_overlapping(rawptr(k_dst), rawptr(k), size_of_k)
			intrinsics.mem_copy_non_overlapping(rawptr(v_dst), rawptr(v), size_of_v)
			hp^ = h

			return result if result != 0 else v_dst
		}

		if probe_distance := map_probe_distance(m^, element_hash, pos); distance > probe_distance {
			if map_hash_is_deleted(element_hash) {
				k_dst := map_cell_index_dynamic(ks, info.ks, pos)
				v_dst := map_cell_index_dynamic(vs, info.vs, pos)
				intrinsics.mem_copy_non_overlapping(rawptr(k_dst), rawptr(k), size_of_k)
				intrinsics.mem_copy_non_overlapping(rawptr(v_dst), rawptr(v), size_of_v)
				hp^ = h

				return result if result != 0 else v_dst
			}
```

If there are no empty slots, elements can fall so that  `distance` goes beyond capacity and falsely inserts into a bad deleted slot.  Essentially, the deleted slots end up stealing some space.  This only becomes a problem when `len` gets close to `cap`, and deleted entries exist.

The solution I used may not be ideal, but I implemented the "backward shift" for erasing keys that was mentioned in the comments:

**Pros**

* likely faster lookups after some keys have been deleted due to smaller distances
* simpler insertion logic

**Cons**

* Likely slower `delete_key`
* Can no longer use `delete_key` while iterating through the map

NOTE: someone should double check the change to the compiler source.

**Other possible solutions**

* Play with parameters like `MAP_LOAD_FACTOR`
* Have `map_insert_hash_dynamic` grow if distance exceeds cap

### DIFF
--- a/core/runtime/dynamic_map_internal.odin
+++ b/core/runtime/dynamic_map_internal.odin
@@ -610,6 +610,7 @@ map_shrink_dynamic :: proc "odin" (#no_alias m: ^Raw_Map, #no_alias info: ^Map_I
 
 		k := map_cell_index_dynamic(ks, info.ks, i)
 		v := map_cell_index_dynamic(vs, info.vs, i)
+		hash = info.key_hasher(rawptr(k), map_seed(shrunk))
 		_ = map_insert_hash_dynamic(&shrunk, info, hash, k, v)
 		// Only need to do this comparison on each actually added pair, so do not
 		// fold it into the for loop comparator as a micro-optimization.
--- a/core/runtime/dynamic_map_internal.odin
+++ b/core/runtime/dynamic_map_internal.odin
@@ -394,32 +394,71 @@ map_insert_hash_dynamic :: proc "odin" (#no_alias m: ^Raw_Map, #no_alias info: ^
 	tk := map_cell_index_dynamic(sk, info.ks, 1)
 	tv := map_cell_index_dynamic(sv, info.vs, 1)
 
-
 	for {
 		hp := &hs[pos]
 		element_hash := hp^
 
 		if map_hash_is_empty(element_hash) {
-			k_dst := map_cell_index_dynamic(ks, info.ks, pos)
-			v_dst := map_cell_index_dynamic(vs, info.vs, pos)
-			intrinsics.mem_copy_non_overlapping(rawptr(k_dst), rawptr(k), size_of_k)
-			intrinsics.mem_copy_non_overlapping(rawptr(v_dst), rawptr(v), size_of_v)
+			kp := map_cell_index_dynamic(ks, info.ks, pos)
+			vp := map_cell_index_dynamic(vs, info.vs, pos)
+			intrinsics.mem_copy_non_overlapping(rawptr(kp), rawptr(k), size_of_k)
+			intrinsics.mem_copy_non_overlapping(rawptr(vp), rawptr(v), size_of_v)
 			hp^ = h
 
-			return result if result != 0 else v_dst
+			return result if result != 0 else vp
+		}
+
+		if map_hash_is_deleted(element_hash) {
+			next_pos := (pos + 1) & mask
+
+			// backward shift
+			for !map_hash_is_empty(hs[next_pos]) {
+				probe_distance := map_probe_distance(m^, hs[next_pos], next_pos)
+				if probe_distance == 0 {
+					break
+				}
+				probe_distance -= 1
+
+				kp := map_cell_index_dynamic(ks, info.ks, pos)
+				vp := map_cell_index_dynamic(vs, info.vs, pos)
+				kn := map_cell_index_dynamic(ks, info.ks, next_pos)
+				vn := map_cell_index_dynamic(vs, info.vs, next_pos)
+
+				if distance > probe_distance {
+					if result == 0 {
+						result = vp
+					}
+					// move stored into pos; store next
+					intrinsics.mem_copy_non_overlapping(rawptr(kp), rawptr(k), size_of_k)
+					intrinsics.mem_copy_non_overlapping(rawptr(vp), rawptr(v), size_of_v)
+					hs[pos] = h
+
+					intrinsics.mem_copy_non_overlapping(rawptr(k), rawptr(kn), size_of_k)
+					intrinsics.mem_copy_non_overlapping(rawptr(v), rawptr(vn), size_of_v)
+					h = hs[next_pos]
+				} else {
+					// move next back 1
+					intrinsics.mem_copy_non_overlapping(rawptr(kp), rawptr(kn), size_of_k)
+					intrinsics.mem_copy_non_overlapping(rawptr(vp), rawptr(vn), size_of_v)
+					hs[pos] = hs[next_pos]
+					distance = probe_distance
+				}
+				hs[next_pos] = 0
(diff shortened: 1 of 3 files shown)


graphitemaster:
As the author of the dynamic map implementation this is very good work. I put off implementing it because I wasn't quite sure how to do it with the dynamic implementation, so good job.

This PR just needs one minor change that the backwards shift deletion approach has in terms of a gotcha and that is when copying one hash map into another, or creating a derived map with the same keys from an existing one (which happens on resize but also if people copy the map). The RH+backwards shift deletion has a uniquely determined layout that only depends on the keys currently in the map (which is good, because it means that the structure is as if any deleted keys had never existed), but that layout is effectively sorted by **increasing hash**, which means that for this implementation, if you iterate over a hash table and create a new hash table with the same keys, they're going to be precisely those same keys that want to map to the first few buckets.

This operation degrades to O(N^2) because you keep having pathologically high collision counts all the way through (until you grow the new hash to the right size to space them out that is, but that happens on the last few grows)

The approach to fix this is to use a seeded hash function for each map. This is what e.g Rust does. It's a good approach. Instead of storing an actual seed though which would increase the size of a map type, we should rely on the fact that the memory address for the backing storage is somewhat random. I suggest using a mixing step though since the address is always 64 byte aligned (lot of zero bits there), maybe splitmix64 or the tempering function in mersenne twister.

Other than that this looks great, good work.

jasonKercher:
Hey graphitemaster, thank you for the kind words.  I sort of only took on the issue because I was curious about how the new map implementation works.  It looked a bit intimidating at first, but once I got more familiar with it, it is actually fairly simple and elegant =]

I was not aware of that collision issue, and I will research it.  I like the idea of getting a seed from the address, and I share the same concern for using it as likely only a couple bytes will change between resizes.  Would we not have space to store a seed behind the swap space?  I noticed it allocates a bit more than what the map was using in total.

Kelimion:
> Would we not have space to store a seed behind the swap space? I noticed it allocates a bit more than what the map was using in total.

You could do that, but that's another cache line that has to be brought in before you can calculate a hash. One that may not otherwise need to be accessed.

gingerBill:
> Can no longer use delete_key while iterating through the map

That's not a good thing 😬

jasonKercher:
Okay, I think it's done now.  `delete_key` is iteration safe, and there is a simple backward shift in the insert, 

Also, the seed is now randomized by the base address of the map data.

Would be nice if someone double checked the changes in `lb_map_set_proc_for_type`.  That function is not currently being used, so ideally, I didn't just leave it broken =]

gingerBill:
The original issue comment for this PR now is out of date however,

PR #2537, opened 2023-05-14
Join URL queries with &

CORDEA:
Hello,

I noticed that `net.join_url` joins queries without a separator. For instance,
```odin
net.join_url("http", "example.com", "", {"a" = "b", "c" = "d"})
```
returns `http://example.com?a=bc=d`. `net.split_url` splits the query with `&`, so I think the return value of `net.join_url` is also expected to have `&` as a separator.

Could you review if my PR is acceptable? :)

### DIFF
--- a/core/net/url.odin
+++ b/core/net/url.odin
@@ -78,13 +78,19 @@ join_url :: proc(scheme, host, path: string, queries: map[string]string, allocat
 	}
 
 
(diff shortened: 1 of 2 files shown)


Kelimion:
Good catch. Thanks for turning it into a test case as well.

PR #2539, opened 2023-05-14
lua MAXSTACK should be 1000000 on 32 bits OR greater

powerc9000:
Code only checks if 4 bytes for `rawptr` size. However lua defines the macro (that I assume the odin code is based on) as:

```
/*
@@ LUAI_IS32INT is true iff 'int' has (at least) 32 bits.
*/
#define LUAI_IS32INT	((UINT_MAX >> 30) >= 3)
```

This error broke `upvalues` because it would be looking for the wrong index for lua to find them at.


gingerBill:
Can you also make this work for all variants of lua too?

PR #2554, opened 2023-05-22
add mysql client bindings

laytan:
Had to write mysql bindings the other day, thought it would fit nicely in vendor.
Also contains a couple wrappers, mainly around time and prepared statements.

I am not sure what the requirements and rules are for adding bindings to vendor, so just close this if you don't want this here.


gingerBill:
I do not think MySQL is appropriate for the `vendor` library collections for reasons I do not want to get into at the moment.

However, I do recommend you posting your bindings to somewhere like https://github.com/odin-lang/Odin/wiki/Odin-Libs.

PR #2557, opened 2023-05-27
fix weird errno returned on darwin

laytan:
The errno being returned was VERY large in the below code, after fiddling with it, this fixed it (not entirely sure why).

I suspect this issue is also in other places for os_darwin but this is the code I tested it with:

```odin
package http

import "core:net"
import "core:os"
import "core:fmt"
import "core:c"

main :: proc() {
   server, err := net.create_socket(.IP4, .TCP)
   assert(err == nil, "create error")

   endpoint := net.Endpoint{
       address = net.IP4_Loopback,
       port = 3131,
   }
   err = net.bind(server, endpoint)
   assert(err == nil, "bind error")

   err = net.set_blocking(server, false)
   assert(err == nil, "set blocking error")

   errn := os.listen(os.Socket(server.(net.TCP_Socket)), 1000)
   assert(errn == 0, "listen error")

   fmt.println(accept(server.(net.TCP_Socket)))
   fmt.println(os.EWOULDBLOCK)
}

accept :: proc(server: net.TCP_Socket) -> (os.Errno) {
	sockaddr: os.SOCKADDR_STORAGE_LH
	sockaddrlen := c.int(size_of(sockaddr))

	// This os.accept call returned 7912543070313250851, after the fix, it returns the expected 35 for EWOULDBLOCK.
	_, err := os.accept(os.Socket(server), cast(^os.SOCKADDR) &sockaddr, &sockaddrlen)
	return err
}
```

### DIFF
--- a/core/os/os_darwin.odin
+++ b/core/os/os_darwin.odin
@@ -413,7 +413,7 @@ F_OK :: 0 // Test for file existance
 F_GETPATH :: 50 // return the full path of the fd
 
 foreign libc {
-	@(link_name="__error") __error :: proc() -> ^int ---
+	@(link_name="__error") __error :: proc() -> ^c.int ---
 
 	@(link_name="open")             _unix_open          :: proc(path: cstring, flags: i32, mode: u16) -> Handle ---
 	@(link_name="close")            _unix_close         :: proc(handle: Handle) -> c.int ---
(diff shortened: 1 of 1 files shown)


Kelimion:
Makes total sense.

jasonKercher:
Hey, that's a good catch.  It seems to be wrong in os_linux.odin, os_openbsd.odin and os_freebsd.odin as well.  Maybe you want to fix those, too?

PR #2563, opened 2023-05-29
WASM: Use BasicC ABI convention for arguments that have `proc "c"`

charles-l:
This patch enables ABI integration with clang-compiled WASM code which uses the calling convention specified in the [BasicC ABI spec](https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md#function-signatures). Most notably, Odin's WASM calling convention passes structs directly if they're smaller than `MAX_DIRECT_STRUCT_SIZE`, but the BasicC convention always passes non-singular structs by reference.

Test case with motivating example:

```odin
// testcase.odin
package main

S :: struct {
    a: i32,
    b: i32,
    c: i32,
}

@export
test_func :: proc "c" (data: S) -> i32 {
    return data.b
}
```

```c
// test.c
#include <stdio.h>

struct S {
    int a;
    int b;
    int c;
};

int test_func(struct S data);

int main() {
    printf("hello world %i\n", test_func((struct S){1,2,3}));
    return 0;
}
```

Without the patch in this PR, the code compiles with a linker warning and executing the code throws an exception. 

```
$ odin build . -no-entry-point -build-mode:obj -target=wasi_wasm32 && emcc test.c testcase.wasm.o -o test.html
wasm-ld: warning: function signature mismatch: test_func
>>> defined as (i32) -> i32 in /tmp/emscripten_temp_q7ego2kl/test_0.o
>>> defined as (i32, i32, i32) -> i32 in testcase.wasm.o
```
```
$ node ./test.js 
/home/nc/src/Odin/testcase/test.js:130
      throw ex;
      ^

RuntimeError: unreachable
    at <anonymous>:wasm-function[2]:0x291
    at <anonymous>:wasm-function[3]:0x314
    at main (<anonymous>:wasm-function[4]:0x352)
    at /home/nc/src/Odin/testcase/test.js:737:22
    at callMain (/home/nc/src/Odin/testcase/test.js:1614:15)
    at doRun (/home/nc/src/Odin/testcase/test.js:1664:23)
    at run (/home/nc/src/Odin/testcase/test.js:1679:5)
    at runCaller (/home/nc/src/Odin/testcase/test.js:1599:19)
    at removeRunDependency (/home/nc/src/Odin/testcase/test.js:647:7)
    at receiveInstance (/home/nc/src/Odin/testcase/test.js:878:5)
```

But with the patch, it compiles and runs as expected:
```
$ ~/dev/Odin/odin build . -no-entry-point -build-mode:obj -target=wasi_wasm32 && emcc test.c testcase.wasm.o -o test.html
$ node test.js 
hello world 2
```

### DIFF
--- a/src/llvm_abi.cpp
+++ b/src/llvm_abi.cpp
@@ -1213,7 +1213,7 @@ namespace lbAbiWasm {
 		            The approach taken optimizes for passing things in multiple
 		            registers/arguments if possible rather than by pointer.
 	*/
-	gb_internal Array<lbArgType> compute_arg_types(LLVMContextRef c, LLVMTypeRef *arg_types, unsigned arg_count);
+	gb_internal Array<lbArgType> compute_arg_types(LLVMContextRef c, LLVMTypeRef *arg_types, unsigned arg_count, ProcCallingConvention calling_convention);
 	gb_internal LB_ABI_COMPUTE_RETURN_TYPE(compute_return_type);
 
 	enum {MAX_DIRECT_STRUCT_SIZE = 32};
@@ -1221,7 +1221,7 @@ namespace lbAbiWasm {
 	gb_internal LB_ABI_INFO(abi_info) {
 		lbFunctionType *ft = gb_alloc_item(permanent_allocator(), lbFunctionType);
 		ft->ctx = c;
-		ft->args = compute_arg_types(c, arg_types, arg_count);
+		ft->args = compute_arg_types(c, arg_types, arg_count, calling_convention);
 		ft->ret = compute_return_type(ft, c, return_type, return_is_defined, return_is_tuple);
 		ft->calling_convention = calling_convention;
 		return ft;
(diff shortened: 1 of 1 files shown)


src/llvm_abi.cpp
```diff
@@ -1258,13 +1258,24 @@ namespace lbAbiWasm {
...
-		if (sz <= MAX_DIRECT_STRUCT_SIZE) {
+		if (calling_convention == ProcCC_CDecl) {
+			// WASM Basic C ABI:
+			// https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md#function-signatures
+			if (kind == LLVMArrayTypeKind) {
+				return false;
+			} else if (kind == LLVMStructTypeKind) {
+				unsigned count = LLVMCountStructElementTypes(type);
+				if (count == 1) {
+					return true;
```
gingerBill, line 1275: You'll need to check this recursively and not just if it has one element.

charles-l: Ah, good point -- updated it to check recursively in the case of nested types.

PR #2566, opened 2023-05-31
Added useful allocators to `mem`. (profiling allocator, lightweight tracking allocator, testing allocator)

jakubtomsu:
I implemented 3 new wrapper allocators which I've been using on my own projects for a while. I think they could be useful to have in the `core` library.

Note: I'm not sure if this PR should've been split up into 3 smaller PRs, but if there are issues with one of the allocators I can just remove them.

# Counting allocator
Counting allocator is basically a lightweight tracking allocator, which increments a counter when allocating and decrements it when freeing. This is considerably simpler than `Tracking_Allocator` when all you need to know is if there were any memory leaks or not.

### My use case:
Tracking allocations for each entity in a game separately, and validating memory leaks when the entity is destroyed. `Tracking_Allocator` per entity would add a lot of overhead since there can be thousands of entities.

# Profiling allocator
Profiling allocator counts total allocations, frees, etc. and measures the total time spent in allocator calls.

### My use case:
- Profiling custom data structures for unnecessary allocations.
- Profiling a whole program to make sure allocations are kept to minimum.

### Usage
When you're done using the allocator:
```odin
for info, mode in prof_allocator.modes {
  printf("[% 16v] count: % 2i, total: % 5.1f us, avg: % 8.3f ns\n", mode, info.count,
            time.duration_microseconds(info.time),
            info.count == 0 ? 0 : f64(time.duration_nanoseconds(info.time)) / f64(info.count))
}
```
Output is something like this:
```
[           Alloc] count: 10, total:  12.9 us, avg: 1290.000 ns
[            Free] count: 70, total: 103.7 us, avg: 1481.429 ns
[        Free_All] count:  0, total:   0.0 us, avg:    0.000 ns
[          Resize] count: 60, total:  79.3 us, avg: 1321.667 ns
[  Query_Features] count:  0, total:   0.0 us, avg:    0.000 ns
[      Query_Info] count:  0, total:   0.0 us, avg:    0.000 ns
[Alloc_Non_Zeroed] count: 60, total:  73.8 us, avg: 1230.000 ns
```

Note: it would be very useful to track each allocation using a real profiler (like Spall), but this is probably better for the `core` library.

Note: this adds a dependency on `core:time`.

# ~Testing allocator~
~Testing allocator is useful for testing edge-case scenarios when `alloc` or `resize` fails. It occasionally returns an error (based on a chance value), which tests the error handling code paths. This can be very useful for making sure some code doesn't crash in weird memory conditions.~

### DIFF
--- a/core/math/rand/rand.odin
+++ b/core/math/rand/rand.odin
@@ -33,6 +33,10 @@ Example:
 		fmt.println(rand.uint64())
 	}
 
+Possible Output:
+
+	10
+
 */
 set_global_seed :: proc(seed: u64) {
 	init(&global_rand, seed)
@@ -55,6 +59,10 @@ Example:
 		my_rand := rand.create(1)
 		fmt.println(rand.uint64(&my_rand))
 	}
+
+Possible Output:
+
+	10
 
 */
 @(require_results)
@@ -82,6 +90,10 @@ Example:
 		fmt.println(rand.uint64(&my_rand))
 	}
 
+Possible Output:
+
+	10
+
 */
 init :: proc(r: ^Rand, seed: u64) {
 	r.state = 0
@@ -98,7 +110,7 @@ On `linux` refer to the `getrandom` syscall.
 On `darwin` refer to `getentropy`.  
 On `windows` refer to `BCryptGenRandom`.
 
-All other platforms wi
+All other platforms are not supported
 
 Inputs:
 - r: The random number generator to use the system random number generator
@@ -114,6 +126,11 @@ Example:
 		rand.init_as_system(&my_rand)
 		fmt.println(rand.uint64(&my_rand))
 	}
+
+Possible Output:
+
+	10
+
 */
 init_as_system :: proc(r: ^Rand) {
 	if !#defined(_system_random) {
@@ -164,6 +181,12 @@ Example:
 		my_rand := rand.create(1)
 		fmt.println(rand.uint32(&my_rand))
 	}
+
+Possible Output:
+
+	10
+	389
+
 */
 @(require_results)
 uint32 :: proc(r: ^Rand = nil) -> (val: u32) { return u32(_random_u64(r)) }
@@ -188,6 +211,12 @@ Example:
 		my_rand := rand.create(1)
 		fmt.println(rand.uint64(&my_rand))
 	}
+
+Possible Output:
+
+	10
+	389
+
 */
 @(require_results)
 uint64 :: proc(r: ^Rand = nil) -> (val: u64) { return _random_u64(r) }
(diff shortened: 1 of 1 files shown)


core/mem/allocators.odin
```diff
@@ -982,3 +984,152 @@ tracking_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
...
+		data = data,
+	}
+}
+
+profiling_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
+                                 size, alignment: int,
+                                 old_memory: rawptr, old_size: int, loc := #caller_location) -> (result: []byte, err: Allocator_Error) {
+	data := cast(^Profiling_Allocator)allocator_data
+
+	start_time := time.now()
```
gingerBill, line 1071: Prefer `time.tick_now()` as that is a high performance timer

core/mem/allocators.odin
```diff
@@ -982,3 +984,152 @@ tracking_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
...
+	return data.backing.procedure(
+		allocator_data = data.backing.data,
+		mode = mode,
+		size = size,
+		alignment = alignment,
+		old_memory = old_memory,
+		old_size = old_size,
+		location = loc,
+	)
+}
```
gingerBill, line 1135: Is this that useful in practice?

jakubtomsu: I think stuff like this is used for large programs to test correctness, but I guess it's probably not useful enough to be in the `core` library. I will remove it from this PR.

core/mem/allocators.odin
jakubtomsu: Ok, thanks for the suggestion.

core/mem/allocators.odin
JesseRMeyer: On the contrary, there is a `testing` package in `core` that could benefit from this.

JesseRMeyer: The ability to specify the "pass test" to user code is probably the better route.  The probabilistic route is neat because it can compress a huge set of possible tests as 1 test, which would complement a large state-space testing problem with decent coverage of explicit tests.

jakubtomsu: Yeah, I guess an utility like this is pretty project specific. It also depends a lot on the error handling strategy, which might be different in each program.

jakubtomsu: Also, I wasn't really thinking about 'tests' when writing this allocator. I feel like it would be much more useful when testing a large program, since that way you can check all of the obscure code paths which might otherwise never get executed.

PR #2569, opened 2023-05-31
[thread] Fix temp allocator not being thread safe when creating new threads

Tetralux:
The default temp allocator is thread safe, but it achieves this via its global state being `@(thread_local)`.
This means that `&this_global_state` must be taken __inside the thread__ that is going to be using it.

Previously, on Windows it was being set to the caller's context, which includes the temp allocator pointer.
It was always setting the `init_context`, which is only intended for when the user specifically wants to use a custom context, and not the default one.
This meant that, __by default__, you'd get a temp allocator that WASN'T thread-safe!

It also wasn't entirely clear how the user could actually set the initial context with `thread.create()`, or how it was meant
to be used. 

This PR addresses this by having `thread.create()` take it as a default parameter as well to match all the other thread-start procs, as well as adding some doc comments on a few things.
It also refactors the handling of `init_context` to hopefully make it clearer what's going on, and thus, harder to bork it up.

Here's a simple program to test it with; on master, it prints out the same pointer both times, but with this PR it prints out different pointers:
```odin
lprintf :: proc(format: string, args: ..any) {
    @static m: sync.Mutex
    sync.lock(&m)
    defer sync.unlock(&m)
    fmt.printf(format, ..args)
}

lprintf("%v %v\n", context.temp_allocator.data, context.temp_allocator.procedure)

t := thread.create(proc(t: ^thread.Thread) {
    lprintf("%v %v\n", context.temp_allocator.data, context.temp_allocator.procedure)
})
thread.start(t)
thread.join(t)
```

### DIFF
--- a/core/thread/thread.odin
+++ b/core/thread/thread.odin
@@ -14,10 +14,36 @@ Thread :: struct {
 	using specific: Thread_Os_Specific,
 	id:             int,
 	procedure:      Thread_Proc,
+
+	/*
+		These are values that the user can set as they wish, after the thread has been created.
+		This data is easily available to the thread proc.
+
+		These fields can be assigned to directly.
+
+		Should be set after the thread is created, but before it is started.
+	*/
 	data:           rawptr,
 	user_index:     int,
 	user_args:      [MAX_USER_ARGUMENTS]rawptr,
 
+	/*
+		The context to be used as 'context' in the thread proc.
+
+		This field must not be changed after the thread has started.
+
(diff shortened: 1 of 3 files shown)


ftphikari:
Can you modify it to account for the change in c194dfb?

Tetralux:
Already did, I think. Unless there's something I missed? 😄

Tetralux:
The bug that this PR was fixing has been fixed separately on latest master.

Tetralux:
Replaced by #2577.

PR #2574, opened 2023-06-02
[core:thread] Fix `run` procedure

ftphikari:
This commit fixes the behavior of `run`. Previously it would do `destroy(t)`, which would attempt to join the thread unto itself, which made no sense, thus it just deadlocked, and `free(t)` was never called.

This change adds `Auto_Cleanup` flag that is set when `run` procedure is called. Thread frees its memory at the end of its life if it finds that flag.

### DIFF
--- a/core/thread/thread_windows.odin
+++ b/core/thread/thread_windows.odin
@@ -7,17 +7,10 @@ import "core:intrinsics"
 import "core:sync"
 import win32 "core:sys/windows"
 
-Thread_State :: enum u8 {
-	Started,
-	Joined,
-	Done,
-}
-
 Thread_Os_Specific :: struct {
 	win32_thread:    win32.HANDLE,
 	win32_thread_id: win32.DWORD,
 	mutex:           sync.Mutex,
-	flags:           bit_set[Thread_State; u8],
 }
 
 _thread_priority_map := [Thread_Priority]i32{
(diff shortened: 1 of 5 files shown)


core/thread/thread_windows.odin
```diff
@@ -44,6 +37,11 @@ _create :: proc(procedure: Thread_Proc, priority := Thread_Priority.Normal) -> ^
 				runtime.default_temp_allocator_destroy(auto_cast context.temp_allocator.data)
 			}
 		}
+
+		if .Auto_Cleanup in t.flags {
+			free(t, t.creation_allocator)
```
Tetralux, line 42: Clear the handle here too?

ftphikari: `win32_thread` you mean? I only did what `_delete` did on Windows, not sure if that's needed. Perhaps it would be a good idea, idk.

Tetralux: Yeah; If you clear it on one, might as well on the other too.

ftphikari: Ok let's first see what gB says.

ftphikari:
@gingerBill @Kelimion does this look good to you? Would it be possible to merge this?

Tetralux:
I think the flag is a decent idea for addressing this problem (that a thread created via `run` will attempt to `destroy` itself).

The bool being passed to the public API of `run` though is a little awkward.

Maybe `_create` can just also take that bool, and `create`, `run`, and the `_with_data` variants, can just pass it through along with the priority?
I use the same approach in my PR that addresses `init_context`. (#2569)

ftphikari:
As I see it, we need input from the maintainers on what they think.

gingerBill:
By only concern is, does `CloseHandle` every get called on Windows?

PR #2579, opened 2023-06-06
Documentation: Tracking_Allocator example is leaking memory

apahl:
Added the missing call to destroy the tracking allocator.


apahl:
Thank you, that was very fast. And indeed I noticed it because valgrind complained.

KR Axel

PR #2587, opened 2023-06-12
fix(dns): don't exit early on no hosts in hosts file

GoNZooo:
If we don't have any hosts specified we'll still not generate any overrides which is fine, but we'll continue onto actually trying to resolve the hostname we came into the function for initially. The early return on no hosts read in the hosts file seems unnecessary and is maybe a copy-paste of the same error for the name server(?).


Kelimion:
Well caught.

PR #2588, opened 2023-06-12
fix fcntl binding on darwin

laytan:
It didn't like the c_vararg when doing absolute_path_from_handle.
Also fixed the buffer for absolute_path_from_handle, which according to docs needs to be of size MAX_PATH which it wasn't.

### DIFF
--- a/core/os/os_darwin.odin
+++ b/core/os/os_darwin.odin
@@ -440,7 +440,7 @@ foreign libc {
 	@(link_name="closedir")         _unix_closedir      :: proc(dirp: Dir) -> c.int ---
 	@(link_name="rewinddir")        _unix_rewinddir     :: proc(dirp: Dir) ---
 
-	@(link_name="__fcntl")          _unix__fcntl        :: proc(fd: Handle, cmd: c.int, #c_vararg args: ..any) -> c.int ---
+	@(link_name="__fcntl")          _unix__fcntl        :: proc(fd: Handle, cmd: c.int, arg: uintptr) -> c.int ---
 
 	@(link_name="rename") _unix_rename :: proc(old: cstring, new: cstring) -> c.int ---
 	@(link_name="remove") _unix_remove :: proc(path: cstring) -> c.int ---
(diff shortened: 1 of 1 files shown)


core/os/os_darwin.odin
```diff
@@ -794,14 +794,14 @@ _readlink :: proc(path: string) -> (string, Errno) {
 }
 
 absolute_path_from_handle :: proc(fd: Handle) -> (string, Errno) {
-	buf : [256]byte
-	res  := _unix__fcntl(fd, F_GETPATH, &buf[0])
-	if	res != 0 {
-		return "", Errno(get_last_error())
+	buf: [DARWIN_MAXPATHLEN]byte
+	_, err := fcntl(int(fd), F_GETPATH, int(uintptr(&buf[0])))
```
Kelimion, line 798: This could probably use the `or_return` treatment.

laytan: That does not seem to work with Errno because it is an int: `'or_return' expects an "optional ok" like value, or an n-valued expression where the last value is either a boolean or can be compared against 'nil', got Errno`

Kelimion: No, I meant after your patch of having it use the wrapper it reads like this?
```odin
absolute_path_from_handle :: proc(fd: Handle) -> (string, Errno) {
	buf: [DARWIN_MAXPATHLEN]byte
	_, err := fcntl(int(fd), F_GETPATH, int(uintptr(&buf[0])))
	if err != ERROR_NONE {
		return "", err
	}

	path := strings.clone_from_cstring(cstring(&buf[0]))
	return path, err
}
```
So `or_return`ing it would be:
```odin
absolute_path_from_handle :: proc(fd: Handle) -> (string, Errno) {
	buf: [DARWIN_MAXPATHLEN]byte
	_ = fcntl(int(fd), F_GETPATH, int(uintptr(&buf[0]))) or_return
	path := strings.clone_from_cstring(cstring(&buf[0]))
	return path, ERROR_NONE
}
```
Does that work?

PR #2590, opened 2023-06-13
[check-type] fix faulty #no_nil variants error

### LINKED ISSUE #2395: Getting error "A union with #no_nil must have at least 2 variants" for a polymorphic union, even when it has 2 variants
## Context

        Odin: dev-2023-03:fe533fb8
        OS:   Windows 10 Home Basic (version: 21H2), build 19044.2728
        CPU:  Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
        RAM:  16246 MiB

## Expected Behavior

This "#no_nil union must have 2+ variants" is only fired in the situation that it describes

## Current Behavior

The following union causes the error, even though it does have two variants:

```
BadUnion :: union($T: typeid) #no_nil {
    int,
    f32,
}
```

As shown above, the error is thrown whenever the union has the (T) parameter - even if the T type isn't actually used in any of the variants.

### Steps to Reproduce

Compile an otherwise-empty project (just a no-op main() procedure) which defines the above BadUnion(T) type

inbelic:
- when checking the variants of a union type we will skip adding the variants if we have an unspecialized polymorphic, hence our union_type variants will be empty and have a count of 0

- so when checking if we violate the #no_nil error, if we are in the unspecialized polymorphic case and there exists at least one variant in the original variants then we should not raise this error

- test checks that we do not raise the error anymore, and that we still detect the #no_nil error in the described circumstances

fixes #2395

### DIFF
--- a/src/check_type.cpp
+++ b/src/check_type.cpp
@@ -729,6 +729,12 @@ gb_internal void check_union_type(CheckerContext *ctx, Type *union_type, Ast *no
 	union_type->Union.kind = ut->kind;
 	switch (ut->kind) {
 	case UnionType_no_nil:
+		if (union_type->Union.is_polymorphic && poly_operands == nullptr) {
+			GB_ASSERT(variants.count == 0);
+			if (ut->variants.count != 1) {
+				break;
+			}
+		}
 		if (variants.count < 2) {
 			error(ut->align, "A union with #no_nil must have at least 2 variants");
 		}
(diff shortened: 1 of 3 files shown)


inbelic:
Maybe not as apparent, but in the case that we use the same type in the first Invalid example, then we will raise the duplicate variant error and cover that case.

The second if statement that checks the `ut->variants.count` is to raise the error directly if the union only consists of the polymorphic type. Otherwise, the error would be raised when we try to define an instance of the Union with the specific type. If that is the desired behaviour then we could just remove that if guard.

gingerBill (approved):

PR #2591, opened 2023-06-14
feat(os_linux): add `poll` & `ppoll`

GoNZooo:
I wanted to play around with `poll` & friends in a project and noticed that we didn't have them in `os_linux` (while we obviously still have Windows equivalents).

It's not necessarily what people will go with in practice (they'll probably want `io_uring`), but I think it's still nice to have access to.

### DIFF
--- a/core/os/os_linux.odin
+++ b/core/os/os_linux.odin
@@ -440,9 +440,7 @@ pollfd :: struct {
 
 nfds_t :: distinct c.uint
 
-sigset_t :: struct {
-  __val: [16]c.ulong,
-}
+sigset_t :: distinct u64
 
 foreign libc {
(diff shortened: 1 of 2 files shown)


jasonKercher:
Linux does not have a `poll` syscall for arm64.  Maybe you can wrap `ppoll` for arm64 there.  

Also, I'm not sure about your definition of `sigset_t`.  Where did you get that.  I *believe* it should only ever be 4 or 8 (if you include the real-time signals).

GoNZooo:
@jasonKercher The builds are passing and the formatting should be cleaned up.

GoNZooo:
The Windows pipeline seems to fail on a completely unrelated test(?)

Kelimion:
> The Windows pipeline seems to fail on a completely unrelated test(?)

It does that sometimes for no clear reason.

PR #2592, opened 2023-06-15
Added generation of stb static libraries for Unix systems

smbryan:
When experimenting with Odin, I received the following error when using `stb`:
```
odin_package:(.text+0x74b7): undefined reference to `stbi_load'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
Upon investigation, the expected static libs were unavailable, but a Makefile was already in place. In the build script, I've added a call to this Makefile for the Unix-based systems (i.e., everything except Windows). 

Since all these OS share the same `lib` directory, I've added a `.gitignore` there to avoid accidentally committing the binaries.


smbryan:
I see now that Darwin's static libraries are tracked in the repo. I'm not sure that static libraries should be generated by default for them.

PR #2603, opened 2023-06-24
Fix fmt_js to match new io.Stream Interface

Byron-Hodel:
After pulling #2584, I noticed that the fmt_js.odin file was not updated to match the new interface.


Kelimion:
We should probably add an `odin check` for the `js` platform to the CI, @gingerBill.

PR #2611, opened 2023-06-27
fix(os): use `setenv` instead of `putenv`

GoNZooo:
`putenv` doesn't copy the value that is put, which means that the previous code had a bug where we free'd the temporary memory and the environment was accidentally cleared right after the function finished.

Using `setenv` the OS instead copies over the values and we're free to deallocate as the original code intended.

### DIFF
--- a/core/os/os_linux.odin
+++ b/core/os/os_linux.odin
@@ -458,6 +458,7 @@ foreign libc {
 	@(link_name="execvp")           _unix_execvp       :: proc(path: cstring, argv: [^]cstring) -> int ---
 	@(link_name="getenv")           _unix_getenv        :: proc(cstring) -> cstring ---
 	@(link_name="putenv")           _unix_putenv        :: proc(cstring) -> c.int ---
+	@(link_name="setenv")           _unix_setenv        :: proc(key: cstring, value: cstring, overwrite: c.int) -> c.int ---
 	@(link_name="realpath")         _unix_realpath      :: proc(path: cstring, resolved_path: rawptr) -> rawptr ---
 
 	@(link_name="exit")             _unix_exit          :: proc(status: c.int) -> ! ---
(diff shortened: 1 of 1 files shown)


core/os/os_linux.odin
```diff
@@ -894,7 +895,10 @@ get_env :: proc(key: string, allocator := context.allocator) -> (value: string)
 set_env :: proc(key, value: string) -> Errno {
 	runtime.DEFAULT_TEMP_ALLOCATOR_TEMP_GUARD()
 	s := strings.concatenate({key, "=", value, "\x00"}, context.temp_allocator)
-	res := _unix_putenv(strings.unsafe_string_to_cstring(s))
+  key_cstring := strings.unsafe_string_to_cstring(strings.concatenate({key, "\x00"}, context.temp_allocator))
```
Kelimion, line 897: `odin check -vet` says `s` is unused. Also, please use tabs to indent the code. Your additions are not aligned properly.

core/os/os_linux.odin
```diff
@@ -893,8 +894,10 @@ get_env :: proc(key: string, allocator := context.allocator) -> (value: string)
 
 set_env :: proc(key, value: string) -> Errno {
 	runtime.DEFAULT_TEMP_ALLOCATOR_TEMP_GUARD()
-	s := strings.concatenate({key, "=", value, "\x00"}, context.temp_allocator)
-	res := _unix_putenv(strings.unsafe_string_to_cstring(s))
+  key_cstring := strings.unsafe_string_to_cstring(strings.concatenate({key, "\x00"}, context.temp_allocator))
```
Kelimion: I would replace these incantations with `strings.clone_to_cstring(str, context.temp_allocator)` instead of hacking together a nul-terminated string and then casting it to a `cstring`.

Kelimion: Once this is addressed it's fine to merge as far as I'm concerned.

GoNZooo: Fixed in f048ad13b52d3c2be2fcd975f1039eb88604360f

core/os/os_linux.odin
GoNZooo: Fixed

core/os/os_linux.odin
Kelimion: Good catch about the putenv thing.

PR #2622, opened 2023-07-02
detect unix color support

laytan:
Adds a check for color support that was previously only on Windows.
`tput colors` outputs the amount of colors a terminal supports, I checked the `error.cpp` file and it seems to use 8 colors, thus the check for more than or equal to 8.

I have not tested the case where `tput` isn't available because it is available on all my available stuff. I would expect it to work because gb_str_to_u64 would return 0 because the output would be the message that it is not available.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2547,12 +2547,24 @@ gb_internal int strip_semicolons(Parser *parser) {
 	return cast(int)failed;
 }
 
-#if defined(GB_SYSTEM_OSX) || defined(GB_SYSTEM_UNIX)
-#include <stdio.h>
-#endif
-
 gb_internal void init_terminal(void) {
 	build_context.has_ansi_terminal_colours = false;
+
+    gbAllocator a = heap_allocator();
+
+    char const *no_color = gb_get_env("NO_COLOR", a);
+    defer (gb_free(a, cast(void *)no_color));
+    if (no_color != nullptr) {
+        return;
+    }
+
+    char const *force_color = gb_get_env("FORCE_COLOR", a);
+    defer (gb_free(a, cast(void *)force_color));
+    if (force_color != nullptr) {
+        build_context.has_ansi_terminal_colours = true;
+        return;
+    }
+
 #if defined(GB_SYSTEM_WINDOWS)
 	HANDLE hnd = GetStdHandle(STD_ERROR_HANDLE);
 	DWORD mode = 0;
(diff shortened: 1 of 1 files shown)


laytan:
Hmm it outputs `tput: No value for $TERM and no -T specified` in github actions, need to find a way to quiet that

laytan:
Note that this will run on every invocation of odin (on unix), is that fine performance wise (I think it is fine)?

Kelimion:
> Note that this will run on every invocation of odin (on unix), is that fine performance wise (I think it is fine)?

Have you timed it with and without `tput` available? And can't we do this by checking some environment variable for the kind of terminal instead?

laytan:
Yeah it does not impact the time significantly, I did do some more research and saw that nodejs, gcc, and python just check that the terminal is a tty, and that the $TERM environment variable is anything but 'dumb'. I will change the approach to that too.

laytan:
Ok, updated to the way that gcc, node, python do the detection instead, which is checking for a tty and $TERM != 'dumb'.
Also added FORCE_COLOR and NO_COLOR as it is relatively standard.

src/main.cpp
```diff
@@ -2549,6 +2549,22 @@ gb_internal int strip_semicolons(Parser *parser) {
 
 gb_internal void init_terminal(void) {
 	build_context.has_ansi_terminal_colours = false;
+
+    gbAllocator a = heap_allocator();
+
+    char const *no_color = gb_get_env("NO_COLOR", a);
```
Kelimion, line 2555: This latest patch looks great, however did you use 4 spaces instead of tabs to indent? It doesn't align as you can see.
image

laytan: Damnit, let me fix that

laytan:
@Kelimion By the way, how did you get GitHub to show you the indentation like that? On my diff the indentation is exactly the same, that's why I didn't catch the difference.

Kelimion:
> @Kelimion By the way, how did you get GitHub to show you the indentation like that? On my diff the indentation is exactly the same, that's why I didn't catch the difference.

No idea? That's just what the Files Changed tab does on github.com, at least for me. Maybe the desktop app behaves differently.

laytan:
> > @Kelimion By the way, how did you get GitHub to show you the indentation like that? On my diff the indentation is exactly the same, that's why I didn't catch the difference.
> 
> No idea? That's just what the Files Changed tab does on github.com, at least for me. Maybe the desktop app behaves differently.

Strange, looks exactly the same for me

Kelimion:
> > > @Kelimion By the way, how did you get GitHub to show you the indentation like that? On my diff the indentation is exactly the same, that's why I didn't catch the difference.
> > 
> > 
> > No idea? That's just what the Files Changed tab does on github.com, at least for me. Maybe the desktop app behaves differently.
> 
> Strange, looks exactly the same for me

It would now that you fixed the indentation. 😁

laytan:
> > > > @Kelimion By the way, how did you get GitHub to show you the indentation like that? On my diff the indentation is exactly the same, that's why I didn't catch the difference.
> > > 
> > > 
> > > No idea? That's just what the Files Changed tab does on github.com, at least for me. Maybe the desktop app behaves differently.
> > 
> > 
> > Strange, looks exactly the same for me
> 
> It would now that you fixed the indentation. 😁

Yes, but even the earlier commits ;)

SentientCoffee:
> Kelimion By the way, how did you get GitHub to show you the indentation like that? On my diff the indentation is exactly the same, that's why I didn't catch the difference.

https://github.com/settings/appearance > Tab size preference

PR #2623, opened 2023-07-02
Fix #2615: can't iterate untyped string

### LINKED ISSUE #2615: Cannot iterate over string literals
The following fails to compile:

```odin
main :: proc() {
    for c in "foo" {
        // ...
    }
}
```
```
Cannot iterate over '"foo"' of type 'untyped string'
        for c in "foo" {
                 ^~~~^
```

as of Odin commit 1ecb4aa

laytan:


Kelimion:
I wonder if it makes sense to add a smoke test under `tests/issues` to let us know if it breaks again.

laytan:
Sure I can add that

laytan:
A non-related thing: Is `odin test` supposed to exit with code 0 when it fails (probably not right)?

Kelimion:
!ytmnd

PR #2634, opened 2023-07-06
[check_expr] add checker error for ! oper

### LINKED ISSUE #2624: Using `!` operator on bool array causes compiler to crash
## Context

```
Odin: dev-2023-07-nightly:1ecb4aa9
OS:   Windows 10 Home Basic (version: 21H1), build 19043.2364
CPU:  Intel(R) Core(TM) i5-8600K CPU @ 3.60GHz
RAM:  16318 MiB
```

## Current Behavior

When using the `!` operator on a boolean array inside an `if` or `for` statement, the compiler crashes with the following message:
```
test.main
lb_emit_conv: src -> dst
Not Identical [10]bool != llvm bool
Not Identical [10]bool != llvm bool
Not Identical 1494cddc040 != 7ff6d927e098
Not Identical 1494cddc040 != 7ff6d927e098
D:\a\Odin\Odin\src\llvm_backend_expr.cpp(2230): Panic: Invalid type conversion: '[10]bool' to 'llvm bool' for procedure 'test.main'
```

When doing the same but inside a `switch` statement, OR when operator is used in combination with a comparison (`if !foo == false {}`, or the second repro), it gives a different crash message:
```
D:\a\Odin\Odin\src\llvm_backend_expr.cpp(119): Assertion Failure: `is_type_array_like(type)`
```

## Expected Behavior

Allowing the operator to be used to invert boolean arrays, or a compiler message says it's not allowed

### Steps to Reproduce

`odin build test.odin -file`
```odin
package test

main :: proc() {
    foo: [10]bool
    if !foo {}
}
```
```odin
main :: proc() {
    foo: [10]bool
    if !foo == false {}
}
```

inbelic:
- the ! is only intended to be used on a bool value and not on an array of booleans, instead we use ~ for the implied functionality

- update our checker to raise an error in the case the ! is used on an array of booleans and raise a suggestion to use the ~ operator

fixes #2624

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -1705,12 +1705,12 @@ gb_internal bool check_unary_op(CheckerContext *c, Operand *o, Token op) {
 		break;
 
 	case Token_Not:
-		if (!is_type_boolean(type)) {
+		if (!is_type_boolean(type) || is_type_array_like(o->type)) {
 			ERROR_BLOCK();
 			str = expr_to_string(o->expr);
 			error(op, "Operator '%.*s' is only allowed on boolean expressions", LIT(op.string));
 			gb_string_free(str);
-			if (is_type_integer(type)) {
+			if (is_type_array_like(o->type) || is_type_integer(type)) {
 				error_line("\tSuggestion: Did you mean to use the bitwise not operator '~'?\n");
 			}
 		} else {


gingerBill:
Just allowing this in the frontend is not technically correct, you also need to change the behaviour in the backend code too. `~x` is equivalent to `-1 ~ x`, which is not the same as you think it is with regards to a boolean.

PR #2638, opened 2023-07-08
Fix #2637

### LINKED ISSUE #2637: testing.expect_value can't compare nils
## Context
* Operating System & Odin Version: Linux, Odin: dev-2023-07:3072479c

## Expected Behavior

testing.expect_value should succeed with two nils as inputs

## Current Behavior

testing.expect_value don't consider two nils as equal

### Steps to Reproduce

```
  Foo :: Maybe(string)
  x: Foo
  testing.expect(t, x == nil) // Succeeds
  testing.expect_value(t, x, nil) // Fails, "expected nil, got nil"
```

ramn:
where testing.expect_value can't compare nils

### DIFF
--- a/core/testing/testing.odin
+++ b/core/testing/testing.odin
@@ -6,6 +6,8 @@ import "core:time"
 import "core:intrinsics"
 import "core:reflect"
 
+_ :: reflect // alias reflect to nothing to force visibility for -vet
+
 // IMPORTANT NOTE: Compiler requires this layout
 Test_Signature :: proc(^T)
 
--- a/core/testing/testing.odin
+++ b/core/testing/testing.odin
@@ -4,6 +4,7 @@ import "core:fmt"
 import "core:io"
 import "core:time"
 import "core:intrinsics"
+import "core:reflect"
 
 // IMPORTANT NOTE: Compiler requires this layout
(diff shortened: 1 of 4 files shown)


Kelimion:
Very good catch. Thanks for the fix and regression test.

Kelimion:
This error is a simple fix. We test the core library with `-vet -strict-style`.
```
D:/a/Odin/Odin/core/testing/testing.odin(7:8) 'reflect' declared but not used
	import "core:reflect"
	       ^
Error: Process completed with exit code 1.
```

ramn:
> 

But reflect is used

Kelimion:
> 

Correct, but they're being used from within a polymorphic procedure and there's no visible specialization that `-vet` can see. To tell `-vet` to shut up, you can do the following:
```odin
import "core:reflect"
_ :: reflect // alias reflect to nothing to force visibility for -vet
```

ramn:
Aha thanks for the explanation

Kelimion:
> Aha thanks for the explanation

Another option would've been:
```odin
is_nil :: reflect.is_nil
```
And then using the aliased `is_nil` without the reflect prefix in the parapoly, because it is again a name visible to the vetter.

PR #2640, opened 2023-07-09
Improve Mutex struct documentation

z64:
Ended up writing this "by accident", figured I would PR it anyway. A basic leveling-up to explain the "what" and "why" of a mutex, without getting into too much detail.

### DIFF
--- a/core/sync/primitives.odin
+++ b/core/sync/primitives.odin
@@ -7,10 +7,21 @@ current_thread_id :: proc "contextless" () -> int {
 	return _current_thread_id()
 }
 
-// A Mutex is a mutual exclusion lock
-// The zero value for a Mutex is an unlocked mutex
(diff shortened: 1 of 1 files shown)


core/sync/primitives.odin
```diff
@@ -7,10 +7,21 @@ current_thread_id :: proc "contextless" () -> int {
 	return _current_thread_id()
 }
 
-// A Mutex is a mutual exclusion lock
-// The zero value for a Mutex is an unlocked mutex
+// A Mutex is a [mutual exclusion lock](https://en.wikipedia.org/wiki/Mutual_exclusion).
```
gingerBill, line 10: Odin's docs are not markdown based but use our own custom format. To add a link, it'll need to be in this form:
```
[[mutual exclusion lock; https://en.wikipedia.org/wiki/Mutual_exclusion]]
```

PR #2642, opened 2023-07-10
core:sys/windows: Add definitions for network byte order conversion procs

ryuukk:


Kelimion:
Not that I mind adding them, but why add API calls for byte order conversion when Odin has endian-specific types? What's the use case? The reason I ask is because we may want to reflect that use case in `core:net` if it's not too obscure of one.

PR #2647, opened 2023-07-11
core:sys/windows: Deprecate htons, htonl, ntohl, ntohs in favor of endian specific integers

ryuukk:
After a lengthily discussion on discord about whether or not the binding is correct, we came to the conclusion that these functions are useless since Odin supports natively endian specific integers, it's being used in the winsock binding, therefore to avoid confusions (like i had + a bug), deprecating them is better than not having them, so the user is properly educated on what is actually happening and what to use instead!


Kelimion:
It's almost as if I predicted this with my question on #2642.

PR #2652, opened 2023-07-11
[vendor:directx/d3d_compiler] Added default include file handler

ftphikari:


vendor/directx/dxgi/dxgidebug.odin
```diff
@@ -139,3 +139,21 @@ IDebug1_VTable :: struct {
 	DisableLeakTrackingForThread:   proc "stdcall" (this: ^IDebug1),
 	IsLeakTrackingEnabledForThread: proc "stdcall" (this: ^IDebug1) -> BOOL,
 }
+
+
+GetDebugInterface: proc "stdcall" (riid: ^IID, ppDebug: ^rawptr) -> HRESULT
+// Call this to start using GetDebugInterface
+debug_interface_init :: proc() -> bool {
```
gingerBill, line 146: Should this follow the same naming conventions as D3D? e.g. `InitDebugInterface`

vendor/directx/dxgi/dxgidebug.odin
```diff
@@ -139,3 +139,21 @@ IDebug1_VTable :: struct {
...
+	if debug_lib == nil {
+		return false
+	}
+
+	ptr := win32.GetProcAddress(debug_lib, "DXGIGetDebugInterface")
+	if ptr == nil {
+		return false
+	}
+
+	GetDebugInterface = auto_cast(ptr)
```
gingerBill, line 157:
```odin
GetDebugInterface = auto_cast ptr
```

vendor/directx/dxgi/dxgidebug.odin
ftphikari: No, it's just a helper function

gingerBill: But why not?

ftphikari: In GL package, helper functions do not follow GL convention, for example

ftphikari: Okay, I'll change it then

ftphikari:
Is there anything preventing this from being merged?

ftphikari:
To be honest I'm not even sure about the entire `GetDebugInterface` commit. `DXGIGetDebugInterface1` is available since Windows 8.1, and I think it's the rule of thumb in Odin that it does not support Windows lower than 8.1, right? So now I'm reconsidering even adding it in the first place, since it's only useful on Windows systems before 8.1.

gingerBill (approved):

PR #2662, opened 2023-07-17
Raymath module bindings for Raylib Vendor

Pix-xiP:
The raymath module was missing from the Raylib vendor code.

I've converted across the entire module based on their cheat sheet on the website.

If this isn't merged may I suggest a comment saying what to use instead somewhere in the raylib source code :)

### DIFF
--- a/vendor/raylib/raylib.odin
+++ b/vendor/raylib/raylib.odin
@@ -475,6 +475,11 @@ FilePathList :: struct {
     paths:    [^]cstring,                 // Filepaths entries
 }
 
+// Raymath Types 
+
+Float3  :: distinct [3]f32
+Float16 :: distinct [16]f32
+
 
 //----------------------------------------------------------------------------------
 // Enumerators Definition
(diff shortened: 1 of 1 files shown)


gingerBill:
Why have the raymath module when `linalg` et al packages are superior in every way, and Odin's vector, matrix and quaternion types are built into the language itself?

Pix-xiP:
Completetionist sake would be my only real answer - also for those who are starting completely fresh and just trying to follow the tutorials it saves them writing in something like:

```
raylib.VectorScale(x, y)
```
And then trying to figure out why its missing, if they included the wrong thing or missed including the something.

Trying to save newer people some pain that I went through being completely new to raylib / graphic related stuff

As I said in the last part of the comment, if its pointless to bring these in because the inbuilt are better in every way, then perhaps a comment into the raylib source code that basically says if looking for raymath - use `linalg` or inbuilts.

Happy to close request if thats the way forward :>

PR #2668, opened 2023-07-19
add flag to disable tracking done tasks in thread pool

laytan:
The `tasks_done` dynamic array is a waste when the user does not need the functionality, which im my experience is most of the time.

I made sure this is backwards compatible, otherwise I would suggest having no tracking be the default.


laytan:
Was this PR forgotten about and is there a chance of this getting merged, or should it be closed?

laytan:
I give up on this pr, the package is being rewritten anyway

Issue #2667, opened 2023-07-18
I think maybe there is a very rare bug in map


xzores:
## Context

I am using map to index a chunk for my game. The index I use is a [3]i32.
Now it seems there there is a very rare chance that the map cannot delete a key.
very rare is like 1 out of 100K to 1M times.

Info:
Odin: dev-2023-07-nightly:3072479c
OS:   Windows 11 Home Basic (version: 22H2), build 22621.1992
CPU:  13th Gen Intel(R) Core(TM) i9-13900HX
RAM:  32507 MiB

## Expected Behavior

I would expect delete_key to remove the element with the given key from the map.

## Current Behavior

It very rarely does not delete the key.

It simply tries to remove the map every frame, but it seems to fail for some reason, my logs:
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]
removing : [-28, 8, 25]

## Failure Information (for bugs)

Ok, since it is a rare phenomenon I will give you all the code I use
```
//This ordering matter
Chunk_state :: enum {
    should_remove,      //This shall be destroyed.
    is_passive,         //Passive is when it is close enough that is should not be destroyed, but not close enough to be generated.
    is_active,          //is_active is when it is close enough that is should generated if it was not.
    generation_ungoing,  //The generation has begun, but is not done yet, and so it shall not be deleted or generated again.
    ungenerated,        //Tells that the chuck has just been created this frame and it needs to be generated.
}

Chunk :: struct {
    blocks : [5][5][5]Block,
    chunk_state : Chunk_state,
}

```

```
Chunk_size_blocks : i32 : 256;
Chunk_size_world : f32 : 4;

[...]

PR #2676, opened 2023-07-21
coalesce tombstones in map insert

jasonKercher:
This fixes issue #2667 and passes internal map testing.  Someone please review the logic in `map_insert_hash_dynamic`.  It feels more complicated than it needs to be and can maybe be compressed?


Kelimion:
I'm going to add [xzores's minimal repro](https://github.com/odin-lang/Odin/issues/2667#issuecomment-1642838661) as an official map test case so we'll be notified if the map tombstone implementation ever breaks again.

Kelimion:
Thanks, Jason.

PR #2689, opened 2023-07-30
GLFW3: static link support for GLFW on Linux

ohlahm2kai:
GLFW version: 3.3.8

Clang:
```
Ubuntu clang version 14.0.0-1ubuntu1.1
Target: x86_64-pc-linux-gnu
Thread model: posix
```

Required libraries were determined by running pkg-config:
```
$ pkg-config --static --libs glfw3
-lglfw3 -lrt -lm -ldl -lX11 -lpthread -lxcb -lXau -lXdmcp
```

Tested to work with both static and dynamic linking.


ohlahm2kai:
@ryuukk I completely agree. Can you point me to where the CI does this? I'm happy to handle it if somebody can show me where this is supposed to happen.

hasanyasin:
They should be here: 

https://github.com/odin-lang/Odin/tree/master/.github/workflows

If this looks unfamiliar, please check GitHub Actions documentation.

About having compiled libraries in releases, here's my five cents: I agree that we should have those available, but not in the main release packs, especially not in the source repo. It makes more sense to have a tool to download those extras (like `odin-deps check` or something) so people who don't already have these dependencies could get those easily.

ohlahm2kai:
@hasanyasin please point me to a file and line number where any GLFW compilation happens for any platform. I can't find anything related to vendor library compilation in the GitHub Actions config or called scripts.

hasanyasin:
GitHub Actions is where all the CI tasks on the repo are defined. The libraries are unfortunately in the code at the moment, as you can see here: https://github.com/odin-lang/Odin/tree/master/vendor/glfw/lib

I'm not sure about the history behind this, but as I stated in the previous message, I hope that we can move these off the code base and have all the dependencies available when/where needed in a convenient way. Until one of us designs and builds such a solution, keeping the libs in the repo seems to be the only convenient way.

ohlahm2kai:
Yeah. All I've done is added a static library for GLFW on Linux to match the current equivalent support for Windows and Darwin.

I agree that checking the binaries into git is not ideal, but that is a project of bigger scope than I am addressing here.

graphitemaster:
Unfortunately you cannot really do this on Linux because of glibc versioning. The static libs won't work on everyone's distribution depending on which version of libc they have, or if they even have say glibc, some distros ship with e.g musl.

ohlahm2kai:
It seems like I've dredged up a lot of problems tangent to this one. All of these other libs have the same problems:

```
./vendor/raylib/linux/libraylib.so
./vendor/raylib/linux/libraylib.so.450
./vendor/raylib/linux/libraylib.so.4.5.0
./vendor/lua/5.1/linux/liblua5.1.so
./vendor/lua/5.2/linux/liblua52.so
./vendor/lua/5.3/linux/liblua53.so
./vendor/lua/5.4/linux/liblua54.so
./libLLVM-11.so.1
./vendor/raylib/linux/libraylib.a
./vendor/lua/5.1/linux/liblua5.1.a
./vendor/lua/5.2/linux/liblua52.a
./vendor/lua/5.3/linux/liblua53.a
./vendor/lua/5.4/linux/liblua54.a
```

This PR was about maintaining the status quo but addressing the lack of static GLFW. I don't really feel like dealing with these larger problems, so I'll just close this.

PR #2719, opened 2023-08-11
add wait group to wait proc group

laytan:


laytan:
mac CI is just generally broken I see

PR #2720, opened 2023-08-11
fix macos CI by upgrading to llvm@14

laytan:


laytan:
Now that I think of it, it was probably pulling llvm 11 because that is the minimum that Odin supports? Might not be the best solution if that is the case.

laytan:
I also created this issue https://github.com/Homebrew/homebrew-core/issues/139339

laytan:
That ticket has been fixed so I assume this is not necessary anymore

PR #2721, opened 2023-08-11
Return .Invalid_Argument in pool allocator to avoid potential segfaults 

Kartoffelsaft:
```odin
pool := mem.Dynamic_Pool{}
// really small pool size for demonstration
mem.dynamic_pool_init(&pool, context.allocator, context.allocator, 16)
context.allocator = mem.dynamic_pool_allocator(&pool)

myVar, err := new(struct{data: [512]byte})
if err != .None do return // now returns here
myVar.data[100] = 5 // previously this would segfault
```

Context: I was trying to wrap `js.page_allocator()` in a pool allocator to port a project to web. Turns out, `microui.Context` is bigger than a single page, leading to some headaches. This is (hopefully) more debuggable.

### DIFF
--- a/core/mem/allocators.odin
+++ b/core/mem/allocators.odin
@@ -749,6 +749,7 @@ dynamic_pool_alloc_bytes :: proc(p: ^Dynamic_Pool, bytes: int) -> ([]byte, Alloc
 	n := bytes
 	extra := p.alignment - (n % p.alignment)
 	n += extra
+    if n > p.block_size do return nil, .Invalid_Argument
 	if n >= p.out_band_size {
 		assert(p.block_allocator.procedure != nil)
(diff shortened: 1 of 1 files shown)


core/mem/allocators.odin
```diff
@@ -749,6 +749,7 @@ dynamic_pool_alloc_bytes :: proc(p: ^Dynamic_Pool, bytes: int) -> ([]byte, Alloc
 	n := bytes
 	extra := p.alignment - (n % p.alignment)
 	n += extra
+    if n > p.block_size do return nil, .Invalid_Argument
```
gingerBill, line 752: Please use tabs for indentation.

Kartoffelsaft: Whoops, that was quite literally invisible on my side. Fixed now.

PR #2739, opened 2023-08-13
Adding "system" prefx to webgl foreign imports for non-Windows systems

scoobery:
Projects including WebGL fail to compile on Linux due to this error. 

`webgl.odin(3:1) Error: File name, , cannot be as a library name as it is not a valid identifier
    foreign import "webgl"`

Seems that it doesn't resolve a file path by just specifying `"webgl"`, but changing it to `"system:webgl"` appears to do the trick!

### DIFF
--- a/vendor/wasm/WebGL/webgl.odin
+++ b/vendor/wasm/WebGL/webgl.odin
@@ -1,6 +1,10 @@
 package webgl
 
-foreign import "webgl"
+when ODIN_OS == .Windows {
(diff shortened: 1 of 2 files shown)


gingerBill:
You shouldn't even be importing WebGL on Linux to begin with. it's a WASM only package.

scoobery:
I see; a misunderstanding on my part. It was just an issue with how the library was designed in this case.

PR #2747, opened 2023-08-15
#caller_location into the strings package

Pix-xiP:
Tetralux and I were talking in the discord after gB came in and gave the tracking allocator as a good way to find leaks.

Not all the makes contain caller locations so you sometimes have to really dig to find out where memory was allocated. So this is more of a QoL change.

We figure to start on something basic like the strings package and then if it seems beneficial, moving it into other core libs if there are not good reasons to not do it.

I've run the tests for the package and got 171/171 passed. Am happy to do more if this seems like a good idea!

### DIFF
--- a/core/strings/builder.odin
+++ b/core/strings/builder.odin
@@ -35,8 +35,8 @@ Returns:
 - res: The new Builder
 - err: An optional allocator error if one occured, `nil` otherwise
 */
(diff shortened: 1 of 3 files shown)


Pix-xiP:
Wild, that it failed on mac which is where I built and tested it first.

Though it looks like the CI has failed to setup the env, so I'm gonna say thats probably not my changes causing a fail :D

PR #2748, opened 2023-08-15
Fix issues with js events not being initialized properly

Caedo:


vendor/wasm/js/events.odin
```diff
@@ -343,7 +345,18 @@ do_event_callback :: proc(user_data: rawptr, callback: proc(e: Event)) {
...
+
+
 		init_event_raw(&event)
+
+		if event.kind == .Key_Up || event.kind == .Key_Down || event.kind == .Key_Press {
+			key := transmute(^runtime.Raw_String) &event.key.key
+			key.data = &event.key._key_buf[0]
+
+			code := transmute(^runtime.Raw_String) &event.key.code
+			code.data = &event.key._code_buf[0]
```
gingerBill, line 357:
```odin
event.key.key = string(event.key._key_buf[:KEY_LENGTH])
event.key.code = string(event.key._code_buf[:CODE_LENGTH])
```

Caedo: Changed it but had to add two more variables to Event struct. Was also thinking of doing:
```go
event.key.key = string(event.key._key_buf[:len(event.key.key])
```
that would save me those two ints but decided against that because it wasn't clear what is going on

PR #2753, opened 2023-08-17
Add vulkan_xcb.h to vulkan headers

JamesDSource:
Also added xcb_connection_t as an opaque struct and added xcb_window_t, xcb_visualid_t as u32s


vendor/vulkan/_gen/create_vulkan_odin_wrapper.py
```diff
@@ -85,6 +87,8 @@ def convert_type(t, prev_name, curr_name):
         "struct BaseInStructure":  "BaseInStructure",
         "struct wl_display": "wl_display",
         "struct wl_surface": "wl_surface",
+	"xcb_window_t": "u32",
+	"xcb_visualid_t": "u32",
```
gingerBill, line 91: Should these be aliases or distinct types?

JamesDSource: XCB tends to use straight aliases a lot. Like, you can pass a xcb_window_t into anything that takes xcb_drawable_t. They're all just generic Ids

PR #2761, opened 2023-08-21
Add README for `core` (UNFINISHED)

Yunky86:
Add a `README.md` for `core:`. It is unfinished and I'll finish it up later.

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -1,0 +1,36 @@
+# Builtin
+package builtin provides documentation for Odin's predeclared identifiers.
+
+# Fmt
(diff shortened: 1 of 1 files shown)


gingerBill:
Why not linked to https://pkg.odin-lang.org/core/ AND fill in the package docs for it like the other ones?

Yunky86:
> Why not have it link to [pkg.odin-lang.org/core](https://pkg.odin-lang.org/core/) AND fill in the package docs for it like the other ones?

Ok linking to the [docs](https://pkg.odin-lang.org/core/) would be Ok.

PR #2782, opened 2023-08-31
Add unix bindings for DXC.

AbdelHameed-Hamed:
PR adding Mac and Linux bindings for [DXC](https://github.com/microsoft/DirectXShaderCompiler) which is included in the VulkanSDK on these platforms.

Mainly switched the calling conventions from `stdcall` to `c` and copied over some type definitions from dxgi and win32 to make it more self-contained.


gingerBill:
If the API is effectively the same between the two files but the calling convention is different, I recommend using `proc "system" (`

AbdelHameed-Hamed:
> If the API is effectively the same between the two files but the calling convention is different, I recommend using `proc "system" (`

Done.

gingerBill (approved):

PR #2784, opened 2023-09-01
Add formatting of bytes into the best unit of measurement

laytan:
I've been doing a lot of logging of memory sizes, and made this procedure to nicely format it. Then I thought this might fit into the `fmt` package nicely so I made this PR.

At first I tried adding this as an implicit formatter for the mem.* constants, this requires them to be of the same type and marked a distinct type though, which breaks a lot of code.


core/fmt/fmt.odin
```diff
@@ -1048,6 +1048,65 @@ _fmt_int_128 :: proc(fi: ^Info, u: u128, base: int, is_signed: bool, bit_size: i
 	fi.zero = false
 	_pad(fi, s)
 }
+// Units of measurements:
+__MEMORY_LOWER := " b kb mb gb tb pb eb"
+__MEMORY_UPPER := " B KB MB GB TB PB EB"
```
gingerBill, line 1053: Maybe prefer using `MiB` etc instead?

laytan: Done!

PR #2805, opened 2023-09-21
Support LLVM 17.0.1

gingerBill:
* Update to allow for LLVM 17.0.1
* Windows Defaults to LLVM 17.0.1
* `-o:aggressive` 
* `-sanitize:address` (windows, linux, darwin)
* `-sanitize:memory` (linux)
* `-sanitize:thread` (linux, darwin)
* Improvement to runtime type info storage 
    * using r(o)data section, by storing everything in a giant packed struct of structs
    * LLVM does not have the concept of a union type, and to initialize one at the global level requires mocking it in memory layout using LLVM's absolutely awful type system


src/build_settings.cpp
```diff
@@ -1738,6 +1747,42 @@ gb_internal bool init_build_paths(String init_filename) {
...
 	}
 
+	if (build_context.sanitizer_flags & SanitizerFlag_Address) {
+		switch (build_context.metrics.os) {
+		case TargetOs_windows:
+		case TargetOs_linux:
+		case TargetOs_darwin:
+			break;
+		default:
+			gb_printf_err("-sanitize:memory is only supported on windows, linux, and darwin\n");
```
laytan, line 1757:
```suggestion
			gb_printf_err("-sanitize:address is only supported on windows, linux, and darwin\n");
```

PR #2811, opened 2023-09-22
Remove trailing zeros after decimal when writing / parsing f64

karl-zylinski:
In `write_f64`: Remove trailing zeroes after decimal point, so it doesn't always write the max amount of zeroes.
In `parse_f64_prefix`: Skip trailing zeroes after decimal point, so it doesn't truncate the number when it wasn't needed.


karl-zylinski:
I realize I made a bug. Closing.

PR #2833, opened 2023-09-30
darwin  enumerate_interfaces

simonwashere:
Implemented darwin enumerate_interfaces basically by copying the code from the linux version (as was suggested in the comment in the code)


Kelimion:
The code style is alright, there's no memory leaks _and_ it works. The trifecta we look for. :-)
Merged.

PR #2835, opened 2023-10-01
added rand_bytes for darwin

simonwashere:
made crypto rand_bytes work on darwin using same code as linux

### DIFF
--- a/core/crypto/rand_darwin.odin
+++ b/core/crypto/rand_darwin.odin
@@ -1,0 +1,12 @@
+package crypto
+
+import "core:c"
+
+foreign import libc "system:c"
+foreign libc {
+	arc4random_buf :: proc "c" (buf: rawptr, nbytes: c.size_t) ---
+}
+
+_rand_bytes :: proc (dst: []byte) {
+	arc4random_buf(raw_data(dst), len(dst))
+}
--- a/core/crypto/rand_generic.odin
+++ b/core/crypto/rand_generic.odin
@@ -1,6 +1,6 @@
 package crypto
 
-when ODIN_OS != .Linux && ODIN_OS != .OpenBSD && ODIN_OS != .Windows {
+when ODIN_OS != .Linux && ODIN_OS != .OpenBSD && ODIN_OS != .Windows && ODIN_OS != .Darwin {
 	_rand_bytes :: proc (dst: []byte) {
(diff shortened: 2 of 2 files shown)


gingerBill:
Is this a good idea to rely on libc? @zhibog

Kelimion:
> Is this a good idea to rely on libc? @zhibog

I can't speak for @zhibog, but (https://pkg.go.dev/crypto/rand)[Go] purports to use `getentropy(2)` on macOS.

> On Linux, FreeBSD, Dragonfly, NetBSD and Solaris, Reader uses getrandom(2) if available, /dev/urandom otherwise. On OpenBSD and macOS, Reader uses getentropy(2). On other Unix-like systems, Reader reads from /dev/urandom. On Windows systems, Reader uses the RtlGenRandom API. On JS/Wasm, Reader uses the Web Crypto API. On WASIP1/Wasm, Reader uses random_get from wasi_snapshot_preview1.

See also (https://support.apple.com/en-kw/guide/security/seca0c73a75b/web)[this developer page] and (https://keith.github.io/xcode-man-pages/getentropy.2.html)[genentropy's man page]. It's defined in (https://opensource.apple.com/source/xnu/xnu-4570.41.2/bsd/sys/random.h.auto.html)[sys/random.h].

simonwashere:
The man page for arc4random on macos says:

`
These functions use a cryptographic pseudo-random number generator to generate high quality random bytes very quickly.  One data pool
     is used for all consumers in a process, so that consumption under program flow can act as additional stirring.  The subsystem is re-
     seeded from the kernel random number subsystem on a regular basis, and also upon fork(2).

     This family of functions provides higher quality random data than those described in rand(3), random(3), and rand48(3).  They can be
     called in almost all environments, including chroot(2), and their use is encouraged over all other standard library functions for
     random numbers.
`

and also:

`
The original version of this random number generator used the RC4 (also known as ARC4) algorithm.  In OS X 10.12 it was replaced with
     the NIST-approved AES cipher, and it may be replaced again in the future as cryptographic techniques advance.  A good mnemonic is “A
     Replacement Call for Random”
`

zhibog:
@gingerBill I'd try to avoid libc as much as I can, but having said that I have little experience developing on/for OSX so I don't have a strong opinion on this.

laytan:
This has been added elsewhere since this PR

PR #2838, opened 2023-10-02
fix(json): return `.Out_Of_Memory` when out of memory on parse

GoNZooo:
`json.parse` would silently return empty or incomplete objects when not having enough memory for a map insert. With this patch we instead are able to return `.Out_Of_Memory` so that it can be handled.

### DIFF
--- a/core/encoding/json/parser.odin
+++ b/core/encoding/json/parser.odin
@@ -264,13 +264,17 @@ parse_object_body :: proc(p: ^Parser, end_token: Token_Kind) -> (obj: Object, er
 			return
 		}
 
-		insert_success := runtime.map_insert(&obj, key, elem)
-		// NOTE(gonz): we'd rather check specifically for an allocation error here but
-		// `map_insert` doesn't differentiate; we can only check for `nil`
-		if insert_success == nil {
-			return nil, .Out_Of_Memory
+		// NOTE(gonz): There are code paths for which this traversal ends up
+		// inserting empty key/values into the object and for those we do not
+		// want to allocate anything
+		if key != "" {
+			reserve_error := reserve(&obj, len(obj) + 1)
+			if reserve_error == mem.Allocator_Error.Out_Of_Memory {
+				return nil, .Out_Of_Memory
+			}
+			obj[key] = elem
 		}
-		
+
 		if parse_comma(p) {
 			break
 		}
--- a/core/encoding/json/parser.odin
+++ b/core/encoding/json/parser.odin
@@ -3,7 +3,6 @@ package json
 import "core:mem"
 import "core:unicode/utf8"
 import "core:unicode/utf16"
-import "core:runtime"
 import "core:strconv"
 
 Parser :: struct {
--- a/core/encoding/json/parser.odin
+++ b/core/encoding/json/parser.odin
@@ -263,6 +263,12 @@ parse_object_body :: proc(p: ^Parser, end_token: Token_Kind) -> (obj: Object, er
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


core/encoding/json/parser.odin
```diff
@@ -263,7 +264,12 @@ parse_object_body :: proc(p: ^Parser, end_token: Token_Kind) -> (obj: Object, er
 			return
 		}
 
-		obj[key] = elem
+		insert_success := runtime.map_insert(&obj, key, elem)
```
GoNZooo, line 267: @Kelimion Do you think it would be advisable here to use `map_grow_...` or `reserve` instead? Is there a good way to get the allocation error more specifically?

There is a test failing in the build because of this added code but it's fairly hard to see why 190 extra entries would be added to the parsed array in the test.

core/encoding/json/parser.odin
```diff
@@ -263,13 +264,10 @@ parse_object_body :: proc(p: ^Parser, end_token: Token_Kind) -> (obj: Object, er
 			return
 		}
 
-		if len(obj) == cap(obj) {
-			reserve_error := reserve(&obj, max(1, cap(obj) * 2))
-			if reserve_error != nil {
-				return nil, .Out_Of_Memory
-			}
+		insert_success := runtime.map_insert(&obj, key, elem)
```
Tetralux: `map_insert` is a globally-available proc, so you shouldn't need to import `core:runtime`.

core/encoding/json/parser.odin
```diff
@@ -263,13 +264,10 @@ parse_object_body :: proc(p: ^Parser, end_token: Token_Kind) -> (obj: Object, er
...
+		if insert_success == nil {
```
Tetralux: Might be worth calling it `inserted_element_pointer` or something, so that the if makes more sense.
Or could just `if map_insert(...) == nil {` instead?

core/encoding/json/parser.odin
```diff
@@ -263,8 +263,17 @@ parse_object_body :: proc(p: ^Parser, end_token: Token_Kind) -> (obj: Object, er
 			return
 		}
 
-		obj[key] = elem
-		
+		// NOTE(gonz): There are code paths for which this traversal ends up
+		// inserting empty key/values into the object and for those we do not
+		// want to allocate anything
+		if key != "" {
```
GoNZooo, line 269: I don't particularly like this solution here but as the note says there are code paths than run through here with `key` being the empty string and these are heavily tested for right now.

core/encoding/json/parser.odin
```diff
@@ -263,8 +263,17 @@ parse_object_body :: proc(p: ^Parser, end_token: Token_Kind) -> (obj: Object, er
...
+			reserve_error := reserve(&obj, len(obj) + 1)
+			if reserve_error == mem.Allocator_Error.Out_Of_Memory {
+				return nil, .Out_Of_Memory
+			}
```
GoNZooo, line 273: An `else` would probably be reasonable here but for now it's just this. We weren't checking *anything* before so even this is better than the status quo.

GoNZooo:
@gingerBill @Kelimion Any word on whether we can get something like this into `master`? I'm not married to the current implementation and can rewrite it if we need to, but I would like for `json.parse` to be able to capture OOM errors instead of failing silently.

core/encoding/json/parser.odin
gingerBill: Any reason this isn't
```odin
reserve(&obj, len(obj) + 1) or_return
```

PR #2850, opened 2023-10-07
Support LLVM >=17.0.1 on Darwin and Linux

jcmdln:
This patch allows conditionally using the new PassBuilder on non-Windows platforms when the version of LLVM is >=17.0.1 as all other Transforms in `llvm-c` have been removed as of this version. Older supported versions of LLVM will continue using the old transforms.

Sanitizers are also now wired up for Darwin and Linux.

Ensure `/path/to/llvm17/bin` is in `PATH` or manually prepend it before building Odin:

```sh
export PATH="/path/to/llvm17/bin:$PATH"
./build_odin.sh
./odin run examples/demo/demo.odin -file -sanitize:address
```

### DIFF
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -117,10 +117,10 @@ run_demo() {
 	./odin run examples/demo/demo.odin -file
 }
 
-if [[ $# -eq 0 ]]; then
+if [ $# -eq 0 ]; then
 	build_odin debug
 	run_demo
-elif [[ $# -eq 1 ]]; then
+elif [ $# -eq 1 ]; then
 	case $1 in
 	report)
 		[ ! -f "./odin" ] && build_odin debug
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -55,7 +55,7 @@ config_darwin() {
 	MAX_LLVM_VERSION=("17.999.999")
 	if [ $(version $($LLVM_CONFIG --version)) -gt $(version $MAX_LLVM_VERSION) ]; then
 		echo "Tried to use " $(which $LLVM_CONFIG) "version" $($LLVM_CONFIG --version)
-		panic "Requirement: llvm-config must be base version smaller than 15"
+		panic "Requirement: llvm-config must be base version smaller than 18"
 	fi
 
 	LDFLAGS="$LDFLAGS -liconv -ldl -framework System"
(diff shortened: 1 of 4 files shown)


laytan:
After changing https://github.com/odin-lang/Odin/blob/cffa035c1b9d36c2639f6b7fd48b6399653244a3/src/main.cpp#L92 it also compiles on OSX

jcmdln:
Current workarounds for using sanitizers on non-Windows platforms:

```sh
export PATH="path/to/llvm/bin:$PATH"
./build_odin.sh
./odin run examples/demo/demo.odin -file -sanitize:address -extra-linker-flags="-fsanitize=address"
```

build_odin.sh
```diff
@@ -52,10 +52,10 @@ config_darwin() {
 		fi
 	fi
 
-	MAX_LLVM_VERSION=("14.999.999")
+	MAX_LLVM_VERSION=("17.999.999")
```
gingerBill, line 55: This is where things get a little complicated.

We supported LLVM 11, 12, 13, and now 17. We didn't support 14 (even if people got it working) and we definitely did not support 15 and 16. So Just a simple "max version" check is not necessarily a good idea.

gingerBill: But maybe this is a non-issue in practice because of the preprocessor check.

jcmdln: Yeah I'm curious why the shell script even needs to separately define dependencies other than to error as early as possible.

jcmdln:
```
$ LLVM_CONFIG=llvm-config-16 ./build_odin.sh 
...
src/llvm_backend.hpp:2:2: error: "LLVM version must be 11, 12, 13, or 17"
    2 | #error "LLVM version must be 11, 12, 13, or 17"
      |  ^
src/llvm_backend.hpp:25:10: fatal error: 'llvm-c/Transforms/AggressiveInstCombine.h' file not found
   25 | #include <llvm-c/Transforms/AggressiveInstCombine.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
```
Hoisting the preprocessor check in `llvm_backend.hpp` could make this a little more clear?

laytan: I think a lot of people are using llvm14 at the moment and making it an error at this point should be avoided

jcmdln: I've opted to clean up `build_odin.sh` instead of relying solely on preprocessor macros.

src/llvm_backend.hpp
```diff
@@ -15,6 +15,9 @@
 #include <llvm-c/Object.h>
 #include <llvm-c/BitWriter.h>
 #include <llvm-c/DebugInfo.h>
+#if LLVM_VERSION_MAJOR >= 17 && (LLVM_VERSION_MINOR > 0 || (LLVM_VERSION_MINOR == 0 && LLVM_VERSION_PATCH > 0))
```
gingerBill, line 18: I think testing for `LLVM_VERSION_MAJOR >= 17` is all that is necessary. Checking for the minor/major distinctions is probably not a good idea.

jcmdln: You're right, it should match the `LB_USE_NEW_PASS_SYSTEM` check below.

Issue #2854, opened 2023-10-08
Compiler produces bad code on `"stdcall"`


seventh-chord:
```
OS:   Windows 10 Professional (version: 22H2), build 19045.3448
CPU:  AMD Ryzen 7 PRO 5850U with Radeon Graphics
RAM:  15182 MiB
```
I've only tested on this machine.

Happens with `dev-2023-10-nightly:2cca0050` (from github releases), and with custom release build of `0c10b951`.
Does not happen with `dev-2023-08-nightly:9453b238` (from github releases).

## Steps to reproduce

Run the following code:

```odin
package main
import "core:runtime"

foo :: proc "stdcall" () {
}

main :: proc() {
  runtime.print_string("Hello\n")
  foo()
  runtime.print_string("Goodbye\n")
}
```

Compiler command line I used:
```
C:\Dev\tools\Odin\odin.exe run main.odin -file -o:speed
```

## Expected Behavior

Program should print `Hello` and then `Goodbye`.

## Current behavior

Program prints `Hello` and then crashes because an `int3` instruction is generated directly after the first call to `print_string`.

## Failure Information (for bugs)

Only happens if I pass `-o:speed`.

I noticed this bug because I was using `CreateWindowW`, and my code stopped working when I enabled `-o:speed` after I updated the compiler.

seventh-chord:
Ran `git bisect`, which shows this:

```
f0ab58dfbb4be6103fb7340ec570d218e2137a4b is the first bad commit
commit f0ab58dfbb4be6103fb7340ec570d218e2137a4b
Author: gingerBill <bill@gingerbill.org>
Date:   Wed Sep 20 10:39:48 2023 +0100

    Begin work with adding the new pass system

 src/llvm_backend.cpp | 54 ++++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 54 insertions(+)
```

seventh-chord:
*I'm seeing if I can solve this bug. Let me know if you'd rather not have me post progress updates :)*

Looks like some optimization pass that is part of `default<O2>` in clang is causing this.

If I run the following commands (with odin built from 0c10b951a9):
```
odin build bug.odin -file -no-crt -keep-temp-files -show-system-calls
clang bug.ll -O2 -S -emit-llvm -o bug_opt.ll
```

I see this in `bug.ll`:
```
define internal x86_stdcallcc void @main.foo() {
decls:
  br label %entry

entry:                                            ; preds = %decls
[...]

PR #2861, opened 2023-10-13
llvm: specify calling convention on call

seventh-chord:
Looks like the compiler only was specifying the calling convention on function declarations, but not on function calls.
But LLVM seems to produce "bad" code when optimizing with level -O2 unless you specify the same calling convention on the call too.

Solves this issue: https://github.com/odin-lang/Odin/issues/2854

Is there some regression test suite I should add a test case to, and which I should run?
I'm not familiar with the odin compiler or with LLVM, so I'm a bit worried I broke something inadvertently.

Also, are there more places where it's necessary to insert `LLVMSetInstructionCallConv`?

## How I figured out this fix

If we look at the LLVM IR generated by odin for this code (using `-keep-temp-files`):

```Odin
package main
import "core:runtime"

foo :: proc "stdcall" () {
}

main :: proc() {
  runtime.print_string("Hello\n")
  foo()
  runtime.print_string("Goodbye\n")
}
```

...we see the call looks like this: `call void @main.foo()`
...but the function declaration looks like this: `define internal x86_stdcallcc void @main.foo() {`

Let's compare what IR clang generates for some simmilar C code:

https://godbolt.org/z/4Ed5v9b1j

Notice how LLVM puts `x86_stdcallcc` both on the call and the function declaration.

From there I experimented with some minimal IR examples, letting clang optimize them:
```llvm
declare x86_stdcallcc i32 @GetLastError()

define internal void @main.bar() {
decls:
  %0 = call i32 @GetLastError()
  ret void
}

define internal x86_stdcallcc void @main.foo() {
decls:
  ret void
}

define internal void @main.main() {
decls:
  br label %entry

  call void @main.bar()
  call void @main.foo()
  ret void
}

; Function Attrs: noinline
define dso_local x86_stdcallcc i32 @mainCRTStartup() noinline {
decls:
  call void @main.main() noinline
  ret i32 0
}
```
Running this through clang with optimizations (`clang bug.ll -O2 -S`) produces this assembly (abriged), which clearly will run straight into the `int3`:
```asm
main.bar:
	jmp	GetLastError                    # TAILCALL
main.main:
	callq	main.bar
	int3
```
But by changing `call void @main.foo()` to `call x86_stdcall void @main.foo()`, and `call i32 @GetLastError()` to `call x86_stdcallcc i32 @GetLastError()`, clang generates sensible assembly instead:
```asm
main.main:
	jmp	GetLastError                    # TAILCALL
```

Godbolt links for the two variants, for covenience: https://godbolt.org/z/Ys1W3vo9z and https://godbolt.org/z/Y8ssorGbj (notice I had to specify `-target x86_64-pc-windows-msvc`...).

Why clang produces garbage when optimizing this I don't know, especially because `x86_stdcall` doesn't appear to do anything when compiling for amd64 on windows (I'm looking at `X86Subtarget.h`, function `isCallingConvWin64`). Is it just a case of garbage-in-garbage-out, or is there something more subtle going on? Understanding this probably would be good, then I could be more confident that the fix I'm submitting here is a proper fix.

### DIFF
--- a/src/llvm_backend_proc.cpp
+++ b/src/llvm_backend_proc.cpp
@@ -891,6 +891,9 @@ gb_internal lbValue lb_emit_call_internal(lbProcedure *p, lbValue value, lbValue
 
 		LLVMValueRef ret = LLVMBuildCall2(p->builder, fnp, fn, args, arg_count, "");
 
+		auto llvm_cc = lb_calling_convention_map[proc_type->Proc.calling_convention];
+		LLVMSetInstructionCallConv(ret, llvm_cc);
+
 		LLVMAttributeIndex param_offset = LLVMAttributeIndex_FirstArgIndex;
 		if (return_ptr.value != nullptr) {
 			param_offset += 1;


JesseRMeyer:
We never really know what a "proper" fix to LLVM is.  This is an unending problem.  Code gen is great for versions X, Y, Z, and then we ship support for W and something random blows up without explaination.

Thanks for the investigation and the PR.  I think this fix is as good and as simple as any.  If you feel especially motivated, you may want to post in the LLVM discord to see if this problem is known in their group, with a follow-up to their github.

seventh-chord:
I looked briefly at the LLVM language reference manual. I think we are just running into "`call` Instruction, Arguments, point 3":
> The optional “cconv” marker indicates which [calling convention](https://llvm.org/docs/LangRef.html#callingconv) the call should use. If none is specified, the call defaults to using C calling conventions. The calling convention of the call must match the calling convention of the target function, or else the behavior is undefined.

https://llvm.org/docs/LangRef.html#id328

PR #2863, opened 2023-10-14
More os2

jasonKercher:
This pull request puts a dent into os2 for Linux:
 - env
 - pipe
 - errors
 - process
 
For process, I changed the API based on my best guess of the original intention.

For `process_signal`, I changed the signature to not take a `^Process` argument.  My guess was this was meant to act like the `signal` function that can be found on Linux and Windows where it installs signal handlers.  Neither of these platforms (as far as I am aware), can install signal handlers to other processes.  I defined a common set of signals and signal handlers that I *believe* both systems support.  Also, I changed out the `Process_Error` for regular `os2.Error`.

Old:
```odin
Signal :: #type proc()

Kill:      Signal = nil 
Interrupt: Signal = nil 

process_signal :: proc(p: ^Process, sig: Signal) -> Process_Error {
        return .None
}
```
New:
```odin
Signal :: enum {
        Abort,
        Floating_Point_Exception,
        Illegal_Instruction,
        Interrupt,
        Segmentation_Fault,
        Termination,
}

Signal_Handler_Proc :: #type proc "c" (c.int)
Signal_Handler_Special :: enum {
        Default,
        Ignore,
}
 
Signal_Handler :: union {
        Signal_Handler_Proc,
        Signal_Handler_Special,
}

process_signal :: proc(sig: Signal, h: Signal_Handler) -> Error {
        return _process_signal(sig, h)
}
```

I added a `process_get_state` proc:
```odin
process_get_state :: proc(p: Process) -> (Process_State, Error) {
        return _process_get_state(p)
}
```

I also added a Duration to `process_wait` as it seems like a common ability between different systems.

I changed the `Process_Attributes` struct to contain explicit pointers to `stdin`, `stdout` and `stderr` instead of `[]^File`. These get populated if they are passed to `process_start` as non-nil.

```odin
Process_Attributes :: struct {
        dir: string,
        env: []string,
        stdin: ^File,
        stdout: ^File,
        stderr: ^File,
        sys: ^Process_Attributes_OS_Specific,
}
```

As far as it goes, I'm not married to any of the API decisions I made.  For example, to open a child process with pipes to stdout and stdin, the code looks like this:

```odin
child_stdout: os2.File
child_stdin: os2.File
attr: os2.Process_Attributes = {
        stdin = &child_stdin,
        stdout = &child_stdout,
}
os2.process_start("cat", nil, &attr)
```

I don't know.. it feels clunky to me.


jasonKercher:
@flysand7 
Thanks dude. I just copied the stat structure out of `core/os/os_linux.odin`. So this change should probably be back-ported. Looks like I jumped

jasonKercher:
Haha. I fat fingered the close button while typing in that reply. Looks like I jumped the gun not making CI pass first anyway. 

> I'm currently working on binding more syscalls in the sys package, potentially adding Odin's types, so that it's a bit easier to do cross-arch stuff

Definitely looking forward to this!

core/os/os2/env_linux.odin
```diff
@@ -1,30 +1,202 @@
 //+private
 package os2
 
+import "core:slice"
 import "core:runtime"
+import "core:strings"
+import "core:intrinsics"
+
+// the environment is a 0 delimited list of <key>=<value> strings
+_env_map: map[string]string
```
gingerBill, line 10: Is there any reason why a `map` is needed in the first place? I know the environment is a delimited `<key>=<value>` array. Does Linux not have something like this already?

jasonKercher: On Linux, the environment is stored/managed in user space and passed into execve. glibc usually handles all the modification. A map will make for faster lookup, but we could just search through it linearly as well. In my head, I just saw key/value and just immediately thought map. I imagine the only time we would see degradation would be if someone made some kind of `make` clone where environment lookup and modification are more important.

jasonKercher:
Please don't merge yet. I want to go the buildah/podman route and get everything working across architectures. Should I just close?

flysand7:
image

In the top-right corner of the page there's a button that makes PR a draft which prevents it from being merged prematurely. I think you can undraft once you're finished

flysand7:
Any updates?

jasonKercher:
> Any updates?

=] I don't have as much free time as you it would seem (full-time job + 4 kids).  I'm sure you are right about the stat struct, but I want to make sure there are not similar issues with other structs.  I'm going to write a bash script to run my tests across all architectures automatically using `buildah`/`podman`.  I have a private repo for it that I will make public when that part works.

core/os/os2/env_linux.odin
flysand7: I was curious about this point too, so I looked it up. Turns out libc mantains `__environ` variable, which it has gotten from loader where the environment map is stored. When `putenv` or `setenv` is called, it reallocates the buffer for the envs and copies the values. There's also some funny stuff with thread locks on env, because multiple threads may be changing the environment at the same time.

Note that when process is forked, the memory is copied, so the child naturally gets a copy of parent's env. So it doesn't matter where the env is stored as long as both parent and the child access it from the same place.

I personally would love to have our custom env map that is accessible publically, but there's two issues concerning that behaviour:
- Direct insertion into the map is not thread-safe (you might need that if you're planning to spin up children). Honestly not that big of a deal, it just sets up the users for bugs and failures, esp since you don't know when some random-ass library modifies env; but this shouldn't occur because:
- There's no sync with libc's environ. There's no way to specify a custom map type with custom procedure for insertion (at least I don't know of a way to do that, since there's no way to overload operators). The best we can do is to mantain this map with calls to libc's `setenv` and `putenv` when we call `os2` functionality. I don't see performance issues here, since it's a rare operation, but that's the cost of mantaining the map PLUS people would have to make sure not to insert anything into that map directly.

TLDR I didn't read the code so it's just a brain fart for anyone who wants to think about this problem.

flysand7:
Ooh that's interesting! I didn't know about buildah/podman, so thanks for name dropping. Something that executes tests in a VM sounds like it could also be useful for my current work with syscalls, as I'm not sure about whether I've done a correct job on arch-dependent stuff in ARM/x86-32

core/os/os2/env_linux.odin
jasonKercher: There is no issue with modifying the map directly because it is private and hidden behind the public API.  It does, however, require a mutex which currently is not in there.  That is "on the list."  I may reverse course and go with a `[]string` that get individually searched.

The idea that we keep track of our own environment is that os2 is supposed to work without `libc`.  If someone is already using `libc.putenv`, then they are probably not then going to use `os2.process_start`, or vice versa, using `os2.put_env` and then calling `libc.system`.  But you raise a good point.  What if a linked library modifies the environment with `libc.putenv` and you want your environments to sync.  I'm not sure how that would work (or if we even care).  If we are going to sync with `libc`, then we should almost just use `libc.putenv/getenv` ourselves.  Maybe we could determine this at compile time with `when NO_CRT` (or whatever it would be called), but that still won't help for third party libraries.

flysand7:
Something I've realised is that there's a difference between handle inheritance between linux and windows that should get addressed. When you create a child process some handles may or may not be inherited by that process. On linux that is controlled by `O_CLOEXEC` flag. On windows that is controlled by `bInheritHandles` value in the security attributes struct.

The difference lies in the default behavior. On linux by default (unless the flag is specified) the handle is inherited. On windows it is **not inherited**. I think the easiest thing to do here is to use the windows behavior as Odin's default, at least for pipes. Could you pass O_CLOEXEC flag when creating linux pipes, or at least set both ends of the pipe to close on exec?

jasonKercher:
@flysand7 Will do!  I just changed some things in sys/linux.  I wanted to ask you about your `Sig_Set` definition.  I thought `sigset_t` was a `u64` as it's just a bit set of the signals.  This is going to be 16 uints?

```odin
@private _SIGSET_NWORDS :: (1024 / (8 * size_of(uint)))
Sig_Set :: [_SIGSET_NWORDS]uint
```

flysand7:
> @flysand7 Will do! I just changed some things in sys/linux. I wanted to ask you about your `Sig_Set` definition. I thought `sigset_t` was a `u64` as it's just a bit set of the signals. This is going to be 16 uints?

I'm not totally sure about it but it should probably be verified with a C compiler, but when I was making this I simply copied the typedefs from libc.

jasonKercher:
Should be good to go.  This passes [these tests](https://github.com/jasonKercher/os2test) on amd64 and arm64 with the aid of Puklaj's fork fix (#3476 ).

I made a [local CI-like runner bash script thing](https://github.com/jasonKercher/odin-archs) to emulate other architectures in case anyone wants to do that on their own machine.

PR #2864, opened 2023-10-14
Allow for shared object linking of vendor libraries

flysand7:
## Context

Recently one other person has ran into an issue where they have been trying to use vendor packages (in this case raylib) with code hotloading, that requires dynamically linking shared two shared objects (for windows users: read "an exe and a dll").

Since each of these two modules gets a copy of the statically-linked library, the global variables and the TLS is duplicated, resulting in incoherent state of the library between two shared objects. Prime offender here is glfw, that [keeps all of it's global state](https://github.com/glfw/glfw/blob/master/src/init.c#L435) in the thread local storage

For applications that build as multiple shared objects and use libraries that require global state it's necessary to provide a way to link to those libraries dynamically.

## The change

These modules in the `vendor` package get's a new config variable:

```odin
<MODULE>_SHARED :: #config(<MODULE>_SHARED, false)
```

If the user specifies it they're can safely assume that the library they're linking to is linked as a shared objects. This means that:

- If the vendor package provides shared builds, these will be linked
- If the vendor package doesn't provide shared builds, and the library is potentially unsafe to link statically, the compilation fails with an appropriate error message.
- If the library is safe to link statically (it doesn't contain any global state or linked as a shared object by default), nothing is done. In reality config for these isn't defined.

## Notes

There was one place where my fixes required some ugly workarounds for the time being

```diff
-       foreign import lib { 
-               "linux/libraygui.a",
-               // "system:dl",
-               // "system:pthread",
+       when RAYGUI_SHARED {
+               // Note(bumbread): can't panic here, because the users might be expecting to
+               // only use raylib. Let's have them get the error at link-time instead..
+               //#panic("Cannot link libraygui.so: not in the vendor collection")
+               // Note(bumbread): unless we import something the rest of the bindings will
+               // make a compile-time error. This is a bit ugly for now, but in the future
+               // raygui probably needs to be in a separate package.
+               foreign import lib {"_"}
+       } else {
+               // #panic("Cannot link libraygui.a: not in the vendor collection")
+               // TODO(bumbread): apparently this one was missing. This might need
+               // to get rebuilt for linux
+               // foreign import lib { 
+               //      "linux/libraygui.a",
+               //      // "system:dl",
+               //      // "system:pthread",
+               // }
+               foreign import lib {"_"}
        }
```

The comments pretty much explain all of it. In the future the bindings for raygui and raylib would need to be separated into different package, because you may want use one without the other and if you're importing a module you're not necessarily subscribing to linking.

This fix can also be, in principle, implemented in a compiler

## Compatibility

This PR breaks source compatibility with apps that used `GLFW_DYNAMIC`. `GLFW_DYNAMIC` was renamed to `GLFW_SHARED` for consistency reasons.

The name `GLFW_SHARED` was picked due to clarity -- you're "sharing" this package between multiple modules.


sepisoad:
I tested this branch locally, however I get the same error

gingerBill (approved):

gingerBill:
I have approved this but I am not merging yet for the issues found.

flysand7:
@sepisoad 

> * if I use -define:RAYLIB_SHARED=true the build shared library, it cannot be loaded by the main executable
> * if I use -define:RAYLIB_SHARED=true the the main executable, then i get the same error as vanila odin

It sounds like you were defining `RAYLIB_SHARED` in one of your modules but not the other. I'll explain why that doesn't work: If you define `RAYLIB_SHARED` it links to the **shared** variant of the library, otherwise it links to the **static** variant of the library.

This leaves you with two modules linked to two different variants of the library, which runs into the same issue as I've described above.

!!! In order to use this fix you need to define `RAYLIB_SHARED` in **both** of your modules.

Unfortunately there's nothing that can be done about vetting this error on the Odin side

sepisoad:
@flysand7 

In fact I used this in both sides, here is my make file:

```makefile
.PHONY: build_game_dll
build_game_dll:
	odin build game -build-mode:shared -define:RAYLIB_SHARED=true

build:
	odin build . -define:RAYLIB_SHARED=true

run:
	odin run . -define:RAYLIB_SHARED=true
```

however as I mentioned above I get error when loading the shared library, this function fails to load the library:
```golang
dynlib.load_library("game.dylib")
```

flysand7:
> btw, are you also on a macbook m1 machine?

Nope, my machine runs arch linux. Do you have a repo I could try building and maybe inspect what is wrong?

flysand7:
After investigating the issue by cloning and running the code on my machine, I've come to the conclusion that the issue doesn't stem from *whether* raylib is linked dynamically or statically, but rather with how dylib interface is being used.

For the context, @sepisoad's issue is that the shared object (`game.so`) is not being unloaded for some reason, before being opened again, even though `dlclose` is called on it. This is probably one of those weird things with RTLD where it doesn't guarantee that the object is actually unloaded from memory...

In short, @gingerBill, this PR is ready to be merged. The `dlopen` stuff will be investigated in a scope of a separate issue, as it is not related to the functionality being added.

sepisoad:
@flysand7 will you open a new issue for the dlopen issue, or maybe is there alreay an issue for it
I just want to track it

flysand7:
> @flysand7 will you open a new issue for the dlopen issue, or maybe is there alreay an issue for it I just want to track it

First of all I don't have the sufficient context to open a new issue, as it doesn't occur on my machine. You might consider simplifying your hotloading code and maybe open an issue.

BTW in your core i noticed that you were accessing state shared between threads without using atomics or synchronization. Probably not an issue, but yeah if you made a really simple example that fits in just two files that would be helpful

PR #2876, opened 2023-10-17
sys/unix implementation

flysand7:
Currently `sys/unix` implements bare minimum necessary for the rest of the core library to run. Some user applications would want to access the underlying OS functionality, without having to bind syscalls.

The idea is to move those bindings to sys/unix package and allow other modules to use those bindings + allow users to access these bindings in a somewhat type-safe way.

## The final view

The sys/unix package implements functionality common to all unix-likes, potentially making a glibc-like wrapper, and this package is split up into subpackages based on the operating system: sys/linux, sys/darwin, sys/openbsd and sys/freebsd. I'll say it upfront, I don't have the other operating systems so the other operating systems would need to be rebound by other people

The bindings for packages inside sys/unix must be somewhat type-safe, and user-accessible, at least to the same degree libc wrappers are. Odin's typesystem allows for better stuff than C, so we can do better things, like replace define chains with bitfields and enums, allow polymorphic procedure groups for some syscalls that take in different parameters depending on the kind of thing it wants to do and more.

## The conventions taken

Where constants define flags, these are packed into a bitfield.

```odin
OFlags_Bits :: enum {
    RDONLY    = 0,
    WRONLY    = 1,
    RDWR      = 2,
    CREAT     = 6,
    EXCL      = 7,
    /* ... */
}

OFlags :: bit_set[OFlags_Bits; u32]
O_RDONLY    :: OFlags{.RDONLY}
O_WRONLY    :: OFlags{.WRONLY}
/* ... */
```

You can see a few things:
- Enum defines the bits for the file and doesn't contain the usual enum prefix (because there's no need for a prefix, since it's namespaced)
- Alongside enum constants, bitfield constants are also defined. This is made largely for consistency purposes, since some flags actually have complex values that overlap with other flag values (`O_SYNC :: OFlags{._15, ._20}`), and also allow you to choose how you want to pass these. Think of bitfield constants as default and enum constants as "sugar" that we're allowing you to do because we're in odin-land.
- The naming scheme is largely preserved and compatible with POSIX. The names of types are converted to Ada_Case anyway (e.g. `Fd`, `Pid`, `Errno`, `Time_Spec`).

This should allow the following call to `open.2` syscall:

```odin
fd, errno := unix.open("myfile", unix.OFlags{.RDONLY, .CLOEXEC, .NOFOLLOW})
```

as well as

```odin
fd, errno := unix.open("myfile", .RDWR);
```

## Other notes

At the time of creation of this PR you're mostly interested in code that's in `sys/unix/linux` as well as some notes I've sprinkled throughout the core library. I do want to have a discussion about the questions raised. Some issues I've found already:

* `examples/all` didn't compile because `vendor:darwin` was imported unconditionally and I had no way of disabling it with a `when` statement. I don't know which one of "import guard onto `vendor:darwin`" and "separating `examples/all` into `examples/all` and `examples/all_<platform>`" is a better idea.

* `core:math` should not depend on `sys` for random, at least I don't think.


Platin21:
The idea behind sys was actually not to unify anything, most of the content should be highly system specific. As OS is the unification layer, It mostly contains say on darwin just syscalls and little convince wrappers around these. Also some Odinified helper procs.

flysand7:
> The idea behind sys was actually not to unify anything

Then I don't get the naming being `sys/unix`. This was what made me think it is meant to unify unix systems. I might not understand the purpose of putting it into the same directory maybe. If those contain platform-specific functionality and included by platform-specific files then whether they are in the same package or in different packages shouldn't matter *unless* you're planning to specify a common interface.

I'll probably not make a POSIX wrapper over syscalls, but if -- as you say -- unifying access to unix system is not required, then wouldn't it make sense to structure the package more like this:

```
sys/
|- linux/*
|- darwin/*
|- openbsd/*
|- netbsd/*
```

flysand7:
I'll reiterate it here:

The `os` package uses a lot of unix definitions to provide functionality like opening files. This makes changing this package extremely hard without potentially breaking a lot of code that uses `os` for no reason.

One example of this is replacing any "flags" kind of argument with a bitfield. Unlike integers, bitfields can not be casted to different integer types (they can be transmuted and then casted). I used a lot of those in my sys bindings.

It would be much easier to deprecate `sys/linux_old` and `os`, and focus on improving `sys/linux` and `os2`. After `os2` is stabilised, the legacy packages can probably be removed.

I don't know Odin's take on source compatibility, but I've seen it break it. It would be nice to get some feedback from other contributors, see what are the thoughts on that decision.

Platin21:
Might actually make sense to break but we intentionally left os like it was because people still using it and we don't have yet complete impl's of anything and also somehow the new impl sometimes does it's own call wrappers. 

Pratically I'm on your side to throw the stuff in the bin :)

core/sys/linux/00_linux.odin
```diff
@@ -0,0 +1,144 @@
...
+c_timespec :: struct {
+	time_sec:  c_ulong,
+	time_nsec: c_ulong,
+}
+
+Time_Spec :: c_timespec
+
+Time_Val :: struct {
+	seconds: i64,
+	microseconds: int,
```
gingerBill, line 51: Is `int` correct here?

core/sys/linux/00_linux.odin
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Note(flysand): In the case of syscall let's get rid of extra
+// casting. First of all, let these syscalls return iptr, because
+// we'll need to check for Errno anyway. Second of all
+// most parameters are going to be trivially-castable to
+// uptr, so we'll have that.
+@private
+syscall0 :: #force_inline proc "contextless" (nr: uptr)->iptr {
```
gingerBill, line 60: Please add spaces either side of the `->`, `) -> iptr`. (See all other cases too)

core/sys/linux/00_linux.odin
```diff
@@ -0,0 +1,144 @@
...
+		hi = c_ulong(no_sign >> 32)
+		lo = c_ulong(no_sign & 0xffff_ffff)
+		return
+	}
+} else {
+	// ... and on 64-bit architectures it's just a long
+	@private compat64_arg_pair :: #force_inline proc "contextless" (a: i64)->(c_ulong) {
+		return c_ulong(a)
+	}
+}
```
gingerBill, line 144: Where is this used, and why do they have different signatures?

Why not this?
```odin
@private compat64_arg_pair :: #force_inline proc "contextless" (a: i64) -> (hi: c_ulong, lo: c_ulong) {
	when LONG_BITS == 32 {
		no_sign := uint(a)
		hi = c_ulong(no_sign >> 32)
		lo = c_ulong(no_sign & 0xffff_ffff)
	} else {
		lo = c_ulong(a)
		hi = 0
	}
	return
}
```

core/sys/linux/file.odin
```diff
@@ -0,0 +1,1697 @@
...
+	NONBLOCK  = 11,
+	DSYNC     = 12,
+	ASYNC     = 13,
+	DIRECT    = 14,
+	_15       = 15,
+	DIRECTORY = 16,
+	NOFOLLOW  = 17,
+	NOATIME   = 18,
+	CLOEXEC   = 19,
+	_20       = 20,
```
gingerBill, line 27: If this is not needed `_ = 20,` can be written

core/sys/linux/file.odin
```diff
@@ -0,0 +1,1697 @@
...
+///
+/// Available since Linux 1.0
+/// For 32-bit systems a different syscall is used that became available since 2.4
+fstat :: proc "contextless" (fd: Fd, stat: ^Stat)->(Errno) {
+	when LONG_BITS == 64 {
+		ret := syscall(SYS_fstat, stat)
+		return Errno(-ret)
+	} else when LONG_BITS == 32 {
+		ret := syscall(SYS_fstat64, stat)
+		return Errno(-ret)
```
gingerBill, line 1010:
```odin
ret := syscall(SYS_fstat when LONG_BITS == 64 else SYS_fstat64, stat)
return Errno(-ret)
```

core/sys/linux/file.odin
```diff
@@ -0,0 +1,1697 @@
...
+		ret := syscall(SYS_truncate, fd, compat64_arg_pair(length))
+		return Errno(-ret)
+	}
+}
+
+/// Open file, get the file descriptor
+/// Available since Linux 1.0
+/// On ARM64 available since Linux 2.6.16
+open :: proc "contextless" (name: cstring, flags: OFlags, mode: Mode = {})->(Fd, Errno) {
+	when ARCH == "arm64" {
```
gingerBill, line 1392: Why not `ODIN_ARCH == .arm64`?

core/sys/linux/00_linux.odin
```diff
@@ -0,0 +1,144 @@
...
+
+/// Architecture to compile for
+ARCH :: ODIN_ARCH_STRING
+
+/// Bits per long
+when size_of(c.long) == 4 {
+	LONG_BITS :: 32
+} else {
+	LONG_BITS :: 64
+}
```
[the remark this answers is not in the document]
flysand7: Right, this will remove the need to prefix the filename with `00`.. I might be more stupid than I thought haha

core/sys/linux/00_linux.odin
```diff
@@ -0,0 +1,144 @@
...
+	iptr :: i32
+} else when size_of(uptr) == 8 {
+	iptr :: i64
+}
+
+// Packages c and intrinsics are going to be used throughout this
+// package, so I'm re-exporting them in the current scope, so that
+// other files can see the identifiers.
+import "core:c"
+import "core:intrinsics"
```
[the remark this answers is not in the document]
flysand7: For sure, I don' know how they got down there...

core/sys/linux/00_linux.odin
flysand7: Technically yes, it's `i64` that's wrong. Though I think `c.long` is better for clarity

core/sys/linux/00_linux.odin
flysand7: Different return types. We don't want to use 7th register when calling `syscall` (I don't remember on the top of my head where that might cause the issue, actually probably won't it's just a waste of 4 bytes on the stack), nor I want to clobber an extra register on a syscall when it gets inlined.

It's not performance-critical but my mind wouldn't feel at ease if I pass extra parameter for no reason. It's the same reason why I'm doing polymorphic procedure groups for syscalls where a parameter is ignored depending on one of the arguments (+ the type safety)

core/sys/linux/file.odin
```diff
@@ -0,0 +1,1697 @@
...
+	CREAT     = 6,
+	EXCL      = 7,
+	NOCTTY    = 8,
+	TRUNC     = 9,
+	APPEND    = 10,
+	NONBLOCK  = 11,
+	DSYNC     = 12,
+	ASYNC     = 13,
+	DIRECT    = 14,
+	_15       = 15,
```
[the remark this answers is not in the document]
flysand7: But if I do that, wouldn't Odin disallow accessing this field by name? I can change it to casting to enum before going into the bitfield, that's probably better, there's no reason to keep an extra enum field, really...

core/sys/linux/file.odin
flysand7: At the time I decided I was doing that, I was probably thinking that I could put in any architecture name I want, even for those not supported by Odin as of the current time, so I could make this package somewhat future-proof. But then I gave up at the idea

Maybe it's best that I change them back to enums, at least those get checked

core/sys/linux/00_linux.odin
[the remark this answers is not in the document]
flysand7: Nvm it doesn't remove that need, I learned that the painful way. I did that refactor either way.

core/sys/linux/file.odin
flysand7: Might have to refuse this suggestion... I don't like long lines, and I feel like its much clearer to pattern-match the difference between two blocks than read an inline when statement

gingerBill: That's fine. It was only a suggestion.

flysand7:
Squashed all the commits, it's ready for review and merge

PR #2877, opened 2023-10-18
[os] Darwin: Add os.set_env() and os.unset_env()

wjlroe:


Platin21:
Does it make still sense to add this to the dead os lib? I dunno if os2 has this at all. As it's sort of non permanent I dunno there shouldn't be two versions of that somehow.

PR #2879, opened 2023-10-18
Ensure required licenses are in distributable bundles

jcmdln:
This PR ensures that all license files are included in an Odin release/nightly distributable bundle.

Odin releases/nightlies are missing Odin's license. This violates clauses 1 and 2 of BSD-3 as `*.odin` files are `source code`, and the `odin` executable is a `binary form`:

```
1. Redistributions of source code must retain the above copyright notice, this
   list of conditions and the following disclaimer.

2. Redistributions in binary form must reproduce the above copyright notice,
   this list of conditions and the following disclaimer in the documentation
   and/or other materials provided with the distribution.
```

From `vendor/`, the only missing required license is for `vendor/botan/bindings/botan-3.lib`. This violates clause 2 of BSD-2 as this is in `binary form`:

```
2. Redistributions in binary form must reproduce the above copyright
   notice, this list of conditions, and the following disclaimer in the
   documentation and/or other materials provided with the distribution.
```

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -26,6 +26,7 @@ jobs:
           rm bin/llvm/windows/LLVM-C.lib
           mkdir dist
           cp odin.exe dist
+          cp LICENSE dist
           cp LLVM-C.dll dist
           cp -r shared dist
           cp -r core dist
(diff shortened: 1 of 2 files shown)


Kelimion:
Good catch.

PR #2883, opened 2023-10-20
Implement stack probing on windows.

flysand7:
Code taken out straight from [ciabatta](https://github.com/flysand7/ciabatta/blob/main/arch/x64_x86-64/chkstk.asm), it's been tested on Windows. This should make `intrinsics.alloca` with large sizes work.


flysand7:
I just realised This code has mixed tabs and spaces...

PR #2884, opened 2023-10-20
Editor config

flysand7:
A few times there's been a precedent with people pull-requesting code with space indentation. Some of the code actually got into the master branch (mostly in `vendor`, botan is formatted exclusively with spaces). I thought it would be a good idea to introduce .editorconfig for people who are contributing and have different editor settings

I didn't touch the stuff in vendor, since the diff would be too large, but this could be fixed

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+
+root = true
+
+[*.{odin,c,h,cpp}]
+indent_style = tab
+end_of_line = lf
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,6 @@
+
+root = true
+
+[*.{odin,c,h,cpp}]
```
Kelimion, line 4: And `py`, for the big math test rig.

flysand7: Maybe I should just say tabs to all files? To not bother with extensions... I don't think anything would break with tabs

PR #2885, opened 2023-10-21
don't disable panic when -disable-assert is set

laytan:
If this is intentional I find it very weird behaviour, imo (and that of most people in the discord) `panic` should always cause a call to the failure proc.

### DIFF
--- a/core/runtime/core_builtin.odin
+++ b/core/runtime/core_builtin.odin
@@ -817,7 +817,6 @@ assert :: proc(condition: bool, message := "", loc := #caller_location) {
 }
 
 @builtin
-@(disabled=ODIN_DISABLE_ASSERT)
 panic :: proc(message: string, loc := #caller_location) -> ! {
 	p := context.assertion_failure_proc
 	if p == nil {


laytan:
Current behaviour is also weird because of the fact that `panic` is a divergent procedure, until you `-disable-assert`

flysand7:
To further clarify the rationale behind this pull request:

Some people insert asserts as "sanity checks" with the idea that these would never happen in a working program, and panics as a way to signify an unrecoverable error where it doesn't make sense for the program to keep working. (I guess it's the same idea of error handling as in Go, except panics are __actually__ unrecoverable, so it's closer to Rust)

Disabling asserts is meant to be somewhat of a performance thing where we free up branch predictor slots to use on something better.

laytan:
Further thinking about divergent procedures and @disabled. Maybe it shouldnt be allowed at all (@disabled on a divergent procedure)?

laytan:
Even though I think this is a mistake, if that is what you wanted I will just create my own panic procedure locally.

I do think it should not be a divergent procedure though, the interaction is weird with the compiler not allowing further statements even though -disable-assert would make them execute. In other words, a call to panic can return to the caller.

flysand7:
@laytan what is a situation where disable-assert causes compilation errors due to it being divergent proc call: I'm curious because so far I wasnt able to imagine a situation where that makes sense

flysand7:
Closing as it is fixed in commit [f3f0ab](https://github.com/odin-lang/Odin/commit/f3f0ab6e2ccac1f69358ea435edec27e3d42efcc)

laytan:
@flysand7 for example:
```odin
foo :: proc() {
    if test {
        panic() // might be compiled out, still want to exit
        os.exit(1) // doesn't compile because no statements may come after the panic call even though this is reasonable to want
        return // even a return isn't allowed
    }
    // important stuff....
}
```

laytan:
Thank you Bill for changing this

PR #2886, opened 2023-10-21
Allow no CRT linking on Linux

flysand7:
Unlike windows there's not much point in doing no-crt builds, but it's nice to have the option anyway, for consistency sake.

One thing users should keep in mind when using this option on linux is that they have no way to create threads because threads are managed by libc. It doesn't help that these threads need to get the TLB and synchronize in a way that depends on a particular dynamic loader installed in the system (typically either musl or glibc)


flysand7:
To let you know, this is ready to merge, I probably won't add anything more to this PR

PR #2888, opened 2023-10-22
Correct loadI64 and storeI64 methods in runtime.js

thetarnav:
Corrects loadI64 and storeI64 methods that were left the same as u64 ones. Just loading the last 32 bits as an i32 instead of u32 seem to do the job well without resolving to using getBigInt64.

### DIFF
--- a/vendor/wasm/js/runtime.js
+++ b/vendor/wasm/js/runtime.js
@@ -63,9 +63,8 @@ class WasmMemoryInterface {
 		return lo + hi*4294967296;
 	};
 	loadI64(addr) {
-		// TODO(bill): loadI64 correctly
 		const lo = this.mem.getUint32(addr + 0, true);
-		const hi = this.mem.getUint32(addr + 4, true);
+		const hi = this.mem.getInt32 (addr + 4, true);
 		return lo + hi*4294967296;
 	};
 	loadF32(addr)  { return this.mem.getFloat32(addr, true); }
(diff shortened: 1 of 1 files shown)


vendor/wasm/js/runtime.js
```diff
@@ -63,9 +63,8 @@ class WasmMemoryInterface {
 		return lo + hi*4294967296;
 	};
 	loadI64(addr) {
-		// TODO(bill): loadI64 correctly
 		const lo = this.mem.getUint32(addr + 0, true);
-		const hi = this.mem.getUint32(addr + 4, true);
+		const hi = this.mem.getInt32 (addr + 4, true);
```
flysand7, line 67: Maybe remove this space before open paren?

flysand7 (commented):
I'm not familiar with js runtime and what in general is expected from it, I wonder if this still counts as "correctly" loading those values... I mean all numbers are 64-bit floats and floats can not fully represent integers, so we're stuck at 52 bits of precision by loading integers as floats. I wonder if getting rid of that imprecision is what this TODO is really about

thetarnav:
There is a [`getBigInt64`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/getBigInt64) method available as well, where the number is loaded as a `bigint` where no precision is lost. The problem is that most other js APIs will still require a `number` so it would have to be converted back anyway, and loose precision. U64 is loaded [in the same way](https://github.com/odin-lang/Odin/blob/master/vendor/wasm/js/runtime.js#L60-L64) btw, instead of using `getBigUint64`. Also i64 methods are currently only being used [here](https://github.com/odin-lang/Odin/blob/master/vendor/wasm/js/runtime.js#L1408-L1417) where the loose of precision is irrelevant as the values are already numbers, and supporting negative values, which this pr does, would be useful.
Right now my feel for it is—stick with numbers for the most numeric valus. Not sure when would you actually care about passing that big numbers between odin and js.

vendor/wasm/js/runtime.js
thetarnav: I was trying to follow the existing style, you can see similar aligning here: https://github.com/odin-lang/Odin/blob/master/vendor/wasm/js/runtime.js#L102-L105

PR #2889, opened 2023-10-22
Fix bit_* calls in `core:simd/x86`

jakubtomsu:
It seems like the simd/x86 package wasn't updated when the bitwise simd intrinsics were renamed, this should fix it.


flysand7 (approved):

flysand7:
Interesting that it wasn't caught by our CI. Should probably add `simd/x86` to `examples/all`, just in case this happens again

Kelimion:
> Interesting that it wasn't caught by our CI. Should probably add `simd/x86` to `examples/all`, just in case this happens again

Good call. That was probably why.

Kelimion:
Thanks, @jakubtomsu.

PR #2891, opened 2023-10-23
GLFW binding fixes

flysand7:
## Fixes implemented

- `GLFWmonitorfun` takes an additional [event parameter](https://www.glfw.org/docs/3.3/group__monitor.html#gab39df645587c8518192aa746c2fb06c3)

- `GLFW_TRUE` and `GLFW_FALSE` were defined as untyped boolean constants, despite the fact that `glfwInit` returns boolean. For consistency and useability in `if` statements, the functions were changed from returning `int` to returning `b32`.

- Replaced `WindowHint` function with a polymorphic procedure group, allowing to take both `c.int` and `b32` parameters for hints.

## Compatibility note

This PR breaks compatibility on any code that calls `glfw.Init()` function and checks for it's return value. In before, the check would have needed to be done like that:

```odin
if glfw.Init() != 0 {
    // log error, exit
}
```

or like that:

```odin
if ! cast(bool) glfw.Init() {
    // log error, exit
}
```

The first variant becomes broken, one needs to remove `!= 0` to make their code compile again

### DIFF
--- a/vendor/glfw/bindings/bindings.odin
+++ b/vendor/glfw/bindings/bindings.odin
@@ -39,7 +39,7 @@ when ODIN_OS == .Windows {
 /*** Functions ***/
 @(default_calling_convention="c", link_prefix="glfw")
 foreign glfw {
-	Init      :: proc() -> bool ---
+	Init      :: proc() -> b32 ---
 	Terminate :: proc() ---
 	
 	InitHint  :: proc(hint, value: c.int) ---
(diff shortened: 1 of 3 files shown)


gingerBill:
Why was compatibility broke with this? Isn't return value a boolean to begin with? It might have been a `c.int`, but was always being used as a boolean.

flysand7:
Note that I specifically mean source compatibility, meaning some of the user's code won't compile if this is merged. It will work if the code is slightly rewritten, it's not a big deal

gingerBill:
Yes. I am asking why are you going for strict source compatibility? C's semantics mean you can do `if (!glfwInit())` no problem because of implicit type conversions.

I'd rather better reflect the intent and usage than the direct types.

flysand7:
Yeah I mean if you're fine with breaking it, sure. I broke it on purpose because the other option is better, as you said.

PR #2893, opened 2023-10-25
Adding LLVM14 - Missing config but says supported in Error message.

Pix-xiP:
As it says on the tin, the build_odin.sh script says 14 is supported in the list of errors, however there is no check in the script to set it. Just popped it in and tested on Arch with the latest build and it worked with 14.

### DIFF
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -36,12 +36,12 @@ if [ -z "$LLVM_CONFIG" ]; then
 	elif [ -n "$(command -v llvm-config12)" ]; then  LLVM_CONFIG="llvm-config-12"
 	elif [ -n "$(command -v llvm-config11)" ]; then  LLVM_CONFIG="llvm-config-11"
 	# fallback
-  elif [ -n "$(command -v llvm-config-14)" ]; then
-    echo "WARN: Odin does not officially support LLVM14"
-    LLVM_CONFIG="llvm-config-14"
-  elif [ -n "$(command -v llvm-config14)" ]; then
-    echo "WARN: Odin does not officially support LLVM14"
-    LLVM_CONFIG="llvm-config14"
+	elif [ -n "$(command -v llvm-config-14)" ]; then
+		echo "WARN: Odin does not officially support LLVM14"
+		LLVM_CONFIG="llvm-config-14"
+	elif [ -n "$(command -v llvm-config14)" ]; then
+		echo "WARN: Odin does not officially support LLVM14"
+		LLVM_CONFIG="llvm-config14"
 	else
 		error "No llvm-config command found. Set LLVM_CONFIG to proceed."
 	fi
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -27,11 +27,13 @@ error() {
 if [ -z "$LLVM_CONFIG" ]; then
 	# darwin, linux, openbsd
 	if   [ -n "$(command -v llvm-config-17)" ]; then LLVM_CONFIG="llvm-config-17"
+	elif [ -n "$(command -v llvm-config-14)" ]; then LLVM_CONFIG="llvm-config-14"
 	elif [ -n "$(command -v llvm-config-13)" ]; then LLVM_CONFIG="llvm-config-13"
 	elif [ -n "$(command -v llvm-config-12)" ]; then LLVM_CONFIG="llvm-config-12"
 	elif [ -n "$(command -v llvm-config-11)" ]; then LLVM_CONFIG="llvm-config-11"
 	# freebsd
 	elif [ -n "$(command -v llvm-config17)" ]; then  LLVM_CONFIG="llvm-config-17"
+	elif [ -n "$(command -v llvm-config14)" ]; then  LLVM_CONFIG="llvm-config-14"
 	elif [ -n "$(command -v llvm-config13)" ]; then  LLVM_CONFIG="llvm-config-13"
 	elif [ -n "$(command -v llvm-config12)" ]; then  LLVM_CONFIG="llvm-config-12"
 	elif [ -n "$(command -v llvm-config11)" ]; then  LLVM_CONFIG="llvm-config-11"
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -27,18 +27,21 @@ error() {
 if [ -z "$LLVM_CONFIG" ]; then
 	# darwin, linux, openbsd
 	if   [ -n "$(command -v llvm-config-17)" ]; then LLVM_CONFIG="llvm-config-17"
-	elif [ -n "$(command -v llvm-config-14)" ]; then LLVM_CONFIG="llvm-config-14"
 	elif [ -n "$(command -v llvm-config-13)" ]; then LLVM_CONFIG="llvm-config-13"
 	elif [ -n "$(command -v llvm-config-12)" ]; then LLVM_CONFIG="llvm-config-12"
 	elif [ -n "$(command -v llvm-config-11)" ]; then LLVM_CONFIG="llvm-config-11"
 	# freebsd
 	elif [ -n "$(command -v llvm-config17)" ]; then  LLVM_CONFIG="llvm-config-17"
-	elif [ -n "$(command -v llvm-config14)" ]; then  LLVM_CONFIG="llvm-config-14"
 	elif [ -n "$(command -v llvm-config13)" ]; then  LLVM_CONFIG="llvm-config-13"
 	elif [ -n "$(command -v llvm-config12)" ]; then  LLVM_CONFIG="llvm-config-12"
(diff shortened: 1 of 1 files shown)


jcmdln:
This has come up a lot so I agree it should just be handled. LLVM-14 is not officially supported so if there are any issues discovered with Odin + LLVM14 they are not likely to be patched, which is why this was made an explicit opt-in.

Personally I would like to see LLVM14 added to the `# fallback` section so that all officially supported versions of LLVM are tried first. Also the FreeBSD commands as they are are wrong and could be corrected:

```diff
diff --git a/build_odin.sh b/build_odin.sh
index 2a2505c9..ea6ca328 100755
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -31,14 +31,16 @@ if [ -z "$LLVM_CONFIG" ]; then
        elif [ -n "$(command -v llvm-config-12)" ]; then LLVM_CONFIG="llvm-config-12"
        elif [ -n "$(command -v llvm-config-11)" ]; then LLVM_CONFIG="llvm-config-11"
        # freebsd
-       elif [ -n "$(command -v llvm-config17)" ]; then  LLVM_CONFIG="llvm-config-17"
-       elif [ -n "$(command -v llvm-config13)" ]; then  LLVM_CONFIG="llvm-config-13"
-       elif [ -n "$(command -v llvm-config12)" ]; then  LLVM_CONFIG="llvm-config-12"
-       elif [ -n "$(command -v llvm-config11)" ]; then  LLVM_CONFIG="llvm-config-11"
+       elif [ -n "$(command -v llvm-config17)" ]; then  LLVM_CONFIG="llvm-config17"
+       elif [ -n "$(command -v llvm-config13)" ]; then  LLVM_CONFIG="llvm-config13"
+       elif [ -n "$(command -v llvm-config12)" ]; then  LLVM_CONFIG="llvm-config12"
+       elif [ -n "$(command -v llvm-config11)" ]; then  LLVM_CONFIG="llvm-config11"
        # fallback
+       elif [ -n "$(command -v llvm-config-14)" ]; then LLVM_CONFIG="llvm-config-14"
+       elif [ -n "$(command -v llvm-config14)" ]; then  LLVM_CONFIG="llvm-config14"
        elif [ -n "$(command -v llvm-config)" ]; then LLVM_CONFIG="llvm-config"
        else
-               error "No llvm-config command found. Set LLVM_CONFIG to proceed."
+               error "No supported llvm-config command found. Set LLVM_CONFIG to proceed."
        fi
 fi
 ```

**Comment by jcmdln** [comment]
To clarify, the ordering is important so that when Odin cuts releases for Linux it uses an officially supported version of LLVM. If LLVM14 is added above LLVM13 then the pipeline would attempt to use LLVM14 which is bundled with GitHub's `actions/runner-images`.

If it's instead added to the `# fallback` section, the Odin pipeline will continue to cut releases as expected and users who have distributions that only provide LLVM14 will be able to build as they currently do, as this version restriction was never enforced in the first place.

**Comment by Pix-xiP** [comment]
Makes sense to me, just figured it should probably exist in there if the message is gonna tell people about it, I pulled 14 cause it mentioned,  builds fine on Arch, but I dunno about other system.

**Comment by jcmdln** [comment]
I agree. Maybe this is a good opportunity to add some kind of warning message?

```
diff --git a/build_odin.sh b/build_odin.sh
index 2a2505c9..025e51cd 100755
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -31,14 +31,20 @@ if [ -z "$LLVM_CONFIG" ]; then
        elif [ -n "$(command -v llvm-config-12)" ]; then LLVM_CONFIG="llvm-config-12"
        elif [ -n "$(command -v llvm-config-11)" ]; then LLVM_CONFIG="llvm-config-11"
        # freebsd
-       elif [ -n "$(command -v llvm-config17)" ]; then  LLVM_CONFIG="llvm-config-17"
-       elif [ -n "$(command -v llvm-config13)" ]; then  LLVM_CONFIG="llvm-config-13"
-       elif [ -n "$(command -v llvm-config12)" ]; then  LLVM_CONFIG="llvm-config-12"
-       elif [ -n "$(command -v llvm-config11)" ]; then  LLVM_CONFIG="llvm-config-11"
+       elif [ -n "$(command -v llvm-config17)" ]; then  LLVM_CONFIG="llvm-config17"
+       elif [ -n "$(command -v llvm-config13)" ]; then  LLVM_CONFIG="llvm-config13"
+       elif [ -n "$(command -v llvm-config12)" ]; then  LLVM_CONFIG="llvm-config12"
+       elif [ -n "$(command -v llvm-config11)" ]; then  LLVM_CONFIG="llvm-config11"
        # fallback
+       elif [ -n "$(command -v llvm-config-14)" ]; then
+               echo "WARN: Odin does not officially support LLVM14"
+               LLVM_CONFIG="llvm-config-14"
+       elif [ -n "$(command -v llvm-config14)" ]; then
+               echo "WARN: Odin does not officially support LLVM14"
+               LLVM_CONFIG="llvm-config14"
        elif [ -n "$(command -v llvm-config)" ]; then LLVM_CONFIG="llvm-config"
        else
-               error "No llvm-config command found. Set LLVM_CONFIG to proceed."
+               error "No supported llvm-config command found. Set LLVM_CONFIG to proceed."
        fi
 fi
 ```

Pix-xiP:
Not opposed to that either, more information is better in this case imo. 

Looks good to me otherwise unless you want me to change anything else

build_odin.sh
```diff
@@ -27,11 +27,13 @@ error() {
 if [ -z "$LLVM_CONFIG" ]; then
 	# darwin, linux, openbsd
 	if   [ -n "$(command -v llvm-config-17)" ]; then LLVM_CONFIG="llvm-config-17"
+	elif [ -n "$(command -v llvm-config-14)" ]; then LLVM_CONFIG="llvm-config-14"
```
jcmdln, line 30: Move this below the `llvm-config-11` condition in this section

build_odin.sh
```diff
@@ -27,11 +27,13 @@ error() {
...
 	elif [ -n "$(command -v llvm-config-13)" ]; then LLVM_CONFIG="llvm-config-13"
 	elif [ -n "$(command -v llvm-config-12)" ]; then LLVM_CONFIG="llvm-config-12"
 	elif [ -n "$(command -v llvm-config-11)" ]; then LLVM_CONFIG="llvm-config-11"
 	# freebsd
 	elif [ -n "$(command -v llvm-config17)" ]; then  LLVM_CONFIG="llvm-config-17"
+	elif [ -n "$(command -v llvm-config14)" ]; then  LLVM_CONFIG="llvm-config-14"
```
jcmdln, line 36: Move this below the `llvm-config11` condition in this section

jcmdln (changes requested):
At minimum, these are the changes that I think should happen.

Without these changes, the Odin pipeline will implicitly use LLVM-14 which is not officially supported, but apparently works. This is visible in the pipeline associated with the current commit: https://github.com/odin-lang/Odin/actions/runs/6642566987/job/18047619005?pr=2893#step:4:5

Pix-xiP:
Mibad, I though you had made edits when I read this before, didn't realise you were just pushing ideas.

I'll add the warning version and see if that vibes

build_odin.sh
```diff
@@ -36,7 +36,12 @@ if [ -z "$LLVM_CONFIG" ]; then
 	elif [ -n "$(command -v llvm-config12)" ]; then  LLVM_CONFIG="llvm-config-12"
 	elif [ -n "$(command -v llvm-config11)" ]; then  LLVM_CONFIG="llvm-config-11"
 	# fallback
-	elif [ -n "$(command -v llvm-config)" ]; then LLVM_CONFIG="llvm-config"
+  elif [ -n "$(command -v llvm-config-14)" ]; then
+    echo "WARN: Odin does not officially support LLVM14"
+    LLVM_CONFIG="llvm-config-14"
+  elif [ -n "$(command -v llvm-config14)" ]; then
+    echo "WARN: Odin does not officially support LLVM14"
+    LLVM_CONFIG="llvm-config14"
```
jcmdln, line 44: Please use tabs instead of spaces

jcmdln:
Thanks for making the changes @Pix-xiP! Sorry I should have explained that those were just ideas I had. Last thing is a minor style issue which I've left a comment for. After that's resolved I think this this is ready to be merged though we'll have to wait for a maintainer to do so.

The pipeline is failing on the `HEAD` of the `master` branch, so ignore the errors regarding it for now as they are unrelated to this PR.

build_odin.sh
Pix-xiP: Done!

Pix-xiP:
Cool, hopefully someone with the big guns gets a free moment to peek this then. 

My bad on the spaces, it didn't even register that it might use tabs.

flysand7:
Related: #3013

laytan:
Looks like this has been added at some point somewhere else

PR #2894, opened 2023-10-25
New built-in procedures for unions

jakubtomsu:
# Overview
Currently unions are kind of a black box when it comes to their structure and memory layout, unless you use RTTI. Other types have some simple compile time reflection intrinsics, like `offset_of`, `type_proc_*`, `type_field_type`, `type_is_specialization_of`, and many more. This makes it difficult to use Unions for more complex or low-level tasks. There were too many times when I had to resort to an enum with `#raw_union` struct to get around these problems...

The aim of this PR is to add a few simple built-in procedures to make unions much more powerful!

Note: any ideas/feedback is very welcome!

## Additions
The following procedures were added as built-in intrinsics:
```odin
// Returns the union tag type. Can be either u8, u16, u32 or u64.
type_union_tag_type :: proc($U: typeid) -> typeid where type_is_union(U) ---

// Returns the tag value offset within the union, in bytes.
type_union_tag_offset :: proc($U: typeid) -> int where type_is_union(U) ---

// Returns the tag of first element in the union. It's 0 when using `#no_nil`, 1 otherwise.
// This is necessarry for converting between variant index and tag.
type_union_base_tag_value :: proc($T: typeid) -> int where type_is_union(U) ---

// Returns total number of variants in the union.
type_union_variant_count :: proc($T: typeid) -> int where type_is_union(T) ---

// Returns a type of an union variant at a given index.
// Note: the index always starts at zero, whereas tag usually starts at 1, unless using `#no_nil`.
//       This makes it simpler to reason about, since 0 is always the first variant.
type_variant_type_of :: proc($U: typeid, $index: int) -> typeid where type_is_union(U) ---

// Returns an index (again, not a tag) of an union variant.
// Note: Tag is equal to `type_union_base_tag_value(U) + type_variant_index_of(U, V)`.
type_variant_index_of :: proc($U, $V: typeid) -> int where type_is_union(U) ---
```
Example:
```odin
Foo :: union {
	f32, // index 0, tag 1
	u16, // index 1, tag 2
	u8,  // index 2, tag 3
}

// Memory layout is equivalent to Foo
Raw_Foo :: struct {
	using variants: struct #raw_union {
		variant_0: f32,
		variant_1: u16,
		variant_2: u8,
	},
	tag:            u32,
}

type_union_tag_type(Foo) == u32
type_union_tag_offset(Foo) == 4
type_union_variant_count(Foo) == 3
type_union_base_tag_value(Foo) == 1
type_variant_type_of(Foo, 1) == u16
type_variant_index_of(Foo, u8) == 2
```

# Usage
Here are just a few examples of how the new features could be useful.
Note: the data structures could be much more powerful when made specifically for a given system, rather than as a generic container.

## Using variant index to replace enums
This would allow union variant indexes to be used almost as an enum. Many entity systems of games written in odin tend to use unions, and I can see a lot of potential for features like this:
```odin
Entity :: struct { variant: Entity_Variant, ... }
Entity_Variant :: union { Player, Enemy, Boss, }
Player_State :: union { Player_Walking, Player_Flying }

// Could be used for filtering scene queries like raycasts, shape overlaps, ...
// Alternative to enum bit_sets.
Entity_Variant_Set :: bit_set[0 ..< intrinsics.type_union_variant_count(Entity_Variant)]

// A constant for some gameplay purposes.
// It's necessarry to use the index instead of the type in cases when you don't store the whole union.
// Alternative to regular enum.
PLAYER_STATE_REQUIRED_TO_ENTER_HOUSE :: intrinsics.type_variant_index_of(Player_State, Walking)

Entity_System :: struct {
  // ... entity data ...
  
  // Counter for entity variants, but this could potentionally store any per-variant data.
  // Alternative to enum arrays.
  variant_counter: [intrinsics.type_union_variant_count(Entity_Variant)]int,
}
```
I've talked about this with @karl-zylinski and has also encountered numerous limitations like this.

## Low-level union manipulation
Pointer arithmetic for doing dangerous stuff :P
```odin
get_tag :: proc(u: $T) -> int where intrinsics.type_is_union(T) {
	u := u
	return int((cast(^intrinsics.type_union_tag(T))(uintptr(&u) + intrinsics.type_union_tag_offset(T)))^)
}

set_tag :: proc(u: ^$T, tag: int) where intrinsics.type_is_union(T) {
	TAG :: intrinsics.type_union_tag(T)
	(cast(^TAG)(uintptr(u) + intrinsics.type_union_tag_offset(T)))^ = TAG(tag)
}

get_variant_index :: proc(u: $T) -> int where intrinsics.type_is_union(T) {
	return get_tag(u) - intrinsics.type_union_base_tag_value(T)
}
```

## Packed union pool
A pool with free list buckets for each one of the variants.
```odin
Packed_Union_Pool :: struct($T: typeid) where intrinsics.type_is_union(T) {
	// Ideally an fast sequential allocator like an arena.
	allocator:  runtime.Allocator,
	free_lists: [intrinsics.type_union_variant_count(T)]rawptr,
}

Packed_Union_Pool_Item :: struct($T: typeid) { data: T, next_free: rawptr, }

packed_union_pool_init :: proc(p: ^$T/Packed_Union_Pool($U), allocator := context.allocator) {
	p^ = { allocator = allocator, }
}

packed_union_pool_insert :: proc(p: ^$T/Packed_Union_Pool($U), val: $V) -> (res: ^V, err: mem.Allocator_Error) #optional_allocator_error {
	ptr := cast(^Packed_Union_Pool_Item(V))p.free_lists[intrinsics.type_variant_index_of(U, V)]
	if ptr == nil {
		ptr = new(Packed_Union_Pool_Item(V), p.allocator) or_return
	} else {
		p.free_lists[intrinsics.type_variant_index_of(U, V)] = ptr.next_free
	}
	ptr^ = { data      = val, next_free = nil, }
	return &ptr.data, nil
}

packed_union_pool_remove :: proc(p: ^$T/Packed_Union_Pool($U), ptr: ^$V) {
	assert(ptr != nil)
	item := cast(^Packed_Union_Pool_Item(V))ptr
	item.next_free = p.free_lists[intrinsics.type_variant_index_of(U, V)]
	p.free_lists[intrinsics.type_variant_index_of(U, V)] = item
}
```

### Packed union arrays
Pack union variants into multiple sub-arrays, to reduce the overhead. This would be ideally implemented specifically for a given system, e.g. game entities.
```odin
Packed_Union_Array :: struct($T: typeid) where intrinsics.type_is_union(T) {
	variants:  [intrinsics.type_union_variant_count(T)]struct { data: []u8, len:  int, },
	allocator: runtime.Allocator,
}

packed_union_array_init :: proc(a: ^$T/Packed_Union_Array($U), cap: int = 32, allocator := context.allocator) {
	a^ = { allocator = allocator, }
	// RTTI hack since a loop cannot produce compile-time constant
	ti := reflect.type_info_base(type_info_of(U)).variant.(reflect.Type_Info_Union)
	for &v, i in a.variants {
		data, _ := mem.alloc_bytes_non_zeroed(cap * ti.variants[i].size)
		v = { data = data, len  = 0 }
	}
}

packed_union_array_append :: proc(a: ^$T/Packed_Union_Array($U), value: $V) where intrinsics.type_is_variant_of(U, V) {
	variants := &a.variants[intrinsics.type_variant_index_of(U, V)]
	if variants.len >= len(variants.data) { /* alloc more */ }
	data := packed_union_array_get_data_buf(a^, V)
	data[variants.len / size_of(V)] = value
	variants.len += size_of(V)
}

packed_union_array_set :: proc(a: $T/Packed_Union_Array($U), index: int, value: $V) where intrinsics.type_is_variant_of(U, V) {
	packed_union_array_get_data(a, V)[index] = value
}

packed_union_array_get :: proc(a: $T/Packed_Union_Array($U), $V: typeid, index: int) -> V where intrinsics.type_is_variant_of(U, V) {
	return packed_union_array_get_data(a, V)[index]
}

packed_union_array_get_data :: proc(a: $T/Packed_Union_Array($U), $V: typeid) -> []V where intrinsics.type_is_variant_of(U, V) {
	vars := a.variants[intrinsics.type_variant_index_of(U, V)]
	return slice.reinterpret([]V, vars.data)[:vars.len / size_of(V)]
}

packed_union_array_get_data_buf :: proc(a: $T/Packed_Union_Array($U), $V: typeid) -> []V where intrinsics.type_is_variant_of(U, V) {
	return slice.reinterpret([]V, a.variants[intrinsics.type_variant_index_of(U, V)].data)
}
```


core/intrinsics/intrinsics.odin
```diff
@@ -162,7 +162,12 @@ type_is_matrix           :: proc($T: typeid) -> bool ---
 type_has_nil :: proc($T: typeid) -> bool ---
 
 type_is_specialization_of :: proc($T, $S: typeid) -> bool ---
+
 type_is_variant_of :: proc($U, $V: typeid) -> bool where type_is_union(U) ---
+type_union_tag :: proc($U: typeid) -> typeid where type_is_union(U) ---
```
karl-zylinski, line 167: Is perhaps the name `type_union_tag` weird? Sounds like it would return a tag (which I think of as an int), but it returns a typeid.

karl-zylinski: (I talked to jakub separately about this and I got my confusion fixed)

jakubtomsu: Yep, but maybe a better name is something like `type_union_tag_type`, similar to `type_field_type` intrinsic for getting struct field type by name.

flysand7: I was just about to comment that `tag_type` souns even weirder: This name suggests that the _type_ of the tag is returned, rather than the value of the tag. Maybe `type_union_tag_value` returning typeid is a better option?

jakubtomsu: But the type of the tag _is_ returned :). You can get the variant index with `type_variant_index_of`, and the tag with `min(U) + type_variant_index_of(U, V)`

jakubtomsu: The variant index/tag is an integer, however `type_union_tag_type` returns compile-time typeid constant.

flysand7: Can you help me understand it, let's say in a union you brought up as an example:

```
Player_State :: union { Walking, Flying }
```

What are the tag types and values in this union?

jakubtomsu: The tag type could be u8, u16, u32, or u64, depending on the number of variants within the union, and the required alignment.

The variant values are like values in enums, so Walking would have index 0 (but tag 1), and Flying index 1 (but tag 2).
The tag starts at 1, since 0 is reserved for nil, unless you use `#no_nil`. The reason why the intrinsics return variant index and not the tag is to make both nil/no_nil consistent.

A better example would be something like the following:
```odin
// size = 8 bytes (4 bytes data, 4 aligned variant index)
Foo :: union {
	u8, // index 0, tag 1
	f32, // index 1, tag 2
	i32, // index 2, tag 3
}

Raw_Foo :: struct {
	using _:       struct #raw_union {
		variant_0: u8,
		variant_1: f32,
		variant_2: i32,
	},
	variant_index: u32,
}

// The following is true:

type_union_tag_type(Foo) == u32
type_union_tag_offset(Foo) == 4

type_variant_index_of(Foo, u8) == 0
type_variant_index_of(Foo, i32) == 2
type_variant_type_of(Foo, 0) == u8
...
```

gingerBill:
I am not a fan allowing `len`. `min`, and `max` on union types. That feels extremely wrong to me, and not at all intuitive.

jakubtomsu:
Ok I understand `min` and `max`, but what feels wrong about `len`? It's just the number of variants, I don't think it could be interpreted in any other way...

gingerBill:
`len` is for array-like things. A union is not an array. I understand you want to get the variant count, but saying the length of a union? What does that even mean. 

I know you are thinking "but allow it for an enum, why not a union?", and that is a great question. For an `enum`, `min` and `max` make perfect sense, and `len` is kind of describable through `max-min+1`, but a `union` has none of those concepts.

jakubtomsu:
Ok, that makes perfect sense. I'll think about this a bit and try to come up with something less confusing.

core/intrinsics/intrinsics.odin
flysand7: Thanks for clarification, I think I misunderstood how unions worked

jakubtomsu:
Alright, I removed `min`, `max`, `len` and `cap` for unions, and replaced them with the 2 following procedures:
- type_union_base_tag_value
- type_union_variant_count

Those cover all my use-cases in terms of functionality, plus it's not as unintuitive :)

See the updated overview for more info.

karl-zylinski:
These things seem very useful. I was wishing there was a `type_union_variant_count` today, which this PR adds!

jakubtomsu:
Thanks @karl-zylinski!

jakubtomsu:
It should be ready to merge, any update on this @gingerBill?

core/intrinsics/intrinsics.odin
```diff
@@ -162,7 +162,14 @@ type_is_matrix           :: proc($T: typeid) -> bool ---
 type_has_nil :: proc($T: typeid) -> bool ---
 
 type_is_specialization_of :: proc($T, $S: typeid) -> bool ---
+
 type_is_variant_of :: proc($U, $V: typeid) -> bool where type_is_union(U) ---
+type_union_tag_type :: proc($T: typeid) -> typeid where type_is_union(T) ---
+type_union_tag_offset :: proc($T: typeid) -> int where type_is_union(T) ---
```
gingerBill, line 168: Maybe change this to `-> uintptr` but otherwise fine.

jakubtomsu: Good idea, done!

PR #2896, opened 2023-10-26
Add system_random and random_bytes for js target

thetarnav:
~~This implements generating random numbers using [`Math.random`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random) and byte slices using [`crypto.getRandomValues`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues).
I'm not sure if `Math.random` is the best choice for generating `u64` because of missing precision, but it's much faster than using crypto for every number.~~

This implements generating random numbers in `rand` and byte slices in `crypto` using [`crypto.getRandomValues`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues).

### DIFF
--- a/core/crypto/rand_generic.odin
+++ b/core/crypto/rand_generic.odin
@@ -1,7 +1,7 @@
 package crypto
 
-when ODIN_OS != .Linux && ODIN_OS != .OpenBSD && ODIN_OS != .Windows {
-	_rand_bytes :: proc (dst: []byte) {
+when ODIN_OS != .Linux && ODIN_OS != .OpenBSD && ODIN_OS != .Windows && ODIN_OS != .JS {
+	_rand_bytes :: proc(dst: []byte) {
 		unimplemented("crypto: rand_bytes not supported on this OS")
 	}
 }
--- a/core/crypto/rand_js.odin
+++ b/core/crypto/rand_js.odin
@@ -1,0 +1,20 @@
+package crypto
+
+foreign import "odin_env"
+foreign odin_env {
+	@(link_name = "rand_bytes")
+	env_rand_bytes :: proc "contextless" (buf: []byte) ---
+}
+
+_MAX_PER_CALL_BYTES :: 65536
+
+_rand_bytes :: proc(dst: []byte) {
+	dst := dst
+
+	for len(dst) > 0 {
+		to_read := min(len(dst), _MAX_PER_CALL_BYTES)
+		env_rand_bytes(dst[:to_read])
+
+		dst = dst[to_read:]
+	}
+}
--- a/core/crypto/rand_js.odin
+++ b/core/crypto/rand_js.odin
@@ -6,7 +6,7 @@ foreign odin_env {
 	env_rand_bytes :: proc "contextless" (buf: []byte) ---
 }
 
-_MAX_PER_CALL_BYTES :: 65536
+_MAX_PER_CALL_BYTES :: 65536 // 64kiB
 
 _rand_bytes :: proc(dst: []byte) {
 	dst := dst
--- a/core/math/rand/system_js.odin
+++ b/core/math/rand/system_js.odin
@@ -2,10 +2,11 @@ package rand
 
 foreign import "odin_env"
 foreign odin_env {
-	rand :: proc "contextless" () -> f64 ---
+	@(link_name = "rand")
+	rand_f64 :: proc "contextless" () -> f64 ---
 }
 
 @(require_results)
 _system_random :: proc() -> u64 {
-	return u64(rand() * 0x1fffffffffffff)
+	return u64(rand_f64() * 0x1fffffffffffff)
 }
--- a/core/math/rand/system_js.odin
+++ b/core/math/rand/system_js.odin
@@ -2,11 +2,13 @@ package rand
 
 foreign import "odin_env"
 foreign odin_env {
-	@(link_name = "rand")
-	rand_f64 :: proc "contextless" () -> f64 ---
+	@(link_name = "rand_bytes")
+	env_rand_bytes :: proc "contextless" (buf: []byte) ---
 }
 
 @(require_results)
 _system_random :: proc() -> u64 {
-	return u64(rand_f64() * 0x1fffffffffffff)
+	buf: [8]u8
+	env_rand_bytes(buf[:])
+	return transmute(u64)buf
 }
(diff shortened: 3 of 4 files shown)


Kelimion:
Is there a technical reason why we wouldn't use `crpyto.getRandomValues` only to seed the current PCG-random implementation via `rand.init_as_system` instead of outright replacing it on Wasm? Then if you explicitly seed it with `rand.set_global_seed` or `rand.create`, you'd get the same output as you would on the other supported platforms.

Kelimion:
Something like this for `system_js.odin`, and forget about `rand_js.odin`.
```odin
package crypto

foreign import "odin_env"
foreign odin_env {
	@(link_name = "rand_bytes")
	env_rand_bytes :: proc "contextless" (buf: []byte) ---
}

@(require_results)
_system_random :: proc() -> (value: u64) {
	buf: [8]u8
	env_rand_bytes(buf[:]) // Can supply a maximum of 64kiB at a time.
	return transmute(u64)buf
}
```

thetarnav:
> Is there a technical reason why we wouldn't use `crpyto.getRandomValues` only to seed the current PCG-random implementation via `rand.init_as_system` instead of outright replacing it on Wasm? Then if you explicitly seed it with `rand.set_global_seed` or `rand.create`, you'd get the same output as you would on the other supported platforms.

Yeah I was using `rand.init_as_system` so every time a new number was to be generated, `_system_random` was called, so I opted for `Math.random` as it was faster then using `crypto` every time.
But youre right, better to generate the seed once, and so call `_system_random` only once. I'll change the `_system_random` to use `rand_bytes` then.

> Something like this for `system_js.odin`, and forget about `rand_js.odin`.

Not sure what do you mean by "and forget about `rand_js.odin`". Should I remove the js implementation from the crypto package, is there something wrong with it?

Kelimion:
> Not sure what do you mean by "and forget about `rand_js.odin`". Should I remove the js implementation from the crypto package, is there something wrong with it?

I mean that `rand_js.odin` would be superfluous to requirements? The binding to `crypto.GetRandomValues` will now move to `system_js.odin` to provide the system seed. There's no `rand_windows.odin` or `rand_darwin.odin` that implements a `_rand_bytes`. We already have `rand.read` or `rand.perm` to generate a slice of random bytes, and if you use the same seed in `rand.create`, it'll generate the same output across the supported platforms.

Kelimion:
If someone really wants to, they can still call into `crypto.GetRandomValues` to fill a slice using `env_rand_bytes` just like `_system_random` does. So for my part I'd add the comment about the 64kiB limit to `system_js.odin` and delete `rand_js.odin` altogether as duplicate in nature.

thetarnav:
So youre saying that there is no point in duplicating code for generating a slice of random bytes?
I guess that makes sense. But then why is `_rand_bytes` implemented at all for some targets, instead of just reusing the `math/rand` package? Also there is a [`_rand_bytes` implementation for windows](https://github.com/odin-lang/Odin/blob/master/core/crypto/rand_windows.odin).
One thing I'm noticing is that `rand.read` will make a sys call for each byte, while `_rand_bytes` in `crypto` does that in bulk.

thetarnav:
Also, by duplication you mean that there are both `_rand_bytes` and `_system_random` implemented, or that `env_rand_bytes` is being imported in two packages?

Kelimion:
> Also, by duplication you mean that there are both _rand_bytes and _system_random implemented, or that env_rand_bytes is being imported in two packages?

Both. Two bindings of the same endpoint in the JS environment. Once used to set the system seed, and once used to implement a `_rand_bytes` that doesn't exist for any other platform under `core:math/rand`. You're correct that there's a `_rand_bytes` for Windows, but that's in `core:crypto`, which has a focus on cryptographically strong randomness. Two packages, two different use cases.

If you want to add `crypto._rand_bytes` to `core:crypto` I won't stand opposed to that, but that didn't seem to be the focus of this PR, and keeping `_rand_bytes` in `core:math/rand` makes no sense when it's not used by that package and has no other implementations.

Like I said, if you want to use `crypto.GetRandomValues` directly to generate randomness in bulk and bypass the PCG random implementation of the package, you can still just call `rand.env_rand_bytes`. It's not private to that file.

Kelimion:
Also, `rand.read` doesn't perform a syscall for each byte at all. It calls `_random_u64`, just like the rest of the package.

thetarnav:
> Also, `rand.read` doesn't perform a syscall for each byte at all. It calls `_random_u64`, just like the rest of the package.

ah youre right 🙈

> Two bindings of the same endpoint in the JS environment. Once used to set the system seed, and once used to implement a `_rand_bytes` that doesn't exist for any other platform under `core:math/rand`. You're correct that there's a `_rand_bytes` for Windows, but that's in `core:crypto`, which has a focus on cryptographically strong randomness. Two packages, two different use cases.
> 
> If you want to add `crypto._rand_bytes` to `core:crypto` I won't stand opposed to that, but that didn't seem to be the focus of this PR, and keeping `_rand_bytes` in `core:math/rand` makes no sense when it's not used by that package and has no other implementations.

Thats the last thing I'm confused about. I'm not adding `_rand_bytes` to `core:math/rand`, but to `core:crypto` package.
Maybe I should make a separate PR for that, and here only focus on `math/rand`? I didn't do that initially because both would need adding `rand_bytes` to `runtime.js`.

Kelimion:
> > Also, `rand.read` doesn't perform a syscall for each byte at all. It calls `_random_u64`, just like the rest of the package.
> 
> ah youre right 🙈
> 
> > Two bindings of the same endpoint in the JS environment. Once used to set the system seed, and once used to implement a `_rand_bytes` that doesn't exist for any other platform under `core:math/rand`. You're correct that there's a `_rand_bytes` for Windows, but that's in `core:crypto`, which has a focus on cryptographically strong randomness. Two packages, two different use cases.
> > If you want to add `crypto._rand_bytes` to `core:crypto` I won't stand opposed to that, but that didn't seem to be the focus of this PR, and keeping `_rand_bytes` in `core:math/rand` makes no sense when it's not used by that package and has no other implementations.
> 
> Thats the last thing I'm confused about. I'm not adding `_rand_bytes` to `core:math/rand`, but to `core:crypto` package. Maybe I should make a separate PR for that, and here only focus on `math/rand`? I didn't do that initially because both would need adding `rand_bytes` to `runtime.js`.

You're right. Github's rather bad at presenting where pieces of a diff live sometimes. It looked like it was part of the math package. My bad. Merged.

PR #2898, opened 2023-10-27
Add ELF64 bindings

flysand7:
Added ELF64 bindings to mirror debug/pe package somewhat. I'll be using this to bind vdso in sys/linux package. Might also be useful for occasional users. For now limited machine-specific extensions to platforms currently supported by Odin.

### DIFF
--- a/examples/all/all_vendor.odin
+++ b/examples/all/all_vendor.odin
@@ -39,10 +39,10 @@ import TTF        "vendor:sdl2/ttf"
 
 import vk         "vendor:vulkan"
 
-import NS         "vendor:darwin/Foundation"
-import MTL        "vendor:darwin/Metal"
-import MTK        "vendor:darwin/MetalKit"
-import CA         "vendor:darwin/QuartzCore"
+// import NS         "vendor:darwin/Foundation"
+// import MTL        "vendor:darwin/Metal"
+// import MTK        "vendor:darwin/MetalKit"
+// import CA         "vendor:darwin/QuartzCore"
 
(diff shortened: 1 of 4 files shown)


examples/all/all_vendor.odin
```diff
@@ -39,10 +39,10 @@ import TTF        "vendor:sdl2/ttf"
...
 import vk         "vendor:vulkan"
 
-import NS         "vendor:darwin/Foundation"
-import MTL        "vendor:darwin/Metal"
-import MTK        "vendor:darwin/MetalKit"
-import CA         "vendor:darwin/QuartzCore"
+// import NS         "vendor:darwin/Foundation"
+// import MTL        "vendor:darwin/Metal"
+// import MTK        "vendor:darwin/MetalKit"
+// import CA         "vendor:darwin/QuartzCore"
```
jakubtomsu, line 45: Did you comment this out intentionally, or was it just for testing?

flysand7: It was very intentional, but not necessary in the long run. The reason is that I was testing for compile errors locally by running:

```
$ odin run examples/all -vet -strict-style; rm all
```

And the `examples/all` package was including a bunch of darwin stuff on my linux machine which resulted in a crap ton of errors. The other PR I have pending also has these lines commented-out.

gingerBill: `examples/all` is technically not meant to use `odin build`, only `odin check`.

flysand7: Thanks for the tip!

No more accidental pushing binaries haha

PR #2900, opened 2023-10-28
Fix `-foreign-error-procedures` flag doing nothing

flysand7:
This flag was introduced in commit [17f32cc](https://github.com/odin-lang/Odin/commit/0fa487f468b1f63d5ec97ae8bbb0da01717f32cc) with the purpose that freestanding targets would define exported symbols that bind to these procedures to allow for custom error handling.

However it seems like in commit [4f12b8](https://github.com/odin-lang/Odin/commit/01162e08b55406578a2972d74c472de8a14f12b8), the code that makes this flag work got removed and my odin-os code wasn't working since then for like a year without me noticing. It wasn't failing to compile, but the error procedures would be crashing the kernel instead of printing the error messages.

### DIFF
--- a/core/runtime/error_checks.odin
+++ b/core/runtime/error_checks.odin
@@ -16,113 +16,130 @@ type_assertion_trap :: proc "contextless" () -> ! {
 	}
 }
 
+when ODIN_FOREIGN_ERROR_PROCEDURES {
+    foreign {
+        bounds_check_error               :: proc "contextless" (file: string, line, column: i32, index, count: int) ---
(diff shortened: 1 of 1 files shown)


flysand7:
After some discussion this is probably not the right way to allow for overriding the error reporting. It ugly and requires "linker hacks", in that you can declare those foreign procedures in the same module. Furthermore, foreign procedures don't actually solve anything because `typeid` is unique to a module, and having type errors be resolved in a different module doesn't make sense.

I'll be working on another way to override the default error reporting mechanism that can be used for freestanding targets, one that makes a little bit more sense. `-foreign-error-procedures` flag should be deprecated and eventually removed.

PR #2902, opened 2023-10-29
Only build vendor:darwin on darwin

flysand7:
Running

```
odin run examples/all
```

Was printing a bunch of errors (something like `msgSend` is only available on `darwin`) on my linux machine. I honestly have no idea how this wasn't caught by the CI...

### DIFF
--- a/examples/all/all_vendor.odin
+++ b/examples/all/all_vendor.odin
@@ -39,11 +39,6 @@ import TTF        "vendor:sdl2/ttf"
 
 import vk         "vendor:vulkan"
 
-import NS         "vendor:darwin/Foundation"
-import MTL        "vendor:darwin/Metal"
-import MTK        "vendor:darwin/MetalKit"
-import CA         "vendor:darwin/QuartzCore"
-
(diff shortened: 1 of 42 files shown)


Kelimion:
I propose you also move these 4 from `vendor_all.odin` to its own `vendor_darwin.odin` under `examples/all`, patterned after `vendor_directx.odin`
```odin
//+build darwin
package all

import NS         "vendor:darwin/Foundation"
import MTL        "vendor:darwin/Metal"
import MTK        "vendor:darwin/MetalKit"
import CA         "vendor:darwin/QuartzCore"

_ :: NS
_ :: MTL
_ :: MTK
_ :: CA
```

flysand7:
Should darwin vendor stuff keep it's build tags, or should I remove them?

Kelimion:
> Should darwin vendor stuff keep it's build tags, or should I remove them?

Keep it. We also do that for `core:sys/windows` after all.

Kelimion:
(That's why I said `also` move and not `ditch this and replace with`. ;-))

PR #2903, opened 2023-10-29
Fix-up inline asm for i386 syscalls emit

flysand7:
Compiling code with `-target:i386` on linux results in an llvm error:

```
error: <inline asm>:1:6: invalid register name
        int %eaxx80
            ^~~~~~~
```

The error was coming from the fact that llvm interprets `$n`, where `n` is a number as a register parameter to the assembly or something like that, but anyway that character needs to be escaped in order to be used to declare constants


Kelimion:
Let's merge this after the current map segfault is fixed, just so this passes cleanly.

flysand7:
yep

PR #2907, opened 2023-10-30
trim trailing zeroes from string in io.write_f64 to prevent json dataloss

karl-zylinski:
When using `io.write_f64` the user has no control over the precision, so it makes sense to trim off trailing zeroes from the resulting string.

Without my changes writing `0.123` will result in `0.1230000000000000`. With my changes writing will result in `0.123`.
If you write `0.0` then it will just write `0`

Without these changes, if you write a json file (using `core:encoding/json`) and then read it back, you may run into trouble due to `io.write_f64` writing lots of trailing zeroes and `strconv.parse_f64_prefix` loosing precision due to taking all those trailing zeroes into consideration. I.e. writing and reading json back and forth will make values drift over time due to the trailing zeroes written and read, essentially resulting in data loss.

Small test program with comments that show how the result is differs with and without my changes:
```
package f64_trim

import "core:strings"
import "core:fmt"
import "core:io"

main :: proc() {
    b, _ := strings.builder_make(context.temp_allocator)
    some_float: f64 = 0.123

    // We have no control of the precision of write_f64. Therefore I have made
    // so that it trims off trailing zeroes from the result.
    io.write_f64(strings.to_writer(&b), f64(some_float))

    // Without my changes, it prints: 0.1230000000000000
    // With my changes, it prints: 0.123
    fmt.println(strings.to_string(b))
}
```

Note: Initially this PR had changes to `parse_f64_prefix`, removing trailing zeroes from the input string, which lessened the precision loss from those trailing zeroes. But that wasn't really necessary for my uses and also a bit scary, so I removed that part.

### DIFF
--- a/core/io/util.odin
+++ b/core/io/util.odin
@@ -115,7 +115,7 @@ trim_decimal_string :: proc "contextless" (s: string) -> string {
 
 	// We have at least one trailing zero. Search backwards and find the rest.
 	trailing_start_idx := len(s)-1
-	trailing_loop: for i := len(s) - 2; i >= 0 ; i -= 1 {
+	for i := len(s) - 2; i >= 0 ; i -= 1 {
 		switch s[i] {
 			case '0':
 				if trailing_start_idx == i + 1 {
--- a/core/io/util.odin
+++ b/core/io/util.odin
@@ -94,7 +94,8 @@ write_f64 :: proc(w: Writer, val: f64, n_written: ^int = nil) -> (n: int, err: E
 		s = s[1:]
 	}
 
-	return write_string(w, string(s), n_written)
+	trimmed_string := strconv.trim_decimal_string(string(s))
+	return write_string(w, trimmed_string, n_written)
 }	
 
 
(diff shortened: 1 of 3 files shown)


flysand7:
well from what i see odin's float parsing is not precise to begin with

karl-zylinski:
@flysand7 this is true, perhaps I should stick to only doing the trimming of the written stuff, since that fixes the json issues without doing scary things to the parser 🤔

karl-zylinski:
I removed the trailing zeroes trimming from parse_f64_prefix, since it wasn't strictly necessary.

karl-zylinski:
@gingerBill @Kelimion I find working with JSON is more or less broken without something like this fix, you just need to write, read and write json again for many floats to get big differences compared to originally written value.

karl-zylinski:
Closing this because https://github.com/odin-lang/Odin/pull/3847 solves this in a more elegant way, and it also works for `fmt.printfln` etc.

PR #2908, opened 2023-10-31
os: tighten up error handling on stream creation

flysand7:


flysand7:
It just hit me how bad it is that different OS'es return different errors of different types. I don't think there's much that can be done except split up `stream.odin` into multiple OS-specific files

flysand7:
The error handling should probably be done on a separate layer, closing this for now

PR #2912, opened 2023-10-31
JSON: Option to sort marshaled maps before outputting + clone_value proc + small improvements

karl-zylinski:
Added `sort_maps_by_key` option when marshaling into JSON. If true, then whenever the marshaler is processing a map, it will temp allocate a list of entries, sort it and then output it. Without doing so, any data that contains maps constantly rearranges, making that data unsuitable for putting into source control.

Also improved the docs on `Marshal_Options` a bit.

Also added a `json.clone_value` proc

Also fixed a bug when marshaling a slice in MJSON mode, in that case it would skip the starting `[` (it confused it with `{`), producing illegal JSON

### DIFF
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -83,6 +83,9 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err: 
 		return
 	}
 
+	// temp guard in case we are sorting map keys, which will use temp allocations
+	runtime.DEFAULT_TEMP_ALLOCATOR_TEMP_GUARD()
+	
 	ti := runtime.type_info_base(type_info_of(v.id))
 	a := any{v.data, ti.id}
 
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -62,6 +62,9 @@ marshal :: proc(v: any, opt: Marshal_Options = {}, allocator := context.allocato
 	defer if err != nil {
 		strings.builder_destroy(&b)
 	}
+	
+	// temp guard in case we are sorting map keys, which will use temp allocations
+	runtime.DEFAULT_TEMP_ALLOCATOR_TEMP_GUARD(ignore = allocator == context.temp_allocator)
 
 	opt := opt
 	marshal_to_builder(&b, v, &opt) or_return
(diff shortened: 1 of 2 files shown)


core/encoding/json/marshal.odin
```diff
@@ -263,36 +275,81 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
...
-					case: return .Unsupported_Type
+						case: return .Unsupported_Type
+						}
 					}
+
+					append(&sorted, Entry { key = name, value = any{value, info.value.id}})
 				}
 
-				marshal_to_writer(w, any{value, info.value.id}, opt) or_return
+                slice.sort_by(sorted[:], proc(i, j: Entry) -> bool { return i.key < j.key })
```
flysand7: Seems like you got spaces indentation on this line (and a few other lines), please change it to tabs

core/encoding/json/marshal.odin
```diff
@@ -263,36 +275,81 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
...
+				Entry :: struct {
+					key: string,
+					value: any,
+				}
 
-				key   := rawptr(runtime.map_cell_index_dynamic(ks, info.map_info.ks, bucket_index))
-				value := rawptr(runtime.map_cell_index_dynamic(vs, info.map_info.vs, bucket_index))
+				// If we are sorting the map by key, then we temp alloc an array
+				// and sort it, then output the result.
+				sorted := make([dynamic]Entry, 0, map_cap, context.temp_allocator)
```
flysand7: With this change a function that was previously allocating no memory would start allocating memory, causing some existing programs to leak memory. Maybe temp allocator guard should be introduced to free the temporary slice.

flysand7 (changes requested):

core/encoding/json/marshal.odin
karl-zylinski: Thanks. I introduced a temp alloc guard, its easy to get used to one's own setup with freeing the temp allocator per-frame.

gingerBill:
@karl-zylinski The temp guard is there for people who don't even use `context.temp_allocator` at all, and there to prevent random leaks.

karl-zylinski:
Right, I think the guard is missing the `ignore = allocator == context.temp_allocator` params, as it trashed my json marshal result when I marshalled using temp allocator. I'll update the PR next week when I have time

gingerBill:
That makes a lot of sense! We need that ignore trick.

PR #2914, opened 2023-11-02
Bump Darwin+Linux to LLVM-17 and test '-o' flags

jcmdln:
This PR builds Odin on LLVM-17 for all platforms and adds missing `-o` flag tests against `examples/demo`.

For Linux, I've dropped the Ubuntu version to 20.04 and used the official LLVM repo for installing LLVM-17. This gets Linux releases one step closer to being compatible with CompilerExplorer which requires all languages run on Ubuntu 20.04 though the real purpose of this PR is catching Linux-specific bugs with LLVM-17.

In my own fork, the Windows and Linux pipelines both succeeded while the MacOS pipeline failed during the `test_trunc_f16` tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - name: Download LLVM, botan and setup PATH
+        # NOTE(jcmdln): Brew will update everything on install without this
+        env: { HOMEBREW_NO_AUTO_UPDATE: 1}
         run: |
           brew update
           brew install llvm@17 botan
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,7 @@ jobs:
       - uses: actions/checkout@v1
       - name: Download LLVM, botan and setup PATH
         run: |
+          brew update
           brew install llvm@17 botan
           echo "/usr/local/opt/llvm@17/bin" >> $GITHUB_PATH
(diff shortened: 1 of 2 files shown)


jcmdln:
Interestingly, in this PR pipeline an older image was used for MacOS so it failed for a different reason:
```
# This PR's image
  Image: macos-12
  Version: 20230921.1

# My fork's image
  Image: macos-12
  Version: 20231017.6
```

jcmdln:
Race condition in Windows? https://github.com/odin-lang/Odin/actions/runs/6727360875/job/18285043404#step:13:468
```
Testing two servers trying to bind to the same endpoint...
Testing client connecting to a closed port...
Testing client connecting to port that doesn't accept...
[D:/a/Odin/Odin/tests/core/net/test_core_net.odin(490:2)] Expected server and client to return `nil` and `Refused`, got nil and Connection_Closed
Testing client sending server data...
```

flysand7:
Haven't looked into the code that much, but it looks like it does `thread.sleep` for 4 seconds to allow the server to get to some state before reading back from the shared state. Technically it's a race condition, because no actual synchronization is being done and it's not guaranteed that server will get to that point in 4 seconds, on a normal day I'd imagine it would work, but in the context of github CI things might be different with regards to that kind of "performance dependency"

jcmdln:
Closing this for now. I'll work on the MacOS pipeline in my fork and resubmit after I tame `brew`.

PR #2923, opened 2023-11-05
Pre-compiled raygui on linux

flysand7:
Also fixed `GuiButton` returning `i32` instead of a boolean


laytan:
Correct me if I'm wrong but a `bool` isn't the same size as an `i32` right? I think you'd use `b32` (32 bit boolean) here.

flysand7:
From the ABI standpoint that's fine because the value is passed in a register and it at most 32 bits, but if that makes you worry I can change it to `b32`

PR #2930, opened 2023-11-06
Add the form data variant of percent_encode/decode

divanburger:
There is a variant of percent encoding used in the body of HTTP requests that differs from the normal percent encoding. Instead of converting spaces to %20 they are instead converted to a +. 
https://en.wikipedia.org/wiki/Percent-encoding#The_application/x-www-form-urlencoded_type

This issue pops up when trying to decode values from a web form as +'s are not decoded to a space.

The form_data_decode can actually decode both variants as a + is always encoded to it's percent encoded version %2B so a + must always be a space. Regardless, I made a separate decode as the form_data variant is more complicated to decode.

A `index_byte_any` method was added to make decoding easier.

### DIFF
--- a/core/net/url.odin
+++ b/core/net/url.odin
@@ -160,6 +160,78 @@ percent_decode :: proc(encoded_string: string, allocator := context.allocator) -
 	return
 }
 
+// For data with content type application/x-www-form-urlencoded
+form_data_encode :: proc(s: string, allocator := context.allocator) -> string {
+	b := strings.builder_make(allocator)
+	strings.builder_grow(&b, len(s) + 16) // NOTE(tetra): A reasonable number to allow for the number of things we need to escape.
+
+	for ch in s {
+		switch ch {
+		case 'A'..='Z', 'a'..='z', '0'..='9', '-', '_', '.', '~':
+			strings.write_rune(&b, ch)
+		case ' ':
+			strings.write_rune(&b, '+')
+		case:
+			bytes, n := utf8.encode_rune(ch)
+			for byte in bytes[:n] {
+				buf: [2]u8 = ---
+				t := strconv.append_int(buf[:], i64(byte), 16)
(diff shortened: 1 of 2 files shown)


Kelimion:
I'm not sure if we can properly call this `// For data with content type application/x-www-form-urlencoded` if it doesn't take in k,v pairs that are then separated by either `&` or `;` (`;` is a holdover from HTML 2).

The [WhatWG](https://url.spec.whatwg.org/#application/x-www-form-urlencoded) spec also talks about tuples.

Further, as per the spec, a literal `%` is allowed: `Otherwise, if byte is 0x25 (%) and the next two bytes after byte in input are not in the ranges 0x30 (0) to 0x39 (9), 0x41 (A) to 0x46 (F), and 0x61 (a) to 0x66 (f), all inclusive, append byte to output.`

Your decoder fails to decode `c=(m^e)%n` to `c=(m^e)%n` for example because it returns early if no two hex digits follow.

As for `+` instead of `%20`, the WhatWG spec says this about `urlencode` (as called the k,v serializer):
```
For each byte of encodeOutput converted to a byte sequence:

If spaceAsPlus is true and byte is 0x20 (SP), then append U+002B (+) to output and [continue](https://infra.spec.whatwg.org/#iteration-continue).
```

The more appropriate action then may be to update `percent_encode` and `percent_decode` (and rename them `urlencode_*`) to take a param `, space_as_plus := false`.

Kelimion:
The thing to do, I think, is to have `urlencode` and `urldecode` with the optional `space_as_plus` param.
Then have a `form_parse` and `form_serialize` or something which take in the `map[string]string`, piece the key and value out to `urlencode` separately and then join it with a `=`, as per the [spec](https://url.spec.whatwg.org/#urlencoded-parsing).

`join_url` and `split_url` can then also use this mechanism for query string.

PR #2931, opened 2023-11-07
Use `#by_ptr` in D3D12 bindings where appropriate

jakubtomsu:
Many D3D12 procedures take a const pointer, but our bindings use regular pointers. Odin doesn't have const pointers since the Odin calling convention can pass large values by reference when appropriate.

I converted all `_In_  const T*` procedure parameters to `#by_ptr` values. This is to better represent the original intention of the API. 

Note: all of optionally nullable const pointers are still pointers.


flysand7:
I'll probably never use D3D11, but I'm not a big fan of this. I would probably be against any use of `#by_ptr` that's not binding to `const &` (const reference) parameters, which is the closest approximation of `#by_ptr` usage.

If the library takes a pointer, the equivalent code in C or C++ would take the address of the value explicitly, I think it's better if this is preserved

jakubtomsu:
`#by_ptr` is specifically for `const T*` in C bindings, as stated in docs. In this case the parameters aren't nullable. And the C version takes a pointer only because the limitations of the C calling convention, to avoid copying large values.

I think `#by_ptr` matches the intention of the API much better, since a pointer would mean the value can be modified by the procedure. Those could also allow certain optimizations, although those aren't very important in this case.

In general I think the pointers are misleading, since they have a slightly different meaning in Odin semantics.

flysand7:
Okay, was just giving a slight pushback. I'm not actually *that* much against the idea

jakubtomsu:
Yeah no worries, thanks for the feedback:D I was also just trying to share my perspective, I hope it didn't come off too argumentative

gingerBill:
My pushback against this idea is mostly that in C (and C++), it is a pointer. There are sometimes cases when `#by_ptr` makes sense, but I don't feel as if this is one of them.

I am not 100% decided on this either way. But I am not sure changing the signatures is necessarily a good idea when people are trying to port code.

jakubtomsu:
That's fair. To be clear I'm also not 100% sure if this is a good idea in general. It's just something I personally would prefer.

PR #2933, opened 2023-11-07
sys/linux: munmap was not using the correct syscall

flga:
tiny typo in munmap :p


flysand7:
lmao, want to avoids memory leaks? how about allocate more memory!

PR #2934, opened 2023-11-09
core/runtime: Use wyhash as the map hash algorithm

Yawning:
FNV64 has the benefit of being easy to implement, but it is possible to do much better in terms of hash quality with not that much more code.  This PR switches the map hash function to wyhash 4.2.

Pros:
- Faster for data sizes > 4 bytes
- Better statistical properties
- At some point, someone could add randomizing the dynamic salt to the runtime initialization

Cons:
- Slower for data sizes <= 3 bytes
- Code is more complicated and bigger
- Maps with C (NUL terminated) strings are probably slower.
- WASM is really sad, because the ISA designers are cretins.

If people disagree with this at the runtime map level, I'll move the wyhash implementation to live entirely in `core/hash` instead.

See:
- https://github.com/wangyi-fudan/wyhash
- https://github.com/rurban/smhasher/blob/master/doc/wyhash.txt

### DIFF
--- a/core/runtime/dynamic_map_hash.odin
+++ b/core/runtime/dynamic_map_hash.odin
@@ -15,13 +15,6 @@ import "core:intrinsics"
 // While wyhash32 exists, it is not recommended for use by the designer,
 // and it would add complexity of having to worry about bad seeds.
 // See: https://github.com/wangyi-fudan/wyhash/issues/92
-//
-// TODO/perf:
-// - Someone should figure out if any supported OS runs AArch64 with
-//   strict alignment enabled, and add `_wy_read_8`/`_wy_read_4`
-//   fast-paths as appropriate.
-// - WASM can also do direct reads, but performance is doomed to suck
-//   on that architecture because the multiply is slow.
 
 // This is free and unencumbered software released into the public domain
 // under The Unlicense (http://unlicense.org/)
@@ -69,22 +62,13 @@ _wy_mix :: #force_inline proc "contextless" (A, B: u64) ->u64 {
 @(private)
 _wy_read_8 :: #force_inline proc "contextless" (p: [^]byte, off: int = 0) -> u64 {
 	pp := p[off:]
-	when ODIN_ARCH == .i386 || ODIN_ARCH == .amd64 {
-		return ([^]u64)(pp)[0]
-	} else {
-		return u64(pp[0]) | u64(pp[1])<<8 | u64(pp[2])<<16 | u64(pp[3])<<24 |
-		    u64(pp[4])<<32 | u64(pp[5])<<40 | u64(pp[6])<<48 | u64(pp[7])<<56
-	}
+	return u64(intrinsics.unaligned_load((^u64le)(pp)))
 }
 
 @(private)
 _wy_read_4 :: #force_inline proc "contextless" (p: [^]byte, off: int = 0) -> u64 {
 	pp := p[off:]
-	when ODIN_ARCH == .i386 || ODIN_ARCH == .amd64 {
-		return u64(([^]u32)(pp)[0])
-	} else {
-		return u64(pp[0]) | u64(pp[1])<<8 | u64(pp[2])<<16 | u64(pp[3])<<24
-	}
+	return u64(intrinsics.unaligned_load((^u32le)(pp)))
 }
 
 @(private)
--- a/core/runtime/dynamic_map_hash.odin
+++ b/core/runtime/dynamic_map_hash.odin
@@ -1,0 +1,147 @@
+package runtime
+
+import "core:intrinsics"
+import "core:math/bits"
+
+// wyhash hash function (4.2, WYHASH_CONDOM=2)
+//
+// This was chosen as it:
+//  - Produces high quality output
+//  - Can include a non-seed dynamic-secret
+//  - Has high performance on reasonable architectures
+//  - Is simple in terms of implementation
+//
+// https://github.com/rurban/smhasher/blob/master/doc/wyhash.txt
+//
+// While wyhash32 exists, it is not recommended for use by the designer,
+// and it would add complexity of having to worry about bad seeds.
+// See: https://github.com/wangyi-fudan/wyhash/issues/92
+//
+// TODO/perf:
+// - Someone should figure out if any supported OS runs AArch64 with
+//   strict alignment enabled, and add `_wy_read_8`/`_wy_read_4`
+//   fast-paths as appropriate.
+// - WASM can also do direct reads, but performance is doomed to suck
+//   on that architecture because the multiply is slow.
+
+// This is free and unencumbered software released into the public domain
+// under The Unlicense (http://unlicense.org/)
+//
+// main repo: https://github.com/wangyi-fudan/wyhash
+// author: 王一 Wang Yi <godspeed_china@yeah.net>
+// contributors: Reini Urban, Dietrich Epp, Joshua Haberman, Tommy Ettinger,
+// Daniel Lemire, Otmar Ertl, cocowalla, leo-yuriev, Diego Barrios Romero,
+// paulie-g, dumblob, Yann Collet, ivte-ms, hyb, James Z.M. Gao,
+// easyaspi314 (Devin), TheOneric
+
+Wyhash_Secret :: distinct [4]u64
+
+_WYHASH_DEFAULT_SECRET := Wyhash_Secret{
+	0x2d358dccaa6c78a5,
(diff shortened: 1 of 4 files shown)


JesseRMeyer:
Please offer before and after performance measures.

core/runtime/dynamic_map_hash.odin
```diff
@@ -0,0 +1,147 @@
+package runtime
+
+import "core:intrinsics"
+import "core:math/bits"
```
gingerBill, line 4: Please do not important anything into `core:runtime` except for `core:intrinsics`. Duplicate the code if necessary.

core/runtime/dynamic_map_hash.odin
```diff
@@ -0,0 +1,147 @@
...
+_WYHASH_DEFAULT_SECRET := Wyhash_Secret{
+	0x2d358dccaa6c78a5,
+	0x8bb84b93962eacc9,
+	0x4b33a62ed433d4a3,
+	0x4d5a2da51de1aa47,
+}
+
+@(private)
+_wy_wide_mul :: #force_inline proc "contextless" (A, B: u64) -> (u64, u64) {
+	hi, lo := bits.mul_u64(A, B)
```
gingerBill, line 48: Reimplement `mul_u64` in `core:runtime`. Do not depend on any other package.

core/runtime/dynamic_map_hash.odin
gingerBill, line 4: Please do not important anything into `core:runtime` except for `core:intrinsics`. Duplicate the code if necessary.

core/runtime/dynamic_map_hash.odin
```diff
@@ -0,0 +1,147 @@
...
+
+@(private)
+_wy_mix :: #force_inline proc "contextless" (A, B: u64) ->u64 {
+	lo, hi := _wy_wide_mul(A, B)
+	return lo ~ hi
+}
+
+@(private)
+_wy_read_8 :: #force_inline proc "contextless" (p: [^]byte, off: int = 0) -> u64 {
+	pp := intrinsics.ptr_offset(p, off)
```
gingerBill, line 71:
```odin
pp := p[off:]
```
use the multi-pointer benefits

core/runtime/dynamic_map_hash.odin
```diff
@@ -0,0 +1,147 @@
...
+		return (transmute([^]u64)(pp))[0]
+	} else {
+		return u64(pp[0]) | u64(pp[1])<<8 | u64(pp[2])<<16 | u64(pp[3])<<24 |
+		    u64(pp[4])<<32 | u64(pp[5])<<40 | u64(pp[6])<<48 | u64(pp[7])<<56
+	}
+}
+
+@(private)
+_wy_read_4 :: #force_inline proc "contextless" (p: [^]byte, off: int = 0) -> u64 {
+	pp := intrinsics.ptr_offset(p, off)
```
gingerBill, line 82:
```odin
pp := p[off:]
```

core/runtime/dynamic_map_hash.odin
```diff
@@ -0,0 +1,147 @@
...
+_wy_mix :: #force_inline proc "contextless" (A, B: u64) ->u64 {
+	lo, hi := _wy_wide_mul(A, B)
+	return lo ~ hi
+}
+
+@(private)
+_wy_read_8 :: #force_inline proc "contextless" (p: [^]byte, off: int = 0) -> u64 {
+	pp := intrinsics.ptr_offset(p, off)
+	when ODIN_ARCH == .i386 || ODIN_ARCH == .amd64 {
+		return (transmute([^]u64)(pp))[0]
```
gingerBill, line 73:
```odin
([^]u64)(pp)[0]
```

core/runtime/dynamic_map_hash.odin
```diff
@@ -0,0 +1,147 @@
...
+		return u64(pp[0]) | u64(pp[1])<<8 | u64(pp[2])<<16 | u64(pp[3])<<24 |
+		    u64(pp[4])<<32 | u64(pp[5])<<40 | u64(pp[6])<<48 | u64(pp[7])<<56
+	}
+}
+
+@(private)
+_wy_read_4 :: #force_inline proc "contextless" (p: [^]byte, off: int = 0) -> u64 {
+	pp := intrinsics.ptr_offset(p, off)
+	when ODIN_ARCH == .i386 || ODIN_ARCH == .amd64 {
+		return u64((transmute([^]u32)(pp))[0])
```
gingerBill, line 84:
```odin
u64(([^]u32)(pp)[0])
```

core/runtime/dynamic_map_hash.odin
```diff
@@ -0,0 +1,147 @@
...
+	// k = 2: p[0] << 16 | p[1] << 8 | p[1]
+	// k = 3: p[0] << 16 | p[1] << 8 | p[2]
+	return ((u64(p[0])) << 16) | ((u64(p[k >> 1])) << 8) | u64(p[k - 1])
+}
+
+@(optimization_mode="speed")
+_wyhash :: proc "contextless" (data: rawptr, n: int, seed: u64, secret: ^Wyhash_Secret) -> u64 {
+	p := ([^]byte)(data)
+	seed_ := seed ~ _wy_mix(seed ~ secret[0], secret[1])
+	a, b: u64 = ---, ---
```
gingerBill, line 104: Don't use `---` unless you can prove it is better. Just do:
```odin
a, b: u64
```

core/runtime/dynamic_map_hash.odin
```diff
@@ -0,0 +1,147 @@
...
+	} else {
+		i := n
+		if intrinsics.expect(i >= 48, false) {
+			see1, see2 := seed_, seed_
+			for {
+				seed_ = _wy_mix(_wy_read_8(p) ~ secret[1], _wy_read_8(p, 8) ~ seed_)
+				see1 = _wy_mix(_wy_read_8(p, 16) ~ secret[2], _wy_read_8(p, 24) ~ see1)
+				see2 = _wy_mix(_wy_read_8(p, 32) ~ secret[3], _wy_read_8(p, 40) ~ see2)
+
+				p = intrinsics.ptr_offset(p, 48)
```
gingerBill, line 125:
```odin
p = p[48:]
```

core/runtime/dynamic_map_hash.odin
```diff
@@ -0,0 +1,147 @@
...
+				if intrinsics.expect(i < 48, false) {
+					break
+				}
+			}
+			seed_ ~= see1 ~ see2
+		}
+		for intrinsics.expect(i > 16, false) {
+			seed_ = _wy_mix(_wy_read_8(p) ~ secret[1], _wy_read_8(p, 8) ~ seed_)
+			i -= 16
+			p = intrinsics.ptr_offset(p, 16)
```
gingerBill, line 136:
```odin
p = p[16:]
```

core/runtime/dynamic_map_internal.odin
```diff
@@ -911,14 +905,12 @@ default_hasher_string :: proc "contextless" (data: rawptr, seed: uintptr) -> uin
...
-	if ptr := (^[^]byte)(data)^; ptr != nil {
-		for ptr[0] != 0 {
-			h = (h ~ u64(ptr[0])) * 0x100000001b3
-			ptr = ptr[1:]
-		}
-	}
-	h &= HASH_MASK
-	return uintptr(h) | uintptr(uintptr(h) == 0)
+	// TODO/perf: len(str) is O(N), and is unavoidable with wyhash.
+	// This is likely insignificant for the typical short key sizes.
```
gingerBill, line 911: But significant for long key sizes. There was a reason the original hash worked the way it did.

gingerBill:
Please provide performance comparisons between the current hash and proposed hash. This PR will not be accepted (even though it is a draft) without such comparisons.

Yawning:
Raw benchmarks of the hash function is somewhat disappointing, at least with llvm 14.0.6.  I see between a 25 to 50% performance discrepancy between the C reference code and the Odin (-o:speed) implementation, so unless having better statistical properties is a huge net gain (unlikely), this isn't worth doing for the time being.

Some takeaways:
- Even odin wyhash does pull ahead, when data size is about 20 to 30 bytes.
- For moderately large data, wyhash predictably dumpsters FNV.  The cstring case for very large strings also comes out in favor of wyhash.

I'll drop the PR and maybe revisit it once getting llvm 17 on a linux system isn't pain/time consuming, or if I happen to actually need a hash-flood resistant map.

PR #2939, opened 2023-11-09
Allow larger thread poly data

laytan:
The poly data currently has the restriction of being less than a pointer's size, but there is much more space in the `Thread.user_args` array which can be utilized, this commit allows you to pass types that are larger than pointer length as long as the total size of the poly data is less than that of the `Thread.user_args`.


laytan:
Think WIndows is OK now, self_cleanup was closing the thread and so was the join_multiple call, resulting in an invalid handle error

PR #2940, opened 2023-11-10
add frame pointer attribute to procs in debug mode

laytan:
The frame pointer attribute is an important attribute when debugging and traversing the call stack of MacOS programs. Without the frame pointer programs are unable to traverse the stack further than the most recent stack frame.

My main motivation for this change is adding MacOS support to [my backtrace package](https://github.com/laytan/obacktracing).

### DIFF
--- a/src/llvm_backend_proc.cpp
+++ b/src/llvm_backend_proc.cpp
@@ -316,7 +316,9 @@ gb_internal lbProcedure *lb_create_procedure(lbModule *m, Entity *entity, bool i
 			lb_set_llvm_metadata(m, p, p->debug_info);
 		}
 
-		lb_add_string_attribute_to_proc(m, p->value, "frame-pointer", "all");
+		if (build_context.metrics.os == TargetOs_darwin) {
+			lb_add_string_attribute_to_proc(m, p->value, "frame-pointer", "non-leaf");
+		}
 	}
 
 	if (p->body && entity->pkg && ((entity->pkg->kind == Package_Normal) || (entity->pkg->kind == Package_Init))) {
--- a/src/llvm_backend_proc.cpp
+++ b/src/llvm_backend_proc.cpp
@@ -315,6 +315,8 @@ gb_internal lbProcedure *lb_create_procedure(lbModule *m, Entity *entity, bool i
 			LLVMSetSubprogram(p->value, p->debug_info);
 			lb_set_llvm_metadata(m, p, p->debug_info);
 		}
+
+		lb_add_string_attribute_to_proc(m, p->value, "frame-pointer", "all");
 	}
 
 	if (p->body && entity->pkg && ((entity->pkg->kind == Package_Normal) || (entity->pkg->kind == Package_Init))) {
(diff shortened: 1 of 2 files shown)


src/llvm_backend_proc.cpp
```diff
@@ -315,6 +315,8 @@ gb_internal lbProcedure *lb_create_procedure(lbModule *m, Entity *entity, bool i
 			LLVMSetSubprogram(p->value, p->debug_info);
 			lb_set_llvm_metadata(m, p, p->debug_info);
 		}
+
+		lb_add_string_attribute_to_proc(m, p->value, "frame-pointer", "all");
```
gingerBill, line 319: This is being added to EVERY procedure, regardless of whether it has a body or not, and regardless of the platform too. Is that a good idea?

laytan: Did some comparisons and searches and it doesn't seem to be a big deal in practice.
- I compiled (with the master branch and this branch) a (pretty big, around 10k loc) program and the difference was 0.7MB in the assembly output and negligible output on the binary and llvm
- There is minor performance differences that are not a big deal (especially in -debug mode): https://stackoverflow.com/questions/13006371/does-omitting-the-frame-pointers-really-have-a-positive-effect-on-performance-an
- Also from that stackoverflow and the sources it links, frame pointer are also used on Windows for better debugging (on top off or besides the pdb) and I don't know about Linux. But in practice only Darwin is not able to do any stack traces at the moment so I would be fine with only doing it on Darwin or putting it behind a compiler flag entirely.

flysand7: 14% slowdown seems like a lot, especially if that slowdown comes from a hot loop or even nested loops. But I would agree it's not a big deal, since you as a developer can manually optimize this stuff if it bothers you.

I don't imagine that it affects bodyless procedures in any way, unless there's a bug with llvm. I think it can be tested and if it doesn't crash its probably fine.

Regarding backtrace, is there some differences from linux? I know that debuggers on linux (x86) generally don't care about frame pointers, they do some buckwild shit like disassembling machine code to figure out the next function pointer.

Otherwise I believe frame pointer emission should probably be controlled by the user.

gingerBill: As I was asking, why are you adding this to EVERY platform rather than just Darwin/macOS that you originally wanted?

Can you just only enable this for Darwin platforms?

laytan: I think I answered why I was adding it to EVERY platform in point 3 of my original reply that Microsoft also says frame-pointers are beneficial for Windows and I said I wasn't sure about Linux but Fedora had an initiative to add them by default in their build commands so it probably helps Linux too.

Anyway I agree that this being only added on Darwin is a good idea because the other platforms don't noticeably need it.

I have made that change, and also made the change to the "non-leaf" variant instead of "all" which reduces the impact a little bit and still works perfectly.

laytan:
Update: I dove into this further and found that the way debuggers get backtraces is using stack unwinding based on the instructions from dwarf CIE and FDE that are either/both embedded in the mach-o (binary file) or the debug information. The `backtrace` API coming from libc and the likes does not do this, it just goes through frame pointers.

I changed my package to use the `libunwind` API's which are actually included when you `-lSystem`, this does do the unwinding correctly and I am able to get frames and am able to translate them to symbols.

This PR is no longer needed for my specific use case. But I do want to leave some info because frame-pointers are generally expected to be there on arm64 macos.

1. Because arm has many more registers than say x86 it is not a big performance loss.
2. And according to Apple themselves (https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms) "The frame pointer register (x29) must always address a valid frame record. Some functions — such as leaf functions or tail calls — may opt not to create an entry in this list. As a result, stack traces are always meaningful, even without debug information.", they even take it a step further than this PR and say it should always be there (not only with `-debug`).

PR #2941, opened 2023-11-10
[runtime]: Add os_specific_linux, that calls the syscall directly.

flysand7:


laytan:
I think the idea with runtime is to import nothing (other than intrinsics) and copy implementations.

I also had a look and it seems writing is the same syscall (number and structure) on (linux, darwin, freebsd, openbsd). So I think the ideal thing here is using the `intrinsics.syscall` and add this in a file `os_specific_unix.odin` with build flags for the before mentioned OS's.

flysand7:
Unifying to unix should not happen. Syscall numbers are different for different unix-like operating systems, and even on the same system they can be the same (on linux: `SYS_read` = 1 on amd64, `4` on arm32 and i386 and 64 on amd64). Also BSD kernels seem handwavy about compatibility, while darwin has broke compatibility a few times (as far as I know), so calling syscalls directly is a paint to maintain which would require turning a simple routine into a giant ODIN_OS + ODIN_ARCH (+OS version?) switch statement.

I'm not against the idea of calling `intrinsics.syscall` and special-casing the architectures on linux. This is basically what sys/linux does, which actually in itself doesn't import anything (I made it importable from anywhere, __especially__ runtime)

On that note, I see why runtime was made to not depend on anything -- it's a good idea to guard against circular imports. But also I feel like packages in `sys` shouldn't depend on runtime (for in reality Odin uses OS functionality to implement it's runtime, not the other way around), but I'm not sure myself how I feel about that.

laytan:
Ah I had the wrong assumption, I looked at a couple syscall tables for the mentioned platforms and it was 4 everywhere. Now I did another search and see they are different indeed, that complicates things.

PR #2942, opened 2023-11-10
core:sys/linux: make Perf_Read_Format a bitset

flga:
The current definition is incompatible with how it's meant to be used, the values assume a bitset but it's used as regular flags.

Convert it to a bitset for consistency with the rest of the api.

cc @flysand7


flysand7 (approved):
Right, that's a blunder. I see how that happened. Yep, this should be a bitset

PR #2944, opened 2023-11-10
Add -microarch:?

Kelimion:
If you don't supply `-microarch` it autodetects as usual.
If you supply `-microarch:?` it prints a list of microarchitectures compatible with `-target` (which can be omitted as usual)
If you supply `-microarch:<unknown>`, Odin tells you the microarchitecture is unrecognized, followed by the list.


DragosPopse:
This could be related. `odin build -help` prints examples of `-microarch` (sandybridge, native), but it's not certain if those are all options. This should however be expanded for `-target` too. The only way to get the available targets is to make `-target` to error by providing an unknown one (`-target:?` works, but it works because the error message prints out all the available targets)

The more "complete" alternative is to either make `-help` provide all the available `-microarch` `-target` and everything else that other flags expect, provide a separate `-help-extra` that does that, or accept a flag like `-help:microarch` that prints a more detailed output of what a specific flag does.

gingerBill (approved):

PR #2950, opened 2023-11-12
fix nil exceptions with incomplete code parse

laytan:
This makes the parser more fault tolerant because the different parse_foo procs return nil when in an invalid state, which is fine most of the time but when creating a node it would crash accessing its position.

### DIFF
--- a/core/odin/ast/clone.odin
+++ b/core/odin/ast/clone.odin
@@ -7,7 +7,7 @@ import "core:reflect"
 import "core:odin/tokenizer"
 _ :: intrinsics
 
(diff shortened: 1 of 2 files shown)


core/odin/ast/clone.odin
```diff
@@ -23,6 +23,15 @@ new :: proc($T: typeid, pos, end: tokenizer.Pos) -> ^T {
 	return n
 }
 
+new_from_pos_and_end_node :: proc($T: typeid, pos: tokenizer.Pos, end: ^Node) -> ^T {
+	return new(T, pos, end != nil ? end.end : {})
```
gingerBill, line 27: Instead of `{}`, use `pos` instead.

PR #2952, opened 2023-11-14
Add various missing windows procedures

Pingar5:
This is far from all of the procedures that are missing, but they are some that I have needed to use so I figured I'd contribute them

### DIFF
--- a/core/sys/windows/kernel32.odin
+++ b/core/sys/windows/kernel32.odin
@@ -290,6 +290,14 @@ foreign kernel32 {
 		hTemplateFile: HANDLE,
 	) -> HANDLE ---
 
+	GetFileTime :: proc(
+		hFile: HANDLE,
+		lpCreationTime: LPFILETIME,
+		lpLastAccessTime: LPFILETIME,
+		lpLastWriteTime: LPFILETIME,
+	) -> BOOL ---
+	CompareFileTime :: proc(lpFileTime1: LPFILETIME, lpFileTime2: LPFILETIME) -> LONG ---
+
 	FindFirstFileW :: proc(fileName: LPCWSTR, findFileData: LPWIN32_FIND_DATAW) -> HANDLE ---
 	FindNextFileW :: proc(findFile: HANDLE, findFileData: LPWIN32_FIND_DATAW) -> BOOL ---
 	FindClose :: proc(findFile: HANDLE) -> BOOL ---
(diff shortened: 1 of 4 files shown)


core/sys/windows/kernel32.odin
```diff
@@ -179,6 +179,11 @@ foreign kernel32 {
 		lpNewFileName: LPCWSTR,
 		bFailIfExists: BOOL,
 	) -> BOOL ---
+	CopyFileA :: proc(
```
Kelimion, line 182: As a rule, we only bind against Unicode-compatible APIs.

See for example the comment in `kernel32.odin`: 	`OutputDebugStringA :: proc(lpOutputString: LPCSTR) --- // The only A thing that is allowed`. There might be a few other exceptions that snuck in by mistake, but we're not adding more.

Kelimion: 1) Odin is built around Unicode support.
2) Consistency reduces bugs.
3) Why add footguns? The `A` variants of some APIs will take UTF-8 input if you're on the right version of Windows, but historically they don't.

@gingerBill might have additional reasons. He decided the rule.

gingerBill: @Pingar5 

Odin does not use ASCII string, it is used UTF-8 strings. As such, we _require_ Unicode-compatible APIs.

flysand7: @Kelimion A-variants on windows only take UTF-8 strings if the application links itself against a manifest requesting utf-8 support, which Odin doesn't do by default anyway, so yeah, not that useful

gingerBill: @flysand7 Not just that, it only works on some versions of Windows 10/11 too. Not all.

gingerBill: @Pingar5 There are numerous ways depending on what you require.

If it's a string literal: `intrinsics.constant_utf16_cstring(...)` or `windows.L(...)` (`core:sys/windows`)

If it's a string value: `windows.utf8_to_wstring(s, allocator)` (`core:sys/windows`) https://github.com/odin-lang/Odin/blob/master/core/sys/windows/util.odin

Pingar5: Okay, I removed the A functions from the PR. Let me know if there is anything else I need to change to get these merged in

PR #2964, opened 2023-11-19
Updated bindings and libs for Raylib 5.0

seanarooni:


vendor/raylib/rlgl.odin
```diff
@@ -425,7 +433,8 @@ foreign lib {
 	rlDisableScissorTest     :: proc() ---                           // Disable scissor test
 	rlScissor                :: proc(x, y, width, height: c.int) --- // Scissor test
 	rlEnableWireMode         :: proc() ---                           // Enable wire mode
-	rlDisableWireMode        :: proc() ---                           // Disable wire mode
+	rlEnablePointMode		 :: proc() --- 							 // Enable point mode
```
flysand7, line 436: Please use tabs for indentation and **spaces** for alignment. There's a few cases of you using tabs to align things which looks off if you set the tab size to 8.

flysand7 (changes requested):
Could you also please add `RAYLIB_VERSION_MAJOR`, etc constants to the package and bind the `raylib_version` symbol from `rcore.c`? They don't really make it public but it's kind of essential.

vendor/raylib/raylib.odin
```diff
@@ -514,6 +514,19 @@ FilePathList :: struct {
     paths:    [^]cstring,                 // Filepaths entries
 }
 
+// Automation event
+AutomationEvent :: struct {
+	frame: 		c.uint,				// Event frame
+	type: 		c.uint,				// Event type (AutomationEventType)
+    params: 	[4]c.int, 			// Event parameters (if required) ---
```
gingerBill, line 521: Please use tabs for indentation and spaces for alignment.

Kelimion:
Are these compiled with JPEG support? The current libraries don't appear to be, and it would be great to have. It's apparently not a default when building Raylib.

PR #2979, opened 2023-11-25
Binary search improvements

rope-hmg:
Modified the algorithm so that the index is either the location of the element if found or the index at which to insert the element to maintain sorted order.

Also added some tests to verify the above claim.

### DIFF
--- a/core/slice/slice.odin
+++ b/core/slice/slice.odin
@@ -117,46 +117,95 @@ linear_search_proc :: proc(array: $A/[]$T, f: proc(T) -> bool) -> (index: int, f
 	return -1, false
 }
 
+/*
+	Binary search searches the given slice for the given element.
+    If the slice is not sorted, the returned index is unspecified and meaningless.
+
+    If the value is found then the returned int is the index of the matching element.
+    If there are multiple matches, then any one of the matches could be returned.
+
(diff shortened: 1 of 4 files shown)


core/slice/slice.odin
```diff
@@ -117,46 +117,95 @@ linear_search_proc :: proc(array: $A/[]$T, f: proc(T) -> bool) -> (index: int, f
 	return -1, false
 }
 
-@(require_results)
-binary_search :: proc(array: $A/[]$T, key: T) -> (index: int, found: bool)
-	where intrinsics.type_is_ordered(T) #no_bounds_check {
+/*
+	Binary search searches the given slice for the given element.
+    If the slice is not sorted, the returned index is unspecified and meaningless.
```
flysand7, line 122: Please use tabs for indentation.

tests/core/slice/test_core_slice.odin
```diff
@@ -180,3 +181,59 @@ test_sort_by_indices :: proc(t: ^testing.T) {
 		}
 	}
 }
+
+@test
+test_binary_search :: proc(t: ^testing.T) {
```
laytan, line 186: The test specific `log` and `expect` procs defined at the top of the file should be used instead of `fmt.printf` and `assert`

PR #2987, opened 2023-11-28
Add `-test-build-only` flag

InKryption:
As the name implies, passing this flag will stop tests from running when invoking `odin test`, only outputting the binary.

This change also makes some small style adjustments for consistency, and removes a TODO comment which has already been resolved.


Kelimion:
This needs a rethink. Sorry it got lost in the churn.

PR #2997, opened 2023-12-02
[core/runtime]: Add append_elem_slice variant for appending values in slices to dynamic arrays

flysand7:
This acts exactly the same as `append_string`, but it is more general variant which I think should be there.


gingerBill:
This causes a bunch of ambiguities, and why the `..` operator is required instead. I will not allow this.

PR #2998, opened 2023-12-02
OS improvements

laytan:
If you currently try to read or write more than about 2GB to/from a file on non-Windows you will probably get an error, this PR adds a max on each call (like Windows already has).
It was currently the case that a big file to os.read_entire_file would fail, this fixes that too and any callers of read/write will automatically handle it now.
This is safe because in practice any user will be checking the returned n and doing this in a loop anyway, and more than a gigabyte will probably never have filled fully without this.

I also added end-of-file propagation from the file streams, which was also already handled on Windows. According to man pages, a 0 byte read/write without an error is an end-of-file.
Because of this the bufio scanner/reader will also become much faster because they would previously do a syscall 128 times (by default) and then return .No_Progress, now this will just be a .EOF after the first empty call.

### DIFF
--- a/core/os/os_darwin.odin
+++ b/core/os/os_darwin.odin
@@ -568,15 +568,24 @@ close :: proc(fd: Handle) -> bool {
 	return _unix_close(fd) == 0
 }
 
+// If you read or write more than `SSIZE_MAX` bytes, most darwin implementations will return `EINVAL`
+// but it is really implementation defined. `SSIZE_MAX` is also implementation defined but usually
+// the max of an i32 on Darwin.
+// In practice a read/write call would probably never read/write these big buffers all at once,
+// which is why the number of bytes is returned and why there are procs that will call this in a
+// loop for you.
+// We set a max of 1GB to keep alignment and to be safe.
 @(private)
-MAX_RW :: 0x7fffffff // The limit on Darwin is max(i32), trying to read/write more than that fails.
+MAX_RW :: 1 << 30
 
 write :: proc(fd: Handle, data: []byte) -> (int, Errno) {
 	if len(data) == 0 {
 		return 0, ERROR_NONE
 	}
 
-	bytes_written := _unix_write(fd, raw_data(data), c.size_t(len(data)))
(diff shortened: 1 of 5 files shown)


core/os/os_darwin.odin
```diff
@@ -568,15 +568,24 @@ close :: proc(fd: Handle) -> bool {
...
 
+// If you read or write more than `SSIZE_MAX` bytes, most darwin implementations will return `EINVAL`
+// but it is really implementation defined. `SSIZE_MAX` is also implementation defined but usually
+// the max of an i32 on Darwin.
+// In practice a read/write call would probably never read/write these big buffers all at once,
+// which is why the number of bytes is returned and why there are procs that will call this in a
+// loop for you.
+// We set a max of 1GB to keep alignment and to be safe.
 @(private)
-MAX_RW :: 0x7fffffff // The limit on Darwin is max(i32), trying to read/write more than that fails.
```
laytan, line 572: This constant was previously unused

core/os/os_linux.odin
```diff
@@ -547,12 +547,23 @@ close :: proc(fd: Handle) -> Errno {
 	return _get_errno(unix.sys_close(int(fd)))
 }
 
+// If you read or write more than `SSIZE_MAX` bytes, result is implementation defined (probably an error).
+// `SSIZE_MAX` is also implementation defined but usually the max of a `ssize_t` which is `max(int)` in Odin.
+// In practice a read/write call would probably never read/write these big buffers all at once,
```
gingerBill, line 552: So we need to wrap the internal `read` functionality then.

because what if you do need to read/write more than 1 GiB at a time? e.g. a very large file. This isn't that uncommon.

laytan: I did think about that and came to the conclusion that you would be calling this, checking the returned int that says how much has been read, and call it again if needed, giving the most amount of control. Which still leaves looped implementations like `read_at_least`, `read_full` and `read_entire_file` for the rest.

If we were to say that `read` does the looping, it would basically be the same as `read_full`.

gingerBill: Very good point!

PR #2999, opened 2023-12-02
add crypto.rand_bytes for Darwin and BSD

laytan:
`arc4random_buf` is implemented on Darwin, FreeBSD and OpenBSD with the same signature.


core/crypto/rand_darwin_and_bsd.odin
```diff
@@ -0,0 +1,12 @@
+//+build freebsd, openbsd, darwin
+package crypto
+
+foreign import libc "system:c"
+
+foreign libc {
+	arc4random_buf :: proc(buf: [^]byte, nbytes: uint) ---
+}
```
gingerBill, line 8: Is `libc` necessary here? Or even a good idea?

core/crypto/rand_darwin.odin
```diff
@@ -0,0 +1,11 @@
+package crypto
+
+import "core:fmt"
+import "core:sys/darwin"
+
+_rand_bytes :: proc(dst: []byte) {
+	res := darwin.SecRandomCopyBytes(count=len(dst), bytes=raw_data(dst))
+	if res != .Success {
+		panic(fmt.tprintf("crypto/rand_bytes: SecRandomCopyBytes returned non-zero result: %v", res))
```
laytan, line 9: I wanted to bind to https://developer.apple.com/documentation/security/1394686-seccopyerrormessagestring to get a string message for the error but this returns a Foundation string, which we currently have bindings for it but those are in vendor, so can't use them

Yawning:
For what it's worth, this looks good to me.  `arc4random` is kind of hard to avoid on non-Darwin BSDs, the getentropy systemcall is tailored around seeding a CSPRNG, and we don't have one presently.  libc tends to be less of a trashfire there as well...

laytan:
Rebased this to fix conflicts, still ready for review

PR #3001, opened 2023-12-04
add non-zeroing append and resize

colrdavidson:
This makes a *huge* difference for Odin memory usage on Linux / OSX.
Odin previously touched and zeroed all pages on resize and allocation, which for big dynamic arrays is incredibly expensive.

I've added non_zero_resize / non_zero_append and friends as an option so that you don't have to zero pages you don't use.

This is probably not exactly what we want, but it's a start.


core/mem/allocators.odin
```diff
@@ -85,7 +85,7 @@ arena_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
 	case .Free_All:
 		arena.offset = 0
 
-	case .Resize:
+	case .Resize, .Resize_Non_Zeroed:
```
flysand7, line 88: Should also return `.Resize_Non_Zeroed` in `Query_Features` request.

flysand7:
`zero` doesn't work well as a verb, so I'm wondering whether using `uninitialized` in its place is better. E.g. `resize_uninitialized`, `reserve_dynamic_array_uninitialized`, `initialize: bool`, et cetera sound like a bit of a more readable names

gingerBill:
I will tentatively accept this after the December Odin release.

core/log/log_allocator.odin
```diff
@@ -80,6 +80,13 @@ log_allocator_proc :: proc(allocator_data: rawptr, mode: runtime.Allocator_Mode,
 				la.prefix, padding, old_memory, old_size, size, alignment,
 				location = location,
 			)
+		case .Resize_Non_Zeroed:
```
laytan, line 83: The log allocator has been changed on master to allow for human readable byte format and to use a stack buffer instead of temp allocating to prevent recursion when logging the temp allocator. Could you apply these changes to this case too?

colrdavidson: yep, will do

PR #3006, opened 2023-12-05
fix strings.last_index_any for single char

hwchen:
Looks like for the case where the haystack was a single char, `strings.last_index_any` was returning its index into the "any" charset, instead of just `0` if found.

Also adds small tests.

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -1792,7 +1792,8 @@ last_index_any :: proc(s, chars: string) -> (res: int) {
 		if r >= utf8.RUNE_SELF {
 			r = utf8.RUNE_ERROR
 		}
-		return index_rune(chars, r)
+		i := index_rune(chars, r)
+		return i if i < 0 else 0
 	}
 	
 	if len(s) > 8 {
--- a/tests/core/strings/test_core_strings.odin
+++ b/tests/core/strings/test_core_strings.odin
@@ -67,6 +67,18 @@ test_index_any_larger_string_found :: proc(t: ^testing.T) {
 	expect(t, index == 8, "index_any should be 8")
 }
 
+@test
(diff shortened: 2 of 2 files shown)


flysand7:
Returning index into the `any` charset is the intended behavior. The documentation of the function explicitly states that:

https://github.com/odin-lang/Odin/blob/d4f4e2f0197d0c39129bba1eea293cd76dd54cd1/core/strings/strings.odin#L1761C8-L1761C68

The idea, I think, is to be able to know which character was found in the string, not just whether it was found.

xtactis:
> Returning index into the `any` charset is the intended behavior. The documentation of the function explicitly states that:

IMO the documentation is actually quite ambiguous about in which string the result index is; regardless the implementation currently sometimes returns indexing into the haystack, and sometimes (as described in this PR) into any charset. That should probably not be the case.

> The idea, I think, is to be able to know which character was found in the string, not just whether it was found.

You get that regardless of which string you index into, if there's a match, the characters will coincide. But the string to index should be the same one regardless of the argument lengths.

core/strings/strings.odin
```diff
@@ -1792,7 +1792,7 @@ last_index_any :: proc(s, chars: string) -> (res: int) {
 		if r >= utf8.RUNE_SELF {
 			r = utf8.RUNE_ERROR
 		}
-		return index_rune(chars, r)
+		return -1 if index_rune(chars, r) == -1 else 0
```
gingerBill, line 1795: I would have done something like this: 
```odin
i := index_rune(chars, r)
return i if i < 0 else 0
```

flysand7:
```odin
return -1 if i<0 else 0
```

hwchen:
Just want to check if there's anything else needed to get the PR merged. I've updated the `return` statement as @gingerBill suggested.

PR #3008, opened 2023-12-06
fix struct stat layout linux arm64

### LINKED ISSUE #2793: filepath.walk does not work on Linux
I believe the main issue is that the directory is not seen as a directory, and thats why it doesn't keep going to the other files.

```odin
package main

import "core:path/filepath"
import "core:fmt"
import "core:os"

main :: proc() {
	filepath.walk(
	".",
	proc(info: os.File_Info, in_err: os.Errno, user_data: rawptr) -> (err: os.Errno, skip_dir: bool) {
		// We only get here once, with the root directory, even though there are more files.

		assert(err == os.ERROR_NONE)

		fmt.println(info.fullpath) // /app (the root)
		fmt.println(info.is_dir)   // false (this is certainly a directory)
		return
	}, nil)
}
```

Dockerfile I used to test it (and make sure it wasn't my os)
```dockerfile
FROM alpine:latest

WORKDIR /odin

ENV PATH "$PATH:/usr/lib/llvm14/bin:/odin"

RUN apk add --no-cache git bash make clang14 llvm14-dev musl-dev linux-headers

RUN git clone --depth=1 https://github.com/odin-lang/Odin . && \
	LLVM_CONFIG=llvm14-config make

WORKDIR /app

# random file, doesn't get a call in the walk
RUN echo "testing" > testing123.txt

# the file above
COPY test.odin test.odin

CMD ["odin", "run", "test.odin", "-file"]
```

laytan:
Fixes #2793

### DIFF
--- a/core/os/os_linux.odin
+++ b/core/os/os_linux.odin
@@ -272,18 +272,17 @@ when ODIN_ARCH == .arm64 {
 		uid:           u32, // User ID of the file's owner
 		gid:           u32, // Group ID of the file's group
 		rdev:          u64, // Device ID, if device
-		 _padding:     u64,
+		_:             u64, // Padding
 		size:          i64, // Size of the file, in bytes
 		block_size:    i32, // Optimal blocksize for I/O
-		_padding_2:    i32,
+		_:             i32, // Padding
 		blocks:        i64, // Number of 512-byte blocks allocated
 
 		last_access:   Unix_File_Time, // Time of last access
 		modified:      Unix_File_Time, // Time of last modification
 		status_change: Unix_File_Time, // Time of last status change
 
-		_reserve1,
-		_reserve2: i32,
+		_reserved:     [2]i32,
 	}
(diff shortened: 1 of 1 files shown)


core/os/os_linux.odin
```diff
@@ -263,26 +263,51 @@ Unix_File_Time :: struct {
...
+when ODIN_ARCH == .arm64 {
+	OS_Stat :: struct {
+		device_id:     u64, // ID of device containing file
+		serial:        u64, // File serial number
+		mode:          u32, // Mode of the file
+		nlink:         u32, // Number of hard links
+		uid:           u32, // User ID of the file's owner
+		gid:           u32, // Group ID of the file's group
+		rdev:          u64, // Device ID, if device
+		 _padding:     u64,
```
flysand7, line 275: Extra space on the start of the line here.

Also as a suggestion instead of `_padding_N` you can just declare padding fields as

```odin
_: u64,
```

flysand7 (approved):

PR #3010, opened 2023-12-07
[vendor/stb/image]: Add pre-compiled binaries for linux

flysand7:
The archive includes three object files, one each for `stb_image`, `stb_image_resize2` and `stb_image_write`. Compiled with default configuration (see below)

```c
#define STB_IMAGE_IMPLEMENTATION
#include "stb_image.h"

#define STB_IMAGE_RESIZE_IMPLEMENTATION
#include "stb_image_resize2.h"

#define STB_IMAGE_WRITE_IMPLEMENTATION
#include "stb_image_write.h"
```


flysand7:
Done.

gingerBill:
How does this work with libc?

flysand7:
My guess is that it doesn't, that was an oversight.

I'm not sure if my thinking is correct here, but maybe overriding libc procedures and making those procedures defined in Odin should work independently of libc?

Issue #3021, opened 2023-12-14
Multiple values for 'assign_at()' may result in an extra element being inserted into the array


lord-andrius:
## Context
using  `assign_at()`  with multiple values will insert an extra element 

* Operating System & Odin Version:
* Please paste `odin report` output:
*    Odin: dev-2023-12-nightly:31b1aef4
        OS:   Windows 10 Professional (version: 22H2), build 19045.3693
        CPU:  Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz
        RAM:  8078 MiB

## Expected Behavior

using  `assign_at()`  with multiple values should only insert the elements that you passed as arguments

## Current Behavior
using  `assign_at()`  with multiple values will insert an extra element 

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.
```
   v := make([dynamic]int)
   assign_at(&v, 0, 33, 1)
   fmt.println(v)
```

PR #3022, opened 2023-12-14
fix issue #3021

lord-andrius:
correcting assign_at_elems() so that it no longer inserts an extra element into the array

### DIFF
--- a/core/runtime/core_builtin.odin
+++ b/core/runtime/core_builtin.odin
@@ -591,7 +591,9 @@ assign_at_elems :: proc(array: ^$T/[dynamic]$E, index: int, args: ..E, loc := #c
 		copy(array[index:], args)
 		ok = true
 	} else {
-		resize(array, index+len(args), loc) or_return
+		new_size := len(array)+len(args)
+		if len(array) == index && len(args) == 0 do new_size += 1
+		resize(array, new_size, loc) or_return
 		copy(array[index:], args)
 		ok = true
 	}
@@ -668,6 +670,7 @@ reserve_dynamic_array :: proc(array: ^$T/[dynamic]$E, capacity: int, loc := #cal
 // Note: Prefer the procedure group `resize`
 @builtin
 resize_dynamic_array :: proc(array: ^$T/[dynamic]$E, length: int, loc := #caller_location) -> Allocator_Error {
+
 	if array == nil {
 		return nil
 	}
--- a/core/runtime/core_builtin.odin
+++ b/core/runtime/core_builtin.odin
@@ -592,7 +592,6 @@ assign_at_elems :: proc(array: ^$T/[dynamic]$E, index: int, args: ..E, loc := #c
 		ok = true
 	} else {
 		new_size := index+len(args)
-		if len(array) == index && len(args) == 0 { new_size += 1 }
 		resize(array, new_size, loc) or_return
 		copy(array[index:], args)
 		ok = true
--- a/core/runtime/core_builtin.odin
+++ b/core/runtime/core_builtin.odin
@@ -591,7 +591,7 @@ assign_at_elems :: proc(array: ^$T/[dynamic]$E, index: int, args: ..E, loc := #c
 		copy(array[index:], args)
 		ok = true
 	} else {
-		resize(array, index+1+len(args), loc) or_return
+		resize(array, index+len(args), loc) or_return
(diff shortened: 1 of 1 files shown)


core/runtime/core_builtin.odin
```diff
@@ -591,7 +591,7 @@ assign_at_elems :: proc(array: ^$T/[dynamic]$E, index: int, args: ..E, loc := #c
 		copy(array[index:], args)
 		ok = true
 	} else {
-		resize(array, index+1+len(args), loc) or_return
+		resize(array, index+len(args), loc) or_return
```
gingerBill, line 594: Does this take into the account when `len(args) == 0`?

lord-andrius: it works

lord-andrius: Thanks for quickly reviewing. it didn't but now it does

core/runtime/core_builtin.odin
```diff
@@ -591,7 +591,9 @@ assign_at_elems :: proc(array: ^$T/[dynamic]$E, index: int, args: ..E, loc := #c
 		copy(array[index:], args)
 		ok = true
 	} else {
-		resize(array, index+1+len(args), loc) or_return
+		new_size := len(array)+len(args)
+		if len(array) == index && len(args) == 0 do new_size += 1
```
laytan, line 595: We don't allow the usage of `do` in core/vendor because the `-disallow-do` flag exists and we don't want core/vendor code to come up with it enabled.
```suggestion
		if len(array) == index && len(args) == 0 { new_size += 1 }
```

lord-andrius: I applied your suggestions

laytan:
This way of determining size actually has a problem too, for example:
```odin
v := make([dynamic]int)
assign_at(&v, 0, 1, 2)
assign_at(&v, 1, 3, 4)
fmt.println(v) // [1, 3, 4, 0], should be [1, 3, 4]
```

I think your original change that just changed the resize to `resize(array, index+len(args), loc) or_return` was actually the way to go here. If `len(args) == 0` (what Bill asked about) would effectively do nothing which is also what would happen in the if branch so that is also consistent.

lord-andrius:
> This way of determining size actually has a problem too, for example:
> 
> ```odin
> v := make([dynamic]int)
> assign_at(&v, 0, 1, 2)
> assign_at(&v, 1, 3, 4)
> fmt.println(v) // [1, 3, 4, 0], should be [1, 3, 4]
> ```
> 
> I think your original change that just changed the resize to `resize(array, index+len(args), loc) or_return` was actually the way to go here. If `len(args) == 0` (what Bill asked about) would effectively do nothing which is also what would happen in the if branch so that is also consistent.

ok, i just changed back

core/runtime/core_builtin.odin
```diff
@@ -591,7 +591,9 @@ assign_at_elems :: proc(array: ^$T/[dynamic]$E, index: int, args: ..E, loc := #c
 		copy(array[index:], args)
 		ok = true
 	} else {
-		resize(array, index+1+len(args), loc) or_return
+		new_size := index+len(args)
+		if len(array) == index && len(args) == 0 { new_size += 1 }
```
laytan, line 595:
```suggestion
```
This line causes an extra unwanted element, eg:
```
v: [dynamic]int
assign_at(&v, 0)
fmt.println(v) // [0] when it should still be empty/[].

v2: [dynamic]int
assign_at(&v2, 0, 1, 2, 3)
assign_at(&v2, 3)
fmt.println(v2) // [1, 2, 3, 0] when it should be [1, 2, 3].
```

lord-andrius: Soory I thought I should create a new position when entering a position that doesn't exist. But if that is not the case then i will except your chnages

laytan: You may be right, lets see what Bill thinks because there aren't any docs/comments on this.

My intuition/reason was assigning nothing at the end should do nothing.
But I can also see wanting the assurance that `assign_at(&v, 4)` with any amount of args makes the array at least of length 4.

laytan (approved):

gingerBill:
I have correctly fixed this now by just unifying the logic by using the same code as the `assign_at_elem_string` code.

PR #3026, opened 2023-12-16
core/crypto/chacha20: Use SIMD

Yawning:
core/crypto/chacha20: Use 128-bit SIMD
    
This is somewhat conservative and only processes 2 blocks at once, when
the supported targets have enough registers to do 3 (amd64) or 4 (arm),
but on the other hand this implementation is such that, as long as LLVM
isn't doing something totally awful, it will never spill.
    
Unscientific benchmarks (Ryzen 5700G, LLVM18):
ChaCha20    64 bytes: 801.586 MiB/s ->  872.280 MiB/s
ChaCha20  1024 bytes: 796.579 MiB/s -> 1372.795 MiB/s
ChaCha20 65536 bytes: 808.675 MiB/s -> 1393.198 MiB/s
    
chacha20poly1305    64 bytes: 172.804 MiB/s -> 181.603 MiB/s
chacha20poly1305  1024 bytes: 497.487 MiB/s -> 679.336 MiB/s
chacha20poly1305 65536 bytes: 590.149 MiB/s -> 848.749 MiB/s


Yawning:
Superseded by #4003

PR #3032, opened 2023-12-18
Runtime arena edge cases

laytan:
There is an edge case in the runtime arena where with very specific sizes and alignments the arena thinks it has enough space so doesn't create a new block and then adding up the alignment causes an out of memory error.

There was also a report https://discord.com/channels/568138951836172421/585072813954564100/1186172289197297745 that the other test case with a big allocation and large alignment fails but I haven't been able to reproduce locally. (Edit: eventually the problem popped up in the Linux CI)

~~This PR just contains test cases for now and is a draft. Fix will come later but I wanted to see if the second case I wasn't able to reproduce does fail CI.~~

### DIFF
--- a/core/mem/virtual/arena.odin
+++ b/core/mem/virtual/arena.odin
@@ -98,15 +98,15 @@ arena_alloc :: proc(arena: ^Arena, size: uint, alignment: uint, loc := #caller_l
 
 	switch arena.kind {
 	case .Growing:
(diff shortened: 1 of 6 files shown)


laytan:
Found another bug where it would return more bytes than requested and added a test case for it.

laytan:
This is fixed and ready for review but I wouldn't be surprised if there is a better way of doing this.

gingerBill (approved):

PR #3034, opened 2023-12-19
Use generated unicode tables to classify runes

flysand7:
- Implemented classification of runes beyond latin-1 unicode range, which should include cyrillic, greek and a bunch of other languages.
- The script that generated XML entities for unicode characters was updated to match the current version of `core:encoding/xml`
- Python script that downloaded `unicode.xml` had mixed indentation (tab,space,tab -- in one spot), which was fixed

---

unicode.xml that test assets download only contains character data up until Unicode 8.0, which is somewhat old. Future someone might rewrite the generator to match Unicode 15.0.


flysand7:
checked out from the wrong branch, I'll make another PR.

PR #3040, opened 2023-12-21
Windows additions

NicknEma:
I started to add missing bindings for user32, and the required types; to make the process easier, I sorted the list alphabetically, that's why it looks like a lot of lines were deleted. I also fixed some incorrect existing declarations (for example, they used int instead of c_int/i32) and added some comments.


core/sys/windows/types.odin
```diff
@@ -297,6 +298,14 @@ REASON_CONTEXT :: struct {
 }
 PREASON_CONTEXT :: ^REASON_CONTEXT
 
+BLENDFUNCTION :: struct {
+  BlendOp: BYTE,
+  BlendFlags: BYTE,
+  SourceConstantAlpha: BYTE,
+  AlphaFormat: BYTE,
+}
```
flysand7, line 306: Make sure to use tabs in your source files. We use tabs for indentation and spaces for alignment. I would recommend installing .editorconfig plugin for whatever editor you're using so that your editor can configure itself in accordance with the styles the projects use.

core/sys/windows/user32.odin
```diff
@@ -5,20 +5,51 @@ foreign import user32 "system:User32.lib"
...
+	BroadcastSystemMessage :: proc(flags: DWORD, lpInfo: ^DWORD, Msg: UINT, wParam: WPARAM, lParam: LPARAM) -> c_long ---
+	BroadcastSystemMessageA :: proc(flags: DWORD, lpInfo: ^DWORD, Msg: UINT, wParam: WPARAM, lParam: LPARAM) -> c_long ---
+	BroadcastSystemMessageExA :: proc(flags: DWORD, lpInfo: ^DWORD, Msg: UINT, wParam: WPARAM, lParam: LPARAM, pbsmInfo: PBSMINFO) -> c_long ---
+	BroadcastSystemMessageExW :: proc(flags: DWORD, lpInfo: LPDWORD, Msg: UINT, wParam: WPARAM, lParam: LPARAM, pbsmInfo: PBSMINFO) -> c_long ---
+	BroadcastSystemMessageW :: proc(flags: DWORD, lpInfo: LPDWORD, Msg: UINT, wParam: WPARAM, lParam: LPARAM) -> c_long ---
+
+	// CalculatePopupWindowPosition
+	// CallMsgFilterA, CallMsgFilterW
+	CallNextHookEx :: proc(hhk: HHOOK, nCode: c_int, wParam: WPARAM, lParam: LPARAM) -> LRESULT ---
+	CallWindowProcA :: proc(lpPrevWndFunc: WNDPROC, hWnd: HWND, Msg: UINT, wParam: WPARAM, lParam: LPARAM) -> LRESULT ---
```
gingerBill, line 31: Please do no include any `*A` versions of the calls. We only accept the `*W` versions.

NicknEma:
No more *A versions of procedures, at least in user32.odin. I also finished adding all those stubs I left commented out before.

PR #3045, opened 2023-12-23
encoding/cbor

laytan:
Package cbor encodes, decodes, marshals and unmarshals types from/into RCF 8949 compatible CBOR binary.
Also provided are conversion to and from JSON and the CBOR diagnostic format.

- [RFC](https://www.rfc-editor.org/rfc/rfc8949.html)


Skytrias:
looking good, read through most of it and looks pretty similar to other un/marshal implementations :+1:

laytan:
@gingerBill Just rebased and updated to be in line with the updates on master, can I get a review on this at some point?

PR #3048, opened 2023-12-27
Fixing a pdb linker error when the path contains spaces

ThomasL81:
Building on Windows, where the build path and the linker path contain spaces, the following error occurs:

LINK : fatal error LNK1201: error writing to program database '<incomplete path>'; check for insufficient disk space, invalid path, or insufficient privilege

This pull request escapes the path with double quotes, which allows for a succesful build


flysand7:
What if the path contains quotes? I don't remember if windows allows double quotes in the filename

PR #3060, opened 2023-12-30
Fix glfw linker flags on linux

raphigaziano:
The generated linker flag was `lglfw` when it should be `lglfw3`.

I also went ahead and added an import for `system:X11`, assuming this would be required in most use cases.

That was enough to compile and run a simple opengl program on my system (Ubuntu 22.04 with glfw installed system wide). Other linux distributions may require additional imports.

I was able to confirm the bug and the fix by running `tests/vendors/glfw`.

### DIFF
--- a/vendor/glfw/bindings/bindings.odin
+++ b/vendor/glfw/bindings/bindings.odin
@@ -22,7 +22,10 @@ when ODIN_OS == .Windows {
 		}
 	}
 } else when ODIN_OS == .Linux {
-	foreign import glfw "system:glfw"
+	foreign import glfw {
+		"system:glfw3",
+		"system:X11",
+	}
 } else when ODIN_OS == .Darwin {
 	foreign import glfw { 
 		"../lib/darwin/libglfw3.a",


flysand7:
The current version of GLFW bindings was working for me (Arch Linux btw). `-lglfw` works as expected, and on the countrary `-lglfw3` does not.

Does `-lglfw` not work on Ubuntu?

Also, glfw imports x11 using dynamic vtables in order to support both x11 and wayland without requiring explicit hard-coded shared object dependencies. Not importing x11 in our bindings should just work, does it not work for you? I reckon it's a bad idea forcing importing x11 when it may not be installed on the systems that exclusively have wayland.

raphigaziano:
I suspected importing X11 might be overkill, however it was somehow still necessaray for me.

I reverted my change and ran the mentionned test again, here's the output:

```
raphi@raphi-MS-7821:~/dev/odin/odin-lang/tests/vendor$ ../../odin run glfw
/usr/bin/ld: cannot find -lglfw: No such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

If i just "fix" the `glfw` import i get:

```
raphi@raphi-MS-7821:~/dev/odin/odin-lang/tests/vendor$ ../../odin run glfw
/usr/bin/ld: cannot find -lglfw: No such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
raphi@raphi-MS-7821:~/dev/odin/odin-lang/tests/vendor$ ../../odin run glfw
/usr/bin/ld: /usr/local/lib/libglfw3.a(x11_init.c.o): in function `createKeyTables':
x11_init.c:(.text+0x668): undefined reference to `XkbGetMap'
/usr/bin/ld: x11_init.c:(.text+0x691): undefined reference to `XkbGetNames'
/usr/bin/ld: x11_init.c:(.text+0x912): undefined reference to `XkbFreeNames'
/usr/bin/ld: x11_init.c:(.text+0x92b): undefined reference to `XkbFreeKeyboard'
/usr/bin/ld: x11_init.c:(.text+0x954): undefined reference to `XDisplayKeycodes'
/usr/bin/ld: x11_init.c:(.text+0x990): undefined reference to `XGetKeyboardMapping'
/usr/bin/ld: x11_init.c:(.text+0xab8): undefined reference to `XFree'
... [a few more pages of undefined references errors]
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The problem may lie with my glfw installation. I compiled it from  source a few months ago, don't remember the details but i don't think i tried anything fancy which would explain the different flag (i do know i need `lglfw3` to compile it with c, though. Just checked it again and i get the same linking error).

raphigaziano:
Googling for glfw linking errors shows both flags being used.

The only reference i could find to the difference between them is at the very end of this page from GLFW's docs, where they say to add the 3 if the library was statically compiled: https://www.glfw.org/docs/latest/build.html

Although the layout seems to imply that this only applies to MacOs, i did check my version and it does seem to be static (`.a` extension). I'll try recompiling as a shared library and see what happens.

flysand7:
What is your output of

```
$ pkg-config --libs glfw3
```

?

raphigaziano:
Sorry, i saw your question after uninstalling and recompiling. Output is now

```
raphi@raphi-MS-7821:~$ pkg-config --libs glfw3
-L/usr/local/lib -lglfw
```
but since i just reinstalled the lib it doens't reflect what was wrong an hour ago.

It does work fine now that the lib is shared, though. I previously installed it with a simple `cmake, make, make install`, which does output a static library by default: https://www.glfw.org/docs/latest/compile.html#BUILD_SHARED_LIBS

You're right about privileging the system package install, which i assume builds the correct version. I'll close the pr and clean things up on my end.

PR #3061, opened 2023-12-31
Fix the declaration of SetJoystickCallback

### LINKED ISSUE #3059: vender:glfw calling SetJoystickCallback causes crash if game controller is (dis)connected
## Context
Odin: dev-2023-12:252fd0e92
        OS:   Windows 11 Home Basic (version: 22H2), build 22621.2861
        CPU:  13th Gen Intel(R) Core(TM) i5-13600K
        RAM:  32475 MiB
Using Playstation 4 controller Model: CUH-ZCT2U

## Expected Behavior
App should not crash when SetJoystickCallback is used to set a callback. It doesn't crash if I (dis)connect without setting a callback through this function. It also doesn't crash when running a basic C raylib application which sets the callback. I used this raylib example code to test: [core_input_gamepad_info.c](https://github.com/raysan5/raylib/blob/master/examples/core/core_input_gamepad_info.c) and when running that the controller connects and shows name and inputs without crashing.

## Current Behavior
Current behavior is if the SetJoystickCallback proc is called, even with an empty proc, (dis)connecting my controller causes the app to crash.

If I create a debug build with 'odin build . -debug' and run and do the (dis)connection. The app crashes after calling glfw.PollEvents() that frame with "Exception thrown at 0x000002CCFCFE5370 in src.exe: 0xC0000005: Access violation executing location 0x000002CCFCFE5370." and checking the call stack shows glfw.PollEvents -> _glfwPlatformPollEvents -> A number of windows dlls (user32, win32u, ntdll) -> glfwGetWin32Window ->_glfwDetectJoystickConnectionWin32 -> dinput8 dll -> _glfwTerminateJoysticksWin32 for the connection case, and a similar path for the disconnection case. 

### Steps to Reproduce
I believe this is minimum reproducible code, I get the crashing behavior when I run it locally and (dis)connect my controller.
```
package main

import "core:c"
import "vendor:glfw"
import "core:fmt"

main :: proc () {
	result := glfw.Init()
	if !result {
		fmt.println("glfw_init_error")
		return
	}

	window := glfw.CreateWindow(100, 100, "title_cstr", nil, nil)
	if window == nil {
		fmt.println("glfw_create_window_error")
		return
	}

	glfw.SetJoystickCallback(window, nil glfw.JoystickProc(JoystickCallback))

	for true {
		glfw.PollEvents()
	}
}

JoystickCallback :: proc (joystick, event: c.int) {

}
```

It's possible this bug is fixed in newer version of glfw, but I haven't updated a vendor lib before so I will try to do that tomorrow and update.

flysand7:
The declaration of SetJoystickCallback does [not take a `^Window` parameter](https://github.com/glfw/glfw/blob/b4c3ef9d0fdf46845f3e81e5d989dab06e71e6c1/src/input.c#L1245).

This PR fixes #3059

### DIFF
--- a/vendor/glfw/bindings/bindings.odin
+++ b/vendor/glfw/bindings/bindings.odin
@@ -181,7 +181,7 @@ foreign glfw {
 	SetCharCallback        :: proc(window: WindowHandle, cbfun: CharProc)        -> CharProc ---
 	SetCharModsCallback    :: proc(window: WindowHandle, cbfun: CharModsProc)    -> CharModsProc ---
 	SetCursorEnterCallback :: proc(window: WindowHandle, cbfun: CursorEnterProc) -> CursorEnterProc ---
-	SetJoystickCallback    :: proc(window: WindowHandle, cbfun: JoystickProc)    -> JoystickProc ---
+	SetJoystickCallback    :: proc(cbfun: JoystickProc)    -> JoystickProc ---
 
 	SetErrorCallback :: proc(cbfun: ErrorProc) -> ErrorProc ---
 }


Andidy:
Thanks for making the PR! But we probably shouldn't be committing our personal `ols.json` files and whatever 'bug' is to the main odin repo.

PR #3065, opened 2024-01-02
Removes macOS min version and supports default latest

Platin21:
This should make the macOS version automatically the target version so there should not be a possible linker verbose information which we get if the version is tinier than the one for which other libs are compiled for.


laytan:
An example of what the current behaviour does (just linking with liblua from `brew install lua`):

Issue #3044, opened 2023-12-22
`-lld` CLI Argument Ignored on Non-Windows Machine


pcleavelin:
## Context

Just to be clear up-front, I'm using the nix package manager to install Odin (which is currently broken, but with workarounds I am able to run the compiler). However, when linking any libraries (in my case the `vendor:raylib` package), it uses `ld` and fails to link some Objective-C symbols. I imagine this works fine through the normal & supported installation method of `brew`, but the issue isn't `ld` not working in my strange environment, its when running the build with the `-lld` argument:
```console
$ odin build <dir> -lld
```

The above command seems to ignore the argument on MacOS (and I think it will on linux as well). Looking at the source, in [src/linker.cpp:L103](https://github.com/odin-lang/Odin/blob/dev-2023-12/src/linker.cpp#L103), a check is made for the `-lld` arg, but only in the branch where the system in compiling on windows, further down in the non-windows systems branch, there is no check made for this argument, and thus `ld` is always the chosen linker when compiling with `clang`.

Since on my machine `lld` links correctly, I've currently made changes locally to make the check for the argument and pass `-fuse-ld=ld64.lld` to `clang` as a workaround.

Please forgive me if I'm missing something, as I'm not familiar with the codebase at all.

## System
```
Odin: dev-2023-12
OS:   macOS Sonoma (build: 23B74, kernel: 23.1.0) (M1 Pro)
CPU:  ARM64
RAM:  16384 MiB
```
## Expected Behavior

Passing `-lld` to `odin build` on MacOS uses `lld` as the linker.

## Current Behavior

On MacOS, `ld` is always used as the linker.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Run `odin build src/ -lld -v`

See that `ld` is used as the linker.

Platin21:
Actually we pass our flags just to clang for linking. But yeah this is not taken into account.

PR #3069, opened 2024-01-04
Respect `-lld` CLI arg on non-windows machines

pcleavelin:
I noticed that it doesn't seem like the `-lld` flag does anything on Linux & MacOS hosts. I just naively added a check and appended `fuse-ld=lld` to the `clang` args.

I'm not sure if this is the best way to do this, or if there are unintended side-effects of this change (such as changing the `system_exec_command_line_app` call from `ld-link` to `lld-link`).

For more context of how I found this, see #3044

### DIFF
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -516,7 +516,12 @@ gb_internal i32 linker_stage(LinkerData *gen) {
 			link_command_line = gb_string_append_fmt(link_command_line, " %.*s ", LIT(build_context.extra_linker_flags));
 			link_command_line = gb_string_append_fmt(link_command_line, " %s ", link_settings);
 
-			result = system_exec_command_line_app("ld-link", link_command_line);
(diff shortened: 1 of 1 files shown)


Platin21:
@Kelimion can be merged.. will see if that makes link times faster on linux/macOS as lld would be a bit faster then default old ld.

laytan:
Works nicely for me, around 250ms faster just trying it out on some projects

norgesSverd:
Glad to see this is getting some attention. New to Odin and I was looking for a way to choose which linker we wanted to use.

Was looking to see if someone already had the idea or not.

Instead of hard-coding a choice between LD & LLD, could we just instead have it accept -fuse-ld= arguments? So that way we can supply our own linker. I'd like to keep using mold, as I use it with my other c/rust projects? unless I am reading this PR wrong and that would be the intention then completely disregard my idiocy.

PR #3089, opened 2024-01-11
Fix for json marshaling of bit_set always returning .Unsupported_Type

karl-zylinski:
When JSON marshaling any kind of bit_set it always failed, this was due to it always returning .Unsupported_Type. Now it only fails if the bit_size is unsupported or if write_u64 fails.


karl-zylinski:
Hm. Seems like the unmarshaler can't handle bit_sets.

flysand7:
I think it wouldn't be too weird if bitsets were marshalled as arrays of strings, just throwing it in

karl-zylinski:
Yeah, I think the code that exists in the marshaler was disabled because there was no unmarshal code and after digging around a bit in this I think you're right @flysand7 , it should be an array of strings

karl-zylinski:
Closing this for now as I don't think this is the right solution

PR #3092, opened 2024-01-13
cpu_features - fix shift direction

jon-lipstate:
The current shift direction was backwards, as such the flags dont ever come back with any information. Now they present correctly, for example, my machine produces `CPU_Features{aes, adx, avx, avx2, bmi1, bmi2, fma, os_xsave, pclmulqdq, popcnt, rdrand, rdseed, sse2, sse3, ssse3, sse41, sse42}`.

### DIFF
--- a/core/sys/info/cpu_intel.odin
+++ b/core/sys/info/cpu_intel.odin
@@ -37,11 +37,11 @@ cpu_name:     Maybe(string)
 
 @(init, private)
 init_cpu_features :: proc "c" () {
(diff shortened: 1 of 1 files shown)


flysand7 (approved):
The shift direction was probably correct, though someone did confuse the order of `value` and the `hwc` values. Perhaps what was meant is

```
value & (1<<hwc) != 0
```

Renaming `hwc` to `bit` makes it slightly more understandable what's going on.

PR #3100, opened 2024-01-16
Draft: Add freestanding aarch64 target

codename-irvin:
* Added target metrics for freestanding aarch64 target
* Only use "kernel" code model for freestanding x86 targets - "kernel" is not supported by LLVM for aarch64


codename-irvin:
Closed because I found some issues that need resolving. Will re-open once resolved.

PR #3102, opened 2024-01-16
Simplify append_elem and append_elems

radekm:
- It seems that `when size_of(E) != 0` was redundant.
- I replaced `max(8, 1)` by `8` even though we lose similarity of `append_elem` with `append_elems`.


gingerBill:
The `when size_of(E) != 0` is not redundant and necessary.

radekm:
I thought that it's redundant because it's inside `else` branch of `when size_of(E) == 0` where `size_of(E) != 0` should always be the case.

flysand7:
They are indeed redundant but bill likes them there

PR #3108, opened 2024-01-17
Add WINDOWS_SUBSYSTEM constant

Kelimion:
true when -subsystem:windows for Windows targets, false otherwise.

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -973,6 +973,8 @@ gb_internal void init_universal(void) {
 	add_global_string_constant("ODIN_ROOT",    bc->ODIN_ROOT);
 	add_global_string_constant("ODIN_BUILD_PROJECT_NAME", bc->ODIN_BUILD_PROJECT_NAME);
 
+	add_global_bool_constant("WINDOWS_SUBSYSTEM", bc->use_subsystem_windows);
+
 	{
 		GlobalEnumValue values[TargetOs_COUNT] = {
 			{"Unknown",      TargetOs_Invalid},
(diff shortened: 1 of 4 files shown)


flysand7:
I think this should be an enum, not a bool.

image

There's more subsystems than just two, though what's true is that users will generally care about just two of them -- Console and Windows. With that I would like to argue that `Efi` is also a subsystem with viable use-case that can surface up when developing EFI bootloaders (OS development stuff).

Even though this value will be unchecked in anything other than console and windowed applications, I still think it's a good idea for consistency's sake..?

Kelimion:
> There's more subsystems than just two, though what's true is that users will generally care about just two of them -- Console and Windows. With that I would like to argue that `Efi` is also a subsystem with viable use-case that can surface up when developing EFI bootloaders (OS development stuff).
> 
> Even though this value will be unchecked in anything other than console and windowed applications, I still think it's a good idea for consistency's sake..?

All of them are supported (by main.cpp's flag parser). Subsystem:Windows and not Subsystem:Windows. ;-)

src/checker.cpp
```diff
@@ -973,6 +973,8 @@ gb_internal void init_universal(void) {
 	add_global_string_constant("ODIN_ROOT",    bc->ODIN_ROOT);
 	add_global_string_constant("ODIN_BUILD_PROJECT_NAME", bc->ODIN_BUILD_PROJECT_NAME);
 
+	add_global_bool_constant("WINDOWS_SUBSYSTEM", bc->use_subsystem_windows);
```
gingerBill, line 976: It makes more sense to do this:
```
add_global_string_constant("ODIN_WINDOWS_SYBSYSTEM", the_string);
```

gingerBill (changes requested):

src/checker.cpp
gingerBill: This also makes it work on non-Windows systems by having it be an empty string constant

Kelimion: Sensible.

PR #3111, opened 2024-01-18
Remove `distinct` from types in `linalg_glsl.odin`

thetarnav:
Since `distinct` has been removed from `math/linalg/specifig.odin` (https://github.com/odin-lang/Odin/commit/b408ec6bac9ba417faf51a2364542e5d9493e193)
And here there is no `distinct` either: https://github.com/odin-lang/Odin/blob/master/core/math/noise/opensimplex2.odin#L12
Why not remove them in `math/linalg/glsl/linalg_glsl.odin` as well?


gingerBill:
No. This is done on purpose in this case. `math/linalg` was a little more generic, compared to `glsl` and `hlsl`. I will keep these being `distinct`.

PR #3113, opened 2024-01-18
Add few mat3 procs to `linalg_glsl.odin`

thetarnav:
Not sure if procs for mat3 operations were omitted on purpose, but that's what I used for some 2d transforms.

### DIFF
--- a/core/math/linalg/glsl/linalg_glsl.odin
+++ b/core/math/linalg/glsl/linalg_glsl.odin
@@ -1357,6 +1357,76 @@ not :: proc{
 @(require_results) identity :: proc "c" ($M: typeid/matrix[$N, N]$T) -> M { return 1 }
 
 @(require_results)
+mat3Translate :: proc "c" (v: vec2) -> mat3 {
+	return {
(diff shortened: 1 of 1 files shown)


core/math/linalg/glsl/linalg_glsl.odin
```diff
@@ -1356,6 +1356,76 @@ not :: proc{
...
+@(require_results)
+mat3Rotate :: proc "c" (radians: f32) -> mat3 {
+	c := cos(radians)
+	s := sin(radians)
+	return {
+		 c, s, 0,
+		-s, c, 0,
+		 0, 0, 1,
+	}
+}
```
gingerBill, line 1384: You've got the -ve sign on the wrong sine components, and this only rotates around one axis. 

This needs a better naming scheme to make it clear it is actually for 2D + 1 homogenous coordinate.

gingerBill (changes requested):
This appears to for 2D + 1 homogenous coordinate stuff, rather than just 3D stuff. I am not sure if is a good idea or not, at least with names.

Issue #3112, opened 2024-01-18
[fmt] left padding of numbers should default to space instead of 0


cridlig:
## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System & Odin Version: MacOS Sonoma on ARM64
* Please paste `odin report` output: dev-2024-01:5961d4b3

## Expected Behavior

fmt.printf("%4d", 92) should print the integer 92 left padded with two spaces

## Current Behavior

What is the current behavior? fmt.printf("%4d", 92) prints 0092

programandala-net:
I just found this problem in a program of mine: the default padding character is zero, unless the space flag is used:

```
fmt.printf("%4d", 92)  // prints "0092"
fmt.printf("%04d", 92) // prints "0092"
fmt.printf("% 4d", 92) // prints "  92" (with 2 leading spaces)
```

Both behaviours (the default padding character and the effect of the space flag) contradict the documentation of the fmt verb flags (in file <core/fmt/doc.odin>):

>    ' '    (space) leave a space for elided sign in numbers (% d)
 >   0      pad with leading zeros rather than spaces

PR #3115, opened 2024-01-18
issue#3112: default to spaces for left padding of % format

cridlig:
Fix issue #3112 [fmt] left padding of numbers should default to space instead of 0

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -439,7 +439,7 @@ wprintf :: proc(w: io.Writer, fmt: string, args: ..any, flush := true) -> int {
 	was_prev_index := false
 
 	loop: for i := 0; i < end; /**/ {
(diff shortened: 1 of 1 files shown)


Kelimion:
`fmt.printf("Hellope '% 4d'\n", 42)` already pads using spaces, so I'll close the PR but leave open the issue because I do feel that `core:fmt`'s padding could be improved.

What I think we should do is a bit more involved:
1) Replace `fi.zero: bool` and `fi.space` with `fi.pad_char: rune`, which if rune(0) means no padding, the default. Maybe even an `fi.pad: Padding_Options` for more flexibility.
2) Update `fmt_write_padding` to honor this.
3) See what we can do about precision prefixes to specify the padding character, now `%4d` (zero) and `% 4d` (space), to better document and make more intuitive.

PR #3121, opened 2024-01-20
core:odin parsing fixes

RilleP:
Fixes:

```
foo(1,
      2,
      3 // New line without trailing comma
)
// Same fix for structs
t: Type = {
	1,
	2,
	3
}

// semicolons for this case
if x  do return y;
else     do return z;
```


laytan:
Hey, thanks! I tested this yesterday and it seems to work great, unfortunately there is a conflict, could you resolve them?

Kelimion:
image
Once you resolve this I'll be able to merge it.

PR #3124, opened 2024-01-22
update LLVM to 17 for CI/releases

laytan:
~~I left the actual CI on the original version as that might make sense for coverage, let me know if you want those updated too.~~

I confirmed that this builds and uses the right version on my fork.


laytan:
There is also an issue with the current builds on Linux, it uses part of llvm-14 and part of llvm-11 to build/distribute.

I confirmed that this is not the case for this PR, so this is also a bug fix.

PR #3125, opened 2024-01-22
sys/linux: Add binding to ioctl syscall + standard fd constants.

marcs-feh:
Add binding to ioctl syscall, due to the vast nature of this syscall adding more device specific request values as constants is possible in the future, I opted to not add them.

Also added the stdin, stdout and stderr, to `constants.odin` for convenience sake when using this syscall


core/sys/linux/sys.odin
```diff
@@ -216,7 +216,17 @@ rt_sigprocmask :: proc "contextless" (mask_kind: Sig_Mask_Kind, new_set: ^Sig_Se
...
 
-// TODO(flysand): ioctl
+/*
+	Control devices. The ioctl syscall is a bit special because
+	its argument is usually a pointer to some driver-specific structure.
+	The request value is device-specific. Consult your LibC implementation's
+	ioctls.h file to learn more.
+	Available since Linux 1.0.
+*/
+ioctl :: proc "contextless" (fd: Fd, request: u64, arg: u64) -> (Errno) {
```
flysand7, line 226: The prototype seems to be wrong. I do not advice using man(2) for figuring out the prototype of syscall functions as man only documents libc, and rarely the underlying kernel implementation.

See [linux source code for ioctl](https://elixir.bootlin.com/linux/latest/source/fs/ioctl.c#L857)  to see the actual prototype for this syscall.

In this case `request` should have type `i32`, and arg should be the size of `uint` or `uintptr`.

Also `u64` and `i64` should probably not appear in parameters for the syscalls, as this won't necessarily apply to 32-bit targets. Syscalls only take parameters in registers, and on 32-bit platforms only have up to 32-bit GP register size.

core/sys/linux/sys.odin
```diff
@@ -216,7 +216,17 @@ rt_sigprocmask :: proc "contextless" (mask_kind: Sig_Mask_Kind, new_set: ^Sig_Se
...
+	ret := syscall(SYS_ioctl, fd, request, arg)
+	return Errno(-ret)
```
flysand7, line 228: Yeah this is one of the reasons I didn't want to do the syscall. Some of ioctl commands return a value, while others don't. I don't feel like bothering making enums and proc groups for this particular syscall is going to be worth it. Maybe splitting it up into groups like [`ioctl_console`](https://man7.org/linux/man-pages/man2/ioctl_console.2.html) and [`ioctl_fat`](https://man7.org/linux/man-pages/man2/ioctl_fat.2.html) (like man pages do) is a better strategy for making the API somewhat sane. It's a ton of work though

marcs-feh: The function signature has been adjusted, I forgot that `man` documentation is straight up wrong sometimes.

The problem with `ioctl` is that it is designed to be extremely flexible, I'm not sure if there is a saner way besides splitting it into many groups and copying all the request codes, and it would still not be complete as some calls are vendor-specific. The main reason I introduced this binding was to do some terminal IO that couldn't be done with just POSIX bindings.

Beefster09:
ayyyy

This will make a number of useful functions in a future core:terminal possible. isatty relies on ioctl calls and other things that detect terminal capabilities on *nix

flysand7:
Make `request` parameter a `u32` and it's good to go.

IllusionMan1212:
I would like to see this merged @marcs-feh if you've got some spare time.

core/sys/linux/sys.odin
```diff
@@ -217,7 +217,18 @@ rt_sigprocmask :: proc "contextless" (mask_kind: Sig_Mask_Kind, new_set: ^Sig_Se
...
-// TODO(flysand): ioctl
+/*
+	Control devices. The ioctl syscall is a bit special because
+	its argument is usually a pointer to some driver-specific structure.
+	The request value is device-specific. Consult your LibC implementation's
+	ioctls.h file to learn more. The returned value of ioctl *may* be an error
+	code value instead of a memory address depending on the request type.
+	Available since Linux 1.0.
+*/
+ioctl :: proc "contextless" (fd: Fd, request: i32, arg: uintptr) -> (uintptr) {
```
laytan, line 228:
```suggestion
ioctl :: proc "contextless" (fd: Fd, request: u32, arg: uintptr) -> (uintptr) {
```

PR #3135, opened 2024-01-24
Add support for basic EGL on Linux

colrdavidson:


vendor/egl/egl.odin
```diff
@@ -0,0 +1,61 @@
+//+build linux
+package egl
+
+NativeDisplayType :: rawptr
+NativeWindowType  :: rawptr
+Display :: rawptr
+Surface :: rawptr
+Config  :: rawptr
+Context :: rawptr
```
gingerBill, line 9: Do these need to be `distinct` or not?

colrdavidson: Probably should. I will sort this out this weekend

PR #3136, opened 2024-01-24
core:crypto/hash - Add and refactor

Yawning:
There is a lot of code duplicated in convenience methods in each hash implementation, and having a generic hash type makes implementing higher-level constructs such as HMAC significantly easier down the road.  Additionally while this code is getting worked on, some improvements to the init/update/final interface can be made.

- Low level API changes
    - "init, update, final" changed to "init, update, final, clone, reset"
    - `reset` will sanitize the internal state of the hash instance
    - `final` now has the option of internally calling clone to easily compute rolling digests
    - `final` will implicitly call `reset`
    - When multiple variants share the same package  (eg: SHA-256, SHA-512) they now have dedicated initialization routines (that MUST be used).
- Add `core:crypto/hash`
    - Low level interface for every hash algorithm with 1 Context data type.
    - Convenience methods moved from being duplicated in each hash implementation to work on `hash.Context`
- Implementation changes
    - SHAKE is no longer exposed as a hash, but instead exposes an API more suited to an XOF.
    - The BLAKE2 initialization no longer allocates scratch space.
    - The BLAKE2 Contexts no longer stores a copy of the config structure.
    - hash_file no longer leaks the buffer if the entire file is loaded at once.
- Test improvements
    - Exercise a good amount of the high-level hash API.
    - Modernize the test harness a bit.
- Documentation changes
    - Relocated the hash function/api documentation to `core:crypto/hash` from the README.md.

- Bonus round:
    - [x] Add `crypto/hmac`


core/crypto/hash/low_level.odin
```diff
@@ -0,0 +1,287 @@
...
+	SHA3_512,
+	SHAKE_128,
+	SHAKE_256,
+	SM3,
+	Insecure_MD5,
+	Insecure_SHA1,
+}
+
+// ALGORITHM_NAMES is the Algorithm to algorithm name string map.
+ALGORITHM_NAMES := map[Algorithm]string {
```
Kelimion, line 31: This could just be an enumerated array, why does this need to be a `map`?
`ALGORITHM_NAMES := [Algorithm]string { ... }`

core/crypto/hash/low_level.odin
```diff
@@ -0,0 +1,287 @@
...
+	.SHA3_512      = "SHA3-512",
+	.SHAKE_128     = "SHAKE-128",
+	.SHAKE_256     = "SHAKE-256",
+	.SM3           = "SM3",
+	.Insecure_MD5  = "MD5",
+	.Insecure_SHA1 = "SHA-1",
+}
+
+// DIGEST_SIZES is the Algorithm to digest size map.
+DIGEST_SIZES := map[Algorithm]int {
```
Kelimion, line 50: Likewise seems like this could be `DIGEST_SIZES := [Algorithm]int{ ... }` instead of a `map`.

Yawning:
Note: The choice to use reflection to specialize the union is because each hash function implementation's context struct is moderately big (500+ bytes depending on the algorithm), and a clear + copy, only to immediately overwrite everything via `init` wastes CPU cyces.

core/crypto/blake2b/blake2b.odin
```diff
@@ -6,122 +13,47 @@ package blake2b
...
-	hash_stream,
-	hash_file,
-	hash_bytes,
-	hash_string,
-	hash_bytes_to_buffer,
-	hash_string_to_buffer,
+// final finalizes the Context, writes the digest to hash, and calls
+// reset on the Context.
+//
+// Iff finalize_clone is set, final will work on a copy of the Context,
```
Skytrias, line 44: **iff**

core/crypto/blake2s/blake2s.odin
```diff
@@ -6,122 +13,47 @@ package blake2s
...
-	_blake2.update(ctx, data)
+// clone clones the Context other into ctx.
+clone :: proc(ctx, other: ^Context) {
+	_blake2.clone(ctx, other)
 }
 
-final :: proc(ctx: ^Context, hash: []byte) {
-	_blake2.final(ctx, hash)
+// reset sanitizes the Context.  The Context must be re-initialzed to
+// be used again.
```
Skytrias, line 56: initialzed -> initialized

Skytrias (approved):
good refactor 👍 a few copy pasted typos but fine

core/crypto/blake2b/blake2b.odin
Yawning: https://mathworld.wolfram.com/Iff.html

core/crypto/blake2s/blake2s.odin
Yawning: Fixed, ty.

PR #3137, opened 2024-01-25
Add flags to show/export defineable values and warn if a -define is unused in the project

laytan:


src/main.cpp
```diff
@@ -1428,6 +1452,106 @@ gb_internal void timings_export_all(Timings *t, Checker *c, bool timings_are_fin
...
+	for_array(i, c->info.defineables) {
+		Defineable *def = &c->info.defineables[i];
+		gb_fprintf(&f,"%.*s,%.*s,%.*s\n", LIT(def->name), LIT(def->default_value_str), LIT(def->pos_str));
+	}
+}
+
+gb_internal void show_defineables(Checker *c) {
+	int max_name_len    = strlen("Defineable");
+	int max_default_len = strlen("Default Value");
+	int max_pos_len     = strlen("Location");
```
gingerBill, line 1527:
```c
	isize max_name_len    = gb_strlen("Defineable");
	isize max_default_len = gb_strlen("Default Value");
	isize max_pos_len     = gb_strlen("Location");
```

laytan: Thanks, resolved!

DragosPopse:
`-show-defines` and `-export-defines`?:)

laytan:
To me, a define is something that is defined, and a defineable something you can define.
I do agree it is not a very ergonomic name though.

laytan:
I've rebased and improved the PR. Now it also collects doc comments tied to the declaration if available, this also needed a change to the way it is shown because those can be multiline.

Here is what it looks like now when I added `-show-defineables` to a `odin test` invocation:
Screenshot 2024-06-07 at 14 54 30

I am open to ideas to improve that.

And maybe the name should be `-show-defineable` instead of with the `s` added, or an entirely different name?

Feoramund:
> And maybe the name should be `-show-defineable` instead of with the `s` added, or an entirely different name?

`-show-config`?

I only just now became aware of this PR, otherwise I would've chimed in earlier. This is a nice feature though.

PR #3138, opened 2024-01-25
Add a testing.expectf proc as a way to avoid a tprintf call.

FourteenBrush:


core/testing/testing.odin
```diff
@@ -91,6 +91,14 @@ expect :: proc(t: ^T, ok: bool, msg: string = "", loc := #caller_location) -> bo
 	}
 	return ok
 }
+
+expectf :: proc(t: ^T, ok: bool, format: string, args: ..any, loc := #caller_location) -> bool {
+	if !ok {
+		errorf(t, format, args, loc=loc)
```
gingerBill, line 97: This needs to be `..args`

PR #3139, opened 2024-01-25
sys/windows: Make `INVALID_FILE_ATTRIBUTES` a `u32` constant

flysand7:
The [GetFileAttributes](http://msdn.microsoft.com/en-us/library/aa364944%28v=vs.85%29.aspx) function returns a DWORD, which introduces a compiler error when trying to directly compare the result of that function with the `INVALID_FILE_ATTRIBUTES` constant, suggesting that the constant has a wrong type.


gingerBill:
Please test things before making a PR.

flysand7:
Done. Linux failure looks like a fluke

PR #3143, opened 2024-01-27
Adding linux & darwin makefile for cgltf

Pariatech:
Hello I wanted to use the vendor binding for cgltf and I found out that it was made only for Windows. 
So I copypasted the Makefile from the stb vendor and tweaked it a little bit to build cgltf for linux and darwinj.  And I also changed cgltf.odin to use the generated library file for those systems. 

Thank you and have a nice day.

### DIFF
--- a/vendor/cgltf/src/Makefile
+++ b/vendor/cgltf/src/Makefile
@@ -1,0 +1,20 @@
+OS=$(shell uname)
+
+ifeq ($(OS), Darwin)
+all: darwin
+else
+all: unix
+endif
+
+unix:
+	mkdir -p ../lib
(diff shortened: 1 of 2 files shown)


vendor/cgltf/src/Makefile
```diff
@@ -0,0 +1,20 @@
...
+endif
+
+unix:
+	mkdir -p ../lib
+	$(CC) -c -O2 -Os -fPIC cgltf.c 	
+	$(AR) rcs ../lib/cgltf.a        cgltf.o
+	rm *.o
+
+darwin:
+	mkdir -p ../lib
```
laytan, line 16:
```suggestion
	mkdir -p ../lib/darwin
```

PR #3147, opened 2024-01-28
`base` library collection work

gingerBill:
* Add `base` library collection
* Move packages
    * Move `core:runtime` to `base:runtime`
    * Move `core:intrinsics` to `base:intrinsics`
    * Move `core:builtin` to `base:builtin`
    * The above are still accessible through `core:*`
* Remove any dependency on `core:os` from `base:runtime`
    * Remove cyclic import hack that allowed for this which makes all import trees strict DAGs
* Implement `heap_allocator` in `base:runtime`, moved from `core:os`
* Unified `heap_allocator` code into one place into `base:runtime`
* Move `builtin` procedures for matrices to `intrinsics`
* Move `@(builtin)` procedures for matrices to:
    * `core:math/linalg`
    * `core:math/linalg/glsl`
    * `core:math/linalg/hlsl`
* `-default-to-panic-allocator` flag, with `DEFAULT_TO_PANIC_ALLOCATOR` constant
* `-default-to-panic-allocator` on wasm targets
* `-default-to-nil-allocator` on freestanding targets
    * Not all freestanding targets may have traps/panics available


colrdavidson (approved):
I was here for this one :P
Looks good to me

PR #3156, opened 2024-01-31
Make spall able to be used in instrumentation procedures

laytan:
This is a draft for now as it is untested on Windows and Linux.

There is also a bug in the instrumentation feature where it does not do anything when in `-debug` mode.

It is a huge pain to work with the instrumentation because basically anything you call or do causes a segfault because of recursion/stack overflow if you are not careful.
This is also why I had to copy/implement the some procedures into the spall package, and why most of these things are #no-bounds-check.

I think adding the `//+no-instrumentation` to the sys/windows and sys/linux packages is a good compromise, otherwise this would be even more of a pain.

Screenshot 2024-01-31 at 21 38 20

### DIFF
--- a/core/prof/spall/doc.odin
+++ b/core/prof/spall/doc.odin
@@ -1,39 +1,38 @@
 /*
-import "core:prof/spall"
+	import "core:prof/spall"
 
-spall_ctx: spall.Context
-spall_buffer: spall.Buffer
+	spall_ctx: spall.Context
+	spall_buffer: spall.Buffer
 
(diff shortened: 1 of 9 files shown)


laytan:
Tested and works nicely on linux, just need to test windows and implement the bsd's

laytan:
Just pushed a fix for Windows and it works correctly there now too, so I am putting this on ready for review.

@colrdavidson already looked it over briefly and didn't see anything wrong.

PR #3161, opened 2024-02-02
fix -no-crt on Darwin

laytan:
Recent changes made `base:runtime` depend on crt again. This PR fixes that.

Before:
```
$ odin run test.odin -file -no-crt -default-to-nil-allocator
Hellope!
$ otool -L ./test
./test:
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1336.61.1
```

After:
```
$ odin run test.odin -file -no-crt -default-to-nil-allocator
Hellope!
$ otool -L ./test
./test:
```

### DIFF
--- a/base/runtime/os_specific_darwin.odin
+++ b/base/runtime/os_specific_darwin.odin
@@ -2,25 +2,14 @@
 //+private
 package runtime
 
-foreign import libc "system:System.framework"
-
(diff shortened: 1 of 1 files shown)


gingerBill:
Is this "stable" to do some on Mac?

PR #3186, opened 2024-02-11
copy_file implemented for Windows, Linux and Darwin

Smilex:
I am unable to test it on darwin, but I assumed it's the same as Linux, so I copied it

Linux tested with WSL
Windows tested on Windows 11

### DIFF
--- a/core/os/file_windows.odin
+++ b/core/os/file_windows.odin
@@ -612,3 +612,14 @@ pipe :: proc() -> (r, w: Handle, err: Errno) {
 	return
 }
 
+copy_file :: proc(src: string, dst: string, overwrite: bool = false) -> Errno {
+	wide_src := win32.utf8_to_wstring(src)
+	wide_dst := win32.utf8_to_wstring(dst)
+	fail_if_exists : win32.BOOL = !overwrite
+	ok := win32.CopyFileW(wide_src, wide_dst, fail_if_exists)
+	if !ok {
+		return Errno(win32.GetLastError())
+	}
+
+	return ERROR_NONE
+}
--- a/core/os/os_darwin.odin
+++ b/core/os/os_darwin.odin
@@ -1118,3 +1118,55 @@ fcntl :: proc(fd: int, cmd: int, arg: int) -> (int, Errno) {
 	}
 	return int(result), ERROR_NONE
 }
+
+copy_file :: proc(src: string, dst: string, overwrite: bool = false) -> Errno {
+	runtime.DEFAULT_TEMP_ALLOCATOR_TEMP_GUARD()
+	src_handle, src_err := open(src)
+	if src_err != ERROR_NONE {
+		return src_err
+	}
+
+	flags := O_WRONLY | O_CREATE
+	if !overwrite {
+		flags |= O_EXCL
+	}
+
+	dst_handle, dst_err := open(dst, flags, 0666)
+	if dst_err != ERROR_NONE {
+		close(src_handle)
+		return dst_err
+	}
+
+	buf, buf_err := make([]u8, 4096, context.temp_allocator)
+	if buf_err != .None {
+		close(src_handle)
+		close(dst_handle)
+		return ENOMEM
+	}
+
+	for {
+		bytes_read, read_err := read(src_handle, buf)
+		if read_err != ERROR_NONE {
+			return read_err
+		} else if (bytes_read == 0) {
+			break;
+		}
+
+		total_written := 0
+		for total_written < bytes_read {
+			bytes_written, write_err := write(dst_handle, buf[total_written:bytes_read])
+			if write_err != ERROR_NONE {
+				close(src_handle)
+				close(dst_handle)
+				return write_err
+			}
+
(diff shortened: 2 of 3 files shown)


core/os/os_darwin.odin
```diff
@@ -1118,3 +1118,50 @@ fcntl :: proc(fd: int, cmd: int, arg: int) -> (int, Errno) {
...
+		bytes_read, read_err := read(src_handle, buf)
+		if read_err != ERROR_NONE {
+			return read_err
+		} else if (bytes_read == 0) {
+			break;
+		}
+
+		total_written := 0
+		for total_written < bytes_read {
+			bytes_written, write_err := write(dst_handle, buf[total_written:bytes_read])
```
Tetralux, line 1156: Might have meant `buf[total_written:][:bytes_read]`.

core/os/file_windows.odin
```diff
@@ -612,3 +612,14 @@ pipe :: proc() -> (r, w: Handle, err: Errno) {
 	return
 }
 
+copy_file :: proc(src: string, dst: string, overwrite: bool = false) -> Errno {
+	wide_src := win32.utf8_to_wstring(src)
```
Tetralux, line 616:
```suggestion
wide_src := win32.utf8_to_wstring(src, context.temp_allocator)
```
Same for `wide_dst` just below.
The guard can be added here too.

core/os/os_linux.odin
```diff
@@ -1144,3 +1144,50 @@ ppoll :: proc(fds: []pollfd, timeout: ^unix.timespec, sigmask: ^sigset_t) -> (in
...
+		bytes_read, read_err := read(src_handle, buf)
+		if read_err != ERROR_NONE {
+			return read_err
+		} else if (bytes_read == 0) {
+			break;
+		}
+
+		total_written := 0
+		for total_written < bytes_read {
+			bytes_written, write_err := write(dst_handle, buf[total_written:bytes_read])
```
Tetralux, line 1182: Same deal here.

core/os/file_windows.odin
```diff
@@ -612,3 +612,14 @@ pipe :: proc() -> (r, w: Handle, err: Errno) {
...
+	wide_dst := win32.utf8_to_wstring(dst)
+	fail_if_exists : win32.BOOL = !overwrite
```
Tetralux, line 618: This might need a cast(?)
```suggestion
fail_if_exists := win32.BOOL(!overwrite)
```

Tetralux (changes requested):
Few things to fix up, otherwise decent.

core/os/os_darwin.odin
Smilex: I want to write from buf + total_written to bytes_read
Isn't my way correct?

Kelimion: Looks correct.

core/os/os_linux.odin
Kelimion: No, it looks correct.

It tries to read 4k at a time and `bytes_read` is how many it actually read.
Then `total_written` is reset to `0` each time it tries to write that 4k block, so Tetra's proposal would cause the `buf` slice to go out of bounds.

Kelimion:
It would be nice if on *nix platforms it'd copy permissions (chmod) and ownership (chown) instead of assigning the new file to the current user with some default permissions, or being able to pass in the desired owner and permission.

Smilex:
> It would be nice if on *nix platforms it'd copy permissions (chmod) and ownership (chown) instead of assigning the new file to the current user with some default permissions, or being able to pass in the desired owner and permission.

I did stat for and copy the permissions at first, but decided to simplify
I'm not sure which I prefer

Smilex:
> It would be nice if on *nix platforms it'd copy permissions (chmod) and ownership (chown) instead of assigning the new file to the current user with some default permissions, or being able to pass in the desired owner and permission.

Should I make it possible to pass chmod and chown?

flysand7:
For a simple implementation copying the original permissions should be enough

Smilex:
> For a simple implementation copying the original permissions should be enough

Ok, now I stat for the mode first and open the new file with that mode

dvrd:
I tested the implementation on MacOS and it works as expected

Smilex:
> I tested the implementation on MacOS and it works as expected

thanks!

Yawning:
Some thoughts:
- The Linux and Darwin code could be simplified by using `os/stream_from_handle` and `io/copy`.
- Should failure during the copy process, remove the partially copied file?  Right now it does.

laytan:
You probably also want `os.O_TRUNC` for the destination file, so it truncates the file if it already exists, otherwise it would just add to it

Tetralux:
> Some thoughts:
> 
> * The Linux and Darwin code could be simplified by using `os/stream_from_handle` and `io/copy`.
> * Should failure during the copy process, remove the partially copied file?  Right now it does.

It currently __doesn't__ remove a partially copied file - at least not explicitly.

PR #3188, opened 2024-02-12
[os] Fix os.change_directory(), os.set_current_directory()

Tetralux:
- set_current_directory() was leaking the wstring path string because it wasn't using the temp allocator.
- change_directory() wasn't calling SetCurrentDirectoryW() under the lock!

There didn't seem to be any actual difference in behaviour between these two procedures, so I've merged them.


Tetralux:
Ready

PR #3197, opened 2024-02-14
Update miniaudio to 0.11.21

laytan:
Also fixes wrong alignment for the pthread types for darwin, the headers don't define alignment on darwin, only on linux.


laytan:
I will compile a new .lib now, thought I already added that but not in diff

raphaelmor:
For what it's worth I was able to test this PR and it fixed my miniaudio issues (both dlopening the framework, and the corrupted device playback info).

PR #3199, opened 2024-02-15
Add into_dynamic_soa, unordered_remove_soa, and ordered_remove_soa

JamesDSource:
Added 3 more procedures to help make #soa[]T and #soa[dynamic]T more versatile.

### DIFF
--- a/base/runtime/core_builtin_soa.odin
+++ b/base/runtime/core_builtin_soa.odin
@@ -446,7 +446,7 @@ into_dynamic_soa :: proc(array: $T/#soa[]$E) -> #soa[dynamic]E {
 	}
 
 	array := array
-	dynamic_data := ([^]rawptr])(&d)[:field_count]
+	dynamic_data := ([^]rawptr)(&d)[:field_count]
 	slice_data   := ([^]rawptr)(&array)[:field_count]
 	copy(dynamic_data, slice_data)
 
--- a/base/runtime/core_builtin_soa.odin
+++ b/base/runtime/core_builtin_soa.odin
@@ -428,7 +428,7 @@ clear_soa :: proc{
 }
 
 // Converts soa slice into a soa dynamic array without cloning or allocating memory
-@(builtin, require_results)
+@(require_results)
 into_dynamic_soa :: proc(array: $T/#soa[]$E) -> #soa[dynamic]E {
 	d: #soa[dynamic]E
 	footer := raw_soa_footer_dynamic_array(&d)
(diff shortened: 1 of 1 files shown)


base/runtime/core_builtin_soa.odin
```diff
@@ -425,4 +425,101 @@ clear_soa_dynamic_array :: proc(array: ^$T/#soa[dynamic]$E) {
...
+	dynamic_data := uintptr(&d)
+	slice_data := uintptr(&array)
+	for _ in 0..<field_count {
+		(^uintptr)(dynamic_data)^ = (^uintptr)(slice_data)^
+		dynamic_data += size_of(rawptr)
+		slice_data += size_of(rawptr)
+	}
+
+	return d
+}
```
gingerBill, line 458: I don't think should be `builtin` whatsoever. The equivalent isn't for "aos" stuff isn't available as being `builtin`.

base/runtime/core_builtin_soa.odin
```diff
@@ -425,4 +425,101 @@ clear_soa_dynamic_array :: proc(array: ^$T/#soa[dynamic]$E) {
...
+	}
+
+	array := array
+	dynamic_data := uintptr(&d)
+	slice_data := uintptr(&array)
+	for _ in 0..<field_count {
+		(^uintptr)(dynamic_data)^ = (^uintptr)(slice_data)^
+		dynamic_data += size_of(rawptr)
+		slice_data += size_of(rawptr)
+	}
```
gingerBill, line 455: This could have been written like this:
```odin
dynamic_data := ([^]rawptr)(&d)[:field_count]
slice_data   := ([^]rawptr)(&array)[:field_count]
copy(dynamic_data, slice_data)
```

base/runtime/core_builtin_soa.odin
JamesDSource: That's a good point. Do you think it should still be in the runtime package then, or would it be better suited for "core:slice"?

base/runtime/core_builtin_soa.odin
JamesDSource: I like that better too. I'll update it to that in a moment.

base/runtime/core_builtin_soa.odin
gingerBill: Maybe a different package. I am not sure yet.

JamesDSource: Okay, I'll keep it in runtime for now. In the future I would support a "core:soa". I actually was thinking about trying to contribute a sort_soa sometime. Combined with soa_zip it might be a pretty good alternative to using sort_with_indices.

PR #3201, opened 2024-02-16
core/container/avl: Add an AVL tree implementation

Yawning:
Also some minor core/test cleanups, primarily for Lenooks.

### DIFF
--- a/core/container/avl/avl.odin
+++ b/core/container/avl/avl.odin
@@ -1,0 +1,678 @@
+/*
+package avl implements an AVL tree.
+
+The implementation is non-intrusive, and non-recursive.
+*/
+package container_avl
+
+import "base:intrinsics"
+import "base:runtime"
+import "core:slice"
+
+_ :: intrinsics
+_ :: runtime
+
+// Originally based on the CC0 implementation by Eric Biggers
+// See: https://github.com/ebiggers/avl_tree/
+
+// Direction specifies the traversal direction for a tree iterator.
+Direction :: enum i8 {
+	// Backward is the in-order backwards direction.
+	Backward = -1,
+	// Forward is the in-order forwards direction.
+	Forward  = 1,
+}
+
+// Ordering specifies order when inserting/finding values into the tree.
+Ordering :: slice.Ordering
+
+// Tree is an AVL tree.
+Tree :: struct($Value: typeid) {
+	// user_data is a parameter that will be passed to the on_remove
+	// callback.
+	user_data: rawptr,
+	// on_remove is an optional callback that can be called immediately
+	// after a node is removed from the tree.
+	on_remove: proc(value: Value, user_data: rawptr),
+
+	_root:           ^Node(Value),
+	_node_allocator: runtime.Allocator,
+	_cmp_fn:         proc(a, b: Value) -> Ordering,
+	_size:           int,
+}
+
+// Node is an AVL tree node.
+//
+// WARNING: It is unsafe to mutate value if the node is part of a tree
+// if doing so will alter the Node's sort position relative to other
+// elements in the tree.
+Node :: struct($Value: typeid) {
+	value: Value,
+
+	_parent:  ^Node(Value),
+	_left:    ^Node(Value),
+	_right:   ^Node(Value),
+	_balance: i8,
+}
+
+// Iterator is a tree iterator.
+//
+// WARNING: It is unsafe to modify the tree while iterating, except via
+// the iterator_remove method.
+Iterator :: struct($Value: typeid) {
+	_tree:        ^Tree(Value),
+	_cur:         ^Node(Value),
+	_next:        ^Node(Value),
+	_direction:   Direction,
+	_called_next: bool,
+}
+
+// init initializes a tree.
+init :: proc {
+	init_ordered,
+	init_cmp,
+}
+
+// init_cmp initializes a tree.
+init_cmp :: proc(
+	t: ^$T/Tree($Value),
+	cmp_fn: proc(a, b: Value) -> Ordering,
+	node_allocator := context.allocator,
+) {
+	t._root = nil
+	t._node_allocator = node_allocator
+	t._cmp_fn = cmp_fn
+	t._size = 0
+}
+
+// init_ordered initializes a tree containing ordered items, with
+// a comparison function that results in an ascending order sort.
+init_ordered :: proc(
+	t: ^$T/Tree($Value),
+	node_allocator := context.allocator,
+) where intrinsics.type_is_ordered_numeric(Value) {
+	init_cmp(t, slice.cmp_proc(Value), node_allocator)
+}
+
+// destroy de-initializes a tree.
+destroy :: proc(t: ^$T/Tree($Value), call_on_remove: bool = true) {
+	iter := iterator(t, Direction.Forward)
+	for _ in iterator_next(&iter) {
+		iterator_remove(&iter, call_on_remove)
+	}
+}
+
+// len returns the number of elements in the tree.
+len :: proc "contextless" (t: ^$T/Tree($Value)) -> int {
+	return t._size
+}
+
+// first returns the first node in the tree (in-order) or nil iff
+// the tree is empty.
+first :: proc "contextless" (t: ^$T/Tree($Value)) -> ^Node(Value) {
+	return tree_first_or_last_in_order(t, Direction.Backward)
+}
+
+// last returns the last element in the tree (in-order) or nil iff
+// the tree is empty.
+last :: proc "contextless" (t: ^$T/Tree($Value)) -> ^Node(Value) {
+	return tree_first_or_last_in_order(t, Direction.Forward)
+}
+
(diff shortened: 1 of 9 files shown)


core/container/avl/avl.odin
```diff
@@ -0,0 +1,591 @@
+/*
+package avl implements an AVL tree.
+
+The impementation is non-intrusive, and non-recursive.
```
Kelimion, line 4: impementation -> implementation

core/container/avl/avl.odin
```diff
@@ -0,0 +1,591 @@
...
+	}
+
+	it._cur = it._next
+	it._next = node_next_or_prev_in_order(it._cur, i8(it._direction))
+
+	return it._cur
+}
+
+@(private)
+tree_first_or_last_in_order :: proc "contextless" (t: ^Tree($Value), sign: i8) -> ^Node(Value) {
```
Kelimion, line 298: Not sure why it takes `sign: i8` instead of `direction: Direction`?

core/container/avl/avl.odin
```diff
@@ -0,0 +1,598 @@
...
+	t._root = nil
+	t._node_allocator = node_allocator
+	t._cmp_fn = cmp_fn
+	t._size = 0
+}
+
+// destroy de-initializes a tree.
+destroy :: proc(t: ^Tree($Value), call_on_remove: bool = true) {
+	iter := iterator(t, Direction.Forward) // TODO: Post-order would be better.
+	for node := iterator_first(&iter); node != nil; node = iterator_next(&iter) {
```
laytan, line 75: When you make `iterator_next` return a `bool` and make `iterator` initialize itself, you can just make this `for node := iterator_next(&iter) {`

tests/core/container/test_core_avl.odin
```diff
@@ -0,0 +1,158 @@
...
+	// Test the forward/backward iterators.
+	insertedValues: [dynamic]int
+	for k in insertedMap {
+		append(&insertedValues, k)
+	}
+	slice.sort(insertedValues[:])
+
+	iter = avl.iterator(&tree, avl.Direction.Forward)
+	visited: int
+	for node := avl.iterator_first(&iter); node != nil; node = avl.iterator_next(&iter) {
```
laytan, line 60: When you make `iterator_next` return a `bool` and make `iterator` initialize itself, you can just make this `for node := iterator_next(&iter) {`

tests/core/container/test_core_avl.odin
laytan, line 1: Make sure to use Odin's casing rules, see some camelCase here.

core/container/avl/avl.odin
```diff
@@ -0,0 +1,598 @@
...
+node_set_child :: proc "contextless" (n: ^Node($Value), sign: i8, child: ^Node(Value)) {
+	if sign < 0 {
+		n._left = child
+	} else {
+		n._right = child
+	}
+}
+
+@(private)
+node_adjust_balance_factor :: proc "contextless" (n: ^Node($Value), amount: i8) {
```
laytan, line 596: I would force inline this (and some of the other helper procs that are called a lot), but it is probably inlined with optimisations on pretty quickly anyway

core/container/avl/avl.odin
```diff
@@ -0,0 +1,598 @@
...
+	_tree:        ^Tree(Value),
+	_cur:         ^Node(Value),
+	_next:        ^Node(Value),
+	_direction:   Direction,
+	_initialized: bool,
+}
+
+// init initializes a tree.  The comparision function MUST return
+// < 0, 0, or > 0 if a < b, a == b, or a > b respectively.
+init :: proc(
```
laytan, line 61: To make it easier to use, a different initialiser for comparable types would be nice (which sets the cmp_proc to a `a - b`.

Suggestion:
```odin
init :: proc {
  init_comparable,
  init_cmp,
}

init_comparable :: proc(t: ^Tree($Value), node_allocator := context.allocator)
	where intrinsics.type_is_comparable(Value) {
	init_cmp(t, proc(a, b: Value) -> int { return a - b }, node_allocator)
}
```

and rename the current `init` to `init_cmp`.

core/container/avl/avl.odin
```diff
@@ -0,0 +1,598 @@
...
+
+	return true
+}
+
+// iterator returns a tree iterator in the specified direction.
+iterator :: proc "contextless" (t: ^Tree($Value), direction: Direction) -> Iterator(Value) {
+	it: Iterator(Value)
+	it._tree = t
+	it._direction = direction
+
```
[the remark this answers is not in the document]
Yawning: I changed this to call iterator_first, but the bool and a special case needs to remain in iterator_next so that the very first call to iterator_next returns the first node instead of advancing.

The bool was renamed to reflect this, and iterator_first was made private to remove a edge case that causes non-intuitive behavior.

core/container/avl/avl.odin
```diff
@@ -0,0 +1,598 @@
...
+	}
+
+	return cur
+}
+
+// insert attempts to insert the value into the tree, and returns a
+// boolean indicating if the value was NOT present, the node, and
+// the node allocator error if relevant.  If the value is already
+// present, the existing node is returned un-altered.
+insert :: proc(t: ^Tree($Value), value: Value) -> (ok: bool, n: ^Node(Value), err: runtime.Allocator_Error) {
```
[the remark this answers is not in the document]
Yawning: Unfortunately the refactor is insufficient (need to set t._root, or parent._left/parent._right), which is what the extra complexity with dereferencing is for.  Adding all the handling without dereferncing quickly turns kind of ugly, so I think the current code ends up being clearer.

laytan:
LGTM. The only thing I might like to see is a rename of the `insert` procedure, the behaviour is more like a `find_or_insert` because it does not insert if it finds the value, and just returns the old one.

core/container/avl/avl.odin
```diff
@@ -0,0 +1,653 @@
...
+	t._root = nil
+	t._node_allocator = node_allocator
+	t._cmp_fn = cmp_fn
+	t._size = 0
+}
+
+// init_comparable initializes a tree containing comparble items, with
+// a comparison function that results in an ascending order sort.
+init_comparable :: proc(
+	t: ^Tree($Value),
```
gingerBill, line 84: Every place you use generic things, prefer doing this to support `distinct` types:
```
t: ^$T/Tree($Value)
```

core/container/avl/avl.odin
```diff
@@ -0,0 +1,653 @@
...
+
+	iterator_first(&it)
+
+	return it
+}
+
+// iterator_get returns the node currently pointed to by the iterator,
+// or nil iff the node has been removed, the tree is empty, or the end
+// of the tree has been reached.
+iterator_get :: proc "contextless" (it: ^Iterator($Value)) -> ^Node(Value) {
```
gingerBill, line 268: `^$I/Iterator($Value)` 
just in case

core/container/avl/avl.odin
```diff
@@ -0,0 +1,653 @@
...
+	t._cmp_fn = cmp_fn
+	t._size = 0
+}
+
+// init_comparable initializes a tree containing comparble items, with
+// a comparison function that results in an ascending order sort.
+init_comparable :: proc(
+	t: ^$T/Tree($Value),
+	node_allocator := context.allocator,
+) where intrinsics.type_is_comparable(Value) {
```
laytan, line 86: This actually needs to be `intrinsics.type_is_ordered`, sorry for giving the wrong info earlier, comparable is for `==`, probably want it to be `init_odered` then too

Yawning: No worries, it's an easy enough change.

core/container/avl/avl.odin
```diff
@@ -0,0 +1,657 @@
...
+	// user_data is a parameter that will be passed to the on_remove
+	// callback.
+	user_data: rawptr,
+	// on_remove is an optional callback that can be called immediately
+	// after a node is removed from the tree.
+	on_remove: proc(value: Value, user_data: rawptr),
+
+	_root:           ^Node(Value),
+	_node_allocator: runtime.Allocator,
+	_cmp_fn:         proc(a, b: Value) -> int,
```
gingerBill, line 36: Why isn't this returning `slice.Ordering`?

core/container/avl/avl.odin
```diff
@@ -0,0 +1,657 @@
...
+	t._size = 0
+}
+
+// init_ordered initializes a tree containing ordered items, with
+// a comparison function that results in an ascending order sort.
+init_ordered :: proc(
+	t: ^$T/Tree($Value),
+	node_allocator := context.allocator,
+) where intrinsics.type_is_ordered_numeric(Value) {
+	init_cmp(t, proc(a, b: Value) -> int {return a - b}, node_allocator)
```
gingerBill, line 91: This is still wrong because `a - b` may not be an `int`. Prefer using `slice.cmp` to get the ordering value correctly.

PR #3206, opened 2024-02-18
[sys/linux]: Fix numeric values for socket fd flags

flysand7:
The values for constants were wrong, stole these from golang source

### DIFF
--- a/core/sys/linux/bits.odin
+++ b/core/sys/linux/bits.odin
@@ -944,8 +944,8 @@ Socket_Type :: enum {
 	Bits for Socket_FD_Flags
 */
 Socket_FD_Flags_Bits :: enum {
-	NONBLOCK  = 14,
-	CLOEXEC   = 25,
+	NONBLOCK  = 11,
+	CLOEXEC   = 19,
 }
 
 /*


gingerBill:
Are you definitely sure that these were wrong?

flysand7:
Yes, it was related to [this](https://discord.com/channels/568138951836172421/568146082396373008/1208703735836971038) issue in discord. I checked with the person having the issue that changing the constants resolved the issue.

Difference under strace, before:

```
socket(AF_UNIX, SOCK_STREAM|0x2000000 /* SOCK_??? */, 0) = -1 EINVAL (Invalid argument)
```

After:

```
socket(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC, 0) = 3
```

zen3ger:
These values are correct.  The definiton for these can be found in `bits/socket_type.h`:
```c
  SOCK_CLOEXEC = 02000000,	/* Atomically set close-on-exec flag for the
				   new descriptor(s).  */
#define SOCK_CLOEXEC SOCK_CLOEXEC
  SOCK_NONBLOCK = 00004000	/* Atomically mark descriptor(s) as
				   non-blocking.  */
#define SOCK_NONBLOCK SOCK_NONBLOCK

```

These are defined as octal, but the previous bits were defined as this was a hex literal.
`0x4000 ==  (1 << 14)` and `0x2000000 == (1 << 25)` instead.

PR #3207, opened 2024-02-18
Add `fmt.*printfln`

Tetralux:
Adds the following procedures, which just call through to the `printf` versions, with `newline = true`; a new parameter
also added in this commit.
In all cases, `wprintf` is the one that ultimately writes the newline, if requested.

- `printfln`
- `fprintfln`
- `eprintfln`
- `aprintfln`
- `tprintfln`
- `bprintfln`
- `caprintfln`
- `ctprintfln`
- `sbprintfln`
- `wprintfln`

These procs are useful as I often find myself switching between `println` and `printf`, and often find myself wanting the newline.

I originally just cutnpasted a version of `*printf` that also added a newline, but after doing so I figured that adding a `newline := false` parameter to the `*printf` procs might be a neater choice, so that's what I've done here.

I also noticed while doing this that something similar could be done with `*print` vs `*println`, but I've left that out of this PR.

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -147,16 +147,30 @@ aprintln :: proc(args: ..any, sep := " ", allocator := context.allocator) -> str
 // 	*Allocates Using Context's Allocator*
 //
 // 	Inputs:
+//	- fmt: A format string with placeholders for the provided arguments.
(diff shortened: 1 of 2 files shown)


Tetralux:
Ready

PR #3212, opened 2024-02-20
Improve net tests

laytan:
Watching the sporadic CI failures it seems to come from these tests a lot of the time, this PR cleans up and simplifies (while testing the same things):

1. Lots of tests were using threads without a need for it
2. Tests had hardcoded `time.sleep` calls which is never a good idea
3. An unclear abstraction was implemented without a real need
4. They weren't being ran on non-windows
5. The `client_connects_to_open_but_not_accepting_port` was not doing what you wanted to test for, the `tcp_server` proc was returning, and then `dial` was called, which meant that the server already closed and you got a refusal error. Now it correctly listens without accepting, which even results in a different error because the kernel buffer would have buffered the send.
6. I have removed the previously mentioned test because it has inconsistent behaviour per OS, on unix it would return a timeout error if the buffer was big enough, but on Windows it would just keep allowing the send. We shouldn't test for something like that.

### DIFF
--- a/tests/core/net/test_core_net.odin
+++ b/tests/core/net/test_core_net.odin
@@ -388,12 +388,8 @@ client_sends_server_data :: proc(t: ^testing.T) {
 	tcp_client :: proc(thread_data: rawptr) {
 		r := transmute(^Thread_Data)thread_data
 
-		log(r.t, "tcp_client entry")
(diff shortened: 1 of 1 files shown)


laytan:
I am pretty sure the latest commit in this chain actually addresses the underlying issue that caused the CI to hang sporadically

PR #3213, opened 2024-02-20
Add backend info to odin report

laytan:
For some reason the line endings in this file were in DOS format. I converted them to the consistent format of the other files in one commit, and made my very simple change in the other.

The change is adding a line to the report, it now reports the LLVM version used, which can later be expanded when we have more backends. Here is what mine looks like now:
```

	Odin:    dev-2024-02:159257597
	OS:      macOS Sonoma 14.2.1 (build: 23D56, kernel: 23.2.0)
	CPU:     ARM64
	RAM:     8192 MiB
	Backend: LLVM 17.0.6
```


laytan:
Let me know if there is a reason for the line endings to stay that way, I can rebase to only add the backend info then.

Issue #3173, opened 2024-02-05
Json Marshaling Crashes on an Empty Union


blob1807:
## Context

Odin: dev-2024-02-nightly:539cec74
Odin: dev-2024-02:19535d87
OS:   Windows 11 Home Basic (version: 22H2), build 22621.3007
CPU:  AMD Ryzen 7 6800HS with Radeon Graphics
RAM:  15608 MiB

Debugged with WinDbg running as it came.
Built with `odin build . -debug -pdb-name:testing.pdb -keep-temp-files`

## Expected Behavior

No crashes & either; an error message (a crash error or return error) or the given code should Marshal to `{ "k" : null }`.
Seeing as you can Unmarshal it I think you should be able to Marshal it as well.

## Current Behavior

Code compiles. When Marshaling an empty Union, program will crash with no error message.
Unmarshaling doesn't have this issue.

## Failure Information (for bugs)

Debugger pointed to below as the part that crashed.
https://github.com/odin-lang/Odin/blob/19535d872162b7968f10822c658b51069cf81e65/core/encoding/json/marshal.odin#L381

### Steps to Reproduce
Running one the following
```odin
package main

import "core:fmt"
import "core:encoding/json"

main :: proc() {
    j: struct{k: union{}}
    fmt.println(j)
    s, em := json.marshal(j)
    fmt.println(string(s), em)
}
```
```odin
package main

import "core:fmt"
import "core:encoding/json"

Some :: union($T: typeid){}

main :: proc() {
    j: struct{k: Some(int)}
    fmt.println(j)
    s, em := json.marshal(j)
    fmt.println(string(s), em)
}
```
Unmarshaling the expected output:
```odin
s := [?]byte{123, 34, 107, 34, 58, 110, 117, 108, 108, 125}
d: struct{k: union{}}
json.unmarshal(s[:], &d)
fmt.println(d)
```

### Failure Logs

Command output from WinDbg.

```
************* Preparing the environment for Debugger Extensions Gallery repositories **************
   ExtensionRepository : Implicit
   UseExperimentalFeatureForNugetShare : true
   AllowNugetExeUpdate : true
   AllowNugetMSCredentialProviderInstall : true
   AllowParallelInitializationOfLocalRepositories : true

   -- Configuring repositories
      ----> Repository : LocalInstalled, Enabled: true
      ----> Repository : UserExtensions, Enabled: true

[...]

PR #3214, opened 2024-02-21
core:encoding/json. Add check for empty unions in Json Marshaling.

blob1807:
Bug fix for https://github.com/odin-lang/Odin/issues/3173   
Issue is caused by `info.tag_type` being `nil` when a union in empty, e.g. `j :: union{}`.  So trying to get its `.id` field causes an access violation.   
To fix it we can either check for `len(info.variants) == 0` or `info.tag_type == nil`. The 1st should only be true for empty unions, so we'll just check for that.   
If it's just `info.tag_type == nil` then that should mean it's a broken type.


core/encoding/json/marshal.odin
```diff
@@ -377,6 +377,14 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
 		opt_write_end(w, opt, '}') or_return
 
 	case runtime.Type_Info_Union:
+		// check for empty unions
+		if len(info.variants) == 0 {
+			io.write_string(w, "null") or_return
+			return
+		} else if info.tag_type == nil {
```
gingerBill, line 384: Do not check for this because of `Maybe(^T)`

PR #3215, opened 2024-02-21
core:encoding/json Add support for writing enum value names

blob1807:
Add support for writing an enum value's name instead of its underlining value.   
Added bool to Marshal_Options to enable this & check for empty enum, e.g. `j :: enum{}`.

Example:
```odin
j :: struct{jj: enum{none, some=2}}
k := j{jj=.some}

d,e := json.marshal(k)
// d == {"jj": 2}

opt := json.Marshal_Options{use_enum_value_names=true}
d,e := json.marshal(k, opt)
// d == {"jj": "some"}


```


gingerBill:
I've added `reflect.enum_name_from_value` to make this easier to implement.

blob1807:
Sorry for the late reply.
I've changed to the new any proc. And thankyou for adding them.

core/encoding/json/marshal.odin
```diff
@@ -404,7 +407,16 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
...
 	case runtime.Type_Info_Enum:
-		return marshal_to_writer(w, any{v.data, info.base.id}, opt)
+		if !opt.use_enum_value_names || len(info.names) == 0 {
+			return marshal_to_writer(w, any{v.data, info.base.id}, opt)
+		} else {
+			name, found := reflect.enum_name_from_value_any(v)
+			if found {
+				return marshal_to_writer(w, name, opt)
+			} else {
+				panic("Unable to find value in enum's values")
```
gingerBill, line 417: Should this panic or return an error? I'm not sure myself.

blob1807: I don't really know either. I would normally log it & return an error. But I was going off the panics in the Union & Bit Set areas here. Where they panic if something's wrong with the type.

blob1807: That's a good point.
I like your suggestion layton. I'll add a fallback.

core/encoding/json/marshal.odin
```diff
@@ -51,6 +51,11 @@ Marshal_Options :: struct {
 	// NOTE: This will temp allocate and sort a list for each map.
 	sort_maps_by_key: bool,
 
+	// Output enum value's name instead of its underlineing value.
```
laytan, line 54:
```suggestion
	// Output enum value's name instead of its underlying value.
```

core/encoding/json/marshal.odin
```diff
@@ -51,6 +51,11 @@ Marshal_Options :: struct {
...
+	//
+	// NOTE: If a name isn't found it'll use the underlineing value.
```
laytan, line 56:
```suggestion
	// NOTE: If a name isn't found it'll use the underlying value.
```

blob1807:
Thankyou for the spelling corrections. I've gone a head & committed them.

blob1807:
I'm happy with the current state of the draft. So I'm changing it to Ready for Review.

PR #3218, opened 2024-02-21
`core:sync` Add recursive mutex functions to `sync` procedure groups

deadwanderer:


core/sync/sync_util.odin
```diff
@@ -6,9 +6,10 @@ Example:
 		...
 	}
 */
-guard :: proc{
+guard :: proc {
```
gingerBill, line 9: Why did you add random spaces?

deadwanderer: That was my VSCode auto-formatting, sorry.

PR #3224, opened 2024-02-22
Make MacOS releases self contained / bundle LLVM

laytan:
This makes all dynamic library dependencies bundle into `dist/libs`.

Just like before, macos will complain that the binary (and now additionally the bundled libraries) aren't signed/trusted, and you need to go to privacy settings to allow it to run.
To be clear, this already happened before this change with the binary.

This PR also makes the action zip up the dist folder, this is to keep the executable permissions when downloading it back, see https://github.com/actions/upload-artifact/issues/38 for why.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -63,11 +63,13 @@ jobs:
           cp -r core dist
           cp -r vendor dist
           cp -r examples dist
+          # Zipping so executable permissions are retained, see https://github.com/actions/upload-artifact/issues/38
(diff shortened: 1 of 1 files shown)


gingerBill:
How much bigger will this be?

laytan:
It actually went down to 60MB because we are zipping it now. Here is a run I did on my fork: https://github.com/laytan/Odin/actions/runs/8010221628 if you want to check out the artifacts.

gingerBill:
Oh wow! This might be a great thing for everyone!

PR #3230, opened 2024-02-26
Add Haiku OS support

avanspector:
core:sync works although has some FIXMEs
core:fmt works
core:sys/haiku is added

generally in Odin Haiku is considered a Unix-like OS, even though technically it is not true

(core:os additions are quite remarkable in their raw quality, albeit being functional enough. As sys/haiku grows, core:os_haiku improves too though)


Skytrias (commented):
small files looking good, does the futex sim have tests or are they implemented the same as something we got already

Skytrias (commented):
rest looks fine, good work 👍

core/sync/futex_haiku.odin
```diff
@@ -0,0 +1,127 @@
...
+	waiter.next.prev = waiter.prev
+
+ 	unix.pthread_sigmask(haiku.SIG_SETMASK, &old_mask, nil)
+
+ 	// FIXME: Add error handling!
+ 	return true
+}
+
+_futex_wait_with_timeout :: proc "contextless" (f: ^Futex, expect: u32, duration: time.Duration) -> bool {
+	// FIXME: Add timeout!
```
gingerBill, line 97: Make sure that you do! This is necessary for the sync primitives

src/path.cpp
```diff
@@ -1,461 +1,461 @@
-/*
-	Path handling utilities.
-*/
-#if !defined(GB_SYSTEM_WINDOWS)
-#include <unistd.h>
-#endif
```
gingerBill, line 6: What happened with this file? What's up with the diff?

src/threading.cpp
```diff
@@ -831,8 +833,119 @@ gb_internal void futex_wait(Futex *f, Footex val) {
...
+	Futex_Wait_Node *prev, *next;	
+};
+ 
+struct Futex_Wait_Queue {
+	std::atomic_flag spinlock;
+	Futex_Wait_Node list;
+ 
+	void lock() {
+		while (spinlock.test_and_set(std::memory_order_acquire)) {
+			; // spin...
```
gingerBill, line 856: At least put a cpu_relax in there of some sort.

core/sync/futex_haiku.odin
```diff
@@ -0,0 +1,127 @@
+//+private
+package sync
+
+import "core:c"
+import "core:c/libc"
```
gingerBill, line 5: Do not use libc here the atomics. package sync has all of the atomics stuff in it already.

core/sync/futex_haiku.odin
```diff
@@ -0,0 +1,127 @@
...
+@(private="file")
+Wait_Queue :: struct {
+	lock: libc.atomic_flag,
+	list: Wait_Node,
+}
+@(private="file")
+waitq_lock :: proc "contextless" (waitq: ^Wait_Queue) {
+	// FIXME: Get rid of context here.
+	context = runtime.default_context()
+	for libc.atomic_flag_test_and_set_explicit(&waitq.lock, .acquire) {
```
gingerBill, line 26: Use the atomics built into the language.

core/sync/futex_haiku.odin
```diff
@@ -0,0 +1,127 @@
...
+			prev = &g_waitq.list,
+			next = &g_waitq.list,
+		},
+	}
+}
+
+@(private="file")
+get_waitq :: #force_inline proc "contextless" (f: ^Futex) -> ^Wait_Queue {
+	_ = f
+	return &g_waitq
```
gingerBill, line 55: If this is private to the file, why even bother with a dummy parameter?

core/sync/futex_haiku.odin
```diff
@@ -0,0 +1,127 @@
...
+
+_futex_broadcast :: proc "contextless" (f: ^Futex) {
+	waitq := get_waitq(f)
+	waitq_lock(waitq)
+	defer waitq_unlock(waitq)
+
+	head := &waitq.list
+	for waiter := head.next; waiter != head; waiter = waiter.next {
+		if waiter.futex == f {
+			unix.pthread_kill(waiter.thread, haiku.SIGCONT)
```
gingerBill, line 124: Is there not a distinction between signal and broadcast on Haiku?

src/threading.cpp
avanspector: is there an alternative for this in the compiler codebase?

src/path.cpp
avanspector: only one `#if defined(GB_SYSTEM_HAIKU)` for `dirent.h` was added, everything else is CRLF conversion

avanspector:
> small files looking good, does the futex sim have tests or are they implemented the same as something we got already

futex for Haiku was simulated using libc, implementation taken from here [](https://tavianator.com/2023/futex.html)
im a novice in the whole futexes/multithreading situation, all test was basically running on the OS, negating chances of deadlocking as much as possible (i'll try reaching out to the haiku devs to make the implementation as reliable as possible since this simulation of futexes works fine on Linux)

src/threading.cpp
gingerBill: `intrinsics.cpu_relax()`

PR #3236, opened 2024-02-29
time: fix leap year maths in datetime_to_time

### LINKED ISSUE #3235: datetime_to_time reports incorrect date
## Context

* Operating System & Odin Version: Linux 6.7.6, Odin commit 7df7fec6f736a52712961a2a54de3747eac4ec84

## Expected Behavior

The following program should output a date of March 1st, but it reports February the 29th.

```odin
package main

import "core:fmt"
import "core:time"

main :: proc() {
	fmt.println(time.datetime_to_time(2024, 3, 1, 0, 0, 0))
}
```

The program prints this:
```
2024-02-29 00:00:00.000000000 +0000 UTC true
```

I think the procedure isn't taking into account leap years with regard to February's extra day, and the logic behind `DAYS_PER_400_YEARS` and the other calculations is sound, but it's starting the arithmetic modulo from the year minus 1970, which would give incorrect results.

widowight:
I caught a bug with `datetime_to_time` where it was failing to calculate days according to leap years.
Also replaced the internal `is_leap_year` proc with a public API that uses bitwise instead of modulo calculations.

I calculated `EPOCH_DAYS` with this program:

```odin
	for i in 1 ..< 1970 {
		days += 365

		if i > 0 && is_leap_year(i) {
			days += 1
		}
	}
```

Fixes #3235

### DIFF
--- a/core/time/time.odin
+++ b/core/time/time.odin
@@ -299,10 +299,6 @@ _time_abs :: proc "contextless" (t: Time) -> u64 {
 
 @(private)
 _abs_date :: proc "contextless" (abs: u64, full: bool) -> (year: int, month: Month, day: int, yday: int) {
-	_is_leap_year :: proc "contextless" (year: int) -> bool {
-		return year%4 == 0 && (year%100 != 0 || year%400 == 0)
-	}
-
 	d := abs / SECONDS_PER_DAY
 
 	// 400 year cycles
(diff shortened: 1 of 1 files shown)


core/time/time.odin
```diff
@@ -360,7 +356,13 @@ _abs_date :: proc "contextless" (abs: u64, full: bool) -> (year: int, month: Mon
 	return
 }
 
+is_leap_year :: proc "contextless" (year: int) -> bool {
+	return !((year & 3 > 0) || (year & 5 > 0) && !(year % 25 == 0))
+}
```
gingerBill, line 361: Why did you remove the clearer logic and replace it with "magic" (I understand what it does)?

gingerBill:
Fixed elsewhere.

core/time/time.odin
widowight: An understanding that partially bitwise arithmetic would be faster than completely modular arithmetic.

PR #3238, opened 2024-02-29
makes raylib and stb_rect_pack free of libc

### LINKED ISSUE #3144: Needless libc import in vendor stb rect pack
https://github.com/odin-lang/Odin/blob/da6edb3764b735a839acdd375328574833d782c1/vendor/stb/rect_pack/stb_rect_pack.odin#L3

Should be import "core:c" because only c types are actually used.

laytan:
May be a bit niche so I would understand not wanting it, but, I was playing around with https://github.com/tsoding/zozlib.js which is a subset of Raylib fully implemented in JS so you don't need emscriptem. With just these changes and some very minor changes in that repo I was able to make it work in Odin.

Also fixed #3144 while I was at it.

### DIFF
--- a/vendor/raylib/raylib.odin
+++ b/vendor/raylib/raylib.odin
@@ -81,7 +81,7 @@ Package vendor:raylib implements bindings for version 5.0 of the raylib library 
 */
 package raylib
 
-import c "core:c"
+import "core:c"
 import "core:fmt"
 import "core:mem"
 import "core:strings"
--- a/vendor/raylib/raylib.odin
+++ b/vendor/raylib/raylib.odin
@@ -81,7 +81,7 @@ Package vendor:raylib implements bindings for version 5.0 of the raylib library 
 */
 package raylib
 
-import c "core:c/libc"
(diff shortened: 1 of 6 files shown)


vendor/raylib/raylib.odin
```diff
@@ -925,13 +925,17 @@ NPatchLayout :: enum c.int {
 	THREE_PATCH_HORIZONTAL,  // Npatch layout: 3x1 tiles
 }
 
-
+// NOTE: Castable to `core:c/libc`'s `va_list`.
+// But some use cases of raylib do not want `libc` imported.
+va_list :: struct #align(16) {
+	_: [4096]u8,
+}
```
gingerBill, line 932: Being castable might be fine, but this seems off.

laytan: What if we put the `va_list` definition in `core:c` and alias it in `core:c/libc`? That seems clean to me

gingerBill: Might be a good idea.

PR #3240, opened 2024-02-29
reduce dynamic library dependencies for macos release

laytan:
With the current releases we bundle 3 dynamic libraries:
- `libLLVM-C.dylib`
- `libLLVM.dylib`
- `libunwind.1.0.dylib`

After this change it will just be `libLLVM.dylib`. I tested and everything still works, I think `libLLVM-C.dylib` is just a subset of `libLLVM.dylib`, and `libunwind` is a system library on macos, but because it also comes with llvm, it was given priority by the linker.

You might say this doesn't matter but it reduces the amount of annoying privacy popups you get (macos shows one for each library).


laytan:
CI fails on the recently flaky crypto test, not an issue with this PR

Issue #3243, opened 2024-03-03
math/big INT_ONE is not set properly


edyu:
## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System & Odin Version:
* Please paste `odin report` output:

        Odin:    dev-2024-03:674bd94f7
        OS:      Ubuntu 22.04.4 LTS, Linux 5.15.133.1-microsoft-standard-WSL2
        CPU:     AMD Ryzen 7 5800HS with Radeon Graphics
        RAM:     11662 MiB
        Backend: LLVM 14.0.0

## Expected Behavior

Please describe the behavior you are expecting

big.INT_ONE should be set properly to `1` just as big.INT_ZERO is set to `0` 

## Current Behavior

What is the current behavior?

big.INT_ONE is basically empty, as you can see if you simply `fmt.println`

```
 &Int{used = 0, digit = [], sign = "Zero_or_Positive", flags = bit_set[Flag; u8]{}}
```

And as a result, `int_one` also doesn't work.
If you attempt to use `int_one`, it would just cause the following error:

```
/Odin/core/math/big/private.odin(3193:54) Index 0 is out of range 0..<0
Illegal instruction
```

## Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

```
    t1: big.Int
    fmt.println("big.INT_ONE", big.INT_ONE)
    big.int_one(&t1)
    fmt.println("big.int_one", t1)
```

### Failure Logs

Please include any relevant log snippets or files here.

PR #3244, opened 2024-03-03
Fix typo for core:math/big.INT_MINUS_INF

edyu:
Found this typo/mistake while trying to fix https://github.com/odin-lang/Odin/issues/3243. This is not directly related but should be fixed anyways.


Kelimion:
Handled in 4c35633e0147b481dd7b2352d6bdb603f78c6dc7.

PR #3247, opened 2024-03-04
@(link_section): accept a comma in the provided link section

BeauSLM:
(only if it's not the first character)

In mach-o binaries, link sections are specified with the format `<SEGMENT>,<SECTION>`. Presently the ',' is literally not allowed, which this change fixes.

e.g. for putting a var in read only memory on m-series mac (which uses mach-o)

```c
@(link_section="__TEXT,__const") // the comma isn't allowed, but required by the mach-o format
foo: int = 0
```

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -1943,6 +1943,7 @@ gb_internal bool is_foreign_name_valid(String const &name) {
 			case '$':
 			case '.':
 			case '_':
+			case ',':
 				break;
 			default:
 				if (!gb_char_is_alphanumeric(cast(char)rune)) {


BeauSLM:
A quick note, attempting to write to `foo` in the above example causes `SIGBUS` to be fired (weird ik) but I make a minimal repro in c and it also fired `SIGBUS`. Weird.

odin:
```c
package main

@(link_section="__TEXT,__const")
foo : int = 23

main :: proc() {
    fmt.println(size_of(int))
    fmt.println(foo)
}


import "core:fmt"
```

```c
#include <stdio.h>

__attribute__((section("__TEXT,__const")))
int myReadOnlyVariable = 42;

int main() {
    int *ptr = &myReadOnlyVariable;
    *ptr = 9;
    printf("Read-only variable value: %d\n", *ptr);
    return 0;
}
```

src/parser.cpp
```diff
@@ -1943,6 +1943,7 @@ gb_internal bool is_foreign_name_valid(String const &name) {
 			case '$':
 			case '.':
 			case '_':
+			case ',':
```
gingerBill, line 1946: Please don't just go an change a random procedure which is used by many other things, just to get this working. 

I will not accept this bodge.

BeauSLM: You're right, I should have considered the issue and filed an issue first. My apologies.

What can I do to help resolve the problem? I'll start by filing an issue that explains the problem and references here, and I'll close the PR.

PR #3258, opened 2024-03-09
Base/runtime (darwin): Incorrect write syscall on Darwin - fixes stderr output.

deckarep:
On macos, Odin was previously swallowing errors that would be reported via stderr. 

I've confirmed with @laytan on Discord that this fixes the issue.


laytan:
Looks good, I made a typo that broke Intel macs when we I ported the syscall to the base collection, curiously it kept working on arm

PR #3270, opened 2024-03-12
Expose internal invmod and int_exponent_mod with more consistent naming

edyu:
A simple patch to expose
existing `internal_invmod` also as `internal_int_invmod` just like all the other `mod` functions such as `addmod`, `divmod`, `mulmod`, `submod` etc.

And more importantly,
expose `internal_int_exponent_mod` as `internal_int_powmod` as the word `pow` is preferred in existing library with all the `pow_f16`, `pow_small`, `int_pow` etc. 

I also renamed `internal_int_exponent_mod` to `internal_int_exponent_modulus` since if we are spelling out exponent already, might as well spell out `mod` as `modulus` like how it is done in `internal_int_inverse_modulus`.
Of course, `internal_int_exponent_mod` now refers to `internal_int_exponent_modulus` to maintain backward compatibility.

### DIFF
--- a/core/math/big/internal.odin
+++ b/core/math/big/internal.odin
@@ -2047,6 +2047,7 @@ internal_int_inverse_modulo :: proc(dest, a, b: ^Int, allocator := context.alloc
 
 	return _private_inverse_modulo(dest, a, b)
 }
+internal_int_invmod :: internal_int_inverse_modulo
 internal_invmod :: proc{ internal_int_inverse_modulo, }
 
 /*
--- a/core/math/big/prime.odin
+++ b/core/math/big/prime.odin
@@ -44,7 +44,7 @@ internal_int_prime_is_divisible :: proc(a: ^Int, allocator := context.allocator)
 	Computes res == G**X mod P.
 	Assumes `res`, `G`, `X` and `P` to not be `nil` and for `G`, `X` and `P` to have been initialized.
 */
-internal_int_exponent_mod :: proc(res, G, X, P: ^Int, allocator := context.allocator) -> (err: Error) {
+internal_int_power_modulo :: proc(res, G, X, P: ^Int, allocator := context.allocator) -> (err: Error) {
 	context.allocator = allocator
 
 	dr: int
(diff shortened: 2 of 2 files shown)


core/math/big/prime.odin
```diff
@@ -44,7 +44,7 @@ internal_int_prime_is_divisible :: proc(a: ^Int, allocator := context.allocator)
 	Computes res == G**X mod P.
 	Assumes `res`, `G`, `X` and `P` to not be `nil` and for `G`, `X` and `P` to have been initialized.
 */
-internal_int_exponent_mod :: proc(res, G, X, P: ^Int, allocator := context.allocator) -> (err: Error) {
+internal_int_exponent_modulus :: proc(res, G, X, P: ^Int, allocator := context.allocator) -> (err: Error) {
```
Kelimion, line 47: In most of the package, the procedures are named `*_modulo`, it's the parameter that's called `modulus`.
image

core/math/big/internal.odin
```diff
@@ -2047,7 +2047,8 @@ internal_int_inverse_modulo :: proc(dest, a, b: ^Int, allocator := context.alloc
 
 	return _private_inverse_modulo(dest, a, b)
 }
-internal_invmod :: proc{ internal_int_inverse_modulo, }
+internal_int_invmod :: internal_int_inverse_modulo
+internal_invmod :: proc{ internal_int_invmod, }
```
Kelimion, line 2051: Why change the proc group? It can remain `internal_invmod :: proc{ internal_int_inverse_modulo, }`.

core/math/big/prime.odin
```diff
@@ -112,6 +112,9 @@ internal_int_exponent_mod :: proc(res, G, X, P: ^Int, allocator := context.alloc
 	*/
 	return _private_int_exponent_mod(res, G, X, P, 0)
 }
+internal_int_exponent_mod :: internal_int_exponent_modulus
+internal_int_powmod :: internal_int_exponent_modulus
+internal_powmod :: proc { internal_int_powmod, }
```
Kelimion, line 117: Same here with the proc group. Just point it to the actual thing:
`internal_powmod :: proc{ internal_int_exponent_modulo, }`

And `_modulo` instead of modulus, because see above.

Kelimion (changes requested):

core/math/big/prime.odin
edyu: You are correct. Let me fix that.

core/math/big/internal.odin
edyu: I was mimicking how you have for example internal_addmod point to internal_int_ addmod process group etc. But I can certainly change that.

edyu:
I made the changes request, except I changed the implementing function name from `internal_int_exponent_mod` to `internal_int_power_modulo` to be more consistent.

PR #3272, opened 2024-03-13
Add getrusage syscall for mac/darwin

iansimonson:
The syscall number existed but the wrapper for calling it did not. Also adds the `RUsage` struct to receive the data.

Naming is kept the same as in sys/linux and sys/unix though happy to change anything.

Happy to add the RUsage_Who enum like in sys/linux also if desired but the darwin syscall helpers seemed to be using primitives everywhere rather than e.g. in sys/linux where it's a "contextless" proc that feels more odin like and forwards to the syscall

### DIFF
--- a/core/sys/darwin/darwin.odin
+++ b/core/sys/darwin/darwin.odin
@@ -3,26 +3,27 @@ package darwin
 
 Bool :: b8
 
-timespec :: struct {
-        seconds:      int,
-        microseconds: int,
+timeval :: struct {
+	tv_sec:  int,
+	tv_usec: int,
 }
 
 RUsage :: struct {
-        utime:         timespec,
-        stime:         timespec,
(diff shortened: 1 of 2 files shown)


iansimonson:
Hmm... using the already defined `__darwin_time_t` is now giving me errors when using  `getrusage`. swapping that to an `int` / `i64` seems to fix it. I'm a bit wary to change the alias to `int` given its use in other areas, but maybe the `darwin` package needs to be set up for 32-bit/64-bit targets

iansimonson:
This should be good now. I went through the type defs and checked a couple sources to confirm, `__darwin_time_t` is defined as `long` which on Mac is equivalent to Odin's `int` not i32

gingerBill:
I would recommend `c.long` vs the raw Odin "int"

PR #3278, opened 2024-03-14
Add start and end parameters to string.index and index_byte functions. 

hodgka:
Might be nice to have start and end parameters for strings.index, similar to [str.find in python](https://docs.python.org/3/library/stdtypes.html#str.find).

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -1398,6 +1398,8 @@ NOTE: Can't find UTF-8 based runes.
 Inputs:
 - s: The input string to search in.
 - c: The byte to search for.
+- start: The index to start searching from. (default: -1)
+- end: The index to stop searching at. (default: -1)
 
 Returns:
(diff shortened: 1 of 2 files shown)


laytan:
I believe this has been suggested before and the person was then pointed to slices making this pretty trivial in Odin:
```odin
i := start + strings.index(str[start:end])
```

hodgka:
Fair enough. Feel free to close

gingerBill:
Why did you not just make a new procedure? Why did you change a pre-existing one which many people might rely on the signature?

hodgka:
I figured it doesn't change the present usage of `strings.index(haystack, needle)`, and doesn't add additional API surface. Does this actually break the existing API somehow? I'm still fairly new to odin, so pardon me if I'm misunderstanding

I could just split it out as like laytan suggested, 
```
index_with_bounds :: proc(haystack, needle: string, start, end: int) {
	return  start + strings.index(haystack[start:end], needle)
}
```

PR #3283, opened 2024-03-15
darwin: use new wait on address API if possible

laytan:
(currently draft, gathering feedback) PR for using the new non-private futex API on darwin (https://developer.apple.com/documentation/os/4316531-os_sync_wait_on_address?changes=_2_8_8&language=objc).

For Odin:
- translated the whole header into Odin at `core:sys/darwin`
- made the already existing `-minimum-os-version` string a global constant (converted to an integer) so you can detect version and switch implementations
- if the minimum version supports the new api, use that instead of the private api's
- An unfortunate thing here is that in order to use the new api, the user has to set the `-minimum-os-version=14.4` or higher (macOS), which is probably not a great idea to do because that is the absolute latest version at the moment

An idea I had was defaulting the `MINIMUM_OS_VERSION` constant to the default that clang (during linking) gives the executable if you don't provide it, but I was unable to figure out what their defaults are at the moment. On my machine it uses `14.0` and my macOS version is `14.4`, but all mentions of defaults in the clang source code seem to mention version in the 10 range. Maybe we should put in a default of our own (which we also pass to clang when linking), but the question then becomes, what should that default be?

For the compiler:
- detect availability using the recommended clang macro `__builtin_available` and use the api if available both when compiling and at runtime
- NOTE: this does at all times still link with the private API because the `__builtin_available` macro detects availability at runtime using weak linking, although this does not seem like a problem to me because we only don't want to use private APIs in the Odin code to avoid being rejected from the app store, the compiler will not be applying for the app store so it should be fine

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -1104,7 +1104,7 @@ gb_internal void init_universal(void) {
 			sscanf(cast(const char *)(build_context.minimum_os_version_string.text), "%d.%d.%d", &major, &minor, &revision);
 			minimum_os_version = (major*10000)+(minor*100)+revision;
 		}
-		add_global_constant("MINIMUM_OS_VERSION", t_untyped_integer, exact_value_i64(minimum_os_version));
+		add_global_constant("ODIN_MINIMUM_OS_VERSION", t_untyped_integer, exact_value_i64(minimum_os_version));
 	}
 
 	add_global_bool_constant("ODIN_DEBUG",                      bc->ODIN_DEBUG);
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -1097,6 +1097,15 @@ gb_internal void init_universal(void) {
 		scope_insert(intrinsics_pkg->scope, t_atomic_memory_order->Named.type_name);
 	}
 
+	{
+		int minimum_os_version = 0;
+		if (build_context.minimum_os_version_string != "") {
+			int major, minor, revision = 0;
+			sscanf(cast(const char *)(build_context.minimum_os_version_string.text), "%d.%d.%d", &major, &minor, &revision);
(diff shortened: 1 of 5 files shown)


src/checker.cpp
```diff
@@ -1097,6 +1097,15 @@ gb_internal void init_universal(void) {
...
 	}
 
+	{
+		int minimum_os_version = 0;
+		if (build_context.minimum_os_version_string != "") {
+			int major, minor, revision = 0;
+			sscanf(cast(const char *)(build_context.minimum_os_version_string.text), "%d.%d.%d", &major, &minor, &revision);
+			minimum_os_version = (major*10000)+(minor*100)+revision;
+		}
+		add_global_constant("MINIMUM_OS_VERSION", t_untyped_integer, exact_value_i64(minimum_os_version));
```
gingerBill, line 1107: I highly recommend prefixing this with `ODIN_` to prevent possible collisions.

laytan:
There was a brief conversation about determining what to call in Odin at runtime, but, I think this would require the compiler to support weak linking, like we do in the C/compiler part, and I don't think we have anything like that currently.

But also, and that is pretty important, because part of this change is to be approved in the app store, I assume the review process just looks at what is linked against, and if we did this it would still link to the private API too, and may be rejected just as before.

P.S: Looking at the stack trace our assumption is also correct that this api just wraps the private API we were using. It also still has the occasional panics that were discussed in Discord.

gingerBill:
We could add weak linking if necessary. This should not be too difficult with LLVM to do, but it would require some tweaking the `foreign import` system e.g. `@(weak)` attribute or something.

gingerBill (approved):

PR #3294, opened 2024-03-18
glfw: update from 3.3.8 to 3.4

laytan:
Changes were done by comparing this: https://github.com/glfw/glfw/compare/3.3.8...3.4

Windows libraries are from their official binary download, and the darwin binary is a fat one for both intel and arm, compiled myself.

There are no breaking changes in the API as could be expected.

Release notes: https://www.glfw.org/docs/latest/news.html


laytan:
CI is not my fault, this is ready to go

PR #3300, opened 2024-03-20
Patch raylib KeyboardKey.MENU typo

matias-eduardo:


matias-eduardo:
Just noticed this was recently fixed in Raylib (post version 5.0). Do we patch it still or wait for next release?

https://github.com/raysan5/raylib/commit/5c25913e095fcf38ecf76abc37917544584c58d3

matias-eduardo:
Nevermind, since we're still in 5.0 this would probably break some stuff.

PR #3304, opened 2024-03-20
Remove entry point when compiled with no-entry-point as shared library

mailgerigk:
Currently the `-no-entry-point` flag gets ignored when building as dynamic library.

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -6320,6 +6320,8 @@ gb_internal void check_parsed_files(Checker *c) {
 
 			error(token, "Undefined entry point procedure 'main'");
 		}
+	} else if (build_context.build_mode == BuildMode_DynamicLibrary && build_context.no_entry_point) {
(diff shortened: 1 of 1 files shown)


gingerBill:
So this is something you have to be very careful with. DLLs do actually have an entry point. On Windows, that is `DllMain`. If you don't want a entry point for a dynamic library, you also have to be very careful and make sure to call `runtime._startup_runtime` and `runtime._cleanup_runtime` where appropriate.

mailgerigk:
The change I made just replaces the [intrinsics.__entry_point()](https://github.com/odin-lang/Odin/blob/fc587c507adce69d7025111ea2a3a40d7e9b373f/base/runtime/entry_windows.odin#L19) call, if a main proc exists in the code and `-no-entry-point` is set, functionally with a no-op. The intend was not to entirely remove the entry-point the operating system is concerned with but the entry-point the user code is concerned with if that makes sense.

PR #3307, opened 2024-03-21
Add ODIN_CRT_MD flag for core:c/libc on windows

mailgerigk:
Opt-in flag to link the ucrt for `core:c/libc` via `ucrt.lib` instead of `libucrt.lib`. This should be equivalent to switching from `/MT` to `/MD`.

### DIFF
--- a/src/llvm_backend_utility.cpp
+++ b/src/llvm_backend_utility.cpp
@@ -2112,11 +2112,10 @@ gb_internal lbAddr lb_handle_objc_find_or_register_selector(lbProcedure *p, Stri
 	}
 
 	if (!entity) {
-		gbString global_name = gb_string_make(temporary_allocator(), "__$objc_SEL$");
+		gbString global_name = gb_string_make(temporary_allocator(), "__$objc_SEL::");
 		global_name = gb_string_append_length(global_name, name.text, name.len);
 
-		lbAddr default_addr = lb_add_global_generated_with_name(
-			default_module, t_objc_SEL, {},
+		lbAddr default_addr = lb_add_global_generated_with_name(default_module, t_objc_SEL, {},
 			make_string(cast(u8 const *)global_name, gb_string_length(global_name)),
 			&entity);
 		string_map_set(&default_module->objc_selectors, name, lbObjcRef{entity, default_addr});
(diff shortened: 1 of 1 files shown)


core/c/libc/complex.odin
```diff
@@ -3,7 +3,11 @@ package libc
 // 7.3 Complex arithmetic
 
 when ODIN_OS == .Windows {
-	foreign import libc "system:libucrt.lib"
+	when ODIN_CRT_MD {
```
gingerBill, line 6: Why is it called `ODIN_CRT_MD`?

mailgerigk: I'm open for a better name. The current name is just the `ODIN` prefix followed by the subject `CRT` and `MD` for the C++ equivalent `/MD` flag.

Apis035: How about `ODIN_DYNAMIC_CRT` which is what `/MD` does as opposed to `/MT` that statically link CRT?

mailgerigk: `DYNAMIC_CRT` is certainly better than `CRT_MD`.

core/c/libc/complex.odin
```diff
@@ -3,7 +3,11 @@ package libc
 // 7.3 Complex arithmetic
 
 when ODIN_OS == .Windows {
-	foreign import libc "system:libucrt.lib"
+	when ODIN_DYNAMIC_CRT {
```
gingerBill, line 6: The problem here is that it kind of implies that other targets have a static one by default, which is not necessarily the case.

I am not saying what you've done is a bad idea, but rather it's got issues with affect other platforms.

mailgerigk: I agree this is a windows only option and having it available on other platforms feels wrong. Is it possible to define a config flag for only one platform? (maybe by using `#config(ODIN_DYNAMIC_CRT, false)` at every use site?)

flysand7: The issue is that the libc is "dynamic" by default on other platforms. Shouldn't it be dynamic on windows by default as well?

gingerBill: No. Don't try to compare Windows to other platforms, that is rarely a good idea, especially in this case. The problem is that Windows has different variants of libc, both dynamic and static, and they do conflict with each other (MD, MT, MDd, MTd, etc).

The one we picked in `libc` is one of the most common, but it seems to conflict with raylib (and it's always raylib that people bring up) every single time. 

The other issue is that people who do this, rarely actually need anything from raylib, but because they are used to C (or copying a tutorial written in C), they think using libc is what they should do, rather than other native Odin things. 

This PR has been left for a long while because it's fundamentally a hard problem, and this solution is just a bad bodge. It doesn't really solve the fundamental problem that it is trying to solve.

I won't close the PR, but it's just not solved by doing this.

PR #3308, opened 2024-03-21
Support LLVM 18 (non-windows targets for now)

laytan:
~~Very much still a draft~~, needs a bunch of testing on all the targets and on earlier version to make sure changes to support llvm 18 don't break older versions.

To test / to do:

- [x] Darwin
    - [x] amd64
        - [x] llvm-18
        - [x] clang-18 (linking)
        - [x] older versions still work
        - [x] Successfully runs CI/all tests with and without optimisations and -sanitize:address
          - [x] test_core_image.run_png_suite:1507 (normal failure) (on a few specific tests in the suite with messages similar to tbyn3p08 test 3 hash is c82111f1, expected b1d45c1e with Options{alpha_add_if_missing, blend_background} - not caused by this upgrade, made #3480    
          - [x] All other tests 
    - [x] arm64
        - [x] llvm-18 
        - [x] clang-18 (linking)
        - [x] older versions still work
        - [x] Successfully runs CI/all tests with and without optimisations and -sanitize:address
- [x] Linux
    - [x] amd64
        - [x] llvm-18
        - [x] clang-18
        - [x] older versions still work
            - [x] 17
            - [x] 14
            - [x] ...
        - [x] Successfully runs CI/all tests with and without optimisations and -sanitize:address
          - [x] test_core_hash.test_xxhash_vectors:323 (segfault) (something about 128 bit ints in relation with packed structs is going wrong)
          - [x] test_core_image.run_png_suite:1507 (normal failure) (on a few specific tests in the suite with messages similar to tbyn3p08 test 3 hash is c82111f1, expected b1d45c1e with Options{alpha_add_if_missing, blend_background}  - not caused by this upgrade, made #3480  
- [x] WASM - Tested manually and by running https://github.com/colrdavidson/spall-web
- [x] FreeBSD - Tested by Discord user siemato 

~~- [ ] OpenBSD~~
~~- [ ] Haiku~~
~~- [ ] Essence~~
~~- [ ] Linux i368~~
~~- [ ] Linux arm64~~
~~- [ ] Linux arm32~~

Striked through targets are targets I don't intend to test, I don't own these, feel free to test and let me know and I will update here

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -1,3 +1,5 @@
+#include <llvm/Config/llvm-config.h>
+
 #if defined(GB_SYSTEM_FREEBSD) || defined(GB_SYSTEM_OPENBSD)
 #include <sys/types.h>
 #include <sys/sysctl.h>
@@ -613,7 +615,6 @@ struct TargetMetrics {
 	isize          max_align;
 	isize          max_simd_align;
 	String         target_triplet;
-	String         target_data_layout;
 	TargetABIKind  abi;
 };
 
@@ -923,7 +924,12 @@ gb_internal isize MAX_ERROR_COLLECTOR_COUNT(void) {
 	return build_context.max_error_count;
 }
 
-
+// NOTE: AMD64 targets had their alignment on 128 bit ints bumped from 8 to 16 (undocumented of course).
+#if LLVM_VERSION_MAJOR >= 18
+	#define AMD64_MAX_ALIGNMENT (16)
+#else
+	#define AMD64_MAX_ALIGNMENT (8)
+#endif
 
 gb_global TargetMetrics target_windows_i386 = {
 	TargetOs_windows,
@@ -934,9 +940,8 @@ gb_global TargetMetrics target_windows_i386 = {
 gb_global TargetMetrics target_windows_amd64 = {
 	TargetOs_windows,
 	TargetArch_amd64,
-	8, 8, 8, 16,
+	8, 8, AMD64_MAX_ALIGNMENT, 16,
 	str_lit("x86_64-pc-windows-msvc"),
-	str_lit("e-m:w-i64:64-f80:128-n8:16:32:64-S128"),
 };
 
 gb_global TargetMetrics target_linux_i386 = {
@@ -949,16 +954,14 @@ gb_global TargetMetrics target_linux_i386 = {
 gb_global TargetMetrics target_linux_amd64 = {
 	TargetOs_linux,
 	TargetArch_amd64,
-	8, 8, 16, 16,
+	8, 8, AMD64_MAX_ALIGNMENT, 16,
 	str_lit("x86_64-pc-linux-gnu"),
-	str_lit("e-m:w-i64:64-f80:128-n8:16:32:64-S128"),
 };
 gb_global TargetMetrics target_linux_arm64 = {
 	TargetOs_linux,
 	TargetArch_arm64,
-	8, 8, 8, 16,
+	8, 8, 16, 16,
 	str_lit("aarch64-linux-elf"),
-	str_lit("e-m:o-p:32:32-Fi8-i64:64-v128:64:128-a:0:32-n32-S64"),
 };
 
 gb_global TargetMetrics target_linux_arm32 = {
@@ -966,15 +969,13 @@ gb_global TargetMetrics target_linux_arm32 = {
 	TargetArch_arm32,
 	4, 4, 4, 8,
 	str_lit("arm-linux-gnu"),
-	str_lit("e-m:o-p:32:32-Fi8-i64:64-v128:64:128-a:0:32-n32-S64"),
 };
 
 gb_global TargetMetrics target_darwin_amd64 = {
 	TargetOs_darwin,
 	TargetArch_amd64,
-	8, 8, 8, 16,
+	8, 8, AMD64_MAX_ALIGNMENT, 16,
 	str_lit("x86_64-apple-macosx"), // NOTE: Changes during initialization based on build flags.
-	str_lit("e-m:o-i64:64-f80:128-n8:16:32:64-S128"),
 };
 
 gb_global TargetMetrics target_darwin_arm64 = {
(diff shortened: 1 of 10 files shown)


JesseRMeyer:
Works on Windows.  An added nicety is that address sanitizer was fixed on windows in llvm 18.

```
gb_global TargetMetrics target_linux_amd64 = {
	TargetOs_linux,
	TargetArch_amd64,
	8, 8, 16, 16,
```
needs to be carried over to windows_amd64 otherwise there is an access-violation accessing a u128 in math/big, which was found by asan running the demo.

With that fix, the demo runs clean as a whistle with and without all optimization modes with and without -debug.

src/build_settings.cpp
```diff
@@ -498,40 +503,35 @@ gb_global TargetMetrics target_linux_i386 = {
...
-	8, 8, 16, 16,
+	8, 8, AMD64_MAX_ALIGNMENT, 16,
 	str_lit("x86_64-pc-linux-gnu"),
-	str_lit("e-m:w-i64:64-f80:128-n8:16:32:64-S128"),
 };
 gb_global TargetMetrics target_linux_arm64 = {
 	TargetOs_linux,
 	TargetArch_arm64,
-	8, 8, 8, 16,
+	8, 8, 16, 16,
```
laytan, line 512: arm targets were already aligning i128 ints on 16 bytes before LLVM 18, so we had those wrong all along

Yawning:
Tried this out with LLVM 18.1.4 compiled from source on Arch (btw).

- There appears to be a regression or change in how `llvm-config` needs to be invoked if the `.a` files are missing (at least I needed to add `--link-shared` to the LDFLAGS invocation).  This may be LLVM bug 83698, but see the note about my screwed up LLVM build. 
- tests/core/hash (make hash_test) fails with the default `-o:speed`, passes with the optimization disabled.
- The other core tests appear to pass.

Note: My LLVM build is held together by duct-tape, string, and tears (`tools/llvm-shlib/typeids.test` and only that test fails when trying to run the regression tests with `ninja check`), and is based on the old llvm17 AUR package.

laytan:
Yep I am going to look at the test failures on Linux with optimizations, if you enable optimizations on all the tests it currently fails on:
- test_core_hash.test_xxhash_vectors:323 (segfault) (looks like this is related to 128 bit ints)
- test_core_image.run_png_suite:1507 (normal failure) (on a few specific tests in the suite with messages similar to `tbyn3p08 test 3 hash is c82111f1, expected b1d45c1e with Options{alpha_add_if_missing, blend_background}`

laytan:
These test failures are real head scratchers, only happening on optimised builds so it is hard to debug, I have ruled out that it is sroa by turning it off and it still fails. Also, doing the slightest change in specific places makes it work or not work, ie printing out a value or putting in a `runtime.trap()` might cause it not to fail anymore.

Also, in the image failure, it is just giving the wrong output, there is no crash or anything, and with the image loading proc being so large it is hard to pin down.

For the image tests I have created a separate issue, this also fails on LLVM 17 so it is not caused by this upgrade, and I haven't figured it out in reasonable time. There is even a comment in the test file here https://github.com/odin-lang/Odin/blob/c72a269b7cfd5f955254f9c3ad8e5e63bac7aa2c/tests/core/image/test_core_image.odin#L1494 by @Kelimion mentioning the failure.

laytan:
Fixed the last test failure, there was an unaligned load because of the packed struct field being loaded on the types original alignment, we know track if a type is from a packed struct and set the alignment of loads to 1.

laytan:
Windows is technically supported here too, @JesseRMeyer has been test driving that this week. It just requires updating the dll and headers when we want to.

laytan:
FreeBSD amd64 has been tested by discord user siemato and confirmed to work on his projects

laytan:
Only thing I did was rebase and now the Linux tests fail

laytan:
Can't reproduce it locally either

laytan:
Seems to be a problem with the new string map, reverting  5969796fbfaa62ea6aba8f0e3915ab8282bb71b5 fixes this @gingerBill

laytan:
Oke put in a better fix to the packed struct loads that fixes the Linux failure and is a much neater fix.

So once again, and now for real, this is ready to go :)

src/checker.cpp
```diff
@@ -4310,7 +4310,7 @@ gb_internal bool correct_single_type_alias(CheckerContext *c, Entity *e) {
 
 gb_internal bool correct_type_alias_in_scope_backwards(CheckerContext *c, Scope *s) {
 	bool correction = false;
-	for (u32 n = s->elements.count, i = n-1; i < n; i--) {
+	for (u32 n = s->elements.capacity, i = n-1; i < n; i--) {
```
laytan, line 4313: This change is unrelated to the LLVM 18 PR but an existing bugfix.

To explain, this array is sparse/has holes so having 3 items doesn't mean it has one at index 1, 2 and 3. It might be at 1, 5, and 8 for example.

This means the entire capacity has to be looped and checked for valid indices inside the body.

src/build_settings.cpp
```diff
@@ -923,7 +922,18 @@ gb_internal isize MAX_ERROR_COLLECTOR_COUNT(void) {
 	return build_context.max_error_count;
 }
 
+#if defined(GB_SYSTEM_WINDOWS)
+	#include <llvm-c/Config/llvm-config.h>
+#else
+	#include <llvm/Config/llvm-config.h>
+#endif
 
+// NOTE: AMD64 targets had their alignment on 128 bit ints bumped from 8 to 16 (undocumented of course).
```
gingerBill, line 931: of course ;)

src/llvm_backend.hpp
```diff
@@ -71,6 +66,7 @@ struct lbProcedure;
 struct lbValue {
 	LLVMValueRef value;
 	Type *type;
+	bool is_packed;
```
gingerBill, line 69: This looks like a massive bodge. Can we not annotate it elsewhere?

laytan: Removed in favor of your approach in https://github.com/odin-lang/Odin/commit/de5ce90fa70c8ae4b859b50704dd0ffacbeedb1e

PR #3316, opened 2024-03-23
darwin: fix ld warnings and set minimum os version

laytan:
- we were basically hardcoding macosx11 as a target regardless of the -minimum-os-version flag (the flag was only passed on to the linker before)
- not setting a version in the triplet (like we did on amd64) results in a linker warning: `ld: warning: no platform load command found in '/Users/agostino/repos/odin/game/game.o', assuming: macOS`
- another linker warning `ld: warning: ignoring duplicate libraries: '-lSystem'` was showing up, I believe this is because clang automatically adds this if you don't specify `-nostdlib` and we should leave it out

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -510,7 +510,7 @@ gb_global TargetMetrics target_darwin_amd64 = {
 	TargetOs_darwin,
 	TargetArch_amd64,
 	8, 8, 8, 16,
-	str_lit("x86_64-apple-darwin"),
+	str_lit("x86_64-apple-macosx"), // NOTE: Changes during initialization based on build flags.
 	str_lit("e-m:o-i64:64-f80:128-n8:16:32:64-S128"),
 };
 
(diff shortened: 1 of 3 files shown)


laytan:
> the first warning is fixed but the second still there ``` ld: warning: no platform load command found in '/Users/agostino/repos/odin/online game/online game.o', assuming: macOS ```

Feedback from discord, it did fix it for me so I will have to look at it more, might be hard because I can't reproduce at this point

src/build_settings.cpp
```diff
@@ -510,15 +510,15 @@ gb_global TargetMetrics target_darwin_amd64 = {
 	TargetOs_darwin,
 	TargetArch_amd64,
 	8, 8, 8, 16,
-	str_lit("x86_64-apple-darwin"),
+	str_lit("x86_64-apple-macosx"), // NOTE: Changes during initialization based on build flags.
```
gingerBill, line 513: We'll need to change this string them when we support `iOS`

laytan: It is already changed to "arm64-apple-ios" or "x86_64-apple-ios" when you use `-subtarget:ios`

PR #3318, opened 2024-03-23
[core/os2]: Fix read_entire_file for reads from pipe and console.

flysand7:
In before the logic for reading from a file was as follows:

- Get the size of the file
- Allocate a buffer of that size
- Read all the data into the buffer

This ignores files that don't have a definite size (`/dev/std*`, `/dev/null`, named and anonymous pipes). In this PR I've added `.No_Size` error to package io and used it to tell whether we should read entire buffer at once (i.e. the old logic) or read it chunk by chunk.

### DIFF
--- a/core/os/os2/errors.odin
+++ b/core/os/os2/errors.odin
@@ -13,6 +13,10 @@ General_Error :: enum u32 {
 
 	Timeout,
 
+	// Indicates that an attempt to retrieve a file's size was made, but the
+	// file doesn't have a size.
+	No_Size,
+
 	Invalid_File,
 	Invalid_Dir,
 	Invalid_Path,
@@ -22,9 +26,15 @@ General_Error :: enum u32 {
 
 Platform_Error :: enum i32 {None=0}
 
+Read_Error :: enum u32 {
+	None,
+	Broken_Pipe,
+}
+
 Error :: union #shared_nil {
 	General_Error,
 	io.Error,
+	Read_Error,
 	runtime.Allocator_Error,
 	Platform_Error,
 }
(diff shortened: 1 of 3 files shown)


core/os/os2/errors.odin
```diff
@@ -22,9 +26,15 @@ General_Error :: enum u32 {
 
 Platform_Error :: enum i32 {None=0}
 
+Read_Error :: enum u32 {
+	None,
+	Broken_Pipe,
+}
```
gingerBill, line 32: Why is this a separate error type?

flysand7: It seems unfit to be `io.Error`, because that one deals with streams in general, also unfit for `General_Error`, because it can only occur during one of the two operations (read, write) and only for pipes, and it's not allocator or some generic platform error. By exclusion I deduced it must live in its own separate type, though now realising `Read_Error` is a bit unfitting. I can't think of a better name as of now.

Is it okay if it lives in`General_Error`?

gingerBill: I'd put this in `General_Error` regardless. Yes it can only happen in two cases, but people will want to handle it "general"ly.

flysand7: Done.

core/os/os2/file_util.odin
```diff
@@ -87,26 +87,50 @@ read_entire_file_from_path :: proc(name: string, allocator: runtime.Allocator) -
...
+				if err == .EOF {
+					err = nil
+				}
+				data = data[:total]
+				return
+			}
+		}
+	} else {
+		buffer: [1024]u8
+		out_buffer := make([dynamic]u8)
```
gingerBill, line 120: `out_buffer := make([dynamic]u8, 0, 0, allocator)`

flysand7: Fixed

gingerBill: Where is this fixed? Did you forget to push?

Issue #3291, opened 2024-03-18
Polymorphic types sometimes mess up the compilation


Adorjani2:
## Context

The following code sometimes doesn't compile and gives me different errors.
https://github.com/Adorjani2/odin-bug/tree/main

* Operating System & Odin Version:
        Odin:    dev-2024-03-nightly:4c35633e0
        OS:      Windows 10 Professional (version: 22H2), build 19045.4170
        CPU:     13th Gen Intel(R) Core(TM) i7-13700K
        RAM:     16139 MiB
        Backend: LLVM 17.0.1

## Expected Behavior

The code compiles 100% of the time.

## Current Behavior

Compilation sometimes fails.

## Failure Information (for bugs)

Compilation fails randomly around 5-10% of the time. I pasted the errors I have got after compiling it 100-120 times below and also in the repository's readme. It's rare, sometimes takes 10-15 tries before any error shows up and the compilation fails.
When the code compiles then it runs correctly, no runtime errors.
It seems like the polymorphic struct is messing things up, but I'm not entirely sure since I wasn't able to shrink the code size further down. When I tried to remove more code the errors stopped showing up.
Limiting the thread count to 1 when building made the errors show up way less but I still got them very very rarely.

### Steps to Reproduce

Build the project i linked above with `odin build .` until an error shows up

### Failure Logs
Error messages I got so far:

1:
-
```
ecs.odin(193:13) Error: '_data[index]' of type 'Component_Wrapper(T)' has no field 'component'
        return &_data[index].component, .None
                ^~~~~~~~~~~^
        Suggestion: Did you mean?
                component
```

2:
-
```
ecs.odin(304:34) Error: Too many values in structure literal, expected 0, got 2
        item := Component_Wrapper(T){ed, component}
```

3:
-
```
ecs.odin(235:23) Error: 'we' of type '^ecs.Component_Wrapper(T)' has no field 'component'
        append(&ret, &we.component)
                      ^^
```

4:
-
```
ecs.odin(235:9) Error: No procedures or ambiguous call for procedure group 'append' that match with the given arguments
        append(&ret, &we.component)
        ^~~~~^
        Given argument types: (^[dynamic]^Perspective_Camera_Component, ecs/ecs.odin(304:34) ^Component_Wrapper($T=Perspective_Camera_Component))
Did you mean to use one of the following:
[...]

Issue #3200, opened 2024-02-15
Race condition compiler bug related to #soa and anonymous procs


DireLines:
## Context

Tried adding a Script component to my game, it caused a compile error to *sometimes* happen. I pared the code down to this single file where the bug always happens. The bug seems to trigger depending on the order in which Odin evaluates the code. If GameObject is defined first, it triggers. If Game is defined first, it does not.

```cpp
package main

main :: proc() {
    game := new_game()
}

//it turns out the frequency with which the race condition is hit
//depends highly on where the definition of GameObject is
// above first mention of the type Game: fails
// below first mention of the type Game: works
//I have gotten half and half but lost what state the code was in when that happened
GameObject :: struct {
    name:          string, //commenting out either of these produces crazy error src/llvm_backend_expr.cpp(2254): Panic: Invalid type conversion: 'Allocator' to 'int' for procedure '_proclit$anon-1'
    component_set: bit_set[Component], //commenting out either of these produces crazy error src/llvm_backend_expr.cpp(2254): Panic: Invalid type conversion: 'Allocator' to 'int' for procedure '_proclit$anon-1'
    script:        Script,
}

Game :: struct {
    systems: [dynamic]System,
    objects: #soa[dynamic]GameObject,
}

new_game :: proc() -> Game {
    game := Game{}
    append(&game.systems, script_runner())
    return game
}

System :: struct {
    needObject: proc(system: ^System, game: ^Game, obj_index: int) -> bool,
}
script_runner :: proc() -> System {
    return System{needObject = proc(system: ^System, game: ^Game, obj_index: int) -> bool {
                obj := game.objects[obj_index] //bug lurks somewhere here
                // removing this line fixes one of the errors:
                // src/array.cpp(61): Assertion Failure: `cast(usize)index < cast(usize)count` Index 5 is out of bounds ranges 0..<3\\n
                return true
            }}
}

Script :: struct {
    awake: proc(self_index: int, game: ^Game),
}

Component :: enum {
    Script,
}
```

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System & Odin Version:
* Please paste `odin report` output:
```
	Odin: dev-2024-02:c5c2a4d09
	OS:   macOS Sonoma 14.2.1 (build: 23C71, kernel: 23.2.0)
	CPU:  ARM64
	RAM:  16384 MiB
```

## Expected Behavior

`odin build main.odin -file` successfully compiles

## Current Behavior

[...]

Issue #2949, opened 2023-11-11
Shared Generic across structs sometimes causes build errors


Skytrias:
## Context
        Odin: dev-2023-11:dd9b0ae4
        OS:   Windows 10 Home Basic (version: 22H2), build 19045.3570
        CPU:  AMD Ryzen 5 3600 6-Core Processor
        RAM:  16310 MiB

## Expected Behavior

Allow to use Generic $T normally across structs in different procedures.

## Current Behavior

crashes about 20% of the builds, otherwhise it compiles as expected. Maybe it's something about the build step and how structs get built internally with generics across threads?

### Steps to Reproduce

```odin
package src

A :: struct($T: typeid) {
    value: T,
}

B :: struct($T: typeid) {
    a: A(T)
}

// second proc is important!
test :: proc() {
    a: A(int) = { 10 }
    b: B(int) = { a }
}

main :: proc() {
    a: A(int) = { 10 }
    b: B(int) = { a }
}
```

### Failure Logs

```
.../main.odin(20:16) Cannot assign value 'a' of type 'A($T=int) (package src)' to 'A($T=int) (package src)' in structure literal
    b: B(int) = { a }
                  ^
    Suggestion: the expression may be directly casted to type A($T=int)
```

or 

```
.../main.odin(13:16) Too many values in structure literal, expected 0, got 1
    b: B(int) = { a }
```

JesseRMeyer:
I ran the compiler with `-fsanitize=thread` using your example and race hazards were detected and reported.  The reports are pretty easy to comprehend, but I don't know enough about the compiler internals to suggest fixes.  I suspect a few well placed semaphores will go a long way!

laytan:
The linked issue has been fixed, can you verify it also fixed this one?

JesseRMeyer:
Still happens, but the error rate isn't as frequent as before.

gingerBill:
Can this be replicated any more on the latest commit?

dozn:
I'm unable to reproduce this on multiple machines (Windows 11/Alpine Linux), ran it about fifty times on each.

Issue #2814, opened 2023-09-24
Frequent compile error with parapoly and/or test file


mtarik34b:
## Context

I have just these two files together in a package: `pkg/main.odin`, `pkg/test.odin`

```odin
/* main.odin */
package main

A :: struct {}
B :: struct {}
C :: struct {}
D :: struct {}
E :: struct {}
F :: struct {}
G :: struct {}

ZZZ :: struct($Key, $Value: typeid) {}

Big_Struct :: struct($T: typeid) {
	field_A: ZZZ(int, A),
	field_B: ZZZ(int, B),
	field_C: ZZZ(int, C),
	field_D: ZZZ(int, D),
	field_E: ZZZ(int, E),
	field_F: ZZZ(int, F),
	field_G: ZZZ(int, G),
}

create :: proc($T: typeid) -> Big_Struct(T) {
	bs: Big_Struct(T)

	zzz: ZZZ(int, A)
	bs.field_A = zzz

	return bs
}

main :: proc() {
	T :: [3]f32
	obj := create(T)
}
```

```odin
/* test.odin */
package main
 
import "core:testing"

@(test)
test :: proc(t: ^testing.T) {
	T :: [3]f32
	obj := create(T)
}
```

* Operating System & Odin Version:
```
Odin: dev-2023-09:c23b5825
OS:   Ubuntu 22.04.3 LTS, Linux 6.2.0-33-generic
CPU:  Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
RAM:  15747 MiB
```
## Expected Behavior

I expect the code to compile successfully always(!).

## Current Behavior

When using `odin run .` or `odin test .` I oftentimes get one of these three compile errors randomly:

```
.../pkg/main.odin(27:2) Error: 'bs' of type 'Big_Struct(T)' has no field 'field_A'
        bs.field_A = zzz
        ^^
```
```
.../pkg/main.odin(27:2) Error: 'bs' of type 'Big_Struct(T)' has no field 'field_A'
        bs.field_A = zzz
        ^^
        Suggestion: Did you mean?
                field_A
                field_B
                field_C
                field_D
                field_E
                field_F
                field_G
```
```
.../pkg/main.odin(27:15) Error: Cannot assign value 'zzz' of type 'ZZZ($Key=int, $Value=A) (package main)' to 'ZZZ($Key=int, $Value=A) (package main)' in assignment
        bs.field_A = zzz
                     ^~^
[...]

PR #3326, opened 2024-03-24
Fix fields_wait_signal futex.

### LINKED ISSUE #2463: Parapoly Types are type checked incorrectly due to race condition
        Odin: dev-2023-04:b42bb5be
        OS:   Windows 10 Home Basic (version: 21H2), build 19044.2728
        CPU:  AMD Ryzen 7 5800X 8-Core Processor
        RAM:  32711 MiB

The type checker will randomly fail with a parapoly type due to a race condition. An error will be reported that a field is missing.
I can only reproduce this with parapoly types that are defined in separate procedures.
Take the following minimal failure case:
```odin
package will_eventually_fail

Foo :: struct($N: int) {
    a: u32,
}

bar :: proc(foo: ^Foo($N)) {
    foo.a = 1
}

a :: proc() {
	foo: Foo(1)
	bar(&foo)
}

b :: proc() {
	foo: Foo(1)
	bar(&foo)
}
```
Running Odin check will pass majority of the time. Every so often (10-100 runs) I will get
```
will_eventually_fail.odin(8:5) 'foo' of type '^Foo(1)' has no field 'a'
        foo.a = 1
        ^~^
```
Which is incorrect.
Now if I define `Foo(1)` in a and make b define `Foo(2)` instead then I never get the error (I have done 10k + runs).
```odin
package fine

Foo :: struct($N: int) {
    a: u32,
}

bar :: proc(foo: ^Foo($N)) {
    foo.a = 1
}

a :: proc() {
	foo: Foo(1)
	bar(&foo)
}

a :: proc() {
	foo: Foo(2) // <-- the only change
	bar(&foo)
}
```

My suspicion is that all the information to resolve `Foo(1)` is being done by one thread when another thread is checking it when the type information is half completed.

### LINKED ISSUE #2679: The new nightly version seems to have some bugs
## Context

I upgraded to the newest version of Odin (nightly) and my old code would not compile.

Odin: dev-2023-07-nightly:5ac7fe45
OS:   Windows 11 Home Basic (version: 22H2), build 22621.1992
RAM:  32507 MiB

## Expected Behavior

I would expect my code to compile, the compiler gives me errors that make no sense.

## Current Behavior

The compiler gives me errors like:
```
Cannot assign value 'entries[index]' of type '
Octant($Holding=Empty, $Index_type=u8) (package octree)
' to '
Octant($Holding=Empty, $Index_type=u8) (package octree)
' in assignment
```

```
Cannot assign value '_get_sort_encoded(Holding, Index_type, "prune")' of type '
proc(Octant($Holding=Empty, $Index_type=u8), Octant($Holding=Empty, $Index_type=u8)) -> bool
' to '
proc(Octant($Holding=Empty, $Index_type=u8), Octant($Holding=Empty, $Index_type=u8)) -> bool
' in procedure argument
```

```
No procedures or ambiguous call for procedure group 'append' that match with the given arguments
        append(&entries, e);
        ^~~~~^
        Given argument types: (^[dynamic]Octant($Holding=Empty, $Index_type=u8), Octant($Holding=Empty, $Index_type=u8))
```

This used to work, and they types match as you can see, but it says they don't

I currently have my directory setup like this:
```
my_game
	-> octree package (includes 2 files a test file and the octree data structure)
	-> network package (has 5 files, 1 is a test file)
	-> ...
	-> ...
```
They bugs only occure in my octree package, and they only occure when I and compiling from my game package (which include the octree package). If I "cd octree" and then compile my tests in there, they is no compliants.

### Steps to Reproduce

I could not tell you as a this only seems to be a bug in some packages and not others, and it only seems to be a bug with `odin run .` or `odin build .` , but not `odin test .`.

EDIT:

As you can see from this failure log, it can acctually compile, sometimes but it just failes like 80% and the time.
I havec experienced this before, but never as freqently as in this new release.
### Failure Logs

```

PS D:\github\Voxel_game> odin run .
D:/github/Voxel_game/Octree/Octree.odin(46:5) Error: No procedures or ambiguous call for procedure group 'append' that match with the given arguments
        append(&entries, Octant(Holding, Index_type){encoding = enc, dat ...
        ^~~~~^
        Given argument types: (^[dynamic]Octant($Holding=Empty, $Index_type=u8), Octant($Holding=Empty, $Index_type=u8))
Did you mean to use one of the following:
        runtime.append_elem        :: proc(array: ^$T/[dynamic]$E, arg: E, loc := #caller_location) -> (n: int, err: Allocator_Error)            at C:/Program Files/Odin/core/runtime/core_builtin.odin(410:1)
        runtime.append_elems       :: proc(array: ^$T/[dynamic]$E, args: ..E, loc := #caller_location) -> (n: int, err: Allocator_Error)         at C:/Program Files/Odin/core/runtime/core_builtin.odin(438:1)
        runtime.append_elem_string :: proc(array: ^$T/[dynamic]$E/u8, arg: $A/string, loc := #caller_location) -> (n: int, err: Allocator_Error) at C:/Program Files/Odin/core/runtime/core_builtin.odin(473:1)

D:/github/Voxel_game/Octree/Octree.odin(130:13) Error: No procedures or ambiguous call for procedure group 'append' that match with the given arguments
        append(&entries, e);
        ^~~~~^
        Given argument types: (^[dynamic]Octant($Holding=Empty, $Index_type=u8), Octant($Holding=Empty, $Index_type=u8))
Did you mean to use one of the following:
        runtime.append_elem        :: proc(array: ^$T/[dynamic]$E, arg: E, loc := #caller_location) -> (n: int, err: Allocator_Error)            at C:/Program Files/Odin/core/runtime/core_builtin.odin(410:1)
        runtime.append_elems       :: proc(array: ^$T/[dynamic]$E, args: ..E, loc := #caller_location) -> (n: int, err: Allocator_Error)         at C:/Program Files/Odin/core/runtime/core_builtin.odin(438:1)
        runtime.append_elem_string :: proc(array: ^$T/[dynamic]$E/u8, arg: $A/string, loc := #caller_location) -> (n: int, err: Allocator_Error) at C:/Program Files/Odin/core/runtime/core_builtin.odin(473:1)

D:/github/Voxel_game/Octree/Octree.odin(139:35) Error: Cannot assign value '_get_sort_encoded(Holding, Index_type, "prune")' of type 'proc(Octant($Holding=Empty, $Index_type=u8), Octant($Holding=Empty, $Index_type=u8)) -> bool' to 'proc(Octant($Holding=Empty, $Index_type=u8), Octant($Holding=Empty, $Index_type=u8)) -> bool' in procedure argument
        ... ies[:], _get_sort_encoded(Holding, Index_type, "prune")); ...
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^

PS D:\github\Voxel_game> cd .\Octree\
PS D:\github\Voxel_game\Octree> odin run .  
D:/github/Voxel_game/Octree/Test.odin(65:27) 
Error: 'o' of type 'Octree(Empty, u8)' has no field 'entries'
        testing.expect(t, len(o.entries) == 0);
                              ^
        Suggestion: Did you mean?
                entries

D:/github/Voxel_game/Octree/Octree.odin(1:2) Error: Undefined entry point procedure 'main'

         ^
PS D:\github\Voxel_game\Octree> odin run .  
D:/github/Voxel_game/Octree/Test.odin(65:27) Error: 'o' of type 'Octree(Empty, u8)' has no field 'entries'
        testing.expect(t, len(o.entries) == 0);
                              ^
        Suggestion: Did you mean?
                entries

D:/github/Voxel_game/Octree/Octree.odin(1:2) Error: Undefined entry point procedure 'main'

         ^
PS D:\github\Voxel_game\Octree> odin run .
D:/github/Voxel_game/Octree/Test.odin(65:27) Error: 'o' of type 'Octree(Empty, u8)' has no field 'entries'
        testing.expect(t, len(o.entries) == 0);
                              ^
        Suggestion: Did you mean?
                entries

D:/github/Voxel_game/Octree/Octree.odin(1:2) Error: Undefined entry point procedure 'main'

         ^
PS D:\github\Voxel_game\Octree> odin run .
D:/github/Voxel_game/Octree/Octree.odin(1:2) Error: Undefined entry point procedure 'main'

         ^
PS D:\github\Voxel_game\Octree> odin test .        
[Package: octree]
[Test: test_1]
[test_1 : SUCCESS]
[Test: test_2]
_bin_search
[test_2 : SUCCESS]
----------------------------------------
2/2 SUCCESSFUL
PS D:\github\Voxel_game\Octree> 
```

rick-masters:
Fixes #2463

Also partially resolves #2679, #2814, #2949, #3200, and #3291.

The missing field problem is a thread concurrency problem with checking structure fields.
One thread is expecting another thread to populate the fields of a structure and the first thread waits until that is done.
But even after waiting, the fields are not populated.

The root cause of this problem is that the function `wait_signal_until_available` which waits for a futex is not functional.

The futex `Struct.fields_wait_signal` is a value that starts at 0 and changes to 1 to indicate completion of `check_struct_fields` for a structure. The futex is waited on by making a `SYS_futex` system call and providing the *current expected* value of the futex. The system call then returns when another thread changes the value. However, if the current futex value is *not* the current expected value that was provided, then the system call returns immediately.

So, the expectation is that the code would call `SYS_futex` with a value of 0.
However, the code is erroneously making the call (via `futex_wait`) with the value of 1 and so the call returns immediately:
https://github.com/odin-lang/Odin/blob/517d7ae0b0fd400ceb6a213e7d644c19b8088bfd/src/threading.cpp#L114-L118

One may wonder how it avoids blocking indefinitely when the futex is already 1. Well, as you can see above, it only makes the system call after checking that the futex is currently zero. I suppose this is an optimization to avoid making an unnecessary system call when the futex is already completed but doing that just so happens to avoid hanging on the completed futex as well.

Unfortunately, fixing the futex reveals that there are problems where the code waits on the futex but the futex is never set. I had to add several calls to set the futex where they were missing. Unfortunately, this means that there may be other situations not yet identified in my testing where a futex wait occurs but the futex is never signaled.

I tried to find any potential lockups by building the demo, running all the tests, and building the [ols](https://github.com/DanielGavin/ols) project. Nevertheless, this fix could possibly reveal new lockups of the compiler. If one experiences a new compiler lockup that is traced to this PR then there is probably another missing `wait_signal_set` call that needs to be tracked down.

Unfortunately, in my testing I identified a design flaw that results in a compiler lock up even when `wait_signal_set` is performed correctly. So, this PR should not be merged until that issue is evaluated.

### DIFF
--- a/src/threading.cpp
+++ b/src/threading.cpp
@@ -113,7 +113,7 @@ struct Wait_Signal {
 
 gb_internal void wait_signal_until_available(Wait_Signal *ws) {
 	if (ws->futex.load() == 0) {
-		futex_wait(&ws->futex, 1);
+		futex_wait(&ws->futex, 0);
 	}
 }
 
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -3393,6 +3393,7 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 			elem->Struct.tags = gb_alloc_array(permanent_allocator(), String, fields.count);
 			elem->Struct.node = dummy_node_struct;
 			type_set_offsets(elem);
+			wait_signal_set(&elem->Struct.fields_wait_signal);
 		}
 
 		Type *soa_type = make_soa_struct_slice(c, dummy_node_soa, nullptr, elem);
@@ -3766,6 +3767,7 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 				soa_struct->Struct.tags[i] = old_struct->Struct.tags[i];
 			}
 		}
+		wait_signal_set(&soa_struct->Struct.fields_wait_signal);
 
 		Token token = {};
 		token.string = str_lit("Base_Type");
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -8873,6 +8873,7 @@ gb_internal ExprKind check_compound_literal(CheckerContext *c, Operand *o, Ast *
 			break;
 		}
 
+		wait_signal_until_available(&t->Struct.fields_wait_signal);
 		isize field_count = t->Struct.fields.count;
 		isize min_field_count = t->Struct.fields.count;
 		for (isize i = min_field_count-1; i >= 0; i--) {
--- a/src/check_type.cpp
+++ b/src/check_type.cpp
@@ -2490,6 +2490,7 @@ gb_internal Type *get_map_cell_type(Type *type) {
 	s->Struct.fields[0] = alloc_entity_field(scope, make_token_ident("v"), alloc_type_array(type, len), false, 0, EntityState_Resolved);
 	s->Struct.fields[1] = alloc_entity_field(scope, make_token_ident("_"), alloc_type_array(t_u8, padding), false, 1, EntityState_Resolved);
 	s->Struct.scope = scope;
+	wait_signal_set(&s->Struct.fields_wait_signal);
 	gb_unused(type_size_of(s));
 
 	return s;
@@ -2520,6 +2521,7 @@ gb_internal void init_map_internal_types(Type *type) {
 	metadata_type->Struct.fields[4] = alloc_entity_field(metadata_scope, make_token_ident("value_cell"), value_cell, false, 4, EntityState_Resolved);
 	metadata_type->Struct.scope = metadata_scope;
 	metadata_type->Struct.node = nullptr;
+	wait_signal_set(&metadata_type->Struct.fields_wait_signal);
 
 	gb_unused(type_size_of(metadata_type));
 
@@ -2537,6 +2539,7 @@ gb_internal void init_map_internal_types(Type *type) {
 	debug_type->Struct.fields[3] = alloc_entity_field(scope, make_token_ident("__metadata"), metadata_type, false, 3, EntityState_Resolved);
 	debug_type->Struct.scope = scope;
 	debug_type->Struct.node = nullptr;
+	wait_signal_set(&debug_type->Struct.fields_wait_signal);
 
 	gb_unused(type_size_of(debug_type));
 
@@ -2832,6 +2835,7 @@ gb_internal Type *make_soa_struct_internal(CheckerContext *ctx, Ast *array_typ_e
 	add_entity(ctx, scope, nullptr, base_type_entity);
 
 	add_type_info_type(ctx, soa_struct);
+	wait_signal_set(&soa_struct->Struct.fields_wait_signal);
 
 	return soa_struct;
 }


rick-masters:
The new issue revealed if this PR is merged is described in Issue #3327

src/threading.cpp
```diff
@@ -113,7 +113,7 @@ struct Wait_Signal {
 
 gb_internal void wait_signal_until_available(Wait_Signal *ws) {
 	if (ws->futex.load() == 0) {
-		futex_wait(&ws->futex, 1);
+		futex_wait(&ws->futex, 0);
```
gingerBill, line 116: Why was this changed to `0` from `1`?

gingerBill:
I have merged this even if it completely breaks: https://github.com/odin-lang/Odin/issues/3327

Issue #3327, opened 2024-03-24
Type check loop with soa struct may result in compiler lock up


rick-masters:
## Context

There is a design flaw that could result in a compiler lock up. However, the lock up is currently hidden by another bug reported in Issue #2463. The `fields_wait_signal` futex does not work, but if that is fixed by merging PR #3326, a new problem is revealed.
The current type checking code for a structure can lock up when #soa structures are combined with a "type check loop".
This is illustrated by the following example derived from Issue #3200:

```
package main
main :: proc() {}
GameObject :: struct {
    game: ^Game
}
Game :: struct {
    objects: #soa[dynamic]GameObject
}
```
In this example the following type checking steps occur:
 1. GameObject is checked by `check_struct_type`
 2. For GameObject, `check_struct_type` calls `check_struct_fields`
 3. For field `game`, the `Game` type is checked.
 4. Game is checked by `check_struct_type`
 5. For Game, `check_struct_type` calls `check_struct_fields`
 6. The field `objects` is "a soa struct dynamic array"
 7. `make_soa_struct_dynamic_array` is called, which calls `make_soa_struct_internal`
 8. `make_soa_struct_internal` calls `wait_signal_until_available(&old_struct->Struct.fields_wait_signal);`

The `old_struct` in step 8 is the GameObject type, and step 2 is currently checking the structure fields.
Therefore `wait_signal_until_available` never returns because it is waiting for the higher level operation/function currently in process.

Note that if `Game` is moved above `GameObject` then the problem does not occur.
This lead me to speculate that the compiler could somehow resolve this by reordering internally.
However, the following contrived example does not have such a solution:
```
package main
main :: proc() {}
GameObject :: struct {
    game_ptr: ^Game,
    objects: #soa[dynamic]Game,
}
Game :: struct {
    gameobj_ptr: ^GameObject,
    objects: #soa[dynamic]GameObject,
}
```

I suppose this second example is invalid. I included it because it might require a different solution to detect cycles (including indirect ones) and report that error rather than locking up.

The first example however seems solvable perhaps by recording all named structure types before trying to validate fields, so that ^Game can be validated superficially rather than attempting a full recursive type check of the "newly encountered" structure type. Such a fix would be sadly well beyond my experience with this code base.

Another course might be to just merge the PR #3326 and put this case back on the user to rearrange the definitions if necessary to avoid a lock up until this issue is resolved.

* Operating System & Odin Version:
```
        Odin:    dev-2024-03:34247b265
[...]

PR #3328, opened 2024-03-24
Fix race between find and add of new polymorphic type.

### LINKED ISSUE #2679: The new nightly version seems to have some bugs

rick-masters:
# Explanation of the PR

Partially fixes #2679, #2814, #2949, and #3291.

This PR fixes a race condition that exists across multiple threads, between calling `find_polymorhic_record_entity` and `add_polymorphic_record_entity`. Two or more threads can create the same type independantly because they may both call `find_polymorphic_record_entity` around the same time, don't find it, and then both proceed to create and then add the same type with `add_polymorphic_record_entity`.

The proposed solution here is for a lock to be held between looking up an entity (and not finding it) and subsequently adding the entity. This will lock out other threads and when another thread finally gets the lock to lookup the entity, it
will already be added and it won't attempt to add the entity again.

Specifically, if `find_polymorphic_record_entity` does not find the entity it leaves the the relevant (local) `gen_types` mutex locked. In combination, if the caller must subsequenty call `add_polymorphic_record_entity` and pass the flag
`local_gen_types_locked = true` so that `add_polymorphic_record_entity` knows that locking the local `gen_types` mutex has already been done.

Note that in `find_polymorphic_record_entity` the global lock must be released before locking the local lock. Otherwise, a deadlock can occur if:
1. Thread 1 takes the local lock in `find_polymorphic_record_entity`
2. Thread 2 goes into `find_polymorphic_record_entity` takes the global lock but can't get the local lock so it stops, holding the global lock.
3. Thread 1 goes into `add_polymorphic_record_entity`, and attempts to get the global lock. But it can't because Thread 1 has it. So thread 1 can't finish its work and then release the local lock. This is a deadlock.

While this is a relatively small fix, the downside is that dividing the lock and unlock into separate functions does make the code a bit more difficult to reason about. Perhaps a better solution can be identified, but from my testing it 100%
fixes this problem which has ben reported several times and at least demonstrates that the problem has been correctly diagnosed.

### DIFF
--- a/src/check_type.cpp
+++ b/src/check_type.cpp
@@ -265,14 +265,15 @@ gb_internal Entity *find_polymorphic_record_entity(CheckerContext *ctx, Type *or
 
 	auto *found_gen_types = map_get(&ctx->info->gen_types, original_type);
 	if (found_gen_types == nullptr) {
-		rw_mutex_shared_unlock(&ctx->info->gen_types_mutex); // @@global
-		return nullptr;
+		GenTypesData gen_types = {};
+		gen_types.types = array_make<Entity *>(heap_allocator());
+		map_set(&ctx->info->gen_types, original_type, gen_types);
+		found_gen_types = map_get(&ctx->info->gen_types, original_type);
 	}
 
-	rw_mutex_shared_lock(&found_gen_types->mutex); // @@local
-	defer (rw_mutex_shared_unlock(&found_gen_types->mutex)); // @@local
-
 	rw_mutex_shared_unlock(&ctx->info->gen_types_mutex); // @@global
+
+	rw_mutex_shared_lock(&found_gen_types->mutex); // @@local
 
 	for (Entity *e : found_gen_types->types) {
 		Type *t = base_type(e->type);
@@ -334,14 +335,16 @@ gb_internal Entity *find_polymorphic_record_entity(CheckerContext *ctx, Type *or
 			}
 		}
 		if (!skip) {
+			rw_mutex_shared_unlock(&found_gen_types->mutex);
 			return e;
 		}
 	}
+	// found_gen_types->mutex will be unlocked by add_polymorphic_record_entity
 	return nullptr;
 }
 
 
-gb_internal void add_polymorphic_record_entity(CheckerContext *ctx, Ast *node, Type *named_type, Type *original_type) {
+gb_internal void add_polymorphic_record_entity(CheckerContext *ctx, Ast *node, Type *named_type, Type *original_type, bool local_gen_types_locked) {
 	GB_ASSERT(is_type_named(named_type));
 	gbAllocator a = heap_allocator();
 	Scope *s = ctx->scope->parent;
@@ -370,7 +373,9 @@ gb_internal void add_polymorphic_record_entity(CheckerContext *ctx, Ast *node, T
 	rw_mutex_lock(&ctx->info->gen_types_mutex);
 	auto *found_gen_types = map_get(&ctx->info->gen_types, original_type);
 	if (found_gen_types) {
-		rw_mutex_lock(&found_gen_types->mutex);
+		if (!local_gen_types_locked) {
+			rw_mutex_lock(&found_gen_types->mutex);
+		}
 		array_add(&found_gen_types->types, e);
 		rw_mutex_unlock(&found_gen_types->mutex);
 	} else {
@@ -613,7 +618,7 @@ gb_internal bool check_record_poly_operand_specialization(CheckerContext *ctx, T
 }
 
 
-gb_internal void check_struct_type(CheckerContext *ctx, Type *struct_type, Ast *node, Array<Operand> *poly_operands, Type *named_type, Type *original_type_for_poly) {
+gb_internal void check_struct_type(CheckerContext *ctx, Type *struct_type, Ast *node, Array<Operand> *poly_operands, Type *named_type, Type *original_type_for_poly, bool local_gen_types_locked) {
 	GB_ASSERT(is_type_struct(struct_type));
 	ast_node(st, StructType, node);
 
@@ -650,7 +655,7 @@ gb_internal void check_struct_type(CheckerContext *ctx, Type *struct_type, Ast *
 	struct_type->Struct.is_poly_specialized = check_record_poly_operand_specialization(ctx, struct_type, poly_operands, &struct_type->Struct.is_polymorphic);
 	if (original_type_for_poly) {
 		GB_ASSERT(named_type != nullptr);
-		add_polymorphic_record_entity(ctx, node, named_type, original_type_for_poly);
+		add_polymorphic_record_entity(ctx, node, named_type, original_type_for_poly, local_gen_types_locked);
 	}
 
 	if (!struct_type->Struct.is_polymorphic) {
@@ -685,7 +690,7 @@ gb_internal void check_struct_type(CheckerContext *ctx, Type *struct_type, Ast *
 
 #undef ST_ALIGN
 }
-gb_internal void check_union_type(CheckerContext *ctx, Type *union_type, Ast *node, Array<Operand> *poly_operands, Type *named_type, Type *original_type_for_poly) {
+gb_internal void check_union_type(CheckerContext *ctx, Type *union_type, Ast *node, Array<Operand> *poly_operands, Type *named_type, Type *original_type_for_poly, bool local_gen_types_locked) {
 	GB_ASSERT(is_type_union(union_type));
 	ast_node(ut, UnionType, node);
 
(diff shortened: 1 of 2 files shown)


gingerBill:
Your changes have caused a dead-lock now on Windows.

src/check_type.cpp
```diff
@@ -370,7 +373,9 @@ gb_internal void add_polymorphic_record_entity(CheckerContext *ctx, Ast *node, T
 	rw_mutex_lock(&ctx->info->gen_types_mutex);
 	auto *found_gen_types = map_get(&ctx->info->gen_types, original_type);
 	if (found_gen_types) {
-		rw_mutex_lock(&found_gen_types->mutex);
+		if (!local_gen_types_locked) {
+			rw_mutex_lock(&found_gen_types->mutex);
+		}
 		array_add(&found_gen_types->types, e);
 		rw_mutex_unlock(&found_gen_types->mutex);
```
gingerBill, line 380: Does this require `local_gen_types_locked` too?

rick-masters:
> Your changes have caused a dead-lock now on Windows.

Unfortunately I don't have a Windows machine handy to look at this right away. Are you able to reproduce this? Strange that it only happens on Windows. I've looked at the futex implementation and it should be the same.

src/check_type.cpp
rick-masters: Sorry, I don't understand. This is the only place it is checked and it's required to avoid locking twice.

gingerBill:
I've just made some changes to see if I could fix it myself. I am not sure if this will fix everything though.

rick-masters:
> I've just made some changes to see if I could fix it myself. I am not sure if this will fix everything though.

OK, I'm curious about what it could be.

I did see a lockup in my testing but it was unrelated to this change. It was actually the broken futex which is now fixed by #3326. In rare cases, one threads confirms that it's zero, another thread sets it to one, then the first thread blocks on it changing from 1. But that lockup was rare and it seems unlikely to be the cause here unless this change somehow caused it to happen more frequently. Anyway, #3326 fixes that so if it doesn't happen again maybe that was it but I doubt it...

gingerBill:
@rick-masters I'm currently focusing on https://github.com/odin-lang/Odin/issues/3327 since that is a proper dead lock due to a cycle.

Issue #3204, opened 2024-02-17
Race condition related to #soa and proc arguments 2


DireLines:
## Context

Tried adding a Script component to my game, it caused a compile error to *sometimes* happen. I pared the code down to this single file where the bug happens a little less than half the time. The bug seems to trigger depending on the order in which Odin evaluates the code. If objects is listed first in the definition of Game, the issue seems to be fixed

```cpp
package main

import "core:time"

main :: proc() {
    game := Game{}
}

Input :: struct {}

Game :: struct {
    input_system: ^Input, //removing this or reordering below objects fixes the bug 
    objects:      #soa[dynamic]GameObject,
    start_tick:   time.Tick, //removing this fixes the bug 
}

GameObject :: struct {
    awake: proc(game: ^Game), //making it so that no args are of type ^Game fixes the bug
}
```

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System & Odin Version:
* Please paste `odin report` output:
```
	Odin: dev-2024-02:c5c2a4d09
	OS:   macOS Sonoma 14.2.1 (build: 23C71, kernel: 23.2.0)
	CPU:  ARM64
	RAM:  16384 MiB
```

## Expected Behavior

`odin build main.odin -file` successfully compiles

## Current Behavior

`Internal Compiler Error: Type_Info for 'Tick' could not be found`
``fish: Job 1, 'odin build main.odin -file' terminated by signal SIGTRAP (Trace or breakpoint trap)``

### Steps to Reproduce

1. save this code to a file, `main.odin`
2. run `odin build main.odin -file` or `odin build .`
3. This seems to be a race condition and the error occurs about 30% of the time, the other 70% build successfully

rick-masters:
I can still reproduce this with https://github.com/odin-lang/Odin/commit/1009182f7b35e38e0fba375ad830fc609a7be831:

```
[rick@hb-ubuntu-20.04 3204]$ ../../Odin-latest/odin build .
Internal Compiler Error: Type_Info for 'Tick' could not be found
Illegal instruction (core dumped)
[rick@hb-ubuntu-20.04 3204]$ ../../Odin-latest/odin report
Where to find more information and get into contact when you encounter a bug:

        Website: https://odin-lang.org
        GitHub:  https://github.com/odin-lang/Odin/issues


Useful information to add to a bug report:

        Odin:    dev-2024-03:1009182f7
[...]

PR #3335, opened 2024-03-26
Don't add type info for incomplete structs.

rick-masters:
Fixes Issue #3204

As part of generating the minimum dependency set, the fields of a structure may not being processed by `add_min_dep_type_info`.

From the example in Issue #3240, the structure type is `Game` and the field is `start_tick` field of type time.Tick.

This can happen because `add_min_dep_type_info` thinks that the Game type has already been processed.
That's because it already found that type's index in `minimum_dependency_type_info_set`.

The problem is that a *different* type was already added to `minimum_dependency_type_info_set` using the same type index.

(In some cases the Game type is processed first so the problem doesn't happen.)

The indexes are into the array `type_info_types`, and these indexes are stored in the `type_info_map` as an optimization.

So, the problem is that two different types have the same index.

The two types where originally added to the `type_info_types` array by `add_type_info_type_internal` in a process initiated by `check_all_global_entities`.

The root cause is very similar to that described by Issue #3327, described as a "type check loop".
This issue could be described as a "type add loop" problem in which the Game type is added to the array before it's fields have been added. So it is added as an empty structure. Later another (actually) empty structure is processed and it appears to be identical so it ends up with the same index.

The proposed solution is to skip structures at this stage that have not had their fields added yet.
Later, the minimum dependency processing will add the type if it hasn't been added yet.
By this time the fields should be defined, especially in light of the recent fix to #3327.

I suspect that there may be a better way to solve this but hopefully this solution at least helps confirm the diagnosis.
For example, maybe there should be `wait_signal_set(&bt->Struct.fields_wait_signal);` before this code:
https://github.com/odin-lang/Odin/blob/1009182f7b35e38e0fba375ad830fc609a7be831/src/checker.cpp#L2049-L2052
However, adding that line will reveal an easily reproducible deadlock that would require a different solution.


rick-masters:
Darn, I accidentally added `tests/core/test_core_time`. I think .gitignore needs that.

src/checker.cpp
```diff
@@ -2028,6 +2028,8 @@ gb_internal void add_type_info_type_internal(CheckerContext *c, Type *t) {
 		break;
 
 	case Type_Struct:
+		if (bt->Struct.fields_wait_signal.futex.load() == 0)
+			return;
```
gingerBill, line 2032: This "might" work, but I worried it'll break something later. But maybe that is the better solution at any rate.

Issue #3343, opened 2024-03-27
Package with large amount of Objective-C wrapper procs exhibits slow compilation time


harold-b:
## Context

The package https://github.com/harold-b/darwodin/ exhibits compilations times of over 3 seconds on an M3 Max 64GB using 16 threads.

The package wraps a handful of Apple Objective-C frameworks and contains a large amount of procs. They are separated over several files, but initially they were contained in a single file per package where the same behavior was observed.

* Operating System & Odin Version: macOS Sonoma 14.4
* Please paste `odin report` output:
```
        Odin:    dev-2024-03:f9bfaeb16
        OS:      macOS Sonoma 14.2.1 (build: 23E214, kernel: 23.2.0)
        CPU:     ARM64
        RAM:     65536 MiB
        Backend: LLVM 14.0.6
        ```

### Steps to Reproduce

```bash
git clone  https://github.com/harold-b/darwodin && cd darwodin
./build -debug
```

### Relevant Logs
[Without `-use-separate-modules` ](https://gist.github.com/harold-b/eb9e2342e59606605d998fde04370630)
[With `-use-separate-modules`](https://gist.github.com/harold-b/52bf1ab918b04b2cd7298fc41ac842ba)

JesseRMeyer:
I'll highlight the egregious

> check procedure bodies                                                 -  2363.033 ms -  72.54%

line item from either timing log.

gingerBill:
There is a huge amount of code.

I'll see what the slow down is.

gingerBill:
if I use `darwodin-macos-lite`, it compiles fast, but with `darwodin-macos` I hit a "Thread Queue Full!" issue. You do have 0.5 Million lines of Odin code in that directory, but that shouldn't be "too much".

laytan:
compile times should look much better now

harold-b:
I'll run these changes here shortly and report back

harold-b:
I've updated the logs linked in gists above and can confirm a significant speed-up. To the point that `check procedure bodies` is no longer consumes the most amount of build time.
Thank you all for your help!

PR #3344, opened 2024-03-28
compiler: speed up path_to_fullpath on Linux/MacOS

laytan:
We/@colrdavidson  did some profiling for #3343 and this seems to be the biggest problem. `realpath` is expensive, and we are locking here for no reason that I can think of.

This improves the "check procedure bodies" timing (of the linked issue) from 2.4s to .4s on my machine (with the compiler in debug mode).


laytan:
There is also a lock in the windows implementation where I don't know why it is there?

PR #3347, opened 2024-03-29
Fix incorrect timings on macOS

harold-b:
This fixes incorrect timings being reported on macOS.

macOS raw timing API is a bit of an outlier compared to the other platforms with regards to frequency handling. This change makes it so that the operations to handle the tick conversion is done at the time `mach_absolute_time()` itself is called, returning the ticks as nanoseconds. `osx_time_stamp__freq()` is then changed to return a constant value to convert from nanoseconds to seconds.

### DIFF
--- a/src/timings.cpp
+++ b/src/timings.cpp
@@ -44,12 +44,12 @@ gb_internal mach_timebase_info_data_t osx_init_timebase_info(void) {
 }
 
 gb_internal u64 osx_time_stamp_time_now(void) {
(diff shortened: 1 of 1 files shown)


harold-b:
I curiously had a look at my own older high-performance timing code and noticed the issue. I  reverted  it back (13b8a5b) to the actual frequency calculation, with the fix in (multiply by `denom` *then* divide by `numer`, as opposed to the other way around. Either commit works, though.

PR #3350, opened 2024-03-29
fix gb.h to be able to use -fsanitize=address

laytan:
With these fixes -fsanitize=address at least works on an empty project.


laytan:
Hmm does something like `__attribute__((disable_sanitizer_instrumentation))` exist for windows?

PR #3365, opened 2024-04-02
Core Foundation and Security core packages.

yay:
This is an attempt to organize bindings kindly added by @laytan in a more idiomatic way and consistent with existing vendor libraries and Apple's API breakdown by framework. This also populates all `OSStatus` error codes.


core/crypto/rand_darwin.odin
```diff
@@ -1,12 +1,16 @@
...
-	res := darwin.SecRandomCopyBytes(count=len(dst), bytes=raw_data(dst))
-	if res != .Success {
-		msg := darwin.CFStringCopyToOdinString(darwin.SecCopyErrorMessageString(res))
-		panic(fmt.tprintf("crypto/rand_bytes: SecRandomCopyBytes returned non-zero result: %v %s", res, msg))
+	err := SecRandomCopyBytes(count=len(dst), bytes=raw_data(dst))
+	if err != 0 {
+		panic(fmt.tprintf("crypto/rand_bytes: SecRandomCopyBytes returned non-zero result: %v", err))
 	}
 }
+
```
yay, line 13: There's a little bit of duplication here to prevent importing `CoreFoundation` vendor library in `core:crypto`, which I hope is fine.

laytan:
IMO the non-objectivec bindings belong in core:sys/darwin and not in vendor

laytan:
I do like the separation into specific framework sub packages though

yay:
@laytan Something like this? https://github.com/odin-lang/Odin/pull/3365/commits/04e4d9a2fd66d69f076d879de3b5c9b329ac0056

vendor/darwin/Foundation/NSString.odin
```diff
@@ -1,39 +1,12 @@
...
-	UTF16             = Unicode,
-
-	UTF16BigEndian    = 0x90000100,
-	UTF16LittleEndian = 0x94000100,
-
-	UTF32             = 0x8c000100,
-	UTF32BigEndian    = 0x98000100,
-	UTF32LittleEndian = 0x9c000100,
-}
-
```
yay, line 36: BTW, are these correct? I see the following in Xcode:
```swift
/* Identifier for character encoding; the values are the same as Text Encoding Converter TextEncoding.
*/
public typealias CFStringEncoding = UInt32

/* Platform-independent built-in encodings; always available on all platforms.
   Call CFStringGetSystemEncoding() to get the default system encoding.
*/
public var kCFStringEncodingInvalidId: UInt32 { get }
public enum CFStringBuiltInEncodings : CFStringEncoding, @unchecked Sendable {

    
    case macRoman = 0

    case windowsLatin1 = 1280 /* ANSI codepage 1252 */

    case isoLatin1 = 513 /* ISO 8859-1 */

    case nextStepLatin = 2817 /* NextStep encoding*/

    case ASCII = 1536 /* 0..127 (in creating CFString, values greater than 0x7F are treated as corresponding Unicode value) */

    case unicode = 256 /* kTextEncodingUnicodeDefault  + kTextEncodingDefaultFormat (aka kUnicode16BitFormat) */

    case UTF8 = 134217984 /* kTextEncodingUnicodeDefault + kUnicodeUTF8Format */

    case nonLossyASCII = 3071 /* 7bit Unicode variants used by Cocoa & Java */

    
    public static var UTF16: CFStringBuiltInEncodings { get } /* kTextEncodingUnicodeDefault + kUnicodeUTF16Format (alias of kCFStringEncodingUnicode) */

    case UTF16BE = 268435712 /* kTextEncodingUnicodeDefault + kUnicodeUTF16BEFormat */

    case UTF16LE = 335544576 /* kTextEncodingUnicodeDefault + kUnicodeUTF16LEFormat */

    
    case UTF32 = 201326848 /* kTextEncodingUnicodeDefault + kUnicodeUTF32Format */

    case UTF32BE = 402653440 /* kTextEncodingUnicodeDefault + kUnicodeUTF32BEFormat */

    case UTF32LE = 469762304 /* kTextEncodingUnicodeDefault + kUnicodeUTF32LEFormat */
}
```

yay: and these in CFString header: https://opensource.apple.com/source/CF/CF-1153.18/CFString.h.auto.html

yay: To answer my own question, it doesn't seem like the ones I linked to are correct. I'm getting `024-04-05 12:53:52.400 cocoa[34665:824994] Incorrect NSStringEncoding value 0x8000100 detected. Assuming NSASCIIStringEncoding. Will stop this compatibility mapping behavior in the near future.` when trying to create an NSString to use an NSWindow title with
```
window_title := NS.String.alloc()->initWithOdinString("Cocoa window 😛")
window->setTitle(window_title)
```
I still wonder why though and where there the `UTF8              = 4,` enum values comes from?

vendor/darwin/Foundation/NSString.odin
```diff
@@ -54,15 +27,6 @@ foreign Foundation {
...
 
-AT :: MakeConstantString
-MakeConstantString :: proc "c" (#const c: cstring) -> ^String {
-	foreign Foundation {
-		__CFStringMakeConstantString :: proc "c" (c: cstring) -> ^String ---
-	}
-	return __CFStringMakeConstantString(c)
-}
-
-
```
yay, line 65: Moved to CoreFoundation package.

vendor/darwin/Foundation/NSString.odin
```diff
@@ -1,39 +1,12 @@
...
-	UTF16             = Unicode,
-
-	UTF16BigEndian    = 0x90000100,
-	UTF16LittleEndian = 0x94000100,
-
-	UTF32             = 0x8c000100,
-	UTF32BigEndian    = 0x98000100,
-	UTF32LittleEndian = 0x9c000100,
-}
```
yay, line 35: Moved to `CFString.odin` in CoreFoundation package.

core/sys/darwin/CoreFoundation/CFString.odin
```diff
@@ -0,0 +1,77 @@
+package CoreFoundation
+
+import "core:runtime"
+
+foreign import CoreFoundation "system:CoreFoundation.framework"
+
+String :: distinct TypeRef // same as CFStringRef
```
yay, line 7: It looks like Apple dropped the `CFStringRef` name in Swift in favor of `CFString`. There isn't a separate `CFString` type in C, just `CFStringRef`, so seems like a safe way to go:
Screenshot 2024-04-05 at 3 33 50 PM

core/sys/darwin/core_foundation.odin
```diff
@@ -1,98 +0,0 @@
...
-
-// Releases a Core Foundation object.
-CFRelease :: proc {
-	CFReleaseString,
-}
-
-// Releases a Core Foundation string.
-CFReleaseString :: #force_inline proc(theString: CFStringRef) {
-	_CFRelease(CFTypeRef(theString))
-}
```
yay, line 83: Does the `CFTypeRef` have to be a `distinct rawptr`?

Seems like we have to do a lot of dancing around with: `@(link_name="CFRelease")`, `_CFRelease`, and `_CFRelease(CFTypeRef(theString))` because of `distinct`

I can see how you may want to have `String` a `distinct TypeRef`, but benefits of having `TypeRef` be `distinct rawptr` not obvious to me.

yay:
@gingerBill @laytan Any thoughts on this guys?

gingerBill:
I am going to do a bit of rearranging of some of the packages, so things in `vendor:darwin` other than `Metal` might be moved to `core:sys/darwin`, which MIGHT break your PR.

vendor/darwin/Foundation/NSString.odin
laytan: I found this documentation that matches up with the current values: https://developer.apple.com/documentation/foundation/1497293-string_encodings/nsutf8stringencoding

And then these docs for the CoreFoundation variants: https://developer.apple.com/documentation/corefoundation/cfstringbuiltinencodings

Now there is a distinction between CoreFoundation (CFString) and Foundation (NSString).

The values you have here, are for CoreFoundation, while the values we have (where utf8 is 4) is for Foundation.

So 2 different enumerations are required here.

laytan (commented):
Looking good! 

You probably have to rebase with master because of what Bill said earlier, he moved some things from vendor into core.

Also have a look at my comment on the string encodings.

vendor/darwin/Foundation/NSString.odin
yay: Ah, yes, that makes sense, thanks. I moved `StringEncoding` back to `NSString.odin` and added `StringBuiltInEncodings` and `StringEncodings` to `CFString.odin`.

core/sys/darwin/CoreFoundation/CFString.odin
```diff
@@ -0,0 +1,199 @@
...
+	StringGetLength :: proc(theString: String) -> Index ---
+
+	// Returns the maximum number of bytes a string of a specified length (in Unicode characters) will take up if encoded in a specified encoding.
+	StringGetMaximumSizeForEncoding :: proc(length: Index, encoding: StringEncoding) -> Index ---
+
+	// Fetches a range of the characters from a string into a byte buffer after converting the characters to a specified encoding.
+	StringGetBytes :: proc(thestring: String, range: Range, encoding: StringEncoding, lossByte: u8, isExternalRepresentation: b8, buffer: [^]byte, maxBufLen: Index, usedBufLen: ^Index) -> Index ---
+
+	@(link_name = "__CFStringMakeConstantString")
+	StringMakeConstantString :: proc "c" (c: cstring) -> String ---
```
yay, line 177: StringMakeConstantString doesn't quite work with Unicode:
```
cf_str := CF.StringMakeConstantString("Cocoa window 😛")
odin_str := CF.StringCopyToOdinString(cf_str)
fmt.println("odin_str", odin_str) // odin_str Cocoa window üòõ
```
It is probably not meant to work properly when not used from ObjC: https://lists.llvm.org/pipermail/cfe-dev/2008-July/002247.html

core/sys/darwin/CoreFoundation/CFString.odin
```diff
@@ -0,0 +1,201 @@
...
+
+StringCopyToOdinString :: proc(
+	theString: String,
+	allocator := context.allocator,
+) -> (
+	str: string,
+	ok: bool,
+) #optional_ok {
+	length := StringGetLength(theString)
+	max := StringGetMaximumSizeForEncoding(length, StringEncoding(StringBuiltInEncodings.UTF8))
```
yay, line 190: I don't really like that we have to do this casting now, but
`CFStringEncoding` is defined as unsigned 32-bit integer, while `CFStringEncodings` as a signed one:
https://opensource.apple.com/source/CF/CF-1153.18/CFStringEncodingExt.h.auto.html

So usage becomes something like this:

```
fmt.println(CF.StringIsEncodingAvailable(CF.StringEncoding(CF.StringBuiltInEncodings.ASCII)))
fmt.println(CF.StringIsEncodingAvailable(CF.StringEncoding(CF.StringEncodings.MacCyrillic)))
```

I guess I could swap `StringEncoding` for `StringBuiltInEncodings`, making it easier for the common case:

```
fmt.println(CF.StringIsEncodingAvailable(.ASCII))
fmt.println(CF.StringIsEncodingAvailable(CF.StringBuiltInEncodings(CF.StringEncodings.MacCyrillic)))
```
But then it becomes inconsistent with ObjC and the fact that you have to cast `CF.StringEncodings` to `CF.StringBuiltInEncodings` is not exactly intuitive.

yay:
@gingerBill Rebased with latest changes from upstream. No conflicts thankfully :) Think it's ready for another look.

core/sys/darwin/CoreFoundation/CFString.odin
laytan: If it doesn't work/isn't intended to, please remove or comment out the binding, maybe with a comment explaining why

core/sys/darwin/CoreFoundation/CFString.odin
laytan: Don't think there is much we can do about this, this is just a weird API which you wouldn't notice in C because of implicit casts

You could define a proc group or generic wrapper that allows passing each of the cases and make the normal one private, but idk.

core/sys/darwin/Foundation/NSString.odin
```diff
@@ -5,7 +5,7 @@ foreign import "system:Foundation.framework"
 @(objc_class="NSString")
 String :: struct {using _: Copying(String)}
 
-StringEncoding :: enum UInteger {
+StringEncoding :: enum u32 {
```
laytan, line 8: This is `UInteger` in my headers, which is an `unsigned long` `c.ulong`.

core/sys/darwin/CoreFoundation/CFString.odin
laytan: Update: This does work but it just doesn't seem to support unicode, I get this warning:
`2024-04-09 20:09:32.384 test[54823:6563359] WARNING: CFSTR("Hello World! \37777777760\37777777637\37777777630\37777777633") has non-7 bit chars, interpreting using MacOS Roman encoding for now, but this will change. Please eliminate usages of non-7 bit chars (including escaped characters above \177 octal) in CFSTR().`

So we can keep it in.

laytan: Probably a good idea to define the common `CFSTR :: StringMakeConstantString` too

core/sys/darwin/Foundation/NSString.odin
```diff
@@ -54,15 +51,6 @@ foreign Foundation {
 	StringFromClass :: proc(cls: Class) -> ^String ---
 }
 
-AT :: MakeConstantString
```
laytan, line 57: This should be kept in, even though it is technically CoreFoundation, it works and is used in a lot of examples and other applications today

core/sys/darwin/CoreFoundation/CFString.odin
```diff
@@ -0,0 +1,201 @@
...
+	// Returns the maximum number of bytes a string of a specified length (in Unicode characters) will take up if encoded in a specified encoding.
+	StringGetMaximumSizeForEncoding :: proc(length: Index, encoding: StringEncoding) -> Index ---
+
+	// Fetches a range of the characters from a string into a byte buffer after converting the characters to a specified encoding.
+	StringGetBytes :: proc(thestring: String, range: Range, encoding: StringEncoding, lossByte: u8, isExternalRepresentation: b8, buffer: [^]byte, maxBufLen: Index, usedBufLen: ^Index) -> Index ---
+
+	StringIsEncodingAvailable :: proc(encoding: StringEncoding) -> bool ---
+
+	@(link_name = "__CFStringMakeConstantString")
+	StringMakeConstantString :: proc "c" (c: cstring) -> String ---
```
laytan, line 179:
```suggestion
	StringMakeConstantString :: proc "c" (#const c: cstring) -> String ---
```

core/sys/darwin/Foundation/NSString.odin
yay: Nice catch! I got confused once I moved this type to CoreFoundation and forgot to revert this change as well.

yay:
@laytan Thanks for reviewing! Addressed all comments.

laytan:
> @laytan Thanks for reviewing! Addressed all comments.

Looks good, last thing, you made the TypeRef distinct again, but didn't put the proc group for releases back allowing `CF.Release(str)` instead of `CF.Release(CF.TypeRef(str))`

If you put that back I am happy with everything here 🚀

yay:
> Looks good, last thing, you made the TypeRef distinct again, but didn't put the proc group for releases back allowing `CF.Release(str)` instead of `CF.Release(CF.TypeRef(str))`

Indeed. Done now.

core/sys/darwin/CoreFoundation/CFString.odin
```diff
@@ -0,0 +1,203 @@
...
+	// Fetches a range of the characters from a string into a byte buffer after converting the characters to a specified encoding.
+	StringGetBytes :: proc(thestring: String, range: Range, encoding: StringEncoding, lossByte: u8, isExternalRepresentation: b8, buffer: [^]byte, maxBufLen: Index, usedBufLen: ^Index) -> Index ---
+
+	StringIsEncodingAvailable :: proc(encoding: StringEncoding) -> bool ---
+
+	@(link_name = "__CFStringMakeConstantString")
+	StringMakeConstantString :: proc "c" (#const c: cstring) -> String ---
+}
+
+STR :: StringMakeConstantString
```
yay, line 182: Renamed from `CFSTR` to `STR` because it's already in the CF package, so usage would be `CF.STR`.

@gingerBill Can you take another look please?

core/sys/darwin/Security/SecBase.odin
```diff
@@ -0,0 +1,386 @@
+package Security
+
+OSStatus :: distinct i32
+
+errSec :: enum OSStatus {
+	Success = 0, // No error.
+	Unimplemented = -4, // Function or operation not implemented.
+	errSecDiskFull = -34, // The disk is full.
```
gingerBill, line 8: I'd probably remove all of the `errSec` prefixes.

core/sys/darwin/Security/SecRandom.odin
```diff
@@ -0,0 +1,19 @@
...
+
+foreign import Security "system:Security.framework"
+
+// A reference to a random number generator.
+RandomRef :: distinct rawptr
+
+@(link_prefix="Sec", default_calling_convention="c")
+foreign Security {
+	// Default random ref for /dev/random. Synonym for nil.
+	@(link_name="kSecRandomDefault") kSecRandomDefault: RandomRef
```
gingerBill, line 13: Why is the `link_name` set if the it has the exact same name?

core/sys/darwin/Security/SecBase.odin
yay: > I'd probably remove all of the `errSec` prefixes.

Good point, thanks.

core/sys/darwin/Security/SecRandom.odin
yay: Because `foreign Security {` has `@(link_prefix="Sec"` on it.
Wouldn't it try to import as `SeckSecRandomDefault` without the `@(link_name="kSecRandomDefault")`?

yay: just like `RandomCopyBytes` and `CopyErrorMessageString` below that are actually `SecRandomCopyBytes` and `SecCopyErrorMessageString`

Issue #3368, opened 2024-04-03
Dynamic library from vendor not loading on Linux


joakin:
## Context

* Operating System & Odin Version: Ubuntu 22.04
* `odin report` output:
```
        Odin:    dev-2024-04:4c35633e0
        OS:      Ubuntu 22.04.4 LTS, Linux 5.15.133.1-microsoft-standard-WSL2
        CPU:     Intel(R) Core(TM) i9-10900KF CPU @ 3.70GHz
        RAM:     15894 MiB
        Backend: LLVM 14.0.0
```

I was trying to compile a small raylib example with `odin build src/main -out:bin/flee -define:RAYLIB_SHARED=true -extra-linker-flags="-v"`, which errors with `/home/linuxbrew/.linuxbrew/bin/ld: cannot find -l/home/joakin/dev/forks/Odin/vendor/raylib/linux/libraylib.so.5.0.0: No such file or directory`

## Expected Behavior

Program should compile normally.

## Current Behavior

The linker arguments seem to be wrong and the program doesn't compile.

## Failure Information (for bugs)

```
flee (master *=) → odin build src/main -out:bin/flee_release -define:RAYLIB_SHARED=true -extra-linker-flags="-v"
Homebrew clang version 17.0.6
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /home/linuxbrew/.linuxbrew/bin
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/11
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/9
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/11
Candidate multilib: .;@m64
Selected multilib: .;@m64
[...]

PR #3369, opened 2024-04-03
Fix vendor dynamic libraries not working on Linux

joakin:
Fixes https://github.com/odin-lang/Odin/issues/3368

Dynamic libraries like libraylib.so.5.0.0 in vendor aren't working on linux, because the name doesn't end in `.so` like the code expected.

### DIFF
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -432,7 +432,7 @@ gb_internal i32 linker_stage(LinkerData *gen) {
 							if (string_ends_with(lib, str_lit(".a")) || string_ends_with(lib, str_lit(".o"))) {
 								// static libs and object files, absolute full path relative to the file in which the lib was imported from
 								lib_str = gb_string_append_fmt(lib_str, " -l:\"%.*s\" ", LIT(lib));
-							} else if (string_ends_with(lib, str_lit(".so"))) {
+							} else if (string_ends_with(lib, str_lit(".so")) || string_contains_string(lib, str_lit(".so."))) {
 								// dynamic lib, relative path to executable
 								// NOTE(vassvik): it is the user's responsibility to make sure the shared library files are visible
 								//                at runtime to the executable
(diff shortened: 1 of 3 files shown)


joakin:
This actually doesn't fix the issue of the library loading, because of this line

https://github.com/joakin/Odin/blob/b632e90eb259e2eec07ffd5ddc4823daa167fddf/src/linker.cpp#L426

```
	lib_str = gb_string_append_fmt(lib_str, " -l:\"%s/%.*s\" ", cwd, LIT(lib));
```

Which is prepending the CWD to the lib path, resulting in a double path where the executable is, and where the vendor library is. In my case, something like:

```
/home/linuxbrew/.linuxbrew/bin/ld: cannot find -l:/home/joakin/dev/flee//home/joakin/dev/forks/Odin/vendor/raylib/linux/libraylib.so.5.0.0: No such file or directory
```

I believe the compiler assumes any dynamic library must be given just the name without the path and be present where the executable will be, but in the case of the vendor library it already has the full path in it.

@laytan What would be the next steps here? If we remove the CWD from it, then maybe it will break dynamic libraries that you place in your executable's path? I'm not sure.

laytan:
I guess you could add an if statement around it checking if the path is already absolute, if it then works?
I know on Windows you need to put all shared libraries next to your executable, I guess this was added because that is also the case for Linux? But if it does work for you this way idk anymore, linking is so stupid.

laytan:
@joakin so I did some looking at this, and adding the cwd seems to be correct, if you ship a binary you also want it based on cwd so you can just ship a directory.

The reason why you have a cwd and then also the odin path is because you missed a spot where the same string contains check has to be done: https://github.com/odin-lang/Odin/blob/a61ae7c861fa301684ee1582507061317b11426b/src/parser.cpp#L5713

After this, it will kinda work, but you'd have to fiddle with your rpath on build through extra linker flags I think.

But it is a move forward, would you want to update the PR with this? At that point we can merge this and see how else to improve the story later.

joakin:
I've also added the same contains check to the parser line, I searched and ".so" is not used anywhere else, good catch.

With this it should work if copying the shared library file to the root of the exe.

Issue #3417, opened 2024-04-12
core:net URL fragment being improperly read


blob1807:
## Context

Odin:    dev-2024-04:aab122ede
OS:      Windows 11 Home Basic (version: 23H2), build 22631.3447
CPU:     AMD Ryzen 7 6800HS with Radeon Graphics
RAM:     15608 MiB
Backend: LLVM 17.0.1

## Expected Behavior
No expections.

## Current Behavior

If you pass an fragment/anchor to net.split_url() it'll get read as part of a query.
```
https://example.com/?t=1#Hellope -> https example.com / map[t=1#Hellope]
https://example.com/?t#Hellope   -> https example.com / map[t#Hellope=]
```

## Expected Behavior
```
https://example.com/?t=1#Hellope -> https example.com / map[t=1] Hellope
https://example.com/?t#Hellope   -> https example.com / map[t=] Hellope
```

## Suggested Solution
Add support for it. A PR will be made to do this.

PR #3418, opened 2024-04-12
core:net Add support for URL fragments

blob1807:
Added support for a URL's fragment/anchor to `split_url` & `join_url` in `core:net` plus new tests to cover it.
Whilst a server should never receive one & client send one. It should still be handled incase it does.
Some apps, likes a browsers, will want the fragment & may choose to use `split_url` to get a URL's parts 

Fixes: https://github.com/odin-lang/Odin/issues/3417

### DIFF
--- a/core/net/url.odin
+++ b/core/net/url.odin
@@ -30,7 +30,7 @@ split_url :: proc(url: string, allocator := context.allocator) -> (scheme, host,
 		s = s[i+3:]
 	}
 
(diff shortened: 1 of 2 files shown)


Kelimion:
Great idea. If you can resolve this error, I'll merge it.
image

blob1807:
I was already on it.

Kelimion:
Thanks!

PR #3419, opened 2024-04-12
Fixup vendor/d3d12 calling convention mistakes

puzzleddev:
This commit updates various D3D12 methods to correct for their strange calling convention, now requiring manually passing in a pointer to where the return value is to be stored.

At this point I do not have a D3D12 project set up, so while I have checked for syntactic correctness, I am not yet able to test the changes in an actual program. I would appreciate anyone who has access to code which uses the vendor/d3d12 package to test these changes before they are applied to the public repository.

### DIFF
--- a/vendor/directx/d3d12/d3d12.odin
+++ b/vendor/directx/d3d12/d3d12.odin
@@ -2283,7 +2283,7 @@ IHeap :: struct #raw_union {
 }
 IHeap_VTable :: struct {
 	using id3d12devicechild_vtable: IDeviceChild_VTable,
-	GetDesc: proc "system" (this: ^IHeap) -> HEAP_DESC,
+	GetDesc: proc "system" (this: ^IHeap, pRetValue: ^HEAP_DESC) -> ^HEAP_DESC,
 }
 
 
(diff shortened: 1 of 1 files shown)


vendor/directx/d3d12/d3d12.odin
```diff
@@ -2297,7 +2297,7 @@ IResource_VTable :: struct {
 	using id3d12devicechild_vtable: IDeviceChild_VTable,
 	Map:                  proc "system" (this: ^IResource, Subresource: u32, pReadRange: ^RANGE, ppData: ^rawptr) -> HRESULT,
 	Unmap:                proc "system" (this: ^IResource, Subresource: u32, pWrittenRange: ^RANGE),
-	GetDesc:              proc "system" (this: ^IResource) -> RESOURCE_DESC,
+	GetDesc:              proc "system" (this: ^IResource, pRetValue: ^RESOURCE_DESC) -> ^RESOURCE_DESC,
```
gingerBill, line 2300: Is it correct that it also returns a pointer?

puzzleddev: While I have not disassembled any D3D12 functions I am basing this on both the x64 calling convention documentation, which states: "Otherwise, the caller must allocate memory for the return value and pass a pointer to it as the first argument. [...] The same pointer must be returned by the callee in RAX." As well as a small sample function which can be seen here on godbolt: https://godbolt.org/z/6Pv7v1exa where on line 24 you can see RAX being set to the pointer to the return value.

blob1807: Yes `GetDesc` should be taking a pointer to `*_DEC`, but it returns a `HRESULT` not a pointer. Though it might only be valid for some VTables.
This is going off MS's Rust Bindings [Link1](https://github.com/microsoft/windows-rs/blob/master/crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/impl.rs#L3170) [Link2](https://github.com/microsoft/windows-rs/blob/master/crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs#L3075) (they're **BIG** files). Shouldn't take them as gospel but they're probably going to be the most right.
winapi-rs (which is what bevy/wgpu use) also agrees with this, it just implemented differently.

puzzleddev: Beyond the evidence I already gave Windows' official D3D12 C bindings also return a pointer to a `D3D12_*_DESC` pointer, as seen here: `D3D12_HEAP_DESC *( STDMETHODCALLTYPE *GetDesc )(ID3D12Heap * This, D3D12_HEAP_DESC * RetVal);`. I personally find my case more compelling since it is based on both maintaining the established calling convention and the "most" official - that is generated from the source of truth '.idl' files - bindings. The only way to truly know would be decompiling a `GetDesc` method which I am not willing to do.

PR #3431, opened 2024-04-15
Fix `big.internal_random_prime` with `Second_MSB_On`

### LINKED ISSUE #3430: internal_random_prime fails when Second_MSB_On is true and size is insufficient
## Context

`big.internal_random_prime`, when asked to create a prime with the `Second_MSB_On` flag, intermittently fails with `.Invalid_Argument` due to an insufficiently small `big.Int`.

I believe this is happening because of the call to `internal_clamp` from within `internal_int_random`, which is called a few lines before the `Second_MSB_On` condition.

	Odin:    dev-2024-04:d510d5e49
	OS:      Arch Linux, Linux 6.8.5-arch1-1
	RAM:     31916 MiB
	Backend: LLVM 14.0.6

The following small program typically fails the assertion within a few runs.
```odin
package main

import "core:fmt"
import "core:math/big"

main :: proc() {
	N :: 128

	for j in 0 ..< 100 {
		a: big.Int
		err := big.internal_random_prime(&a, N, -1, {.Second_MSB_On})
		assert(err == nil)
		fmt.println(j)
	}
}
```

Feoramund:
Fixes #3430

### DIFF
--- a/core/math/big/prime.odin
+++ b/core/math/big/prime.odin
@@ -1247,6 +1247,20 @@ internal_random_prime :: proc(a: ^Int, size_in_bits: int, trials: int, flags := 
 			a.digit[0] |= 3
 		}
 		if .Second_MSB_On in flags {
+			/*
+				Ensure there's enough space for the bit to be set.
+			*/
+			if a.used * _DIGIT_BITS < size_in_bits - 1 {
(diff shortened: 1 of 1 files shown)


Kelimion (commented):
Solid. Merged.

PR #3433, opened 2024-04-15
Fix `text/table` using `len()` for strings

### LINKED ISSUE #3432: text/table uses bytes instead of runes for length
## Context
The `text/table` library is using `len` where `utf8.rune_count` would probably be more appropriate. See the following program and its output.

```odin
package main

import "core:os"
import "core:text/table"

main :: proc() {
    t: table.Table
    table.init(&t)
    table.caption(&t, "Tɛst")
    table.row(&t, "Hellope!")
    table.row(&t, "Hɛllope!")

    stdout := os.stream_from_handle(os.stdout)
    table.write_ascii_table(stdout, &t)
}
```

The `ɛ` is a two-byte UTF-8 rune here.

## Result
```
+---------+
|  Tɛst  |
+---------+
|Hellope! |
|Hɛllope!|
+---------+
```

## Expected
```
+--------+
|  Tɛst  |
+--------+
|Hellope!|
|Hɛllope!|
+--------+
```

Feoramund:
Fixes #3432

### DIFF
--- a/core/text/table/table.odin
+++ b/core/text/table/table.odin
@@ -12,6 +12,7 @@ import "core:io"
 import "core:fmt"
 import "core:mem"
 import "core:mem/virtual"
+import "core:unicode/utf8"
 import "base:runtime"
 
 Cell :: struct {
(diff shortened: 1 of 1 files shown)


gingerBill:
Note that because the procedure was named `write_ascii_table`, you will need change its name to better reflect these changes.

oskarnp:
> Note that because the procedure was named `write_ascii_table`, you will need change its name to better reflect these changes.

The name is OK if we redefine it as referring to the delimiter characters used, as opposed to something like Unicode box-drawing characters, for example.

I'll review and test it later today.

oskarnp:
I would reject this PR because it doesn't really solve much. See my comment in https://github.com/odin-lang/Odin/issues/3432#issuecomment-2057519705.

oskarnp (commented):
I would reject this PR because it doesn't really solve much. See my comment in https://github.com/odin-lang/Odin/issues/3432#issuecomment-2057519705.

PR #3439, opened 2024-04-16
NetBSD support

andreas-jonsson:
This is my attempt at an initial port of Odin to NetBSD. It targets NetBSD 10.0 (AMD64) with Clang/LLVM 17 from ~~the [pkgsrc wip](https://pkgsrc.se/wip/llvm) section~~ [pkgsrc](https://pkgsrc.se/lang/llvm). If you don't want to taint your system with ~~WIP~~ libraries you can build LLVM using the [mksandbox](https://wiki.netbsd.org/tutorials/pkgsrc/pbulk/) tool.

The current goal ~~(at the moment) is to have the compiler fully working on NetBSD as well as enough of the standard library to run `demo.odin`~~ is to have the NetBSD port be on par with the other BSD's.

Right now the compiler seems happy and the demo is running, ~~but there are concurrency issues in the runtime. This needs more investigation. One possible issue could be the thread id, that I do not know how to get a hold of atm and I'm currently investigating the issue. Another culprit could be the futex implementation that is pretty much the same as on Haiku.~~

There is also room for some more cleanup and possible joining or refactoring some code that could be shared with other BSDs.
Generally this port follow the patterns outlined by the other BSD's.


andreas-jonsson:
I think we are at a pretty good state right now. Would love to get some feedback.

Yawning:
Add the appropriate build tag to `core/crypto/rand_bsd.odin`.

Yawning:
Ah sorry, you also need to exclude netbsd in the generic `core/crypto/rand_generic.odin`.  The tests under `tests/core` will fail to compile unless you do.

laytan:
Are you sure netbsd doesn't have native futexes?

andreas-jonsson:
> Are you sure netbsd doesn't have native futexes?

I don't think so. :(
I was not able to find any documentation about it at least. Also, was not able to find any headers hinting of it's being supported.

andreas-jonsson:
I think I can rewrite the futex implementation using raw syscalls. Will give it a go.

andreas-jonsson:
Native futex support is in. Plz let me know if you have any other feedback.

Yawning:
> Native futex support is in. Plz let me know if you have any other feedback.

I looked into this a bit and came across this thread, which is concerning.  Though the followup by thorpej seems to indicate that it may be usable in certain cases.
- https://mail-index.netbsd.org/tech-kern/2023/12/11/msg029334.html
- https://mail-index.netbsd.org/tech-kern/2023/12/11/msg029335.html

otoh Mesa doesn't use futexes on NetBSD when I peeked at the sourcecode, so I don't know if/how the followup progressed.

https://man.netbsd.org/_lwp_park.2 and co might be the "right" thing to do, but NetBSD appears to be unique among the BSDs here.

andreas-jonsson:
Hm.. yea.
That is an interesting read. I hope the NetBSD folks fix it because it is convenient for porting purposes to follow Linux here.

As for the current state of things I have hammered this code quite a bit and I have not found any issues. (That is not to say there is none.) But it could also be that this works well for Odin since the use of the API is quite simplistic, and it could be that most of the issues the Mesa team is experiencing comes from using more "complicated" patterns. Perhaps there use of futexes are not process private for example?

I would prefer keeping the code as is and do wider testing. If we see any problems it is only a few lines of code to switch to the _lwp_park API. Most of the work would just be to read up on the API documentation.

If this work, the codebase also gets cleaner because most code is just shared with Linux. :)

andreas-jonsson:
There is now a working CI for Odin on NetBSD in a separate PR here #3482
Result can be seen here: https://github.com/andreas-jonsson/Odin/actions/runs/8925922741/job/24515753510

core/sys/unix/pthread_netbsd.odin
```diff
@@ -0,0 +1,102 @@
...
+PTHREAD_RWLOCKATTR_T_SIZE  :: 16
+PTHREAD_BARRIERATTR_T_SIZE :: 16
+
+PTHREAD_COND_T_SIZE    :: 40
+PTHREAD_MUTEX_T_SIZE   :: 48
+PTHREAD_RWLOCK_T_SIZE  :: 64
+PTHREAD_BARRIER_T_SIZE :: 48
+PTHREAD_ATTR_T_SIZE    :: 16
+
+pthread_cond_t :: struct #align(16) {
```
laytan, line 20: Are you sure this is `#align(16)`, some other targets have also copied this from previous targets where this wasn't actually the case which caused problems.

src/build_settings.cpp
```diff
@@ -557,6 +559,14 @@ gb_global TargetMetrics target_openbsd_amd64 = {
 	str_lit("e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"),
 };
 
+gb_global TargetMetrics target_netbsd_amd64 = {
+	TargetOs_netbsd,
+	TargetArch_amd64,
+	8, 8, 8, 16,
```
laytan, line 565: With LLVM 18 being merged in, this should probably be `8, 8, AMD_64_MAX_ALIGNMENT, 16`

src/build_settings.cpp
```diff
@@ -557,6 +559,14 @@ gb_global TargetMetrics target_openbsd_amd64 = {
...
+	str_lit("x86_64-unknown-netbsd-elf"),
+	str_lit("e-m:w-i64:64-f80:128-n8:16:32:64-S128"),
```
laytan, line 567: Also with LLVM 18, this line can be removed

laytan (commented):
Looks great in general!

src/build_settings.cpp
andreas-jonsson: Got it.

But how should this be handled when building with LLVM17? I did not see any branches on LLVM version.
It will take a while before 18 lands in pkgsrc. 17 just landed in current and there wont be binaries until Q2 2024.

core/sys/unix/pthread_netbsd.odin
andreas-jonsson: I will double check this. Thanks

andreas-jonsson: You are right. It should be 8.

src/build_settings.cpp
laytan: Ah I was not clear, the llvm 18 branch that was merged removed this field from the struct completely, for all versions of llvm. This field was actually dead code all along, it didn't do anything.

andreas-jonsson: Yea.. I just figured it out. :D
Sorry. I missed that 18 was merged.

core/sys/info/cpu_intel.odin
```diff
@@ -67,7 +67,7 @@ init_cpu_features :: proc "c" () {
 	try_set(&set, .os_xsave,  27, ecx1)
 	try_set(&set, .rdrand,    30, ecx1)
 
-	when ODIN_OS == .FreeBSD || ODIN_OS == .OpenBSD {
+	when ODIN_OS == .FreeBSD || ODIN_OS == .OpenBSD || ODIN_OS == .NetBSD {
```
Yawning, line 70: Maybe edit the comment to indicate that this isn't allowed under netbsd as well.

core/crypto/rand_bsd.odin
```diff
@@ -1,4 +1,4 @@
-//+build freebsd, openbsd
+//+build freebsd, openbsd, netbsd
```
Yawning, line 1: Sorry about the churn here.

core/os/os_netbsd.odin
```diff
@@ -0,0 +1,780 @@
...
+
+close :: proc(fd: Handle) -> Errno {
+	result := _unix_close(fd)
+	if result == -1 {
+		return Errno(get_last_error())
+	}
+	return ERROR_NONE
+}
+
+// We set a max of 1GB to keep alignment and to be safe.
```
Yawning, line 359: I know the `os` package is slated to be replaced so this code is kind of whatever, but should read/write loop to try to satisfy the full request?

core/os/os_netbsd.odin
```diff
@@ -0,0 +1,780 @@
...
+S_IRWXO :: 0o0007 // RWX mask for other
+S_IROTH :: 0o0004 // R for other
+S_IWOTH :: 0o0002 // W for other
+S_IXOTH :: 0o0001 // X for other
+
+S_ISUID :: 0o4000 // Set user id on execution
+S_ISGID :: 0o2000 // Set group id on execution
+S_ISVTX :: 0o1000 // Directory restrcted delete
+
+S_ISLNK  :: #force_inline proc(m: mode_t) -> bool { return (m & S_IFMT) == S_IFLNK  }
```
Yawning, line 272: Maybe make these contextless as well?

Yawning (commented):
Mostly nits, free feel to ignore.

core/os/os_netbsd.odin
laytan: Other platforms don't right? We have `read_full` etc. for that

andreas-jonsson: Yea. I have just followed the other BSDs here.

PR #3462, opened 2024-04-21
core:math/rand: Use `crypto.rand_bytes()` for the system RNG

Yawning:
There is a bit of code duplication, when doing things this way expands the system RNG support to every supported target.  This still also preserves the panic on initialization when unsupported behavior, and adds relevant documentation on how to check if the current target is supported at compile time and runtime.


core/crypto/crypto.odin
```diff
@@ -57,7 +57,7 @@ rand_bytes :: proc (dst: []byte) {
 }
 
 // has_rand_bytes returns true iff the target has support for accessing the
-// system entropty source.
+// system entropy source.
 has_rand_bytes :: proc () -> bool {
-	return _has_rand_bytes()
```
gingerBill, line 62: What's the rationale from going to a constant to a call? Seems counter intuitive to me.

Yawning: "I added the call a short while ago, and felt bad about breaking compatibility."

I can get rid of the call entirely if that is better.

PR #3475, opened 2024-04-24
Add support for iOS/iPhoneSimulator target

harold-b:
Attempts to add support for building for iOS and iPhoneOS simulator targets.

The basic changes for compiler support are few, mainly on the CLI frontend and on the linker side.

iOS support is only available targeting `17.4` and up because [os_sync_wait_on_address](https://developer.apple.com/documentation/os/4316531-os_sync_wait_on_address?changes=_2_8_8) is only available starting with said version. The fallback `ulock` APIs are disallowed in the Appstore.

Opened in draft mode as an additional layer to avoid syscalls on iOS must be added too, as these would also be disallowed in the Appstore.

### DIFF
--- a/base/runtime/core.odin
+++ b/base/runtime/core.odin
@@ -531,6 +531,7 @@ Odin_Endian_Type :: type_of(ODIN_ENDIAN)
 	Odin_Platform_Subtarget_Type :: enum int {
 		Default,
 		iOS,
+		iPhoneSimulator,
 	}
 */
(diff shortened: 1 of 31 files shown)


herohiralal:
Is this actively being worked upon?

harold-b:
Yes, but I am attempting to get a personal project moving forward a bit more so I can properly dogfood it on device before making significant changes. Need a little bit more time

gingerBill:
I'm just closing this because it's been a long time since it has been touched.

PR #3476, opened 2024-04-24
[sys/linux] Fix fork and execve syscalls on arm64

### LINKED ISSUE #3472: linux.fork is broken on arm64
## Context

	Odin:    dev-2024-04:efc3f9916
	OS:      Arch Linux ARM, Linux 6.7.9-1-MANJARO-ARM
	CPU:     ARM64
	RAM:     3862 MiB
	Backend: LLVM 14.0.6

## Expected Behavior

```odin
package main

import "core:sys/linux"
import "core:fmt"

main :: proc() {
    using linux

    pid, pid_err := fork()
    fmt.println("after fork")
    assert(pid_err == .NONE)

    if pid == 0 {
        argv := []cstring{
            "/usr/bin/env", nil,
        }
        envp := []cstring{
            "execy_env=1234", nil,
        }
        name := cstring("/usr/bin/env")

        fmt.println("in child")

        e_err := execve(name, raw_data(argv), raw_data(envp))
        fmt.eprintln("unreachable")
        assert(e_err == .NONE)
    } else {
        fmt.println("in parent")
        status: i32
        w_err := syscall(SYS_wait4, Pid(-1), &status, cast(u32)0, cast(rawptr) nil)
        assert(w_err >= 0)

        fmt.printfln("status: {}", status)
    }
}
```

This program should work, but it sometimes does not.

## Current Behavior

This is the syscall for fork on arm64:
```odin
syscall(SYS_clone, Signal.SIGCHLD)
```

The current behaviour is that it sometimes fails to fork.

The clone syscall is [documented](https://arm64.syscall.sh/) as follows:
```
220	clone	man/ cs/	DC	unsigned long	unsigned long	int*	   int*	unsigned long
```

The syscall should set all parameters.

### Steps to Reproduce

1. step 1
2. step 2
3. you get it...

### Failure Logs

### LINKED ISSUE #3487: Bug in linux execveat proc call
## Context

```bash
$ uname -svprio
Linux 6.6.26-1-MANJARO #1 SMP PREEMPT_DYNAMIC Wed Apr 10 20:11:08 UTC 2024 unknown unknown GNU/Linux

$ odin version
dev-2024-04-nightly:aab122ede
```

```

	Odin:    dev-2024-04-nightly:aab122ede
	OS:      Manjaro Linux, Linux 6.6.26-1-MANJARO
	CPU:     Intel(R) Core(TM) i5-8300H CPU @ 2.30GHz
	RAM:     15837 MiB
	Backend: LLVM 17.0.6

```

## Expected Behavior

Compile and run the following code:

```odin
package execveat

import "core:fmt"
import "core:os"
import "core:sys/linux"


// odin run execveat.odin -file

main :: proc() {
	name :: "date"
	flags: linux.Open_Flags = {linux.Open_Flags_Bits.RDONLY, linux.Open_Flags.DIRECTORY}
	argv: [^]cstring
	envp: [^]cstring

	fd, fd_err := linux.open("/usr/bin", flags)
	defer linux.close(fd)

	if fd_err != .NONE {
		message := fmt.tprintf("error file desc: %d", fd_err)
		fmt.eprintfln(message)
		os.exit(-1)
	}

	err_exec := linux.execveat(fd, name, argv, envp)
	if err_exec != .NONE {
		message := fmt.tprintf("error exec: %d", err_exec)
		fmt.eprintfln(message)
		os.exit(-1)
	}

}

```

Expected behavior:
```bash
$ odin run execveat.odin -file            
Fri Apr 26 14:21:10 EEST 2024
```

## Current Behavior

```bash
$ odin run execveat.odin -file            
error exec: 22
```

## Suggestion

Add missing `flags` parameter to the `execveat` proc.

Current implementation: [core/sys/linux/sys.odin#L2806-L2810](https://github.com/odin-lang/Odin/blob/f745fff640ab1582bdfdd18a7239c58fa37db753/core/sys/linux/sys.odin#L2806-L2810) has **4** parameters, but C suggests it has to have **5** parameters:
```c
int execveat(int dirfd, const char *pathname, char *const _Nullable argv[],
             char *const _Nullable envp[], int flags);
```

By manually adding missing fifth param I was able to run example with `execveat` call:

```odin
execveat :: proc "contextless" (dirfd: Fd, name: cstring, argv: [^]cstring, envp: [^]cstring) -> (Errno) {
	ret := syscall(SYS_execveat, dirfd, cast(rawptr) name, cast(rawptr) argv, cast(rawptr) envp, cast(rawptr)nil)
	return Errno(-ret)
}
```

PucklaJ:
This PR does the following things:
+ Adds flag parameter to execveat syscall
+ Adds clone syscall
+ Uses execveat procedure in execve syscall on arm64
+ Uses clone procedure in fork syscall on arm64

Fixes #3472
Fixes #3487

The syscalls have been implemented based on the [ChromiumOS docs](https://www.chromium.org/chromium-os/developer-library/reference/linux-constants/syscalls)


core/sys/linux/sys.odin
```diff
@@ -729,18 +729,32 @@ getsockopt :: proc {
...
 */
 fork :: proc "contextless" () -> (Pid, Errno) {
 	when ODIN_ARCH == .arm64 {
-		// Note(flysand): this syscall is not documented, but the bottom 8 bits of flags
-		// are for exit signal
-		ret := syscall(SYS_clone, Signal.SIGCHLD)
-		return errno_unwrap(ret, Pid)
+		ret, err := clone(u64(Signal.SIGCHLD), nil, nil, nil, 0)
+		if err != .NONE do return Pid(ret), err
+		return Pid(ret), .NONE
```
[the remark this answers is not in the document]
PucklaJ: That's just my coding style. My thought process always is that it is slightly more optimized, because it does not need to retrieve the value of a variable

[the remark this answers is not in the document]
gingerBill: Well that thought process is just wrong because you are "retrieving" (i.e. a load) `err` already in the `if`. 

Please change it to what I suggested to match the rest of the codebase's style.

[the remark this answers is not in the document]
PucklaJ: Sure. I updated it.

core/sys/linux/sys.odin
```diff
@@ -2803,8 +2816,8 @@ getrandom :: proc "contextless" (buf: []u8, flags: Get_Random_Flags) -> (int, Er
 	Execute program relative to a directory file descriptor.
 	Available since Linux 3.19.
 */
-execveat :: proc "contextless" (dirfd: Fd, name: cstring, argv: [^]cstring, envp: [^]cstring) -> (Errno) {
-	ret := syscall(SYS_execveat, dirfd, cast(rawptr) name, cast(rawptr) argv, cast(rawptr) envp)
+execveat :: proc "contextless" (dirfd: Fd, name: cstring, argv: [^]cstring, envp: [^]cstring, flags: i32) -> (Errno) {
```
flysand7, line 2819: Since you're adding flags could you make this value a bitset instead of an `i32`? It would help people interfacing with this syscall.

image

core/sys/linux/sys.odin
```diff
@@ -729,18 +729,32 @@ getsockopt :: proc {
...
+	Creates a new ("child") process, in a manner similar to fork.
+	Available since Linux 1.0? (<2.4)
+
+	Note(flysand): this syscall is not documented, but the bottom 8 bits of flags
+	are for exit signal
+*/
+clone :: proc "contextless" (flags: u64, stack: rawptr, parent_tid, child_tid: ^i32, tls: u64) -> (i64, Errno) {
+	when ODIN_ARCH == .amd64 {
+		ret := syscall(SYS_clone, flags, stack, parent_tid, child_tid, tls)
+		return errno_unwrap(ret, i64)
```
flysand7, line 742: I'll explain why I didn't want to make clone syscall part of the API. If one calls this function, providing a `stack` parameter that is not NULL, then this line will crash the child process, because child's stack doesn't have the value for RIP on the stack.

Usually this syscall is used for creating threads, and for threads the ideal API is a callback. The way one would set this up is by calling the `clone` syscall directly, and if we found ourselves as the child process, we jump to (not `call`) an assembly routine that sets up the stack, copies the TLS image and does a few things, before calling the thread callback with needed parameters.

But calling a callback, let alone having a separate assembly routine is too much of an API change, which goes against the way this package is supposed to work. That's why I didn't want to provide clone as part of this sys/linux API.

Even if you manage to create a thread you need to know the exact TLS layout for the given libc the program is compiled with. And the TLS is defined by the dynamic linker, which makes things so much harder because there's no API's that let you get the TLS image / struct from a dynamic linker and call it done. Having a different TLS layout can lead to bugs, from what I've heard nvidia drivers on linux expect glibc's TLS layout and don't work with e.g. musl.

So I think this procedure needs to be removed and syscall called directly in the `fork()` procedure.

core/sys/linux/sys.odin
```diff
@@ -729,18 +729,32 @@ getsockopt :: proc {
...
 */
 fork :: proc "contextless" () -> (Pid, Errno) {
 	when ODIN_ARCH == .arm64 {
-		// Note(flysand): this syscall is not documented, but the bottom 8 bits of flags
-		// are for exit signal
-		ret := syscall(SYS_clone, Signal.SIGCHLD)
-		return errno_unwrap(ret, Pid)
+		ret, err := clone(u64(Signal.SIGCHLD), nil, nil, nil, 0)
+		if err != .NONE do return Pid(ret), err
+		return Pid(ret), err
```
flysand7, line 757: This could be simplified...

core/sys/linux/sys.odin
```diff
@@ -769,8 +783,7 @@ execve :: proc "contextless" (name: cstring, argv: [^]cstring, envp: [^]cstring)
 		ret := syscall(SYS_execve, cast(rawptr) name, cast(rawptr) argv, cast(rawptr) envp)
 		return Errno(-ret)
 	} else {
-		ret := syscall(SYS_execveat, AT_FDCWD, cast(rawptr) name, cast(rawptr) argv, cast(rawptr) envp)
-		return Errno(-ret)
+		return execveat(AT_FDCWD, name, argv, envp, 0)
```
flysand7, line 786: It's a small thing so I was wondering as to whether I want to make this remark, but I don't want to call procedures that call syscalls as opposed to calling syscalls directly, because they might have additional logic that maybe you don't want to duplicate. Let's keep this package as plain as possible

core/sys/linux/constants.odin
```diff
@@ -5,6 +5,8 @@ package linux
 	that relative paths are relative to current directory.
 */
 AT_FDCWD :: Fd(-100)
+AT_EMPTY_PATH :: 0x1000
```
flysand7, line 8: Might be a wrong place to put them, these are flags, not file descriptors.

PucklaJ:
Is there a particular reason as to why the `execve` syscall is not used on arm64? It does exist and run completely fun on my Pinebook Pro. I considered changing the `execveat` syscall to `execve` on arm64 in the execve procedure.

core/sys/linux/sys.odin
PucklaJ: I suppose this is not relevant anymore?

jasonKercher:
> Is there a particular reason as to why the `execve` syscall is not used on arm64? It does exist and run completely fine on my Pinebook Pro. I considered changing the `execveat` syscall to `execve` on arm64 in the execve procedure.

I believe this one may have been assumed =]
It's not uncommon for arm64 to only have the `*at` version of the syscall.  I wonder if we should just use the `*at` functions for everything as it would get rid of all the `when` business.  Everyone has those functions.  Here is the entirety of the difference between `open` and `openat` in the Linux source code:

```c
SYSCALL_DEFINE3(open, const char __user *, filename, int, flags, umode_t, mode)
{
        if (force_o_largefile())
                flags |= O_LARGEFILE;
        return do_sys_open(AT_FDCWD, filename, flags, mode);
}

SYSCALL_DEFINE4(openat, int, dfd, const char __user *, filename, int, flags,
                umode_t, mode)
{
        if (force_o_largefile())
                flags |= O_LARGEFILE;
        return do_sys_open(dfd, filename, flags, mode);
}
```

PucklaJ:
@flysand7 I made the requested changes. Is there still something that needs to be changed?

flysand7:
@gingerBill looks good to merge

gingerBill:
@flysand7 What should that type be in the conflict?

flysand7:
@gingerBill I can't see what the conflict is, so... no idea

flysand7 (commented):
After the merge `execveat` seems to be using the wrong type for its flags, it should be `Execve_Flags`. Could you change that?

PucklaJ:
`Execve_Flags` does not exist. Should I create a bit_set that uses `FD_Flags_Bits`?

flysand7:
https://github.com/odin-lang/Odin/pull/3476/files#diff-988793090a86cbbc5126181f239031996412c074c814e90738c6aa08e57b5d25R1310

PucklaJ:
@flysand7 done

PR #3484, opened 2024-04-25
debug: write variant type name as part of variant name

laytan:
`v1` -> `v1(int)`
`v2` -> `v2(Foo)`

This aids debuggers that do not show the associated type of the field (like vscode with lldb).


gingerBill:
But how do you then access it like a normal struct? Seems like it might make it too verbose (or even impossible) to type.

Or is the idea to pointer cast the `union`?

PR #3486, opened 2024-04-26
Fix memory leak in `sync/chan`

Feoramund:
`chan.destroy` wasn't freeing the memory allocated by the `create_*` procs because the allocator was not assigned.


gingerBill:
It should be noted, that the `chan` package is still mostly in development.

PR #3506, opened 2024-04-28
core/sys/windows: added various procedures and constants for the hid library

IllusionMan1212:
This PR adds more procedures and constants for the hid library on windows. 

I tried following the structure of the Windows SDK header files, let me know if this is desirable or if I should put everything under `hidpi.odin`


core/sys/windows/hidsdi.odin
```diff
@@ -0,0 +1,41 @@
+// +build windows
+package sys_windows
```
gingerBill, line 2: Why is this in a separate file to the rest when they use the same system library `hid.lib`?

IllusionMan1212: No particular reason, I was just following the header files for the SDK, wasn't sure about the implications of importing the hid library more than once.

gingerBill: I'd recommend moving them to one file, rather than doing the same thing as the header files.

core/sys/windows/hidpi.odin
```diff
@@ -2,8 +2,20 @@
...
+    RingBufferSize: ULONG,
+}
+PHIDD_CONFIGURATION :: ^HIDD_CONFIGURATION
+
+HIDD_ATTRIBUTES :: struct {
+    Size: ULONG,
+    VendorID: USHORT,
+    ProductID: USHORT,
+    VersionNumber: USHORT,
+}
```
gingerBill, line 17: Please use tabs for indentation everywhere.

gingerBill (changes requested):

gingerBill (approved):

PR #3509, opened 2024-04-28
Make more `core:unicode` procs contextless

Feoramund:


gingerBill:
Please don't make random callbacks contextless. This is not a good change.

PR #3540, opened 2024-05-06
Handle fixed soa arrays

VictorSohier:
The panic was being hit on #soa types of fixed length in a procedure type arguments and variable declaration. Tested on Linux.

This is to fix the following case:
```odin
package main;

import "core:fmt"

Bar :: struct
{
	a: i32,
	b: f32,
}

main :: proc()
{
	foo: #soa[8]Bar;
	fmt.printfln("%#v", foo);
	baz(#soa[8]Bar, foo);
}

baz :: proc($T: typeid/#soa[$N]$E, param: T)
{
	fmt.printfln("%#v", param);
}
```
This does not manifest when you manually resolve the polyparametric.

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -102,6 +102,7 @@ gb_internal Type *   check_init_variable            (CheckerContext *c, Entity *
 gb_internal void check_assignment_error_suggestion(CheckerContext *c, Operand *o, Type *type, i64 max_bit_size=0);
 gb_internal void add_map_key_type_dependencies(CheckerContext *ctx, Type *key);
 
+gb_internal Type *make_soa_struct_fixed(CheckerContext *ctx, Ast *array_typ_expr, Ast *elem_expr, Type *elem, i64 count, Type *generic_type);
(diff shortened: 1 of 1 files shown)


src/check_expr.cpp
```diff
@@ -1409,11 +1410,16 @@ gb_internal bool is_polymorphic_type_assignable(CheckerContext *c, Type *poly, T
...
 				bool ok = is_polymorphic_type_assignable(c, poly->Struct.soa_elem, source->Struct.soa_elem, true, modify_type);
 				if (ok) switch (source->Struct.soa_kind) {
-				case StructSoa_Fixed:
 				default:
 					GB_PANIC("Unhandled SOA Kind");
 					break;
-
+				case StructSoa_Fixed:
+					if (modify_type) {
+						bool breakpoint = true;
```
JesseRMeyer, line 1418: Please remove debug code.

VictorSohier: Now that I look at it, there is an unhandled case remaining. I think might actually better if I keep it there and explicitly put in the unhandled case of `StructSoa_None` instead (See types.cpp:125). Even though the rest of the cases are still now handled, it might be better to have that variant to be explicit in this switch. If it is an invalid case, crashing here would retain the existing behavior for that case, even if it may never be hit.

PR #3541, opened 2024-05-06
src: Re-enable the new StringMap

Yawning:
- Make `CheckerInfo.files` an array as it needs to be sorted and nothing queries by key.
- Re-enable the new StringMap (this includes the fix for the backward type alias correction)
- Remove the old StringMap

Whew.  This makes the test cases pass instead of crash on my system(s).  If this should be done another way, feel free to close the branch, but I'm satisfied that I root-caused the crashes I was seeing.

### DIFF
--- a/src/string_map.cpp
+++ b/src/string_map.cpp
@@ -25,7 +25,7 @@ gb_internal gb_inline StringHashKey string_hash_string(String const &s) {
 }
 
 
-#if 1 /* old string map */
+// #if 1 /* old string map */
 
 template <typename T>
 struct OldStringMapEntry {
@@ -312,7 +312,7 @@ gb_internal OldStringMapEntry<T> const *end(OldStringMap<T> const &m) noexcept {
 	return m.entries + m.count;
 }
 
-#else /* new string map */
+// #else /* new string map */
 
 template <typename T>
 struct StringMapEntry {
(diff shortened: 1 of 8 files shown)


src/string_map.cpp
```diff
@@ -25,28 +25,26 @@ gb_internal gb_inline StringHashKey string_hash_string(String const &s) {
...
+	u32    hash;
+	T      value;
 };
 
 template <typename T>
 struct StringMap {
-	MapIndex *         hashes;
-	usize              hashes_count;
 	StringMapEntry<T> *entries;
+	Array<T>           ordered_entries;
```
gingerBill, line 38: That's an absolutely horrible bodge.

gingerBill (changes requested):
I don't think I will accept this bodge, and I think it needs to be solves in a much better way.

Yawning:
Changed to just use `c->parser->packages` and `packages->files` which are conveniently sorted in the correct order.

src/string_map.cpp
```diff
@@ -25,295 +25,6 @@ gb_internal gb_inline StringHashKey string_hash_string(String const &s) {
 }
 
 
-#if 1 /* old string map */
```
gingerBill, line 28: Might be better to just set this to `0` rather than delete the code

Yawning: Done.

PR #3547, opened 2024-05-06
sys/linux: Pass size to epoll_create

hnakamur:
Without this fix, you can see a random size value is passed to epoll_create by using strace.

### DIFF
--- a/core/sys/linux/sys.odin
+++ b/core/sys/linux/sys.odin
@@ -2314,7 +2314,7 @@ futex :: proc {
 */
 epoll_create :: proc(size: i32 = 1) -> (Fd, Errno) {
 	when ODIN_ARCH != .arm64 {
-		ret := syscall(SYS_epoll_create)
+		ret := syscall(SYS_epoll_create, i32(1))
 		return errno_unwrap(ret, Fd)
 	} else {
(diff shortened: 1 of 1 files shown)


flysand7:
I realised the `size` parameter to `epoll_create` procedure is obsolete and not needed, maybe we should just call `epoll_create1` and have the `flags` parameter optionally allow `CLOEXEC`

hnakamur:
According to [Ubuntu Manpage: epoll_create, epoll_create1 - open an epoll file descriptor](https://manpages.ubuntu.com/manpages/noble/en/man2/epoll_create1.2.html), `epoll_create1` were added to Linux 2.6.27.

Since we have [epoll_create1](https://github.com/odin-lang/Odin/blob/8e263de4aa2a48cb26bf235bdad162de5c1547c8/core/sys/linux/sys.odin#L2695-L2698), maybe we can just deprecate [epoll_create](https://github.com/odin-lang/Odin/blob/8e263de4aa2a48cb26bf235bdad162de5c1547c8/core/sys/linux/sys.odin#L2315-L2323), or let `epoll_create` call always `epoll_create` syscall.
I don't understand why the current implementation `epoll_create` calls `epoll_create1` when `ODIN_ARCH == .arm64`

PR #3564, opened 2024-05-09
fix a couple of `-vet-style` failures 

laytan:
after f54977336b27c32eab52b77d94e7b1610f4350cf


laytan:
That's a very, very weird test failure

PR #3571, opened 2024-05-10
build: add llvm18 support

chenrui333:
- https://github.com/Homebrew/homebrew-core/pull/171372

cc @laytan

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,14 +5,17 @@ jobs:
   build_linux:
     name: Ubuntu Build, Check, and Test
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        llvm_version: [17, 18]
     steps:
       - uses: actions/checkout@v1
       - name: Download LLVM
         run: |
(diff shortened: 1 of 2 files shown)


laytan:
This is a weird one? Why would we need to change things here?

We don't want to run on the latest and greatest llvm in our ci because we also support older versions, and generally wait a bit on upgrading CI/releases.

chenrui333:
> This is a weird one? Why would we need to change things here?

to match with the homebrew build?

chenrui333:
what is the concern here?

laytan:
> > This is a weird one? Why would we need to change things here?
> 
> to match with the homebrew build?

Alright, but for the reasons I stated in the last comment we don't necessarily want that.

chenrui333:
that is fine, let's do matrix build

---

note, this is some common practice I am doing to build alliance between upstream and downstream/homebrew

laytan:
I would be fine with a matrix, I don't have final say here though

chenrui333:
updated to use the matrix build

PR #3572, opened 2024-05-10
Fix joining non-`Started` threads from blocking main thread

### LINKED ISSUE #2746: thread.pool_destroy halt if no task was added

Feoramund:
While investigating #2746, I noticed this was happening for UNIX threads too. I was able to fix and test it for threads on Linux, but I do not have a Windows machine to test my fix there, so if someone can try it out, that would be great.

Here's the program I tested with:
```odin
package main

import "core:fmt"
import "core:thread"

main :: proc () {
	p: thread.Pool
	fmt.println("pool_init:")
	thread.pool_init(&p, context.allocator, 1)
	fmt.println("pool_destroy:")
	thread.pool_destroy(&p)
	fmt.println("done")
}
```

Fixes #2746

### DIFF
--- a/core/thread/thread_unix.odin
+++ b/core/thread/thread_unix.odin
@@ -36,6 +36,10 @@ _create :: proc(procedure: Thread_Proc, priority: Thread_Priority) -> ^Thread {
 			sync.wait(&t.cond, &t.mutex)
 		}
 
+		if .Joined in t.flags {
+			return nil
+		}
+
 		when ODIN_OS != .Darwin {
 			// Enable thread's cancelability.
 			if can_set_thread_cancel_state {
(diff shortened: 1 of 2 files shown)


gingerBill:
This doesn't work on Windows, but I am merging it anyway to fix Unix threads.

gingerBill:
I have now fixed this on Windows, it was due to a deadlock in the mutex when trying to call `_start` within `_join`.

PR #3575, opened 2024-05-11
vendor/glfw: add native linux bindings

JopStro:
Some basic bindings to glfw native linux platform functions, I used rawptr to represent the wayland handles since there isn't a vendor library for wayland currently (that I could see). I didn't add bindings for the xrandr specific functions as I personally didn't see a need to, but they could potentially be added in a similar way to the wayland specific functions if needed.

This pr also adds a static library for linux (built from the 3.4 release), since it was pretty straightforward to do so.


JopStro:
Addressed by #3740

PR #3583, opened 2024-05-13
Fix #3578

### LINKED ISSUE #3578: Invalid code compiles and produces a binary that crashes with SIGSEGV
## Context

```
	Odin:    dev-2024-05:8e263de4a
	OS:      Fedora Linux 40 (Workstation Edition), Linux 6.8.9-300.fc40.x86_64
	CPU:     AMD Ryzen 9 5900X 12-Core Processor            
	RAM:     128721 MiB
	Backend: LLVM 18.1.1
```

The following code shoud not compile but it does (it's missing a return in `foo`):
```
package main

import "core:fmt"

foo :: proc() -> []byte {
    loop: for {
        break loop
    }
}


main :: proc() {
    fmt.println(foo())
}
```

The resulting binary either crashes with a SIGSEGV or prints garbage:

## Expected Behavior

I expect the code to not compile

## Current Behavior

The code compiles and produces a buggy binary.

### Steps to Reproduce

Simply run the code with: `odin run main.odin -file`

FrancisTheCat:
Fix checking for infinite loops to account for labels


gingerBill:
I've fixed this in another commit, and this one was subtly wrong too (even if it worked).

PR #3592, opened 2024-05-16
add a default heap/general purpose allocator for wasm to `base:runtime`

laytan:
I've tested this a lot with some random allocation patterns and sizes, on both wasm32 and wasm64p32. It has also been in use on one of my projects for a few weeks now.

There are cases where the wasm environment does not provide memory, but in that case the allocator will panic on the first allocation, which I feel like is fine because the panic allocator (previous default) would do that too.

This started as a port of emmalloc (an emscripten allocator), but I had to make quite a few modifications to make it work well and nicely in Odin, the main logic is still the same.

### DIFF
--- a/base/runtime/wasm_allocator.odin
+++ b/base/runtime/wasm_allocator.odin
@@ -1,0 +1,870 @@
+//+build wasm32, wasm64p32
+package runtime
+
+import "base:intrinsics"
+
+/*
+Port of emmalloc, modified for use in Odin.
+
+Invariants:
+ - Per-allocation header overhead is 8 bytes, smallest allocated payload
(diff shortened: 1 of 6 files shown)


gingerBill (approved):

base/runtime/wasm_allocator.odin
```diff
@@ -0,0 +1,870 @@
...
+				if ok {
+					return
+				}
+
+				if state == .Waiting {
+					break spin_lock
+				}
+
+				for i := min(spin+1, 32); i > 0; i -= 1 {
+					intrinsics.cpu_relax()
```
gingerBill, line 288: On wasm, this doesn't do anything, IIRC.

gingerBill (approved):

PR #3594, opened 2024-05-16
correctly load/store uint&int in the runtime.js for wasm64p32

laytan:


vendor/wasm/js/runtime.js
```diff
@@ -13,7 +13,8 @@ function stripNewline(str) {
     return str.replace(/\n/, ' ')
 }
 
-const STRING_SIZE = 2*4;
+const INT_SIZE    = 4; // NOTE: set to `8` if the target has 64 bit ints (`wasm64p32` for example).
+const STRING_SIZE = 4+INT_SIZE;
```
gingerBill, line 17: That's still wrong. `STRING_SIZE` is `2*INT_SIZE` because there will be padding between the data pointer and the `len`.

laytan: Right, dumb mistake, corrected

Issue #2974, opened 2023-11-24
LLVM failure when using `intrinsics.atomic_*` operations


iansimonson:
## Context

This is actually two different bugs, one is platform specific (intel Mac) the other is platform agnostic. It's all related to the following code snippet from a port of Odin's MPSC queue found in `queue.cpp`

```odin
enqueue_node :: proc(q: ^Queue($T), node: ^Node(T)) -> int {
	intrinsics.atomic_store_explicit(&node.next, nil, .Release)
        prev := intrinsics.atomic_exchange_explicit(&q.head, node, .Acq_Rel)
	intrinsics.atomic_store_explicit(&prev.next, node, .Release)
	count := 1 + intrinsics.atomic_add_explicit(&q.count, 1, .Release)
	return count
}

dequeue :: proc(q: ^Queue($T)) -> (value: T, has_item: bool) {
	tail := intrinsics.atomic_load_explicit(&q.tail, .Relaxed)
	next := intrinsics.atomic_load_explicit(&tail.next, .Relaxed)
	if next != nil {
		context.allocator = q.allocator
		intrinsics.atomic_store_explicit(&q.tail, next, .Relaxed)
		value = next.value
		intrinsics.atomic_sub_explicit(&q.count, 1, .Relaxed)
		if tail != &q.sentinel {
			//free(tail)
		}
		return value, true
	}

	assert(intrinsics.atomic_load_explicit(&q.count, .Acquire) == 0)
	return {}, false
}
```

### Odin Report
For the mac only bug:
```txt
Odin: dev-2023-11:3c021f9c5
OS:   macOS Monterey 12.5 (build: 21G83, kernel: 21.6.0)
CPU:  Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
RAM:  8192 MiB
```

for the bug that affects all systems:
```txt
Odin: dev-2023-11:042440414
OS:   Windows 11 Professional (version: 22H2), build 22621.2715
CPU:  AMD Ryzen 7 5800X 8-Core Processor
RAM:  32696 MiB
```

## Expected Behavior

(mac bug): in the second line of `enqueue_node`, `intrinsics.atomic_exchange_explicit` should be able to take a `^^Node(T)` in a parapoly proc and function properly. It can be worked around by casting to a `^uintptr` and back

(all platforms) In `dequeu`, the `intrinsics.atomic_sub_explicit` should work with `.Relaxed` memory ordering

## Current Behavior

[...]

PR #3607, opened 2024-05-20
Fix Odin to LLVM memory order mapping for .Relaxed and .Consume

Szwagi:
This fixes how Odin's atomic memory order is mapped to LLVM. Now it's the same as Clang.

Before this, .Relaxed threw an LLVM error in some situations and .Consume was more relaxed than it should be.
```rust
sync.atomic_add_explicit(&foo.ref_count, 1, .Relaxed);
```
```llvm
atomicrmw instructions cannot be unordered.
  %31 = atomicrmw volatile add ptr %30, i64 1 unordered, align 8
```

#2974

### DIFF
--- a/src/llvm_backend_utility.cpp
+++ b/src/llvm_backend_utility.cpp
@@ -2223,8 +2223,8 @@ gb_internal LLVMAtomicOrdering llvm_atomic_ordering_from_odin(ExactValue const &
 	GB_ASSERT(value.kind == ExactValue_Integer);
 	i64 v = exact_value_to_i64(value);
 	switch (v) {
-	case OdinAtomicMemoryOrder_relaxed: return LLVMAtomicOrderingUnordered;
-	case OdinAtomicMemoryOrder_consume: return LLVMAtomicOrderingMonotonic;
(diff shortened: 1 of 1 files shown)


src/llvm_backend_utility.cpp
```diff
@@ -2223,8 +2223,8 @@ gb_internal LLVMAtomicOrdering llvm_atomic_ordering_from_odin(ExactValue const &
 	GB_ASSERT(value.kind == ExactValue_Integer);
 	i64 v = exact_value_to_i64(value);
 	switch (v) {
-	case OdinAtomicMemoryOrder_relaxed: return LLVMAtomicOrderingUnordered;
-	case OdinAtomicMemoryOrder_consume: return LLVMAtomicOrderingMonotonic;
+	case OdinAtomicMemoryOrder_relaxed: return LLVMAtomicOrderingMonotonic;
+	case OdinAtomicMemoryOrder_consume: return LLVMAtomicOrderingAcquire;
```
gingerBill, line 2227: But that isn't correct. That is just a bodge for your problem.

Feoramund: Did this turn out to not be a bodge?

PR #3611, opened 2024-05-22
vendor/microui: nicer textbox

hstormo:
This makes microui textboxes a bit nicer to use, adding selection manipulation and cut/copy/paste using the text_edit helpers.

I've left out undo/redo since it would introduce allocation, but let me know if you think there's a nice way to do it.

### DIFF
--- a/vendor/microui/microui.odin
+++ b/vendor/microui/microui.odin
@@ -984,16 +984,6 @@ checkbox :: proc(ctx: ^Context, label: string, state: ^bool) -> (res: Result_Set
 }
 
 textbox_raw :: proc(ctx: ^Context, textbuf: []u8, textlen: ^int, id: Id, r: Rect, opt := Options{}) -> (res: Result_Set) {
-	try_input_string :: proc(state: ^textedit.State, textbuf: []u8, textlen: ^int, s: string) -> bool {
-		n := min(len(textbuf) - textlen^, len(s))
-		if n > 0 {
-			textedit.input_text(state, s[:n])
-			textlen^ = strings.builder_len(state.builder^)
-			return true
-		}
-		return false
-	}
-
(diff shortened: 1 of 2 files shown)


vendor/microui/microui.odin
```diff
@@ -965,49 +984,149 @@ checkbox :: proc(ctx: ^Context, label: string, state: ^bool) -> (res: Result_Set
...
 		if .RETURN in ctx.key_pressed_bits {
 			set_focus(ctx, 0)
 			res += {.SUBMIT}
 		}
+
+		/* handle click/drag */
+		if .LEFT in ctx.mouse_down_bits {
+			idx := textlen^
+			for i in 0..<textlen^ {
+				if textbuf[i] >= 0x80 && textbuf[i] < 0xc0 do continue // skip continuation bytes
```
gingerBill, line 1097: Try not do use `do` here if you can.

vendor/microui/microui.odin
```diff
@@ -965,49 +984,149 @@ checkbox :: proc(ctx: ^Context, label: string, state: ^bool) -> (res: Result_Set
...
+			textedit.copy(&ctx.textbox_state)
+		}
+		/* handle ctrl+v */
+		if .V in ctx.key_pressed_bits && .CTRL in ctx.key_down_bits && .ALT not_in ctx.key_down_bits {
+			if ctx.textbox_state.get_clipboard != nil {
+				if s, ok := ctx.textbox_state.get_clipboard(ctx.textbox_state.clipboard_user_data); ok {
+					if try_input_string(&ctx.textbox_state, textbuf, textlen, s) do res += {.CHANGE}
 				}
 			}
+		}
```
gingerBill, line 1041: Why are you not using the copy and paste commands?

gingerBill: Then the text edit paste needs fixing then.

PR #3614, opened 2024-05-23
Add `core:container/rbtree`

Kelimion:
Add a Red-Black tree with configurable `$Key` and `$Value`,  as well as an `Iterator` and optional `on_remove` callback.
Also includes tests that verify it maintains Red-Black and Iterator invariants and doesn't leak.

Originally based on the CC0 implementation from literateprograms.org. But reworked to the same API used by @Yawning's excellent `core:container/avl` for ease of use.

### DIFF
--- a/core/container/rbtree/rbtree.odin
+++ b/core/container/rbtree/rbtree.odin
@@ -1,5 +1,5 @@
 // This package implements a red-black tree
-package container_intrusive_rbtree
+package container_rbtree
 
 @(require) import "base:intrinsics"
 @(require) import "base:runtime"
--- a/core/container/rbtree/rbtree.odin
+++ b/core/container/rbtree/rbtree.odin
@@ -131,7 +131,7 @@ find_value :: proc(t: ^$T/Tree($Key, $Value), key: Key) -> (value: Value, ok: bo
 // find_or_insert attempts to insert the value into the tree, and returns
 // the node, a boolean indicating if the value was inserted, and the
 // node allocator error if relevant.  If the value is already present, the existing node is updated.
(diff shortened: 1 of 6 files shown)


core/container/rbtree/rbtree.odin
```diff
@@ -0,0 +1,567 @@
...
+		case .Less:
+			n_ptr = &n._left
+		case .Greater:
+			n_ptr = &n._right
+		case .Equal:
+			return
+		}
+	}
+	_parent := n
+	n = new_clone(Node(Key, Value){key=key, value=value, _parent=_parent, _color=.Red}, t._node_allocator) // or_return
```
gingerBill, line 148: Why not do?
```odin
err: runtime.Allocator_Error
n, err = new_clone(...)
if err != nil {
    return
}
```

core/container/rbtree/rbtree.odin
```diff
@@ -0,0 +1,567 @@
...
+	}
+	free(node, t._node_allocator)
+	t._size -= 1
+	return true
+}
+
+// iterator returns a tree iterator in the specified direction.
+iterator :: proc "contextless" (t: ^$T/Tree($Key, $Value), direction: Direction) -> Iterator(Key, Value) {
+	it: Iterator(Key, Value)
+	it._tree      = transmute(^Tree(Key, Value))t
```
gingerBill, line 211: A normal `cast` should do.

core/container/rbtree/rbtree.odin
Kelimion: I forgot to add `err: runtime.Allocator_Error` to the return values, like `core:container/avl` has, so I added that and updated the test to `_` the err.

core/container/rbtree/rbtree.odin
Kelimion: Done.

PR #3615, opened 2024-05-23
Add build flag to export dependencies.

puzzleddev:
Adds the `-export-dependencies` build flag. This takes a file to which dependency information about the current build will be written in Make format. This is roughly compatible with GCCs `-MD -MF` flag and help projects which contain multiple independent Odin builds, only some of which should be rebuilt during any given build.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -235,6 +235,7 @@ enum BuildFlagKind {
 	BuildFlag_ExportTimings,
 	BuildFlag_ExportTimingsFile,
 	BuildFlag_ExportDependencies,
+	BuildFlag_ExportDependenciesFile,
 	BuildFlag_ShowSystemCalls,
 	BuildFlag_ThreadCount,
 	BuildFlag_KeepTempFiles,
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -2916,10 +2935,54 @@ int main(int arg_count, char const **arg_ptr) {
...
+				for (isize k = 0; k < file->fullpath.len; k++) {
+					char part = file->fullpath.text[k];
+					if (part == ' ') {
+						gb_file_write(&f, "\\", 1);
+						current_line_length++;
+					}
+					gb_file_write(&f, &part, 1);
+					current_line_length++;
+				}
+			}
```
gingerBill, line 2979: It might make sense to export it to a .json file, because it'll need more than just the .odin files in the future, maybe even the linking libraries and other things.

puzzleddev: I want to maintain the ability to output in Make format for compatibility with GCC and Clang, and existing build tools. But I agree that it is not a great format for new tools and will implement a way to output either as Make, or a JSON file.

puzzleddev:
The dependency format can now be set like timing export format. Currently both only consider the parsed source files for compatibility with GCC/Clangs `-MD -MF` options, but if desired other types like libraries and windows resources files can be included as well.

PR #3616, opened 2024-05-24
Fix os_sync_wake_by* variants taking wrong flags type.

harold-b:
As the title says -- small fix.

### DIFF
--- a/core/sys/darwin/sync.odin
+++ b/core/sys/darwin/sync.odin
@@ -283,7 +283,7 @@ foreign system {
 	//				and the shared memory specification
 	//				(See os_sync_wake_by_address_flags_t).
 	// ENOENT	:	No waiter(s) found waiting on the @addr.
-	os_sync_wake_by_address_any :: proc(addr: rawptr, size: uint, flags: os_sync_wait_on_address_flags) -> i32 ---
+	os_sync_wake_by_address_any :: proc(addr: rawptr, size: uint, flags: os_sync_wake_by_address_flag) -> i32 ---
 
 	// This function is a variant of os_sync_wake_by_address_any that wakes up all waiters
 	// blocked in os_sync_wait_on_address or its variants.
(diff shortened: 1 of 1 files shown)


core/sys/darwin/sync.odin
```diff
@@ -283,7 +283,7 @@ foreign system {
 	//				and the shared memory specification
 	//				(See os_sync_wake_by_address_flags_t).
 	// ENOENT	:	No waiter(s) found waiting on the @addr.
-	os_sync_wake_by_address_any :: proc(addr: rawptr, size: uint, flags: os_sync_wait_on_address_flags) -> i32 ---
+	os_sync_wake_by_address_any :: proc(addr: rawptr, size: uint, flags: os_sync_wake_by_address_flag) -> i32 ---
```
gingerBill, line 286: This is actually the fix, the real fix would be to fix the enums directly.

harold-b:
This is still wrong -- the point is that the wake_**by** variants are incorrectly taking the flags of the  wake_**on** variants (even if they are the same under the hood).

That is, `os_sync_wake_by_address_any` and `os_sync_wake_by_address_all` are supposed to take `os_sync_wake_by_address_flag` not `os_sync_wait_on_address_flag`, as it currently does.

gingerBill:
Again, no. It should take the `flags` `bit_set` not the `flag` `enum`.

harold-b:
The problem is not bit_set vs enum, it is that the wrong bit_set *type* is being taken as a parameter. `wake_on` should take `wake_on` bit_sets, `wake_by` tshould ake `wake_by` bit_sets. But right now the the `wake_by` procs are taking the `wake_on` bit_sets:

```odin
os_sync_wake_by_address_all :: proc(addr: rawptr, size: uint, flags: os_sync_wait_on_address_flags) -> i32 ---

flags: os_sync_wait_on_address_flags
                           to
flags: os_sync_wake_by_address_flags
```

gingerBill:
I'll fix the type. It doesn't really make much a difference in practice because the memory layout and names are IDENTICAL.

harold-b:
Thanks - - Yes, I acknowledges such here 
 > (even if they are the same under the hood)

But it would still be good to have the correct type on the proc

PR #3617, opened 2024-05-24
Clearer function parameters for core:os:read_dir()

romero-kenny:
I was confused when initially using read_dir and had to read the code to figure out what n controlled. Decided to make it more explicit that n was the size of the fi return array and display the default size.

### DIFF
--- a/core/os/dir_bsd.odin
+++ b/core/os/dir_bsd.odin
@@ -3,7 +3,7 @@ package os
 
 import "core:mem"
 
-read_dir :: proc(fd: Handle, n: int, allocator := context.allocator) -> (fi: []File_Info, err: Errno) {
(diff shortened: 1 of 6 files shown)


core/os/dir_bsd.odin
```diff
@@ -3,7 +3,7 @@ package os
 
 import "core:mem"
 
-read_dir :: proc(fd: Handle, n: int, allocator := context.allocator) -> (fi: []File_Info, err: Errno) {
+read_dir :: proc(fd: Handle, fi_size: int = 100, allocator := context.allocator) -> (fi: []File_Info, err: Errno) {
```
gingerBill, line 6: Please no. If it was confusing, you should have improved the documentation, not add a default parameter.

laytan:
This changes the behaviour quite significantly, previously if `n` was `<= 0` it would set the initial capacity of the array to 100 BUT it could return more than that, now it is just a maximum of 100

PR #3619, opened 2024-05-24
Add iterator_next(&r) to CSV.

Kelimion:
See also `csv/example.odin` with 3 usage scenarios.

### DIFF
--- a/core/encoding/csv/reader.odin
+++ b/core/encoding/csv/reader.odin
@@ -57,6 +57,9 @@ Reader :: struct {
 	field_indices: [dynamic]int,
 	last_record:   [dynamic]string,
 	sr: strings.Reader, // used by reader_init_with_string
+
+	// Set and used by the iterator. Query using `iterator_last_error`
+	last_iterator_error: Error,
 }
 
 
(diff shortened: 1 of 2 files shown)


core/encoding/csv/reader.odin
```diff
@@ -121,6 +124,25 @@ reader_destroy :: proc(r: ^Reader) {
...
+*/
+iterator_next :: proc(r: ^Reader) -> (record: []string, idx: int, err: Error, more: bool) {
+	record, r.last_iterator_error = read(r)
+	return record, r.line_count - 1, r.last_iterator_error, r.last_iterator_error == nil
+}
+
+// Get last error if we the iterator
+iterator_last_error :: proc(r: Reader) -> (err: Error) {
+	return r.last_iterator_error
+}
```
gingerBill, line 144: Does this need to be stored any more if we also return it from `iterator_next`?

Kelimion: Yes: `for rec, idx in ...`  and `for rec in ...` don't have access to the error otherwise.

PR #3624, opened 2024-05-24
Allow for lower-case z in RFC 3339.

yay:
https://www.rfc-editor.org/rfc/rfc3339#section-5.6


Kelimion:
Just added the same change.

PR #3628, opened 2024-05-24
fix(core/encoding/json): remove Unsupported_Type return when marshalling bit_set

IllusionMan1212:
This seems to have been left here accidentally


Kelimion:
If you're up for it, can you also add a test to `tests\core\encoding\json` that includes marshaling and unmarshaling `bit_set`s?

PR #3631, opened 2024-05-25
fix/test(encoding/json): support unmarshalling `bit_set`s

IllusionMan1212:
Also added a test for marshalling and unmarshalling `bit_set`

### DIFF
--- a/core/encoding/json/unmarshal.odin
+++ b/core/encoding/json/unmarshal.odin
@@ -86,6 +86,12 @@ assign_bool :: proc(val: any, b: bool) -> bool {
 @(private)
 assign_int :: proc(val: any, i: $T) -> bool {
 	v := reflect.any_core(val)
+	ti := runtime.type_info_base(type_info_of(v.id))
+
+	if bs, ok := ti.variant.(runtime.Type_Info_Bit_Set); ok {
+			v.id = reflect.type_info_core(bs.elem).id
+	}
+
 	switch &dst in v {
 	case i8:      dst = i8     (i)
 	case i16:     dst = i16    (i)
--- a/tests/core/encoding/json/test_core_json.odin
+++ b/tests/core/encoding/json/test_core_json.odin
@@ -35,6 +35,7 @@ main :: proc() {
 	unmarshal_json(&t)
 	surrogate(&t)
 	utf8_string_of_multibyte_characters(&t)
+	marshal_and_unmarshal_bit_set(&t)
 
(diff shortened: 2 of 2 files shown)


core/encoding/json/unmarshal.odin
```diff
@@ -86,6 +86,12 @@ assign_bool :: proc(val: any, b: bool) -> bool {
 @(private)
 assign_int :: proc(val: any, i: $T) -> bool {
 	v := reflect.any_core(val)
+	ti := runtime.type_info_base(type_info_of(v.id))
+
+	if bs, ok := ti.variant.(runtime.Type_Info_Bit_Set); ok {
+			v.id = reflect.type_info_core(bs.elem).id
+	}
+
```
gingerBill, line 94: I am not sure if this is a good solution or not. Is it better to save it as an integer or to save it as an array of values?

Kelimion: If saved as an array of values, then it would keep working if you later renumber the enum's iota.

IllusionMan1212: I'm not sure I understand how an array of values will keep the unmarshalling consistent if the enum was reordered or renumbered. Can you give an example?

Kelimion:
```cpp
Enum :: enum{.A, .B, .C, .D, .E}
Bit_Set :: bit_set[Enum;u8]
v := Bit_Set{.A, .C, .E}
```
So far it's just bit 0, 2, and 4 set, right, so we'd encode it as 1 + 4 + 16 = 21.
Now unmarshal after we do this:
```cpp
Enum :: enum{.A, .A2, .B, .B2, .C, .D, .E}
```
A C E in -> 21 -> A B C out.

Now marshal `{.A, .C, .E}` as `"A,C,E"`. Rearrange the enum to `{.E, .F, .A, .C, .B}`, but `"A,C,E"` still unmarshals to the same `{.A, .C, .E}`.

Kelimion: And better to get an unmarshal error because a value "Foo" is missing in the enum since you marshaled it, then to unmarshal A = 5 to Z = 5 later without being able to notice it was wrong.

IllusionMan1212: Right, I see what you mean now, I thought an array of values meant an array of 0s and 1s. That does seem better than an integer.
Should this get merged first and then we can change bit_set (un)marshalling to be an array or should I close this PR?

PR #3648, opened 2024-05-28
Remove unused code in `internal_random_prime`

Feoramund:
It looks like it's there for a reason, but it's never referred to again.


Kelimion:
You found the easter egg. Tea for you and tea for me.

(No, I don't know why it was there either.)

PR #3663, opened 2024-06-03
Add missing FlushConsoleInputBuffer procedure to core/sys/windows/kernel32.odin

ProChaser:
While working on a project, I noticed that the procedure was missing.  The Microsoft documentation for `FlushConsoleInputBuffer` can be found [here](https://learn.microsoft.com/en-us/windows/console/flushconsoleinputbuffer).


Kelimion:
The entire old-school console API family is on life support.
image

Microsoft themselves strongly recommend against using this particular API in new projects.
image

As such we won't be adding it to our win32 bindings. If you still have an irreplaceable need for it, you should be able to bind it in your project no problem. Linking to kernel32 twice won't introduce any conflicts.

PR #3664, opened 2024-06-03
Fix signatures of two Win32 functions

reduf:
The signature of `CreateIoCompletionPort` and `GetQueuedCompletionStatusEx` are not quite the same as in C. The `ULONG_PTR` is an integer and not a pointer.

Seems to have been incorrectly changed in 0f1c5b3891a578b65280e78cb9c1831190d496d5.


reduf:
Make sense, I updated the PR to use those.

PR #3666, opened 2024-06-03
Add missing `Raw_*` complex and quaternion types

Feoramund:
I'm not sure why those types were defined in `core:mem`, but I moved them to `base:runtime` where the other `Raw_*` types live.

To be clear, this adds the following types to `base:runtime`:
```odin
Raw_Complex32     :: struct {real, imag: f16}
Raw_Quaternion64  :: struct {imag, jmag, kmag: f16, real: f16}
Raw_Quaternion64_Vector_Scalar  :: struct {vector: [3]f16, scalar: f16}
```

### DIFF
--- a/core/mem/raw.odin
+++ b/core/mem/raw.odin
@@ -11,13 +11,6 @@ Raw_Dynamic_Array :: runtime.Raw_Dynamic_Array
 Raw_Map           :: runtime.Raw_Map
 Raw_Soa_Pointer   :: runtime.Raw_Soa_Pointer
 
-Raw_Complex64     :: struct {real, imag: f32}
-Raw_Complex128    :: struct {real, imag: f64}
-Raw_Quaternion128 :: struct {imag, jmag, kmag: f32, real: f32}
-Raw_Quaternion256 :: struct {imag, jmag, kmag: f64, real: f64}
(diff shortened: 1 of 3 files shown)


core/mem/raw.odin
```diff
@@ -11,13 +11,6 @@ Raw_Dynamic_Array :: runtime.Raw_Dynamic_Array
 Raw_Map           :: runtime.Raw_Map
 Raw_Soa_Pointer   :: runtime.Raw_Soa_Pointer
 
-Raw_Complex64     :: struct {real, imag: f32}
-Raw_Complex128    :: struct {real, imag: f64}
-Raw_Quaternion128 :: struct {imag, jmag, kmag: f32, real: f32}
-Raw_Quaternion256 :: struct {imag, jmag, kmag: f64, real: f64}
-Raw_Quaternion128_Vector_Scalar :: struct {vector: [3]f32, scalar: f32}
-Raw_Quaternion256_Vector_Scalar :: struct {vector: [3]f64, scalar: f64}
```
gingerBill, line 19: I'd still keep these by just having them be the aliases of the `runtime` versions.

gingerBill (changes requested):

core/mem/raw.odin
Feoramund: All set.

PR #3673, opened 2024-06-04
Implement `#exists(path)` and use it to provide good errors for common missing vendor libraries

laytan:


src/check_expr.cpp
```diff
@@ -8329,6 +8330,7 @@ gb_internal ExprKind check_basic_directive_expr(CheckerContext *c, Operand *o, A
 		    name == "assert" ||
 		    name == "defined" ||
 		    name == "config" ||
+			name == "exists" ||
```
gingerBill, line 8333: Minor indentation issue

PR #3684, opened 2024-06-05
`core:mem/tlsf` - "Two-Level Segregated Fit" memory allocator

gingerBill:


core/mem/tlsf/tlsf.odin
```diff
@@ -0,0 +1,156 @@
+/*
+	Copyright 2024 Jeroen van Rijn <nom@duclavier.com>.
+	Made available under Odin's BSD-3 license.
+
+	List of contributors:
+		Matt Conte:      Original C implementation, see LICENSE file in this package
```
Kelimion, line 6: We do need to add back the original LICENSE file somewhere, per the "this must be reproduced" language. Could also add it to `tlsf_internal.odin`'s end?

Kelimion (approved):
Other than the C version's license needing to be added back, it looks great.

core/mem/tlsf/tlsf.odin
gingerBill: I think adding the LICENSE file directly should be fine.

PR #3690, opened 2024-06-05
Runtime support for orca

laytan:
### Make assertions, panics, type checks, bounds checks etc. go through orca's APIs

This looks a bit sketchy because our runtime is designed around being able to do writes to stderr like it is done on all other targets, but orca has APIs for logging, aborting etc which you can't just write one part of the string to and later another part. + these APIs take (null terminated) cstrings.

For this reason, I added a buffering mechanism that will flush to orca's log once a newline is found.

### Set up our entrypoint

We now set up an internal `oc_on_init` and `oc_on_terminate` where it sets up the runtime, calls `@init` procs, and then calls the user's `main` proc. Same deal with `oc_on_terminate` calling `@fini`.

Users can opt-out of our entry point with `-no-entry-point` and can then provide these exports themselves.

### Set up allocators

Orca provides a libc malloc API we can easily use

### Linking

Uses the output of `orca sdk-path` to automatically link to the needed things for a correct wasm module.

## What works now

(Tested on Windows and MacOS)

```odin

package main

import "base:runtime"

main :: proc() {
    runtime.print_string("Hello Orca\n")
}
```

`odin build . -target:orca_wasm32 -out:module.wasm`
`orca bundle module.wasm`

results in a valid Orca app that prints "Hello Orca" to the log and shows a blank screen.

@Skytrias Is working on auto-generated bindings for `core:sys/orca` to get to the point of building apps/games with it.

I used the in-progress bindings to port the breakout example of Orca and that ran smoothly, so it should just be a matter of adding them.


laytan:
One, maybe weird thing here is that the end of `main` is not the end of the program, this is actually also the case in the `js_wasm` targets but worth noting.

Skytrias:
Lovely work, much appreciated laytan 💪 
Auto bindings generator is at https://github.com/Skytrias/orca-odin which can be used for testing till we merge the generated file in

PR #3717, opened 2024-06-10
Add permutation & combination procs to `core:math/big`

Feoramund:
While making this PR, I noticed the test setup in `math/big` involves a library build and checking against Python's implementation. I renamed the package to conform to the convention of the rest of the tests, but I've otherwise left it alone. I'm not sure if the library build should be moved, since it does not have any `@test` procs on it.

### DIFF
--- a/core/math/big/combinatorics.odin
+++ b/core/math/big/combinatorics.odin
@@ -1,0 +1,60 @@
+package math_big
+
+/*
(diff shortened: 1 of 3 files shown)


Kelimion:
After several years of `math/big` it's tested well enough that it could be used as an oracle for later self-tests, but when I wrote it it made sense to test against Python. :-)

PR #3721, opened 2024-06-10
Add compilation-related constants

### LINKED ISSUE #2240: Add more `ODIN_*` global constants.
When benchmarking the influence of compiler versions on a piece of code, it would be nice to print the compiler version (and options) it was compiled with.

Presently this is limited to:
```odin
fmt.printf("Compiled with Odin %v.\n\t- Asserts disabled: %v\n\n", ODIN_VERSION, ODIN_DISABLE_ASSERT)
```
```
Compiled with Odin dev-2022-12.
    - Asserts disabled: false
```

To be added:
- `ODIN_VERSION_HASH` (either git hash or `-dev` if unknown)
- `ODIN_OPTIMIZATION_MODE` (`speed`, `size`, `minimal`, from `-o:`)
- `ODIN_DISABLE_BOUNDS_CHECK` (`false` by default, `true` with `-no-bounds-check`)
- `ODIN_MICROARCH_STRING` (e.g. `native`, from `-microarch:`)

Feoramund:
`ODIN_VERSION_HASH` is the `git` SHA hash of the commit the Odin compiler was built with.

`ODIN_MICROARCH_STRING` is the string passed to `-microarch` when the program was built.

`ODIN_OPTIMIZATION_MODE` is an enum value of which optimization mode was used to build the program.

Resolves #2240

(`ODIN_NO_BOUNDS_CHECK` already exists at this time, compared to the then-proposed `ODIN_DISABLE_BOUNDS_CHECK`.)

### DIFF
--- a/base/runtime/core.odin
+++ b/base/runtime/core.odin
@@ -560,6 +560,19 @@ Odin_Platform_Subtarget_Type :: type_of(ODIN_PLATFORM_SUBTARGET)
 */
 Odin_Sanitizer_Flags :: type_of(ODIN_SANITIZER_FLAGS)
 
+/*
+	// Defined internally by the compiler
+	Odin_Optimization_Mode :: enum int {
+		None       = -1,
+		Minimal    =  0,
+		Size       =  1,
+		Speed      =  2,
+		Aggressive =  3,
+	}
+
+	ODIN_OPTIMIZATION_MODE // is a constant
+*/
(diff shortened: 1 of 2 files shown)


Kelimion:
The microarch string should probably use `get_final_microarchitecture` so we know what it is other than `native`?

Feoramund:
I can fix that.

PR #3722, opened 2024-06-10
Fix `core:odin/parser` not parsing `label: #reverse`

### LINKED ISSUE #3715: `core:odin/parser` package doesn't support putting labels on `#reverse for in` loops
Doing `label: #reverse for item in list` will cause errors when using `core:odin/parser`.

Similar issue was raised for the compiler here: https://github.com/odin-lang/Odin/issues/3319
But only the compiler got fixed, not the core package.

CODE

```odin
list := []string{"a"}
label: #reverse for a in list {
    fmt.println(a)
}

pkg, ok := parser.collect_package(".")
if !ok do panic("Failed to collect package\n")

ok = parser.parse_package(pkg)
if !ok do panic("Failed to parse package\n")
```

OUTPUT

```
$ odin run .
a
/odin-bug/main.odin(9:13): expected ';', got for
/odin-bug/main.odin(13:16): expected 2 expressions on the right-hand side, got 1
/odin-bug/main.odin(18:1): expected a statement, got }
```

REPORT

```
	Odin:    dev-2024-06:7c529e990
	OS:      Ubuntu 22.04.4 LTS, Linux 6.5.0-35-generic
	RAM:     7631 MiB
	Backend: LLVM 17.0.6
```

Feoramund:
I had a look at the AST with `ast.inspect` using print statements to make sure the loop is parsed as a `reverse` loop on the right line and column, tried something like `label: #force_inline` to make sure it still erred out, and I think it's working properly - as in no unexpected errors, but I haven't a good way to visualize and verify that the AST is in a precisely correct construction, just as an advisory.

Fixes #3715

**EDIT**: Well, that's unfortunate. I didn't run tests locally. I suppose it was too good to be true to be so easy to fix.

### DIFF
--- a/core/odin/parser/parser.odin
+++ b/core/odin/parser/parser.odin
@@ -3514,9 +3514,21 @@ parse_simple_stmt :: proc(p: ^Parser, flags: Stmt_Allow_Flags) -> ^ast.Stmt {
 		}
 	case op.kind == .Colon:
 		expect_token_after(p, .Colon, "identifier list")
-		if .Label in flags && len(lhs) == 1 {
+		colon_label_block: if .Label in flags && len(lhs) == 1 {
 			#partial switch p.curr_tok.kind {
-			case .Open_Brace, .If, .For, .Switch:
+			case .Open_Brace, .If, .For, .Switch, .Hash:
+				if p.curr_tok.kind == .Hash {
+					name := peek_token(p)
+					switch name.text {
+					case "reverse":
+						expect_token(p, .Hash)
+						expect_token(p, .Ident)
+						break
+					case:
+						break colon_label_block
+					}
+				}
+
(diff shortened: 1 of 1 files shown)


Kelimion:
How coincidental. @gingerBill mentioned wanting to fix that on the Discord last night, to which you're most welcome, of course.

image

Kelimion:
>  but I haven't a good way to visualize and verify that the AST is in a precisely correct construction, just as an advisory.

I'll leave this one for Bill. I'm not familiar enough with the AST to be sure either.

Feoramund:
All tests are passing now. At the very least, it's parsing the demo.

gingerBill:
I have improved this in both the C++ side and Odin side in 1945218f6

PR #3725, opened 2024-06-10
add forced shutdown to new test runner

laytan:
Currently, a Ctrl+c starts a graceful shutdown of the tests and runner. Sometimes tests get stuck and this would never complete.

This simply adds an extra step, if Ctrl+c is given for the second time, just `os.exit` right away.


Feoramund:
I suspect the stalled thread is not being terminated properly, and the runner is joining with it. More rarely, it may have to do with the join issue I've mentioned before, depending on the platform in question.

The test runner is written to immediately terminate all threads upon that signal, then join with them. This will not work if thread termination is not implemented on a particular platform, like Darwin for example.

PR #3732, opened 2024-06-11
add missing functions to core:sys/windows

FrancisTheCat:
Also made SUCEEDED contextless


core/sys/windows/user32.odin
```diff
@@ -17,6 +17,17 @@ foreign user32 {
 
 	GetClassNameW :: proc(hWnd: HWND, lpClassName: LPWSTR, nMaxCount: c_int) -> c_int ---
 
+	GetParent :: proc(hWnd: HWND) -> HWND ---
+	IsWindowVisible :: proc(hWnd: HWND) -> BOOL ---
+	SetWinEventHook :: proc(
+		eventMin, eventMax: DWORD,
+		hmodWinEventProc: HMODULE,
+		pfnWinEvenProc: WINEVENTPROC,
+		idProcess, idThread, dwmFlags: DWORD,
```
Kelimion, line 26: I'd add the enum for `dwmFlags`, or even make it a `bit_set`.
```cpp
        WINEVENT_INCONTEXT = 4 // 2 in bit_set
        WINEVENT_OUTOFCONTEXT = 0  // not present in a bit_set
        WINEVENT_SKIPOWNPROCESS = 2 // 1 in bit_set
        WINEVENT_SKIPOWNTHREAD = 1 // 0 in bit_set
```
Keep in mind that for the `bit_set`, passing `WINEVENT_OUTOFCONTEXT` would mean an empty bit set `{}`

Kelimion: On a related note that's not your fault, I find it a bit sloppy that [Microsoft](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwineventhook) uses `DWORD` and `UINT` interchangeably, using `DWORD` in the signature, but then mixing `UINT` and `DWORD` under `Parameters`. It's ultimately the same type, so whatever, but sloppy documentation work.

core/sys/windows/kernel32.odin
```diff
@@ -1210,6 +1213,15 @@ SYSTEM_LOGICAL_PROCESSOR_INFORMATION :: struct {
 	DummyUnion: DUMMYUNIONNAME_u,
 }
 
+SYSTEM_POWER_STATUS :: struct {
+	ACLineStatus:        BYTE,
```
Kelimion, line 1217: Maybe replace these `BYTE`s with `enum`s or `bit_set`s as appropriate, e.g.
```odin
AC_Line_Status :: enum BYTE {
   Offline = 0,
   Online  = 1,
   Unknown = 255,
}

// enums used for a bit set are log2 of the usual value
Battery_Flag :: enum BYTE {
    High     = 0,
    Low      = 1,
    Critical = 2,
    Charging = 3,
    No_Battery = 7, 
}
Battery_Flags :: bit_set[Battery_Flag; BYTE]
```

core/sys/windows/kernel32.odin
```diff
@@ -1210,6 +1213,15 @@ SYSTEM_LOGICAL_PROCESSOR_INFORMATION :: struct {
...
+	BatteryFlag:         BYTE,
+	BatteryLifePercent:  BYTE,
+	SystemStatusFlag:    BYTE,
+	BatteryLifeTime:     DWORD,
+	BatteryFullLifeTime: DWORD,
```
Kelimion, line 1222: `The number of seconds of battery life remaining, or –1 if remaining seconds are unknown or if the device is connected to AC power.`

Strange that they're they're using an unsigned type (`DWORD`), but insist that `-1` is used for `unknown`. Does that imply it's actually signed, or is it only `max(DWORD)` that has a special meaning?

Again not your fault. I just wish that the documentation was clearer.

FrancisTheCat: Thanks for the suggestions, it should be quite a bit nicer to use now. Btw I didnt add an enum/bitset for the system status flag, because I dont really think it really helps since there are only two possible values 0/1.

core/sys/windows/user32.odin
```diff
@@ -568,3 +580,12 @@ RedrawWindowFlags :: enum UINT {
 	RDW_FRAME           = 0x0400,
 	RDW_NOFRAME         = 0x0800,
 }
+
+WinEventFlags :: bit_set[WinEventFlag; DWORD]
+
+WinEventFlag :: enum DWORD {
+    WINEVENT_OUTOFCONTEXT   = 0,
```
Kelimion, line 587: I should've been clearer that bit sets use 1 << flag, where `0` becomes `1`, and so forth. That's why I suggested that `WINEVENT_OUTOFCONTEXT` would not be present if used in a `bit_set`. The equivalent to `WINEVENT_OUTOFCONTEXT` as the zero value would be `{}`.

`WINEVENT_OUTOFCONTEXT | WINEVENT_SKIPOWNTHREAD` would become `{.SKIPOWNTHREAD}`, with out of context implied.

As such the enum should be:
```cpp
// OUTOFCONTEXT is the zero value, use {}
WinEventFlag :: enum DWORD {
    SKIPOWNTHREAD  = 0,
    SKIPOWNPROCESS = 1,
    INCONTEXT      = 2,
}
```

The Battery status `bit_set` is correct. I'd already made the adjustment in the example code I suggested.

FrancisTheCat: oh that makes sense, I overlooked that

Kelimion (approved):

PR #3741, opened 2024-06-12
Added missing constants and enums in rlgl.odin

karl-zylinski:
Many of these are identical to raylib.odin. But that's how rlgl.h works, it redefines many of the same things to make sure that it is independent from raylib.


karl-zylinski:
Scrapping this. I'll make a PR where I put in an explicit raylib dependency and pull in enums instead. It's nice if rlgl is _truly_ independent, but since we don't have a separate `rlgl.lib` etc it won't be independent from the other files anyways.

PR #3752, opened 2024-06-13
corrected the wrong value of CREATE_FACTORY_FLAG.DEBUG

chromedays:
CREATE_FACTORY_FLAG.DEBUG should be 1 as noted in the original docs:
https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_3/nf-dxgi1_3-createdxgifactory2
```
Valid values include the DXGI_CREATE_FACTORY_DEBUG (0x01) flag, and zero.
```

I also removed the integer constant that is redundant and incompatible to the CREATE_FACTORY bitset


laytan:
An enum used by a bit set will not have the same values as they would be in C. In Odin each enum value is basically the bit index of the flags, aka the log2 of the value in C.
So in this case `log2(0x01) = 0` meaning the current value of 0 should be correct.

chromedays:
Ohh... didn't notice that. Thank you for clarifying.

PR #3772, opened 2024-06-16
Fix `bit_set` construction crash on constant non-integer field

Feoramund:
Found this when I switched a `struct` to be a `bit_set`. The following program fails an assertion:
```odin
package main

main :: proc() {
	E :: enum {
		Alpha,
		Beta,
	}

	BS :: bit_set[E]

	f := BS{ true }
}
```

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -9819,7 +9819,9 @@ gb_internal ExprKind check_compound_literal(CheckerContext *c, Operand *o, Ast *
 				if (tav.mode != Addressing_Constant) {
 					continue;
 				}
-				GB_ASSERT(tav.value.kind == ExactValue_Integer);
+				if (tav.value.kind != ExactValue_Integer) {
+					continue;
+				}
 				i64 v = big_int_to_i64(&tav.value.value_integer);
 				i64 lower = bt->BitSet.lower;
 				u64 index = cast(u64)(v-lower);


oskarnp:
I don't think this is supposed to compile.

Kelimion:
> I don't think this is supposed to compile.

I don't think so either, but neither should it assert. An error message that you can't use `bool` in an `int` bit set would be appropriate here.

Feoramund:
> I don't think this is supposed to compile.

The PR does not make it compile.

> An error message that you can't use `bool` in an `int` bit set would be appropriate here.
```
/tmp/odin/crash/crash.odin(11:11) Error: Cannot convert untyped value 'true' to 'E' from 'untyped bool'
	f := BS{ true }
	         ^~~^
```
Something more specific than what is already being displayed?

oskarnp:
> > I don't think this is supposed to compile.
> 
> The PR does not make it compile.

My bad! Sorry.

> 
> > An error message that you can't use `bool` in an `int` bit set would be appropriate here.
> 
> ```
> /tmp/odin/crash/crash.odin(11:11) Error: Cannot convert untyped value 'true' to 'E' from 'untyped bool'
> 	f := BS{ true }
> 	         ^~~^
> ```
> 
> Something more specific than what is already being displayed?

This seems fine then.

Kelimion:
> > I don't think this is supposed to compile.
> 
> The PR does not make it compile.
> 
> > An error message that you can't use `bool` in an `int` bit set would be appropriate here.
> 
> ```
> /tmp/odin/crash/crash.odin(11:11) Error: Cannot convert untyped value 'true' to 'E' from 'untyped bool'
> 	f := BS{ true }
> 	         ^~~^
> ```
> 
> Something more specific than what is already being displayed?

No, this is great. Thanks.

PR #3776, opened 2024-06-18
fix(os2): check for 0 bytes read and return EOF

IllusionMan1212:
Current os2 `file_stream` implementation for `read` on linux never returns an EOF error when reaching the end of file.

Linux's read syscall returns 0 read bytes if the end of file is reached: https://man7.org/linux/man-pages/man2/read.2.html

This also matches the implementation of the `os` package as evident by [`core/os/stream.odin:30`](https://github.com/odin-lang/Odin/blob/f280ba8511b97bb1d2f5173933a6179c2aba231b/core/os/stream.odin#L30)

### DIFF
--- a/core/os/os2/file_linux.odin
+++ b/core/os/os2/file_linux.odin
@@ -121,6 +121,9 @@ _read :: proc(f: ^File, p: []byte) -> (i64, Error) {
 	if n < 0 {
 		return -1, _get_platform_error(n)
 	}
+	if n == 0 {
+		return 0, .EOF
+	}
 	return i64(n), nil
 }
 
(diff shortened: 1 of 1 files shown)


gingerBill:
This doesn't handle the case where `len(p) == 0`.

The question is, how should that be handled?

IllusionMan1212:
> This doesn't handle the case where `len(p) == 0`.
> 
> The question is, how should that be handled?

It seems like the os package doesn't handle `len(p) == 0` either.

Maybe `len(p) == 0` should return `io.Error.Short_Buffer`. What do you think?

gingerBill:
Ignore `package os`, it's awful.

But is that a short buffer or just an empty one? It's not necessarily an error.

laytan:
IMO that should just `return 0, nil`

Kelimion:
Reading into an empty buffer should always succeed, imo.

IllusionMan1212:
Personally I think an `.Empty_Buffer` error would be better since you're not getting anything useful out of reading into an empty buffer. But if everyone thinks an empty read should succeed then I'm ok with that too.

I could update the condition to be `if len(p) != 0 && n == 0 && ferr == nil` to distinguish EOF and empty reads.

gingerBill:
It should be `return 0, nil` on an empty buffer.

IllusionMan1212:
Changed the check for EOF to be in the `_read` and `_read_at` functions. This should be good to go now.

PR #3779, opened 2024-06-18
Clarify when the arena in base:runtime should be used

karl-zylinski:
This PR clarifies that the arena in `base:runtime` shouldn't be used for anything but the default temp allocator. This is done by:
- Renaming the file in which it lives: `default_allocators_arena.odin => default_temp_allocator_arena.odin`
- Improve the comment above `Arena :: struct {`.

This should avoid some confusion where people end up using the Arena in`base:runtime` because they thought it was the 'default arena', as the old file name `default_allocators_arena.odin` might have lured them into believing.

### DIFF
--- a/base/runtime/default_temp_allocator_arena.odin
+++ b/base/runtime/default_temp_allocator_arena.odin
@@ -12,7 +12,8 @@ Memory_Block :: struct {
 	capacity:  uint,
 }
 
-// NOTE: This is for internal use, prefer `Arena` from `core:mem/virtual` if necessary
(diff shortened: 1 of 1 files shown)


karl-zylinski:
I guess if one wanted to take this further and make it even clearer, then we could also rename `Arena` in `base:runtime` to `Temp_Arena` etc. I can do that if it's seen as a good idea.

That would lessen the "jump to symbol" confusion with there being 3 different `Arena` in `base` and `core` a bit smaller.

PR #3780, opened 2024-06-18
Loosen constraint on the Key type in rb & avl tree

destroycomputers:
The current constraint on the `init_ordered` function accepts only numeric types. While one still can init a non-numerically-keyed tree by passing a comparator explicitly, the reason to disallow non-numeric ordered types is nowhere to be found.


laytan:
Note that the where clause will now allow pointer types through, which are not comparable with less/greater than. It is probably fine though, it will still error somewhere else.

PR #3781, opened 2024-06-18
Add missing imports to `core` test suite

Feoramund:
I forgot about this.

### DIFF
--- a/tests/core/flags/test_core_flags.odin
+++ b/tests/core/flags/test_core_flags.odin
@@ -1113,6 +1113,13 @@ test_if_map_cstrings_get_freed :: proc(t: ^testing.T) {
 
 @(test)
 test_os_handle :: proc(t: ^testing.T) {
+	defer if !testing.failed(t) {
+		// Delete the file now that we're done.
+		//
+		// This is not done all the time, just in case the file is useful to debugging.
+		testing.expect_value(t, os.remove(TEMPORARY_FILENAME), os.ERROR_NONE)
+	}
+
 	TEMPORARY_FILENAME :: "test_core_flags_write_test_output_data"
 
 	test_data := "Hellope!"
@@ -1147,14 +1154,6 @@ test_os_handle :: proc(t: ^testing.T) {
 	testing.expect_value(t, read_ok, true)
 	file_contents_equal := 0 == bytes.compare(transmute([]u8)test_data, data)
 	testing.expectf(t, file_contents_equal, "expected file contents to be the same, got %v", data)
-
-	if file_contents_equal {
-		// Delete the file now that we're done.
-		//
-		// This is not done as a defer or all the time, just in case the file
-		// is useful to debugging.
-		testing.expect_value(t, os.remove(TEMPORARY_FILENAME), os.ERROR_NONE)
-	}
 }
 
 @(test)
--- a/tests/core/normal.odin
+++ b/tests/core/normal.odin
@@ -19,6 +19,7 @@ download_assets :: proc() {
 @(require) import "encoding/json"
 @(require) import "encoding/varint"
 @(require) import "encoding/xml"
+@(require) import "flags"
 @(require) import "fmt"
 @(require) import "math"
 @(require) import "math/big"
@@ -37,3 +38,4 @@ download_assets :: proc() {
 @(require) import "text/match"
 @(require) import "thread"
 @(require) import "time"
+@(require) import "unicode"


Feoramund:
The `core:flags` Windows test failed for error code 32.

```
ERROR_SHARING_VIOLATION

32 (0x20)

The process cannot access the file because it is being used by another process.
```

It failed on the cleanup part of checking if `os.Handle` is parsed correctly where it deletes the temporary file. I'm wondering if I should strike that out, since I'm not sure in what way this error is arising, and it's not highly necessary to remove it since it's only a few bytes.

laytan:
I am pretty sure that error arises there because you have not closed your file handles before trying to delete the file, the defers run after your attempted deletion.

Feoramund:
Oh, that makes perfect sense. I can just defer the removal too. Why this is allowed on every other platform is beyond me.

tests/core/flags/test_core_flags.odin
```diff
@@ -1113,6 +1113,13 @@ test_if_map_cstrings_get_freed :: proc(t: ^testing.T) {
 
 @(test)
 test_os_handle :: proc(t: ^testing.T) {
+	defer if !testing.failed(t) {
```
laytan, line 1116: Should be
```odin
defer {
    if !testing.failed(t) {
        ...
    }
}
```
right?

Currently it will check if the test failed at the beginning of the test, which is always `false`.

Feoramund: I made a discussion for this since it seems like something that could be confusing. #3782

laytan: Oh strange I could've sworn it was a thing used to defer something IF the condition is currently true.
Because
```odin
if true {
    defer blah()
}
```
Would execute blah at the end of that if block, not at the end of the wanted scope.

I really thought `defer if` was created to get around that, @gingerBill ?

Feoramund: I did a quick peek through `git log` for `defer if` and all I could find was 118ab60588832904e6a980c712702f71b4f7ec15 by @Kelimion.

laytan: I think I'm just mistaken and I have to look over some of my codebases for bugs :laughing:

Kelimion: Seems pretty solid evidence that `defer if ...` is just shorthand for `defer { if ... }`, because it's used prodigiously for clean-up only iff errors are (not) present, as appropriate.

image

Kelimion: And really that's no different than the difference between `if cond do single_statement()` and `if cond { single_statement() }`. Really a way to avoid opening two blocks on one line, i.e. `defer { if cond { do_stuff() } }`.

Feoramund: Of note, `defer` works with blocks optionally in general. I.e. `defer a = true` works just fine, same as the common and idiomatic `defer delete(x)`. So I believe the `defer if` distinction might be human language mis-parsing on our parts.

Kelimion: Right, if you want a conditional defer: `if defer`
Deferred conditional: `defer if`

laytan: > Right, if you want a conditional defer: `if defer` Deferred conditional: `defer if`

`if defer` won't do what you want though, the if creates a new scope that the defer is ran after.

Kelimion: > `if defer` won't do what you want though, the if creates a new scope that the defer is ran after.

Then `if defer` might be a thing the compiler should error on as being a pointless and confusing construct.

Kelimion: But mainly I wrote it out that way because just by reading left to right we can rephrase the Odin code into English that highlights we want `defer if` (deferred conditional), not `if defer` (conditional defer) if we're unsure which of the two we should write.

That there's the confusion to begin with and that the 2nd phrasing is nonsensical in practice suggests it should be made an error.

Feoramund: I can't get `if defer a { /* ... */ }` to compile, if this was meant as a concrete and not a hypothetical syntax example.

Kelimion: Confirmed. Well, nothing to worry about then.

laytan: I thought we all meant `if x do defer ...` here

Kelimion: Yes, I tried `if a defer` and it doesn't compile.

PR #3785, opened 2024-06-19
Separate I18N calls for immutable and for pluraliseable strings. Fixes #3687

### LINKED ISSUE #3687: core:text/i18n: separate get() calls requiring a quantity parameter from calls not requiring it (proposal)
### Discussed in https://github.com/odin-lang/Odin/discussions/3501

This is not about a bug, but the outcome of the discussion quoted at the top (and also on Discord), advancing a proposal.

In summary, the idea is that the message catalogues used by `core:text/i18n` (primarily for `.mo`-based catalogues, but QtLinguist-based ones would also benefit from it.) distinguish — as two different forms of entries — between translating strings which never change (regardless of the quantity of item involved, for instance: "All temporary files have been deleted") and which do not require a quantity numeric parameter and of strings which may vary according to this quantity and thus require such a parameter.

In the current API, both the quantity and the source catalogue are optional and the compiler is obviously perfectly able to distinguish between , for instance:
    `i18n.get("Some key", number)`
and
    `i18n.get("some key", catalogue)`

A tool for collecting translatable strings from Odin sources — as GNU `xgettext` or Qt Linguist do for C/C++ sources — which needs to output two different types of entries from them, would have much more troubles distinguishing that the first is a pluraliseable string (with a numeric parameter) and the second is immutable (the second parameter is a pointer to a catalogue), as the surface syntax is parallel (a literal string and an identifier).

So, in part to ease the job of such tools (without which the current i18n API is largely unusable) and in part to reflect the nature of underlying things, the proposal is to split the current i18n API for translations in two:

- a `get_n()` proc group which, in the current procs `get_single_section` and `get_by_section`, renamed to `get_single_section_w_plural` and `get_by_section_w_plural`, the numeric parameter is no longer optional, but required;
- a `get()` proc group which, in the current procs `get_single_section` and  `get_by_section`, which keep their names, the numeric parameter is absent.

In summary of summary (!):
```
    get_single_section :: proc(key: string, catalog: ^Translation = ACTIVE) -> (value: string) {...}
    get_by_section :: proc(section, key: string, catalog: ^Translation = ACTIVE) -> (value: string) {...}
    get :: proc{get_single_section, get_by_section}

    get_single_section_w_plural :: proc(key: string, quantity: int, catalog: ^Translation = ACTIVE) -> (value: string) {...}
    get_by_section_w_plural :: proc(section, key: string, quantity: int, catalog: ^Translation = ACTIVE) -> (value: string) {...}
    get_n :: proc{get_single_section_w_plural, get_by_section_w_plural}
```
I have a PR ready to be submitted implementing the proposed changes in `core/text/i18n/doc.odin`, `core/text/i18n/i18n.odin` and `tests/core/text/i18n/test_core_text_i18n.odin`.

Comments and/or objections are welcome!

mgavioli:
Fixes #3687

This PR:

- refactors `get_single_section()` and `get_by_section()` — grouped into `get()` group — by removing the `number` parameter (previously optional)
- adds `get_single_section_w_plural()` and `get_by_section_w_plural()` — grouped into `get_n()` — by adding a mandatory `quantity: int` parameter (analogous to the previous `number` parameter but no longer optional).

Tests in `tests/core/text/i18n/test_core_text_i18n.odin` have also been updated.

The rationale of this change has been analysed in the #3687 issue and previously in the #3501 discussion.

### DIFF
--- a/core/text/i18n/doc.odin
+++ b/core/text/i18n/doc.odin
@@ -1,31 +1,44 @@
 
 /*
-The `i18n` package is flexible and easy to use.
+The `i18n` package is a flexible and easy to use way to localise applications.
 
-It has one call to get a translation: `get`, which the user can alias into something like `T`.
+It has two calls to get a translation: `get()` amd `get_n()`, which the user can alias into something like `T` and `Tn`
+with statements like:
+	T :: i18n.get
+	Tn :: i18n.get_n.
 
-`get`, referred to as `T` here, has a few different signatures.
-All of them will return the key if the entry can't be found in the active translation catalog.
+`get()` is used for retrieving the translation of sentences which **never**change in form,
+like for instance "Connection established" or "All temporary files have been deleted".
+Note that the number (singular, dual, plural, whatever else) is not relevant: the semtece is fixed and it will have only one possible translation in any other language.
 
-- `T(key)`              returns the translation of `key`.
-- `T(key, n)`           returns a pluralized translation of `key` according to value `n`.
+`get_n()` is used for retrieving the translations of sentences which change according to the number of items referenced.
+The various signatures of `get_n()` have one more parameter, `n`, which will receive that number,
+and which be used to select the correct form according to the pluraliser attached to the message catalogue when initially loaded;
+for instance, to summarise a rather complex matter, some languages use the singular form when reerring to 0 items and some use the (only in their case) plural forms;
+also, languages may have more or less quantifier forms than a single singular form an a universal plural form:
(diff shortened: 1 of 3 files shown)


Kelimion:
Looks good in principle, but you've inadvertently introduced a *lot* of typos. I'll do a pass over it.

mgavioli:
Thanks for merging.

I am sorry for the typos: there was really a lot of them!

VSCode, which I use to work in Odin, does not seem to have a spell checker (and English is not my native language...); I hope it wasn't too long and too boring! In the future, I'll run a spell checking step...

PR #3788, opened 2024-06-19
Add to_cstring for Builders

Beefster09:
When using external libs that use cstrings a lot, it can be very helpful to be able to output cstrings from a builder.

It's not straightforward to think about the null terminator manually and then feed it through a series of non-allocating casts.

### DIFF
--- a/core/strings/builder.odin
+++ b/core/strings/builder.odin
@@ -286,6 +286,33 @@ to_string :: proc(b: Builder) -> (res: string) {
 	return string(b.buf[:])
 }
 /*
+Appends a trailing null byte if missing and then casts the Builder byte buffer to cstring
+
+Inputs:
+- b: A pointer to builder
+
+Returns:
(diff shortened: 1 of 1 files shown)


core/strings/builder.odin
```diff
@@ -286,6 +286,20 @@ to_string :: proc(b: Builder) -> (res: string) {
...
+
+Inputs:
+- b: A pointer to builder
+
+Returns:
+- res: A cstring of the Builder's buffer
+*/
+to_cstring :: proc(b: ^Builder) -> (res: cstring) {
+	append(&b.buf, 0)
+	defer pop(&p.buf)
```
gingerBill, line 299: I wouldn't even do a `defer` here.

Beefster09: Yeah, I guess you're right. It'll still be there even after the pop.

PR #3791, opened 2024-06-20
Refactor `core:text/table`

### LINKED ISSUE #3432: text/table uses bytes instead of runes for length

Feoramund:
_Several kilobytes of lookup tables and one grapheme segmentation algorithm later..._

unicode-table-alignment

Resolves #3432

### DIFF
--- a/core/text/table/table.odin
+++ b/core/text/table/table.odin
@@ -324,9 +324,8 @@ write_markdown_table :: proc(w: io.Writer, tbl: ^Table) {
 		build(tbl, WIDTH_PROC)
 	}
 
-	for row in 0..<tbl.nr_rows {
-		alignment: Cell_Alignment = .Center if tbl.has_header_row && row == header_row(tbl) else .Left
-		for col in 0..<tbl.nr_cols {
+	write_row :: proc(w: io.Writer, tbl: ^Table, row: int, alignment: Cell_Alignment = .Left) {
+		for col in 0 ..< tbl.nr_cols {
 			cell := get_cell(tbl, row, col)
 			if col == 0 {
 				io.write_byte(w, '|')
@@ -335,30 +334,42 @@ write_markdown_table :: proc(w: io.Writer, tbl: ^Table) {
 			io.write_string(w, "|")
 		}
 		io.write_byte(w, '\n')
+	}
 
-		if tbl.has_header_row && row == header_row(tbl) {
-			for col in 0..<tbl.nr_cols {
-				cell := get_cell(tbl, row, col)
-				if col == 0 {
-					io.write_byte(w, '|')
-				}
-				divider_width := tbl.colw[col] + tbl.lpad + tbl.rpad - 1
-				switch cell.alignment {
-				case .Left:
-					io.write_byte(w, ':')
-					write_byte_repeat(w, max(1, divider_width), '-')
-				case .Center:
-					io.write_byte(w, ':')
-					write_byte_repeat(w, max(1, divider_width-1), '-')
-					io.write_byte(w, ':')
-				case .Right:
-					write_byte_repeat(w, max(1, divider_width), '-')
-					io.write_byte(w, ':')
-				}
+	start := 0
+
+	if tbl.has_header_row {
+		row := header_row(tbl)
+
+		write_row(w, tbl, row, .Center)
+
+		for col in 0 ..< tbl.nr_cols {
+			cell := get_cell(tbl, row, col)
+			if col == 0 {
 				io.write_byte(w, '|')
 			}
-			io.write_byte(w, '\n')
+			divider_width := tbl.colw[col] + tbl.lpad + tbl.rpad - 1
+			switch cell.alignment {
+			case .Left:
+				io.write_byte(w, ':')
+				write_byte_repeat(w, max(1, divider_width), '-')
+			case .Center:
+				io.write_byte(w, ':')
+				write_byte_repeat(w, max(1, divider_width - 1), '-')
+				io.write_byte(w, ':')
+			case .Right:
+				write_byte_repeat(w, max(1, divider_width), '-')
+				io.write_byte(w, ':')
+			}
+			io.write_byte(w, '|')
 		}
+		io.write_byte(w, '\n')
+
+		start += row + 1
+	}
+
+	for row in start ..< tbl.nr_rows {
+		write_row(w, tbl, row)
 	}
 }
 
--- a/core/text/table/table.odin
+++ b/core/text/table/table.odin
@@ -4,6 +4,7 @@
 
 	List of contributors:
 		oskarnp: Initial implementation.
+		Feoramund: Unicode support.
 */
 
 package text_table
@@ -12,10 +13,12 @@ import "core:io"
 import "core:fmt"
 import "core:mem"
 import "core:mem/virtual"
+import "core:unicode/utf8"
 import "base:runtime"
 
 Cell :: struct {
 	text: string,
+	width: int,
 	alignment: Cell_Alignment,
 }
 
@@ -24,6 +27,9 @@ Cell_Alignment :: enum {
 	Center,
 	Right,
 }
+
+// Determines the width of a string used in the table for alignment purposes.
+Width_Proc :: #type proc(str: string) -> int
 
 Table :: struct {
 	lpad, rpad: int, // Cell padding (left/right)
@@ -37,8 +43,17 @@ Table :: struct {
 	dirty: bool, // True if build() needs to be called before rendering
 
 	// The following are computed on build()
-	colw: [dynamic]int, // Width of each column (including padding, excluding borders)
+	colw: [dynamic]int, // Width of each column (excluding padding and borders)
 	tblw: int,          // Width of entire table (including padding, excluding borders)
+}
+
+ascii_width_proc :: proc(str: string) -> int {
+	return len(str)
+}
+
+unicode_width_proc :: proc(str: string) -> (width: int) {
+	_, _, width = #force_inline utf8.grapheme_count(str)
+	return
 }
 
 init :: proc{init_with_allocator, init_with_virtual_arena, init_with_mem_arena}
@@ -165,7 +180,7 @@ first_row :: proc(tbl: ^Table) -> int {
 	return header_row(tbl)+1 if tbl.has_header_row else 0
 }
 
-build :: proc(tbl: ^Table) {
+build :: proc(tbl: ^Table, width_proc: Width_Proc) {
 	tbl.dirty = false
 
 	resize(&tbl.colw, tbl.nr_cols)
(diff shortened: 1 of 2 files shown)


oskarnp:
This is awesome. However, I am just not a fan of breaking the API unless it is neccessary?

oskarnp:
I'll review next week.

Feoramund:
I changed `colw` to make calculations easier, `set_cell_value_and_alignment` because it was inconsistent with `set_cell_value` and only took a string (not an `any`), `build` to be the place where `Width_Proc` is specified (though this could have a default value if needed), renamed the proc from `ascii` to `plain` per the discussion in the original issue, and `write_text_align` to take a `space` argument instead of `colw`.

The arguments were rotated on `write_text_align` to purposefully break the API because the semantics of the arguments changed but the types did not.

core/text/table/table.odin
```diff
@@ -37,10 +43,19 @@ Table :: struct {
...
 
 	// The following are computed on build()
-	colw: [dynamic]int, // Width of each column (including padding, excluding borders)
+	colw: [dynamic]int, // Width of each column (excluding padding and borders)
 	tblw: int,          // Width of entire table (including padding, excluding borders)
 }
 
+ascii_width_proc :: proc(str: string) -> int {
+	return len(str)
+}
```
oskarnp, line 58: Is this really needed, considering ascii is a subset of utf-8?

core/text/table/table.odin
```diff
@@ -165,7 +180,7 @@ first_row :: proc(tbl: ^Table) -> int {
 	return header_row(tbl)+1 if tbl.has_header_row else 0
 }
 
-build :: proc(tbl: ^Table) {
+build :: proc(tbl: ^Table, width_proc: Width_Proc) {
```
oskarnp, line 290: I think I would prefer a default Width_Proc here. Default to `unicode_width_proc`.

But considering previous comment about ascii being a subset, is Width_Proc needed at all? Any use case for custom width procs you can think of?

core/text/table/table.odin
```diff
@@ -251,8 +265,10 @@ write_html_table :: proc(w: io.Writer, tbl: ^Table) {
 }
 
 write_plain_table :: proc(w: io.Writer, tbl: ^Table) {
+	WIDTH_PROC :: unicode_width_proc
+
 	if tbl.dirty {
-		build(tbl)
+		build(tbl, WIDTH_PROC)
```
oskarnp, line 378: If we are to support custom Width_Proc's they should be stored on the `Table` struct at the very least to avoid this situation. Or just remove custom Width_Proc's. Again, is there a use case?

core/text/table/table.odin
```diff
@@ -152,22 +164,116 @@ header :: proc(tbl: ^Table, values: ..any, loc := #caller_location) {
...
 			panic("Cannot create empty row unless the number of columns is known in advance")
 		} else {
 			tbl.nr_cols = len(values)
 		}
 	}
+
 	tbl.nr_rows += 1
-	for col in 0..<tbl.nr_cols {
+
+	for col in 0 ..< tbl.nr_cols {
```
oskarnp: `a..<b` (without surrounding space) is the preferred Odin style.

core/text/table/table.odin
```diff
@@ -152,22 +164,116 @@ header :: proc(tbl: ^Table, values: ..any, loc := #caller_location) {
...
+		if len(aligned_values) == 0 {
+			panic("Cannot create empty row unless the number of columns is known in advance")
+		} else {
+			tbl.nr_cols = len(aligned_values)
+		}
 	}
+
+	tbl.nr_rows += 1
+
+	for col in 0 ..< tbl.nr_cols {
```
oskarnp, line 258: `a..<b` (without surrounding space) is the preferred Odin style.

core/text/table/table.odin
```diff
@@ -152,22 +164,116 @@ header :: proc(tbl: ^Table, values: ..any, loc := #caller_location) {
...
+		if len(values) == 0 {
+			panic("Cannot create empty row unless the number of columns is known in advance")
+		} else {
+			tbl.nr_cols = len(values)
+		}
+	}
+
+	tbl.nr_rows += 1
+
+	for col in 0 ..< tbl.nr_cols {
```
oskarnp, line 220: `a..<b` (without surrounding space) is the preferred Odin style.

core/text/table/table.odin
```diff
@@ -37,10 +49,19 @@ Table :: struct {
...
+	colw: [dynamic]int, // Width of each column (excluding padding and borders)
 	tblw: int,          // Width of entire table (including padding, excluding borders)
 }
 
+ascii_width_proc :: proc(str: string) -> int {
+	return len(str)
+}
+
+unicode_width_proc :: proc(str: string) -> (width: int) {
+	_, _, width = #force_inline utf8.grapheme_count(str)
```
oskarnp, line 61: `utf8.grapheme_count()` calls `utf8.decode_grapheme_clusters()` which allocates implicitly using `context.allocator`. Isn't this a hidden allocation and a memory leak?
All allocators used should be specified up front at `init`.

core/text/table/table.odin
```diff
@@ -137,22 +165,116 @@ header :: proc(tbl: ^Table, values: ..any, loc := #caller_location) {
...
+		if len(values) == 0 {
+			panic("Cannot create empty row unless the number of columns is known in advance")
+		} else {
+			tbl.nr_cols = len(values)
+		}
+	}
+
+	tbl.nr_rows += 1
+
+	for col in 0 ..< tbl.nr_cols {
```
oskarnp, line 239: `a..<b` (without surrounding space) is the preferred Odin style.

core/text/table/table.odin
```diff
@@ -306,45 +425,58 @@ write_ascii_table :: proc(w: io.Writer, tbl: ^Table) {
...
-					io.write_byte(w, ':')
-				}
+	start := 0
+
+	if tbl.has_header_row {
+		row := header_row(tbl)
+
+		write_row(w, tbl, row, .Center)
+
+		for col in 0 ..< tbl.nr_cols {
```
oskarnp, line 453: `a..<b` (without surrounding space) is the preferred Odin style.

I'll stop pointing these out now. Please fix them all :)

oskarnp (changes requested):

core/text/table/table.odin
Kelimion: I suppose that if you have a specific font in mind, you could parse its kerning table and glyph info to be more specific. But are you still working in plain text using this package or doing tables in a graphical context?

core/text/table/table.odin
Feoramund: Yes. `grapheme_count` has to iterate by UTF-8 codepoints then address several lookup tables by binary search. It's not fast at all. If someone knows in advance that all the text they're handling is ASCII, then it's better to use this proc rather than the Unicode version.

core/text/table/table.odin
Feoramund: I did consider if the proc should be on the table or not. I concluded that it shouldn't be on the table, because how the table is printed is separate from what data is stored on the table and HTML tables do not measure width.

core/text/table/table.odin
oskarnp: Given your comment about `grapheme_count` I change my mind here. Leave it as it is.

core/text/table/table.odin
oskarnp: Maybe we should consider `if tbl.dirty` at that point to be an error instead to avoid this. 🤔

core/text/table/table.odin
Feoramund: I find this style `a ..< b` to be more readable and tend to use it. I did a search to confirm your statement. At the very least, it's more common (~450 instances of `a..<b` versus ~50 of `a ..< b`). I can change it if it's an issue.

core/text/table/table.odin
Feoramund:
```
Inputs:
- str: The input string.
- track_graphemes: Whether or not to allocate and return `graphemes` with extra data about each grapheme.
- allocator: (default: context.allocator)

```

```
grapheme_count :: proc(str: string) -> (graphemes, runes, width: int) {
	_, graphemes, runes, width = decode_grapheme_clusters(str, false)
```

```
				if track_graphemes {
					append(&graphemes, Grapheme{
						byte_index,
						rune_count,
						width - last_width,
					})
				}
```

This interpretation is incorrect. In addition, `grapheme_count` is tested under the test suite, and if a memory leak occurs, it would show up in the warnings.

oskarnp: I stand corrected. Ignore my comment.

oskarnp (changes requested):
I think `doc.odin` needs some more love. Could you include the example from your PR? And also showcase using different Width_Proc's.

core/text/table/table.odin
Feoramund: Do you want `tbl.dirty` done away with?

oskarnp: What would you suggest?

Feoramund: It's a caching situation. The advantage of having `tbl.dirty` around is that it allows re-printing of the table without needing to recalculate column widths. This is assuming there's a situation where a table has been built and is not often changed, but is shown on more than one or two occasions. However, I think if someone wants to really save time, they're better off writing the table once and storing that as a string and avoiding all of the calculations altogether.

In summary, `dirty` avoids recalculating column widths, but in the end, you're still doing all the work to do the printing, which is the bulk of the work being done (all the calls to `io`'s API which potentially has to branch with each call).

My suggestion is to remove `dirty` and offer string caching as a way to save time, if that's a goal for the programmer.

oskarnp: If you don't mind making those changes, then yes I agree that would be a better solution and we should get rid of it.

oskarnp: The user could easily do the string caching themselves passing a custom `io.Writer` so I think we can skip that and just remove `dirty`.

Feoramund:
All set.

oskarnp:
I can't get the docs.odin to render that unicode table properly. Both GitHub and Sublime Text shows it incorrectly.
Did you paste that properly? Or are they both wrong.

Feoramund:
I have confirmed that Alacritty, kitty, WezTerm (with Thai out of alignment for some reason but otherwise fine), Konsole, QTerminal (same Thai issue), LXTerminal, and MATE terminal render it properly. I may not have good all-around Thai fonts installed, but it's otherwise fine on my end.

GitHub's display is browser-dependent, and it certainly doesn't render properly on my end, but I wouldn't expect it to.

I found a Sublime Text editor package for my distribution of Linux and installed it to have a look. With the default settings, all of the Unicode lines are out of alignment (except the actual Younger Futhark runes, amusingly enough). I'm able to get some of them to align if I set certain monospace fonts (Ubuntu Mono), but as this is the first time I've ever used Sublime Text before, I am uncertain if there may be some setting I'd have to toggle to get it to work properly.

It's very strange that it's getting CJK wrong, as they're the most easily predictable in terms of width. Even with a CJK font set, it's still not aligned. It also has issues with rendering the excessive diacritics line, as in, glyph placeholders are appearing.

I tried manually inserting spaces to see if I could align the Chinese row, and even with extra spaces, it's mis-aligned. I.e. there's no amount of space characters I can manually put in that would make the `|` line up. I'm inclined to say this is a rendering issue.

oskarnp:
@gingerBill @Kelimion LGTM

Great work, @Feoramund

Kelimion:
Thanks as always, @Feoramund.

PR #3792, opened 2024-06-21
Add `core:encoding/uuid`

Feoramund:
This is the `odin-uuid` package I wrote a while back, updated to use the new `context`-based random generator, along with some minor documentation touch-ups. It implements RFC 4122 compliant unique identifiers with generators for versions 3, 4, and 5.


core/encoding/uuid/definitions.odin
```diff
@@ -0,0 +1,59 @@
+package uuid
+
+// A RFC 4122 Universally Unique Identifier
+Identifier :: struct #raw_union {
+	integer: u128be,
+	bytes:   [16]u8,
+}
```
gingerBill, line 7: I am not sure if it's better to have it be a raw union like this or just:
```odin
Identifier :: distinct [16]u8
```

Kelimion: The only thing `.integer` appears to be used for is for the pseudo-random variant's generation, to use `rand_u128`. I think that `Identifier` should just be `distinct [16]u8`. `generate_v4` can transmute the u128 to `[16]u8`.

Also, because it's pseudorandom, enforcing that you're putting the random bits in in a big endian order instead of machine endian is a bit overkill, so `u128` -> `[16]u8` is just fine.

gingerBill: And even then, `rand.read` is a much better idea too.
```
_ = rand.read(identifier[:])
```

Kelimion:
Incidentally, I think we can relatively trivially extend the supported UIID versions to also include v1 and v6, which per [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) allow replacing the MAC address with random bytes for privacy. v6 is v1 with the time bits reversed.

So a `generate_v1 :: proc(mac: Maybe([6]u8)) -> Identifier`? On platforms that have it, the caller can then use `net.enumerate_interfaces` to get a real MAC if that interests them.

Not that the PR should be held up for those. They can be added after merger.

core/encoding/uuid/definitions.odin
```diff
@@ -0,0 +1,59 @@
...
+// Name string is a URL.
+Namespace_DNS := Identifier {
+	bytes = {
+		0x6b, 0xa7, 0xb8, 0x10, 0x9d, 0xad, 0x11, 0xd1,
+		0x80, 0xb4, 0x00, 0xc0, 0x4f, 0xd4, 0x30, 0xc8,
+	},
+}
+
+// Name string is a fully-qualified domain name.
+Namespace_URL := Identifier {
```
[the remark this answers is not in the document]
Feoramund: Yes. Looks like a mistake from my original implementation. I can fix that.

core/encoding/uuid/definitions.odin
Feoramund: I can address both concerns.

Feoramund:
One upside of making the `Identifer` be only a distinct type is that I was able to make the namespaces `@(rodata)` now, which is nice.

I'll have to give this RFC a read. I wasn't aware of it before.

Kelimion:
> One upside of making the `Identifer` be only a distinct type is that I was able to make the namespaces `@(rodata)` now, which is nice.

Indeed!
 
> I'll have to give this RFC a read. I wasn't aware of it before.

I'm not surprised. It was published only last month.

Kelimion (approved):

Kelimion:
Merge it, or do you want to tinker with it some more in light of the RFC update?

Feoramund:
> Merge it, or do you want to tinker with it some more in light of the RFC update?

I'm working on v6 at the moment, then I'll add another test or two and clean up. Might get it all wrapped up in the next few hours, depending. I may implement v8 just as a version/variant stamp on top of a user-provided u128/array slice for good measure.

If you could check over the v1 proc for me, that'd be great. The endianness had me confused for a bit, because I know they're supposed to be stored in big-endian, but the time fields are flipped and also separated, so I just resorted to doing it byte-by-byte for sanity's sake.

Kelimion:
> If you could check over the v1 proc for me, that'd be great. The endianness had me confused for a bit, because I know they're supposed to be stored in big-endian, but the time fields are flipped and also separated, so I just resorted to doing it byte-by-byte for sanity's sake.

Sure thing. I'll give it a thorough look tomorrow.

core/encoding/uuid/generation.odin
```diff
@@ -0,0 +1,288 @@
...
+/*
+Generate a version 4 UUID.
+
+This UUID will be pseudorandom, save for 6 pre-determined version and variant bits.
+
+Returns:
+- result: The generated UUID.
+*/
+generate_v4 :: proc() -> (result: Identifier) {
+	bytes_generated := rand.read(result[:])
```
laytan, line 127: According to the RFC this (and all other randomness) SHOULD be cryptographically secure, so this should either use crypto.rand_bytes or assert that the context's random is cryptographic.

core/encoding/uuid/generation.odin
```diff
@@ -0,0 +1,288 @@
...
+bits and a 48 bit timestamp.
+
+It is designed with time-based sorting in mind, such as for database usage, as
+the highest bits are allocated from the timestamp of when it is created.
+
+Returns:
+- result: The generated UUID.
+*/
+generate_v7 :: proc() -> (result: Identifier) {
+	unix_time_in_milliseconds := time.to_unix_nanoseconds(time.now()) / 1e6
```
laytan, line 220: Could the time be a parameter here (and the other versions that take the time), retrieving the time is costly and I can see use cases for the caller caching it for the time that the precision allows.
It could be a `Maybe(time.Time)` where you do a `or_else time.now()` or mandatory with a wrapper?

laytan (commented):
I would like to see a `to_string` where you pass your own byte buffer (because the length is constant callers can use it instead of allocating or messing around with the writers themselves).

Also imo version 3 and 5 should either be removed or be called something like `generate_insecure_v3` because they use insecure hashing algorithms that should be avoided.

core/encoding/uuid/generation.odin
Kelimion: That's a good shout, because you may also want to retroactively (re)create UUID from an existing source of OLTP data with timestamps as part of a migration, for example.

Creativty: > Also imo version 3 and 5 should either be removed or be called something like generate_insecure_v3 because they use insecure hashing algorithms that should be avoided.

I think instead of renaming it, using the new `#warning` builtin comp-time procedure should be used instead.

Kelimion: That's a strange way to ask for a `#warning` directive to be added. There's currently no such thing.

Kelimion: I think you mean `@(warning="Uses deprecated cryptography. Only use this for backward compatibility.")`

Feoramund: > That's a strange way to ask for a `#warning` directive to be added. There's currently no such thing.

Or is there? :)

Feoramund:
There's been some good suggestions brought up. I should be able to handle them today.

Kelimion:
> There's been some good suggestions brought up. I should be able to handle them today.

The consensus we came to within the team was to adopt the same strategy as `core:crypto`, and to add a `core:encoding/uuid/legacy` for the variants using MD5 and SHA1, themselves from `core:crypto/legacy`.

Then there's no need for a compile-time warning of any sort. We'd just have a large banner at the top of the file, same as in `md5.odin`.

e.g.
```cpp
package encoding_uuid_legacy

import "core:encoding/uuid"
import "core:crypto/legacy/md5"

Identifier :: uuid.Identifier
(etc)

generate_v3_bytes :: proc(
    namespace: Identifier,
    name: []byte,
) -> (
    result: Identifier,
)
```

Feoramund:
If we're restricting usage to cryptographic generators, I think that should be fine so long as I include an example of how to do that, since the new context `rand` API is still quite fresh.

I read in the newer RFC that they suggest you could make your own v8 with an up to date hashing algorithm as an alternative to v3 and v5. We could provide one of those too, with the stipulation that v8 is the "make up your own" ID and that ours isn't an official implementation. `generate_v8_sha2` or such. I'd need to take a look at the `crypto` API again. I forget if it's possible to specify the algorithm by enum but last I looked, I think it was. If that's the case, `generate_v8_hash` could work, with an enum argument.

`uuid/legacy` strikes me as better than `generate_insecure_v3`, too.

core/encoding/uuid/generation.odin
Feoramund: I prefer the idea of using the user's RNG and not forcing an implementation, so I went with the assert.

Feoramund:
@Kelimion I believe I've handled all the ideas brought up so far, other than where to put the package. I found myself accidentally typing `core:uuid` sometimes. Is that a better place for it?

Kelimion:
> @Kelimion I believe I've handled all the ideas brought up so far, other than where to put the package. I found myself accidentally typing `core:uuid` sometimes. Is that a better place for it?

I feel like that would set a bad precedent. The package is too narrowly focused for top-level billing. The top-level packages are either categories, or fundamental. I think of `uuid` as more of a `core:encoding/uuid` thing than something fundamental. No offence. :-)

core/encoding/uuid/reading.odin
```diff
@@ -151,16 +152,31 @@ time_v1 :: proc "contextless" (id: Identifier) -> (timestamp: u64) {
...
+Get the timestamp of a version 1 UUID.
+
+Inputs:
+- id: The identifier.
+
+Returns:
+- timestamp: The timestamp of the UUID.
+*/
+time_v1 :: proc "contextless" (id: Identifier) -> (timestamp: time.Time) {
+	delta := cast(time.Duration)(raw_time_v1(id) - HNS_INTERVALS_BETWEEN_GREG_AND_UNIX) * 100
```
Kelimion, line 166: This should probably just be a straight up `Time{_nsec=...}` because subtraction notwithstanding, we're converting a moment between epochs, not calculating an interval.

core/encoding/uuid/reading.odin
```diff
@@ -170,19 +186,46 @@ time_v6 :: proc "contextless" (id: Identifier) -> (timestamp: u64) {
...
+
+Inputs:
+- id: The identifier.
+
+Returns:
+- timestamp: The timestamp, in 100-nanosecond intervals since 1582-10-15.
+*/
+time_v6 :: proc "contextless" (id: Identifier) -> (timestamp: time.Time) {
+	delta := cast(time.Duration)(raw_time_v6(id) - HNS_INTERVALS_BETWEEN_GREG_AND_UNIX) * 100
+	return time.time_add({}, delta)
```
Kelimion, line 199: This should probably just be a straight up `Time{_nsec=...}` because subtraction notwithstanding, we're converting a moment between epochs, not calculating an interval.

core/encoding/uuid/reading.odin
```diff
@@ -170,19 +186,46 @@ time_v6 :: proc "contextless" (id: Identifier) -> (timestamp: u64) {
...
+Get the timestamp of a version 7 UUID.
+
+Inputs:
+- id: The identifier.
+
+Returns:
+- timestamp: The timestamp, in milliseconds since the UNIX epoch.
+*/
+time_v7 :: proc "contextless" (id: Identifier) -> (timestamp: time.Time) {
+	return time.time_add({}, cast(time.Duration)(raw_time_v7(id) * 1e6))
```
Kelimion, line 226: This should probably just be a straight up `Time{_nsec=...}` because subtraction notwithstanding, we're converting a moment between epochs, not calculating an interval.

tests/core/encoding/uuid/test_core_uuid.odin
```diff
@@ -67,87 +67,86 @@ test_legacy_namespaced_uuids :: proc(t: ^testing.T) {
...
+	testing.expect(t, time.diff(time_a, time_b) > 0, "The time on the later-generated v1 UUID is earlier than its successor.")
+	testing.expect(t, time.diff(time_b, time_c) > 0, "The time on the later-generated v1 UUID is earlier than its successor.")
+	testing.expect(t, time.diff(time_a, time_c) > 0, "The time on the later-generated v1 UUID is earlier than its successor.")
 }
 
 @(test)
 test_v6 :: proc(t: ^testing.T) {
 	context.random_generator = crypto.random_generator()
 
+	point_a := time.time_add({}, 1 * time.Second)
```
Kelimion, line 182: That's a really strange way to construct `Time`.

Kelimion (changes requested):

tests/core/encoding/uuid/test_core_uuid.odin
Feoramund: I favored using the public API rather than the `_nsec`, under the assumption that a `_` prefixed variable could be changed in the future.

Kelimion: If it ever does we'll update the users in the core libraries, but if you want to use a constructor, `time.unix` is closer in spirit to what you're doing than `time_add`.

Feoramund:
That ought to do it.

Kelimion:
> That ought to do it.

The thing is that it's `_nsec` because it's "opaque" and it allows changing how it's represented in the future. Unlikely for that to happen at this point, but that's to deter *users* of the core library from poking at it directly. If we merge this PR, then the UUID package is part of the core library packages - not a downstream user - that can do so legitimately.

Kelimion (approved):

Feoramund:
> Unlikely for that to happen at this point, but that's to deter _users_ of the core library from poking at it directly. If we merge this PR, then the UUID package is part of the core library packages - not a downstream user - that can do so legitimately.

That all makes sense to me. I tend to err on the side of caution with these sort of things in the aim of not introducing more of a burden to future maintenance.

Kelimion:
> > Unlikely for that to happen at this point, but that's to deter _users_ of the core library from poking at it directly. If we merge this PR, then the UUID package is part of the core library packages - not a downstream user - that can do so legitimately.
> 
> That all makes sense to me. I tend to err on the side of caution with these sort of things in the aim of not introducing more of a burden to future maintenance.

We can introduce an even lighter `time.time_from_nanoseconds :: proc(nsec: i64)` that's a barely disguised cast that'll inline to the same for people to use for "outside" epoch conversions and Time construction that'll be safe to use in the event of `Time` updates, but I'd consider that outside of the scope of this PR.

Kelimion:
I've just added `time.from_nanoseconds`.

Feoramund:
Not a bad idea at all. I'd have to rebase to get access to it, which will overwrite all the past commits. I _think_ GitHub correctly tracks the differences in force-push rebasing, but I'm not 100% sure. I can do that here and now in this PR if you'd like.

Kelimion:
> Not a bad idea at all. I'd have to rebase to get access to it, which will overwrite all the past commits. I _think_ GitHub correctly tracks the differences in force-push rebasing, but I'm not 100% sure. I can do that here and now in this PR if you'd like.

It should do, especially because it's a simple re-graft on top of the same branch and you're introducing an entirely new package. Up to you. Can also update to use the new call after merge if you want.

Waiting for Bill to give the new package his blessing and merge it. It has mine.

Feoramund:
There we go. Looks like the PR survived intact.

PR #3795, opened 2024-06-22
Fix `#warning` directive

Feoramund:
`#warning` was added in dab3c832e00a3186bddc79585d9ba7ee10d39eaf but cannot actually be used, because it's an unrecognized directive. This should fix it. I did run into one issue though, which is where if I try to assign a `#warning` to a variable, the warning itself shadows the underlying error that would arise. I verified this with `-ignore-warnings` and ran into an assertion failure, which I have also handled in this PR.

### DIFF
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -1727,7 +1727,7 @@ gb_internal bool check_builtin_procedure_directive(CheckerContext *c, Operand *o
 			return false;
 		}
 		warning(call, "%.*s", LIT(operand->value.value_string));
-		if (c->proc_name != "") {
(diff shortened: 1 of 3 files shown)


Kelimion:
We're discussing whether to even keep this and `@(warning)`, as most places where you'd use a warning you'd be better off with an error, and the one place you're not we already have `@(deprecated)` for.

Feoramund:
Warnings are vanishingly rare in the Odin compiler, I'll admit. I'm surprised to see one, if I ever do.

Kelimion:
@gingerBill decided to remove `warning` in c0987394840d63fae627c1623ea2661c31adc9b9, so I'm closing this.

PR #3800, opened 2024-06-23
Invalid Function Signature of proc CmdSetBlendConstants in vendor:vulkan

tadeohepperle:
vulkan_core.h specifies:
```
typedef void (VKAPI_PTR *PFN_vkCmdSetBlendConstants)(VkCommandBuffer commandBuffer, const float blendConstants[4]);
```
but in vendor/vulkan/procedures.odin the function signature is:
```
ProcCmdSetBlendConstants :: #type proc "system" (commandBuffer: CommandBuffer)
```
I think the signature in odin should be 
```
ProcCmdSetBlendConstants :: #type proc "system" (commandBuffer: CommandBuffer, blendConstants: [4]f32)
```


gingerBill:
All of the vulkan headers are _generated_ from the headers directly using a python script. If something is wrong, the byte script need to change. 

Also the corrected signature you gave is still wrong. It should be `[^]f32` or `^[4]f32`.

PR #3804, opened 2024-06-24
Add strings.builder_replace

Beefster09:
In some cases, it can be preferable to replace substrings in a builder in-place. While this is less efficient than generating a new string in many cases, when the lengths of the strings are the same, it may sometimes be faster or the intent may be clearer.

### DIFF
--- a/core/strings/builder.odin
+++ b/core/strings/builder.odin
@@ -812,3 +812,53 @@ Returns:
 write_int :: proc(b: ^Builder, i: int, base: int = 10) -> (n: int) {
 	return write_i64(b, i64(i), base)
 }
+
+/*
+Replaces all non-overlapping instances of `old` with `new`, in-place.
+May be less efficient than `replace` if `new` and `old` are different lengths, but avoids unnecessary allocations.
+
+Inputs:
+- b: A pointer to the Builder
+- old: the substring to replace
+- new: the string replacing it
+- n: the maximum number of replacements (or -1 if unlimited)
+
+NOTE: The backing dynamic array may be fixed in capacity or fail to resize, `replaced_count` states the number of replacements actually made,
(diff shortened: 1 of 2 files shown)


Kelimion:
Can you motivate this addition with a practical use case?

Why, when building a string, would I write "hello" 10 or 20 times and then go back to replace them all with "hellope". Why wouldn't I write "hellope" to begin with?

As an example, I could write `<placeholder beefster>` and then replace that with `Beefster <beefster email>` in a git log after the fact so I don't have to make that lookup while building the string, but I could do one pass over the git log to get all committer names and only then build out the log. Or I could finalize the string with the placeholders and then do a `string.replace`.

This new call feels out of place with how the string builder is used, and I'm also unclear about a solid use case where this would be preferable over waiting for the needle to be known instead of replacing it, or replacing it when the string is built.

gingerBill:
I am not going to accept this. It's kind of just enabling really bad practices from GC languages.

Also, you have clearly not kept to the coding style of the rest of the core library.

Beefster09:
The use case that motivated this was replacing the underscores in an enum string.

I'm not going to argue further for its inclusion, but I am wondering where exactly I deviated from the core coding style so that I can know for next time.

gingerBill:
@Beefster09 If it's an enum string, that means it is a compile time variable, and probably just want to call one of the `strings.to_*_case` calls.

Kelimion:
Or, and I'm just throwing this out there: Bake it into an enumerated array.
```odin
	Foo :: enum {
		A_lot_of_dashes,
	}

	FOO_STRING :: [Foo]string{
		.A_lot_of_dashes = "A lot of dashes",
	}

	fs := FOO_STRING
	fmt.println(fs[.A_lot_of_dashes])
```

PR #3807, opened 2024-06-25
Added fmt.ctprint

karl-zylinski:
There is a `fmt.tprint` but no `fmt.ctprint`, so I added it.

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -368,6 +368,25 @@ caprintf :: proc(format: string, args: ..any, newline := false) -> cstring {
 caprintfln :: proc(format: string, args: ..any) -> cstring {
 	return caprintf(format, ..args, newline=true)
 }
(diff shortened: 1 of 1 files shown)


Kelimion:
There's something bothering me about the procedure's description that it shares with `fmt.print`, `fmt.tprint`, etc, in that it's stated to _format_ the arguments, but takes no format string.

Separating the incoming arguments with a user-definable separator stretches the definition of formatting for me.

Maybe - and I consider this outside the scope of this PR - we should literally just say that: "Separates the arguments with a user-defined separator (space by default) and ...", where ... is "and prints the result to stdout" or "and returns this as a cstring", as appropriate.

karl-zylinski:
@Kelimion maybe it's also because you can send in a value of any type as arg that someone thought of it as "formatting". But I agree, I found it a bit confusing too.

Kelimion:
That's a good point.

PR #3810, opened 2024-06-26
Port `core:net` to FreeBSD

Feoramund:
I'd like to get some extra eyes on this. This is a minimal syscall-based implementation to get FreeBSD working with the `net` package.

I've made sure tests pass, and I have a simple echo server working in TCP with telnet, but I haven't done anything more rigorous as far as testing goes, yet. If there's something out there written in Odin that could stress test the networking side of things, I'd like to know about it. Of note, I've tested this both on AMD64 and ARM64.

Some important notes: I wrote this mostly from the ground up while keeping an eye on what Linux, Darwin, and Windows do for their internal API with regards to `net`. I broke from convention and implemented all of this using only the new `sys/freebsd` package I added, instead of trying to finagle anything with the `os` package, given my understanding that 1) `os` is not a good package in Bill's words and 2) it is due to be replaced by `os2` at some point. Easily sidestepped all of that by keeping things close to the metal, as it were. I opted not to touch `sys/unix` either, since that package seems to be taking on too much responsibility as it is, given how little commonality there is between the UNIX-likes aside from perhaps POSIX `libc` overlap.

During implementation, I noticed that the different platforms tended to have different error enums exposed in their various `Network_Error`s. This is understandable given how systems will work differently. I implemented FreeBSD's on the basis of man page descriptions per the underlying syscalls and tried to make them fit error names that were already present, but in cases where I felt the names were not clear enough, I wrote more verbosely. This can be changed to conform with others if needed.

I'm wondering if this is something that should be standardized in the future, since it means dealing with network errors is different per-platform right now, if you actually need to know the specific error, that is.

I also had to fix the FreeBSD implementation of futexes. There was an incorrect usage of the system API in that it did not pass the size of the `timeout` structure, nor did it specify `WAIT_UINT` when it was using an unsigned integer as the underlying futex data. This was required to get `net` tests to pass, otherwise `Wait_Group` was failing with a complaint about simultaneous waiting and adding. I went ahead and removed any dependency on `libc` there by using direct syscalls, too.

Pinging @andreas-jonsson because I know you are interested in this for NetBSD. I haven't tinkered much with NetBSD, but I'm optimistic that this implementation could serve as a good starting point for that platform. I aimed to keep documentation and organization regarding system types and calls as helpful as possible for future maintenance or porting.

I am not so sure about OpenBSD though. I have heard that Go has had some troubles with their syscall pinning security restrictions, but it's not something I have thoroughly researched. My understanding is that OpenBSD is assertive about `libc` being the API to the kernel, whereas FreeBSD at least has [an entire chapter](https://docs.freebsd.org/en/books/developers-handbook/x86/) in their manual dedicated to programming for the kernel in assembly with syscalls.

- [x] Test TCP.
- [x] Test UDP.
- [x] Test DNS.
- [x] Test non-blocking sockets.
- [x] Test other socket options.
- [x] Review error names.
- [x] Implement more socket options if needed.
- [x] Carefully check every `!freebsd` build tag to see if it can be enabled now (`core:flags` is one I know for sure).

### DIFF
--- a/core/sys/freebsd/syscalls.odin
+++ b/core/sys/freebsd/syscalls.odin
@@ -58,7 +58,7 @@ close :: proc "contextless" (fd: Fd) -> Errno {
 //
 // The getpid() function appeared in Version 7 AT&T UNIX.
 getpid :: proc "contextless" () -> pid_t {
-	// always succeeds
+	// This always succeeds.
 	result, _ := intrinsics.syscall_bsd(SYS_getpid)
 	return cast(pid_t)result
 }
@@ -468,7 +468,7 @@ accept4_T :: proc "contextless" (s: Fd, sockaddr: ^$T, flags: Socket_Flags = {})
 where
 	intrinsics.type_is_subtype_of(T, Socket_Address_Header)
 {
-	// sockaddr must contain a valid pointer, or this will segfault because
+	// `sockaddr` must contain a valid pointer, or this will segfault because
 	// we're telling the syscall that there's memory available to write to.
 	addrlen: u32 = size_of(T)
 
--- a/core/sys/freebsd/syscalls.odin
+++ b/core/sys/freebsd/syscalls.odin
@@ -79,7 +79,10 @@ where
 		cast(uintptr)len(buf),
 		cast(uintptr)flags,
 		cast(uintptr)from,
-		cast(uintptr)fromlen)
+		cast(uintptr)&fromlen)
+
+	// `from.len` will be modified by the syscall, so we shouldn't need to pass
+	// `fromlen` back from this API.
 
 	if !ok {
 		return 0, cast(Errno)result
--- a/core/sys/freebsd/syscalls.odin
+++ b/core/sys/freebsd/syscalls.odin
@@ -1,5 +1,13 @@
 package sys_freebsd
 
+/*
+	(c) Copyright 2024 Feoramund <rune@swevencraft.org>.
+	Made available under Odin's BSD-3 license.
+
+	List of contributors:
+		Feoramund: Initial implementation.
+*/
+
 import "base:intrinsics"
 import "core:c"
 
--- a/core/sys/freebsd/syscalls.odin
+++ b/core/sys/freebsd/syscalls.odin
@@ -450,7 +450,7 @@ sysctl :: proc "contextless" (mib: []MIB_Identifier, oldp: rawptr, oldlenp: ^c.s
 // The _umtx_op() system call is non-standard and is used by the 1:1 Threading
 // Library (libthr, -lthr) to implement IEEE Std 1003.1-2001 (“POSIX.1”)
 // pthread(3) functionality.
-_umtx_op :: proc "contextless" (obj: rawptr, op: Mutex_Operation, val: c.ulong, uaddr, uaddr2: rawptr) -> Errno {
+_umtx_op :: proc "contextless" (obj: rawptr, op: Userland_Mutex_Operation, val: c.ulong, uaddr, uaddr2: rawptr) -> Errno {
 	result, _ := intrinsics.syscall_bsd(SYS__umtx_op,
 		cast(uintptr)obj,
 		cast(uintptr)op,
--- a/core/sys/freebsd/syscalls.odin
+++ b/core/sys/freebsd/syscalls.odin
@@ -207,7 +207,7 @@ fcntl_setfd :: proc "contextless" (fd: Fd, close_on_exec: bool) -> Errno {
 	result, _ := intrinsics.syscall_bsd(SYS_fcntl,
 		cast(uintptr)fd,
 		cast(uintptr)File_Control_Command.SETFD,
-		cast(uintptr)(close_on_exec ? FD_CLOEXEC : 0))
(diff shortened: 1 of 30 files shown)


core/sys/freebsd/syscalls.odin
```diff
@@ -0,0 +1,504 @@
...
+
+	return cast(Errno)result
+}
+
+// Accept a connection on a socket.
+//
+// The accept4() system call appeared in FreeBSD 10.0.
+accept4_T :: proc "contextless" (s: Fd, sockaddr: ^$T, flags: Socket_Flags = {}) -> (Fd, Errno)
+where
+	intrinsics.type_is_subtype_of(T, Socket_Address_Header)
```
JesseRMeyer, line 466: I must have missed it but where is `Socket_Address_Header` defined?

flysand7:
Do you plan on mantaining syscall tables for every version of freebsd? Because as far as I know, unlike linux, freebsd's syscall interface is not stable and no one guarantees that the numbers or the prototypes will stay the same.

Feoramund:
Every version? No.

However, I don't think it's unreasonable to support the officially supported release and stable versions of FreeBSD.

laytan:
How about calling the stable libc wrappers instead? Is there much to gain from doing syscalls directly?

Syscalls is a good idea on Linux because they are stable while libc is a mess there, I don't think it's the case on freebsd?

JesseRMeyer:
According to this source, syscalls on FreeBSD are stable: https://lobste.rs/s/yjd59n/crt_free_2023_tips_tricks#c_viakfm

This has led to Zig adopting syscalls on FreeBSD as opposed to a trip through libc.

laytan:
Aha, great, disregard my comment then :+1:

siemato:
I came across this PR, after running into limits on FreeBSD Current, then realized that Core isn't exactly looking great when trying to use OLS. While my issue doesn't exactly have anything to do with `net` but `os`, and `sys`.

It would make more sense for me to build on top of this PR, so is there anything I can help with to push this forward?

Feoramund:
The PR is more or less finished. I just have to give it a final review, which I may be able to do this weekend. In lieu of stress-testing a real networked program, I may write an additional benchmark.

Feoramund:
I dug into this more than I originally planned to, but it was good. I unearthed some important information: it is imperative that `NOSIGPIPE` is set on FreeBSD sockets. Otherwise, the kernel will raise `SIGPIPE` when writing to a closed socket (currently this silently crashes the test runner - something unrelated I'll have to look into).

Fixed a couple other bugs while I was at it. No stress-testing benchmark, but I think that can be for another PR since it won't be FreeBSD-specific.

(The FreeBSD test suite failed for strange GitHub reasons, but it succeeded on my repo's CI.)

Kelimion:
> (The FreeBSD test suite failed for strange GitHub reasons, but it succeeded on my repo's CI.)

This is what it failed on. Doesn't look like a strange GH reason?
```
+ ./odin test tests/core/normal.odin -file -all-packages '-define:ODIN_TEST_FANCY=false'
  /home/runner/work/Odin/Odin/tests/core/net/test_core_net.odin(615:42) Error: Type 'net.Accept_Error' has no field 'Would_Block' 
  	... ting.expect_value(t, accept_err, net.Accept_Error.Would_Block) { 
  	                                     ^~~~~~~~~~~~~~~^ 
  	                                     
```

JesseRMeyer:
Why is writing to a closed socket without (kernel) error desirable?

Feoramund:
> This is what it failed on. Doesn't look like a strange GH reason?
Aha, so that was buried in the logs. I see it now that you pointed it out. Usually the errors are closer to the bottom of the log, and for some reason it didn't error on my own CI, so I thought it was a glitch.

> Why is writing to a closed socket without (kernel) error desirable?
It still returns an error, but it won't raise `SIGPIPE` and crash your entire program if you don't have a signal handler enabled. Do the other implementations of `core:net` fail in such a catastrophic way? I was under the impression that they didn't.

laytan:
> Do the other implementations of core:net fail in such a catastrophic way? I was under the impression that they didn't.

Other targets also have this signal but it is afaik disabled in the specific send calls with the NOSIGNAL option.

Feoramund:
I see it only for Linux, as you've pointed out. I don't see any handling for the Darwin implementation though. Does that mean `core:net` could raise `SIGPIPE` on Darwin and crash? I don't have a Darwin system to test with. Looks like it goes straight from `net.send` to `os.send` to `_unix_send` with a zero flags argument.

laytan:
> Does that mean core:net could raise SIGPIPE on Darwin and crash?

Yes, and we should remedy that (Darwin also has the NOSIGNAL flag).

Feoramund:
Got a PR up for it. #4029

Feoramund:
I got the tests passing again. It turns out that the `odin-lang/Odin` CI was somehow running with the tests from the future that I implemented based on my work with this PR (mind that this branch is behind quite a bit and I have not rebased), but my personal repo's CI wasn't. Strange, but it was helpful to catch that error early.

Kelimion (approved):

PR #3812, opened 2024-06-26
Add `write_decorated_table` to `text/table`

### LINKED ISSUE #3805: core:text/table - customizable table separators
It would be nice to be able to customize the table separators for text/table, including no borders at all. This can be very useful for aligning text into columns and could be used in the core:flags implementation for usage messages.

Based on the existing code, it doesn't look very hard to add and could be customized fairly easily with a struct containing a few runes.

Feoramund:
Anybody enjoy MS-DOS box drawing graphics?

box_drawing_table

Resolves #3805


Kelimion:
So much yes.

Does this mean we can soon expect to import `core:tui/borland` and render our parsed Odin code in a mock Turbo C IDE?

Feoramund:
I might have to learn how to port Odin to MS-DOS first, but I believe it's possible.

PR #3821, opened 2024-06-27
Allow `ODIN_CLANG_PATH=clang-path` overrides for link stage.

### LINKED ISSUE #3783: Clang won't be found if not aliased?
## Context

```bash
Odin:    dev-2024-06-nightly:f745a1c47
OS:      Pop!_OS 22.04 LTS, Linux 6.8.0-76060800daily20240311-generic
CPU:     AMD Ryzen 9 3900X 12-Core Processor            
RAM:     15907 MiB
Backend: LLVM 17.0.6
```

## Expected Behavior

After installing llvm via `apt.llvm` in any version it should use `clang-version` and not `clang`, as when intalling a specific version it won't add clang as alias.

I would expect odin to add if not present already, a way to tell the odin compiler where to look for the clang executable.

## Current Behavior

The current behaviour is failing when trying to run odin at all, it compiles but fails to execute.

### Steps to Reproduce

1. install llvm-17
2. `odin run .` or `odin build .`

### Failure Logs

`clang: command not found`

Kelimion:
Fixes #3783

### DIFF
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -578,9 +578,16 @@ gb_internal i32 linker_stage(LinkerData *gen) {
 				}
 			}
 
-			gbString link_command_line = gb_string_make(heap_allocator(), "clang -Wno-unused-command-line-argument ");
+			// Link using `clang`, unless overridden by `ODIN_CLANG_PATH` environment variable.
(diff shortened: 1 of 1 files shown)


gingerBill:
I don't like the name `ODIN_CC`. It should probably be something else to be clear what `CC` means.

Issue #3493, opened 2024-04-27
`linux.ppoll` always returns `E_INVAL` due to invalid `sigsetsize`


Feoramund:
## Context

It seems the Linux kernel is expecting a specific value for `sigsetsize` (which glibc handles), despite what the man pages say about this syscall. In Odin, we pass `size_of(Sig_Set)`, which is 128 bytes, compared to the expected 8 bytes (on a 64-bit system, at least).

```
   C library/kernel differences

       [...]

       The raw ppoll() system call has a  fifth  argument,  size_t  sigsetsize,
       which  specifies  the  size in bytes of the sigmask argument.  The glibc
       ppoll() wrapper function specifies this argument as a fixed value (equal
       to sizeof(kernel_sigset_t)).  See sigprocmask(2) for a discussion on the
       differences between the kernel and the libc notion of the sigset.
```

There's an interesting report over on the zig project related to this issue, including a link to the Linux kernel source where this is checked. https://github.com/ziglang/zig/issues/12715

Using `size_of(c.long)` may fix it, but then I wonder what to do about the `Sig_Set` structure as far as the Odin user-facing API is concerned.

	Odin:    dev-2024-04:8fd318ea7
	OS:      Arch Linux, Linux 6.8.7-arch1-1
	CPU:     12th Gen Intel(R) Core(TM) i7-12700K
	RAM:     31916 MiB
	Backend: LLVM 14.0.6

flysand7:
It appears that our definition of the `Sig_Set` type is wrong then. In the issue you have linked, someone pointed out that on 64-bit architectures the size of the structure is 64 bits in x86-64. I don't know how I found the number 128 for the size of this structure, but it felt off all this time.

As for the user-facing API i've been considering making sigset a bitset, there was just an issue where first signal would correspond to bit 0, so I didn't want to tie it to a different type. Maybe I'll reconsider this now that I know its not more than register size.

Feoramund:
I'm assuming our `Sig_Set` was derived from `<bits/types/__sigset_t.h>`, but it looks like the `__NSIG_BYTES` macro (also from glibc) is what the Linux kernel expects in this instance.

```c
#define _SIGSET_NWORDS (1024 / (8 * sizeof (unsigned long int)))
typedef struct
{
  unsigned long int __val[_SIGSET_NWORDS];
} __sigset_t;
```

[...]

PR #3822, opened 2024-06-27
os2 linux round 2

jasonKercher:
This is basically the same as pull request #2863 but with a more narrow scope.  That last one is over a year old, and had gone from trying to expand `sys/unix`, to undoing all that work and using `sys/linux`.  There were random leftover modifications all over the place and the history was complete garbage.

This version only touches `sys/linux` and `os/os2`.  Overview:
 - fixes and additions to `sys/linux` (mostly arm stuff)
 - convert all of os2 to use `sys/linux`
 - new implementations for pipe, process and env

This passes a set of tests [here](https://github.com/jasonKercher/os2test) on amd64 and arm64.


core/sys/linux/types.odin
```diff
@@ -495,16 +495,15 @@ Pid_FD_Flags :: bit_set[Pid_FD_Flags_Bits; i32]
 //  1. Odin's bitfields start from 0, whereas signals start from 1
 //  2. It's unclear how bitfields act in terms of ABI (are they an array of ints or an array of longs?).
 //     it makes a difference because ARM is big endian.
-@private _SIGSET_NWORDS :: (1024 / (8 * size_of(uint)))
+@private _SIGSET_NWORDS :: (8 / size_of(uint))
 Sig_Set :: [_SIGSET_NWORDS]uint
```
Feoramund, line 499: Looks like this will do away with #3493 when `os2` becomes `os`. Glad that's getting fixed.

core/os/os2/env_linux.odin
```diff
@@ -2,29 +2,230 @@
...
+	}
 
+	_env = make(type_of(_env), heap_allocator())
+	cstring_env := _get_original_env()
+	_org_env_begin = uintptr(rawptr(cstring_env[0]))
+	for i := 0; cstring_env[i] != nil; i += 1 {
+		bytes := ([^]u8)(cstring_env[i])
+		n := len(cstring_env[i])
+		_org_env_end = uintptr(&bytes[n])
+		append(&_env, string(bytes[:n]))
```
gingerBill, line 203: Which allocator does `_env` use? Relying on `context.allocator` is not a good idea.

core/os/os2/errors.odin
```diff
@@ -99,3 +99,19 @@ error_string :: proc(ferr: Error) -> string {
 
 	return "unknown error"
 }
+
+print_error :: proc(ferr: Error, msg: string) {
```
gingerBill, line 103:
```odin
print_error :: proc(f: ^File, msg: string, err: Error) {
```

core/os/os2/errors_linux.odin
```diff
@@ -1,145 +1,164 @@
...
+	linux.Errno.EPIPE           = "Broken pipe",
+	linux.Errno.EDOM            = "Numerical argument out of domain",
+	linux.Errno.ERANGE          = "Numerical result out of range",
+	linux.Errno.EDEADLK         = "Resource deadlock avoided",
+	linux.Errno.ENAMETOOLONG    = "File name too long",
+	linux.Errno.ENOLCK          = "No locks available",
+	linux.Errno.ENOSYS          = "Function not implemented",
+	linux.Errno.ENOTEMPTY       = "Directory not empty",
+	linux.Errno.ELOOP           = "Too many levels of symbolic links",
+	41                          = "Unknown Error (41)",
```
gingerBill, line 48: Maybe it's a good idea to add a pseudo-error in `sys/linux` instead for these missing gaps.

core/os/os2/errors_linux.odin
```diff
@@ -1,145 +1,164 @@
...
+	linux.Errno.EUNATCH         = "Protocol driver not attached",
+	linux.Errno.ENOCSI          = "No CSI structure available",
+	linux.Errno.EL2HLT          = "Level 2 halted",
+	linux.Errno.EBADE           = "Invalid exchange",
+	linux.Errno.EBADR           = "Invalid request descriptor",
+	linux.Errno.EXFULL          = "Exchange full",
+	linux.Errno.ENOANO          = "No anode",
+	linux.Errno.EBADRQC         = "Invalid request code",
+	linux.Errno.EBADSLT         = "Invalid slot",
+	58                          = "Unknown Error (58)",
```
gingerBill, line 65: Same again with the pseudo-error

core/os/os2/errors_linux.odin
```diff
@@ -1,145 +1,164 @@
...
-EKEYREJECTED   :: 129  /* Key was rejected by service */
-EOWNERDEAD     :: 130  /* Owner died */
-ENOTRECOVERABLE:: 131  /* State not recoverable */
-ERFKILL        :: 132  /* Operation not possible due to RF-kill */
-EHWPOISON      :: 133  /* Memory page has hardware error */
-
-_get_platform_error :: proc(res: int) -> Error {
-	errno := unix.get_errno(res)
-	return Platform_Error(i32(errno))
+_errno_strings : [int(max(linux.Errno)) + 1]string = {
```
gingerBill, line 6: If the pseudo errors are done, then this can be changed to an enumerated array instead, and then even have `@(rodata, static)` to it.

core/os/os2/file.odin
```diff
@@ -46,10 +46,9 @@ O_SPARSE  :: File_Flags{.Sparse}
 O_CLOEXEC :: File_Flags{.Close_On_Exec}
 
 
-
-stdin:  ^File = nil // OS-Specific
-stdout: ^File = nil // OS-Specific
-stderr: ^File = nil // OS-Specific
+stdin:  ^File = &_stdin
+stdout: ^File = &_stdout
+stderr: ^File = &_stderr
```
gingerBill, line 51: If you see in file_windows.odin, `@(init)` and `@(fini)` calls are used initialize these variables. I would not rely on the global variables, especially since they will not be closed on (successful) program exit.

core/os/os2/process.odin
```diff
@@ -2,36 +2,42 @@ package os2
```
gingerBill, line 2: I wouldn't worry too much about this part because it needs a designing properly. All of this was dummy placeholder code.

core/os/os2/env_linux.odin
jasonKercher: Hey Bill, thank you for taking the time to review the PR.  `_env` uses the `heap_allocator` which is explicit in `_build_env`.  There is no path to `append` without getting to `_build_env` first, so this should be fine.

core/os/os2/process.odin
jasonKercher: Do you want me to just delete the implementation so people don't rely on it?  I could just roll the nuts and bolts into @flysand7's `process` API that he is working on.

core/os/os2/errors_linux.odin
jasonKercher: Addressed in new commit...

core/os/os2/file.odin
jasonKercher: Also handled in new commit.

core/os/os2/process.odin
gingerBill: Good question. Maybe start working with @flysand7 on getting the two ideas in there.

jasonKercher: I just reverted all of the process changes.  No sense adding function to an API that won't be around in the future.  Flysand is better at API design than me anyway. =]

PR #3825, opened 2024-06-28
wgpu_native fixes

Valakor:
1. `LogCallback` should have the "c" calling convention and takes a userdata pointer like other callbacks.
2. `DevicePoll`'s `wrappedSubmissionIndex` argument is an optional in/out parameter. When not specified the function does not query the status of any queue, but if specified the function polls the status of the passed (valid) queue. The previous wrapper method was broken since it passed a non-nil structure to the underlying API with invalid members.
3. `SupportedLimitsExtras.chain` should be a `ChainedStructOut` to properly chain from `SupportedLimits`


laytan:
Thanks for fixing my mistakes :wink:

PR #3828, opened 2024-06-28
re-raise signals from the `odin run` binary

laytan:
Makes `odin run` behave more like `odin build test && ./test` in that it causes (on most shells) the "segmentation fault" message to display, eg:
```
$ odin run test.odin -file
[1]    53445 segmentation fault  odin run test.odin -file
$ odin build test.odin -file && ./test
[1]    54063 segmentation fault  ./test
```

Previously, the former `odin run` variant would not cause the shell to print the message.

Note that this PR has no effect on Windows, because, on the shell I tried, doing `odin build test.odin -file && .\test.exe` does not print anything, so there is nothing to gain there.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -61,6 +61,7 @@ gb_global Timings global_timings = {0};
 #include "llvm-c/Types.h"
 #else
 #include <llvm-c/Types.h>
+#include <signal.h>
 #endif
 
 #include "parser.hpp"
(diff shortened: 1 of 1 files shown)


laytan:
A way to improve this further on Windows would be to detect segfaults specifically and have the compiler print its own message for them, you wouldn't want this on any non-zero exit code though, if somebody does `os.exit(69)` it might be perfectly valid.

Tetralux:
From some pottying about, this seems to do something useful:
```diff
diff --git a/src/main.cpp b/src/main.cpp
index 1896dd0b7..9947abdcb 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -131,6 +131,12 @@ gb_internal i32 system_exec_command_line_app(char const *name, char const *fmt,

                CloseHandle(pi.hProcess);
                CloseHandle(pi.hThread);
+
+               #define NT_ERROR(status) (status >= 0xC0000000 && status <= 0xFFFFFFFF)
+               if (NT_ERROR(exit_code)) {
+                       gb_printf_err("error: child process terminated with error; exit code %d.\n", exit_code);
+                       exit(1);
+               }
        } else {
                // NOTE(bill): failed to create process
                gb_printf_err("Failed to execute command:\n\t%s\n", cmd_line);
```

PR #3831, opened 2024-06-28
Let Darwin safely panic in a test

### LINKED ISSUE #3830: Test runner tends to swallow latest logs before exiting
```
	Odin:    dev-2024-06:9d5bf063e
	OS:      macOS Sonoma 14.5 (build: 23F79, kernel: 23.5.0)
	CPU:     Apple M1
	RAM:     8192 MiB
	Backend: LLVM 18.1.4
```

The new test runner tends to swallow latest logs/messages before exiting, for example a `panic` that stops the test might not actually show up in the output at all.

For example with this test:
```odin
@test
test_panic :: proc(t: ^testing.T) {
	log.panicf("HMMMM")
}
```

Just running it a couple times and I get different results each time:
```
$ odin test tests -define:ODIN_TEST_NAMES=test_panic -define:ODIN_TEST_LOG_LEVEL=warning
[FATAL] --- [2024-06-28 21:51:43] [client.odin:214:test_panic()] HMMMM
[FATAL] --- [2024-06-28 21:51:43] [client.odin:214:test_panic()] panic: log.panicf

$ odin test tests -define:ODIN_TEST_NAMES=test_panic -define:ODIN_TEST_LOG_LEVEL=warning
[FATAL] --- [2024-06-28 21:51:45] [client.odin:214:test_panic()] HMMMM
[FATAL] --- [2024-06-28 21:51:45] [client.odin:214:test_panic()] panic: log.panicf
tests_client  [|                       ]    0/   1 :: test_panic
1 thread                                    0/   1 :: total

$ odin test tests -define:ODIN_TEST_NAMES=test_panic -define:ODIN_TEST_LOG_LEVEL=warning
[FATAL] --- [2024-06-28 21:51:47] [client.odin:214:test_panic()] HMMMM
```

I first thought this might be because it panics as soon as it starts, but even putting a `time.sleep()` before the panic shows some of the same issues.

Feoramund:
I made sure that even with the signal handler catching `SIGTRAP` that I was able to catch a call to `intrinsics.debug_trap()` with `lldb` and continue on merrily without the runner taking control.

This PR also lets Darwin call `pthread_cancel` again, with an advisory comment based on my experiences working with FreeBSD. It is possible that the previous unreliable behavior was based off of a lack of cancelation points being used.

Fixes #3830 

**EDIT:** One moment while I look up what `SIGTRAP` is on Windows...

### DIFF
--- a/core/c/libc/signal.odin
+++ b/core/c/libc/signal.odin
@@ -34,20 +34,7 @@ when ODIN_OS == .Windows {
 	SIGTERM :: 15
 }
 
-when ODIN_OS == .Linux || ODIN_OS == .FreeBSD || ODIN_OS == .Haiku || ODIN_OS == .OpenBSD || ODIN_OS == .NetBSD {
-	SIG_ERR  :: rawptr(~uintptr(0))
-	SIG_DFL  :: rawptr(uintptr(0))
-	SIG_IGN  :: rawptr(uintptr(1)) 
-
-	SIGABRT  :: 6
-	SIGFPE   :: 8
-	SIGILL   :: 4
-	SIGINT   :: 2
-	SIGSEGV  :: 11
-	SIGTERM  :: 15
-}
-
-when ODIN_OS == .Darwin {
+when ODIN_OS == .Linux || ODIN_OS == .FreeBSD || ODIN_OS == .Haiku || ODIN_OS == .OpenBSD || ODIN_OS == .NetBSD || ODIN_OS == .Darwin {
 	SIG_ERR  :: rawptr(~uintptr(0))
 	SIG_DFL  :: rawptr(uintptr(0))
 	SIG_IGN  :: rawptr(uintptr(1)) 
--- a/core/testing/signal_handler.odin
+++ b/core/testing/signal_handler.odin
@@ -9,6 +9,7 @@ Stop_Reason :: enum {
 	Illegal_Instruction,
 	Arithmetic_Error,
 	Segmentation_Fault,
+	Unhandled_Trap,
 }
 
 test_assertion_failure_proc :: proc(prefix, message: string, loc: runtime.Source_Code_Location) -> ! {
--- a/core/testing/signal_handler_libc.odin
+++ b/core/testing/signal_handler_libc.odin
@@ -19,6 +19,11 @@ import "core:os"
 @(private="file", thread_local)
 local_test_index: libc.sig_atomic_t
 
+// Windows does not appear to have a SIGTRAP, so this is defined here, instead
+// of in the libc package, just so there's no confusion about it being
+// available there.
+SIGTRAP :: 5
+
 @(private="file")
 stop_runner_callback :: proc "c" (sig: libc.int) {
 	prev := intrinsics.atomic_add(&stop_runner_flag, 1)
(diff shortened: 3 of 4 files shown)


core/c/libc/signal.odin
```diff
@@ -58,4 +45,5 @@ when ODIN_OS == .Darwin {
 	SIGINT   :: 2
 	SIGSEGV  :: 11
 	SIGTERM  :: 15
+	SIGTRAP  :: 5
```
laytan, line 48: If this is added here, it would also need to be there for Windows, with the `libc` package I believe we want every symbol to be available on all targets.

If windows does not have a `SIGTRAP`, just defining it inline in the test signal handler is probably good too?

laytan: Oh just saw your edit ;)

Feoramund: I did some searching, and it looks like Windows doesn't have one. I'll figure out a fix.

Feoramund:
I think the MacOS Intel test is stalling on the thread pool termination in the demo, if I had to take a guess.

laytan:
> I think the MacOS Intel test is stalling on the thread pool termination in the demo, if I had to take a guess.

hmm, now we know why it was "unreliable" I guess?

Although I do wonder if this is a 100% thing or a failure that we sometimes have in ci

Feoramund:
This idea might be starting to creep beyond the scope of this PR, but I think `thread` either needs a new API for threads to call to check a cancelation point to satisfy pthread (a no-op on other platforms), or thread termination should be defined as forceful termination and not cancellation, possibly with a split in the API, one proc for terminate/kill (immediate) and one for cancel (negotiated).

I imagine forceful termination for relevant platforms could be handled as a thread-specific signal handler in the thread preamble, with `thread.terminate` sending a signal to that thread, similar to how it's done in the test runner except not process-wide.

laytan:
> This idea might be starting to creep beyond the scope of this PR, but I think `thread` either needs a new API for threads to call to check a cancelation point to satisfy pthread (a no-op on other platforms), or thread termination should be defined as forceful termination and not cancellation, possibly with a split in the API, one proc for terminate/kill (immediate) and one for cancel (negotiated).
> 
> I imagine forceful termination for relevant platforms could be handled as a thread-specific signal handler in the thread preamble, with `thread.terminate` sending a signal to that thread, similar to how it's done in the test runner except not process-wide.

Yeah this is starting to get into things the thread rewrite will probably cover. I just hoped this was a quick fix we could add for now so I/macos users can run tests without pulling their hair out after a panic.

Feoramund:
> I just hoped this was a quick fix we could add for now so I/macos users can run tests without pulling their hair out after a panic.

I hoped it was going to be quick too, but it looks like we stumbled into the mire on this one.

laytan:
Hmm there are some other when statements regarding cancelling further up in thread_unix, these probably need to be removed too, maybe it will work then?

Otherwise, yeah, out of (quick) ideas, thanks for the quick response and PR though, at least it works locally on my end.

Feoramund:
> Hmm there are some other when statements regarding cancelling further up in thread_unix, these probably need to be removed too, maybe it will work then?

You're right. Well, it's worth a shot. I've pushed again, so we'll see in a few minutes.

laytan:
> > Hmm there are some other when statements regarding cancelling further up in thread_unix, these probably need to be removed too, maybe it will work then?
> 
> You're right. Well, it's worth a shot. I've pushed again, so we'll see in a few minutes.

Looks like it worked and is ready to go now! iMO a great change pending the thread rewrite.

PR #3832, opened 2024-06-29
Add `-vet-identical-cast`

Feoramund:
This is a feature I've wanted for a while, and I think I've finally figured out how to get it working. I had to disable this check on polymorphic procedures, because the information that a type was originally polymorphic didn't seem to be stored anywhere, as far as I could find. (`is_type_polymorphic` wasn't working for me on the individual types, at least, but it did work on the contextual proc.)

This should be a good feature for refactoring code when changing one type to another.

I would have done this for `transmute` too, but I saw the usage of a compile-time type-switching paradigm in `base:runtime/internal.odin` where `__float16` is declared to be one type or another, and this was causing the identical transmute prevention to trigger. I wasn't sure what to do about that, so I left it as-is and didn't make `transmute` also check for identical types, for the time being until I get some feedback.

### DIFF
--- a/core/encoding/uuid/generation.odin
+++ b/core/encoding/uuid/generation.odin
@@ -106,8 +106,8 @@ generate_v6 :: proc(clock_seq: Maybe(u16) = nil, node: Maybe([6]u8) = nil, times
 		temporary: [2]u8
 		bytes_generated := rand.read(temporary[:])
 		assert(bytes_generated == 2, "RNG failed to generate 2 bytes for UUID v1.")
-		result[8] |= cast(u8)temporary[0] & 0x3F
-		result[9]  = cast(u8)temporary[1]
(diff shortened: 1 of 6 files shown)


gingerBill:
So I experimented with this before, but it might have too many false positives because of polymorphic procedures.

I think this might still have that problem but for nested procedures within polymorphic ones.

```odin
foo :: proc(x: $T) {
    bar :: proc(x: T) {
        ... // this isn't a polymorphic procedure to the compiler, per se
    }
}
```

gingerBill:
I am going to merge this and then solve that specific problem.

Feoramund:
This looks awesome. Thanks for figuring out the rest of it. :) The check for unneeded transmutes is really cool too.

PR #3835, opened 2024-06-29
Add `raw_data` proc to `base/builtin/builtin.odin`

thetarnav:
`raw_data` was missing, but it's nice to be able to see it's rough definition.
Had to split it into a proc group, so the other procs are made private *because they are not real anyway*


gingerBill:
These are not even procedures. That's just going to confuse people further.

PR #3846, opened 2024-06-30
Add more `core:net` tests

Feoramund:
I wrote these for my ongoing work with FreeBSD.

### DIFF
--- a/tests/core/net/test_core_net.odin
+++ b/tests/core/net/test_core_net.odin
@@ -18,6 +18,7 @@ import "core:sync"
 import "core:time"
 import "core:thread"
 import "core:fmt"
+import "core:log"
 
 @test
 address_parsing_test :: proc(t: ^testing.T) {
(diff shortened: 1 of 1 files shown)


tests/core/net/test_core_net.odin
```diff
@@ -516,6 +519,102 @@ join_url_test :: proc(t: ^testing.T) {
...
+	testing.expect_value(t, msg, transmute(string)buf[:bytes_read])
+}
+
+@test
+test_dns_resolve :: proc(t: ^testing.T) {
+	// NOTE: This test depends on external factors, so if it fails, an IP
+	// address may have changed or become unavailable.
+
+	// The net API returns only one address per protocol version, and DNS
+	// records can store many, so we'll have to check all possibilities.
```
Kelimion, line 569: We could add a `resolve_iterate` that iterates over DNS response classes until there's no more entries.
`resolve` could then use that and just cap at 1.

PR #3849, opened 2024-07-01
Don't factor trailing zeroes into mantissa division

Feoramund:
This should fix issues where `N00 / (pow+2)` results in a different number than `N / pow`.

@karl-zylinski Would you be able to try this out and let me know if it resolves your issue with floating point precision loss from #3847?

### DIFF
--- a/core/strconv/strconv.odin
+++ b/core/strconv/strconv.odin
@@ -932,6 +932,7 @@ parse_f64_prefix :: proc(str: string) -> (value: f64, nr: int, ok: bool) {
 		nd := 0
 		nd_mant := 0
 		decimal_point := 0
+		trailing_zeroes_nd := -1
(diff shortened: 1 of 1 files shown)


karl-zylinski:
This resolves my issues. Thank you so much! You're such a great contributor to this project!

For reference, this program:
```
v: f64 = 1
b := strings.builder_make()
strings.write_f64(&b, v, 'f')
vs := strings.to_string(b)
vv, _ := strconv.parse_f64(vs)
fmt.println(vv)
```
outputs `1.976262583364986e-308` without these changes.
With these changes it outputs just `1`.

Similarly, with `v: f64 = 1.2` it outputs `1.1102230246251565` without these changes. But with these changes it outputs a perfect `1.2`.

✨✨✨✨

Issue #3856, opened 2024-07-02
src/llvm_backend_const.cpp(412): Assertion Failure: `sz >= max_count` max_count: 7, sz: 4, written: 0, type int


laytan:
On 32 bit targets (wasm32, arm32) the following code raises the assertion failure in the title of this issue:

```odin
package main

import "core:math"

main :: proc() {
	x := math.round_f64(1)
}
```

laytan:
Looks like the constant math is acting as if the destination is int, which the value doesn't fit in on 32 bit so it asserts there
Doing this "fixes" it:
```patch
diff --git a/core/math/math.odin b/core/math/math.odin
index 3d0ab3c4e..1a76d5244 100644
--- a/core/math/math.odin
+++ b/core/math/math.odin
@@ -805,8 +805,8 @@ round_f64 :: proc "contextless" (x: f64) -> f64 {
             bits |= transmute(u64)f64(1)
         }
     } else if e < bias + shift {
-        half     :: 1 << (shift - 1)
-        mantissa :: (1 << shift) - 1
+        half     :: u64(1) << (shift - u64(1))
+        mantissa :: (u64(1) << shift) - u64(1)
         e -= bias
         bits += half >> e
         bits &~= mantissa >> e
 ```

PR #3859, opened 2024-07-02
wasm: support `vendor:stb/truetype` and `vendor:fontstash`

laytan:
- Adds linking of wasm object files to the compiler and linker, very simply looks at the file extension being `.o`, I am open for better ideas but we already have a bunch of file extension checks so it should fit the codebase
- Makes the `core:os` package work much better on `wasi`, previously it would take file descriptor number 3 and use that as the relative base for every `os.open`, this only works (and only with relative paths) if the environment makes number 3 the current directory. Now we use the preopen functionality correctly by retrieving them all to check paths given to `os.open` against
- Adds a separate compilation of a wasm object file to `vendor:stb/truetype` and redefines the stdlib symbols it would normally use to Odin's definitions

I found a bug #3856 which makes the `fmod` definition not work, I added a panic for now, which in my use-case wasn't hit because this is only used in a single exported function.


laytan:
I might move the preopen things out of `@(init)` so it is not compiled in when you don't use `os.open`

PR #3861, opened 2024-07-02
Remove @(require_results) from `arena_init_X` in `virtual.odin`

karl-zylinski:
Since you can init a growing arena like this (which is how I usually do it):
```
arena: vmem.Arena
arena_allocator := vmem.arena_allocator(&arena)
```
And then use for example a dynamic array to grow that arena without having to handle any allocation error, then you shouldn't be required to handle the initial error of the arena init either.


gingerBill:
No. I added that on purpose.

PR #3881, opened 2024-07-06
[core:strings] split_by_byte_iterator add nil pointer & empty string check

blob1807:
add nil check to prevent deferring a nil pointer and empty string check to stop iterating early.

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -1059,6 +1059,11 @@ Output:
 
 */
 split_by_byte_iterator :: proc(s: ^string, sep: u8) -> (res: string, ok: bool) {
+	// stop once the string is empty or nil
+	if s == nil || len(s^) == 0 {
+		return
+	}
+	
 	m := index_byte(s^, sep)
 	if m < 0 {
 		// not found


Kelimion:
I'm not seeing it iterate on an empty string or crash because of a `nil` dereference:
```cpp
package bug

import "core:mem"
import "core:fmt"
import "core:strings"

_main :: proc() {
	fmt.println("--- 4-piece string ---")
	{
		i := 0; foo := "foo,bar,baz,blarz"
		for piece in strings.split_by_byte_iterator(&foo, ',') {
			defer i += 1
			fmt.println(i, piece)
		}
	}

	fmt.println("--- empty string ---")
	{
		i := 0; foo := ""
		for piece in strings.split_by_byte_iterator(&foo, ',') {
			defer i += 1
			fmt.println(i, piece)
		}
	}

	fmt.println("--- nil string ---")
	{
		_foo: []u8
		i := 0; foo := string(_foo)
		for piece in strings.split_by_byte_iterator(&foo, ',') {
			defer i += 1
			fmt.println(i, piece)
		}
	}
	fmt.println("That's all folks!")
}
```

Can you share usage code that illustrates this nil pointer dereference? I'm not seeing anything to fix that this fix addresses in practice.

Kelimion:
I can think of this, but while I can see a nil data pointer in a string as above, how would you run into `^string  = nil` without the caller writing buggy code?
```cpp
fmt.println("--- nil ^string pointer ---")
{
	i := 0; foo: ^string
	for piece in strings.split_by_byte_iterator(foo, ',') {
		defer i += 1
		fmt.println(i, piece)
	}
}
```

Kelimion:
I'm inclined not to merge it for several reasons:
1) A crash because of a nil deference is a useful signal to the programmer that they goofed up. Your patch would allow buggy code (for passing a nil string pointer to begin with) to continue execution and wreak more havoc down the line. That's ultimately not in the programmer's interest.

2) A `for in` iteration of the same nil pointer crashes, so this fix introduces new semantics to the language.
```cpp
	foo: ^string
	for ch, i in foo^ {
		fmt.println(i, ch)
	}
	fmt.println("That's all folks!")
```

blob1807:
> 1. A crash because of a nil deference is a useful signal to the programmer that they goofed up. Your patch would allow buggy code (for passing a nil string pointer to begin with) to continue execution and wreak more havoc down the line. That's ultimately not in the programmer's interest.

I should have pointed out that this keeps it in line with how the normal split_iterator works.
https://github.com/odin-lang/Odin/blob/80d122f531bb8534f9885598db7d788a4c0f02c9/core/strings/strings.odin#L1003-L1007

If this is the right solution. then shouldn't split_iterator be changed as well?

Kelimion:
> > 1. A crash because of a nil deference is a useful signal to the programmer that they goofed up. Your patch would allow buggy code (for passing a nil string pointer to begin with) to continue execution and wreak more havoc down the line. That's ultimately not in the programmer's interest.
> 
> I should have pointed out that this keeps it in line with how the normal split_iterator works.
> 
> https://github.com/odin-lang/Odin/blob/80d122f531bb8534f9885598db7d788a4c0f02c9/core/strings/strings.odin#L1003-L1007
> 
> If this is the right solution. then shouldn't split_iterator be changed as well?

That's a good point. I'll remove it there as well.

Kelimion:
Rather than passing garbage to a procedure and then expecting it not to throw up, we should really try not to pass garbage in the first place.

It's what we do with untrusted user input as well: validate, then use. Now you could argue that the split iterator here would be doing that validation before doing useful work, but it's already been pushed down the stack for it to get that far.

PR #3890, opened 2024-07-08
ci: make releases use LLVM 18

laytan:
Since we updated to LLVM 18 on Windows now, we should make the releases for non-windows use LLVM 18 too, this way all releases use the same version.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -50,8 +50,8 @@ jobs:
         run: |
           wget https://apt.llvm.org/llvm.sh
           chmod +x llvm.sh
-          sudo ./llvm.sh 17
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -50,8 +50,8 @@ jobs:
         run: |
           wget https://apt.llvm.org/llvm.sh
           chmod +x llvm.sh
-          sudo ./llvm.sh 17
-          echo "/usr/lib/llvm-17/bin" >> $GITHUB_PATH
+          sudo ./llvm.sh 19
```
gingerBill, line 53: Is that meant to be 18?

laytan: It is, just noticed it myself too

PR #3893, opened 2024-07-08
implement `temporary_directory` on non-windows

laytan:


src/string.cpp
```diff
@@ -544,11 +544,10 @@ gb_internal String string16_to_string(gbAllocator a, String16 s) {
 
 
 gb_internal String temporary_directory(gbAllocator allocator) {
-	String res = {};
 #if defined(GB_SYSTEM_WINDOWS)
 	DWORD n = GetTempPathW(0, nullptr);
 	if (n == 0) {
-		return res;
+		return String{0};
```
gingerBill, line 550: That should be identical at any rate.

PR #3904, opened 2024-07-10
make instrumentation "work" on wasm

laytan:
Using instrumentation on WASM causes it to complain about the llvm.returnaddress instrinsic. This PR could be considered a "hack" but makes this work by just passing `nil` instead of the return address.


gingerBill:
This is a bodge, but it is WASM so bodges will be needed.

PR #3908, opened 2024-07-12
Write docs for `core:container/intrusive/list` package

NicknEma:
Since I found it very hard to understand how this package worked without any example, I added some, plus some general description of the procedures and their parameters.
All the examples have been compiled and tested.

### DIFF
--- a/core/container/intrusive/list/doc.odin
+++ b/core/container/intrusive/list/doc.odin
@@ -8,6 +8,13 @@ An intrusive container requires a `Node` to be embedded in your own structure, l
 		value: string,
 	}
 
+Embedding the members of a `list.Node` in your structure with the `using` keyword is also allowed:
+
+	My_String :: struct {
+		using node: list.Node,
+		value: string,
+	}
+
 Here is a full example:
 
 	package test
--- a/core/container/intrusive/list/doc.odin
+++ b/core/container/intrusive/list/doc.odin
@@ -1,0 +1,39 @@
+/*
+Package list implements an intrusive doubly-linked list.
+
+An intrusive container requires a `Node` to be embedded in your own structure, like this:
+
+	My_String :: struct {
+		node:  list.Node,
+		value: string,
+	}
+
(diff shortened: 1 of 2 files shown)


core/container/intrusive/list/intrusive_list.odin
```diff
@@ -104,12 +136,26 @@ remove_by_proc_contextless :: proc(list: ^List, to_erase: proc "contextless" (^N
 	}
 }
 
+/*
+Checks if the given list does not contain any element.
```
Kelimion, line 140: Minor nitpick: Prefer `Checks whether` instead of `Checks if`, because the latter says the check is conditional on the list not containing any element, rather than whether that's the case.

core/container/intrusive/list/intrusive_list.odin
```diff
@@ -61,7 +81,13 @@ remove :: proc "contextless" (list: ^List, node: ^Node) {
 		}
 	}
 }
+/*
+Removes from the given list all elements that satisfy a condition.
```
Kelimion, line 85: Implied, but given you mentioned O(1) complexity elsewhere, perhaps mention O(N) complexity here and for the contextless variant.

core/container/intrusive/list/doc.odin
```diff
@@ -0,0 +1,39 @@
+/*
+Package list implements an intrusive doubly-linked list.
+
+An intrusive container requires a `Node` to be embedded in your own structure, like this:
+
+	My_String :: struct {
+		node:  list.Node,
```
Kelimion, line 7: A minor point, but it can be helpful to know that `using node: list.Node,` is also allowed thanks to the `container_of` magic. A more naïve implementation of an intrusive list may have not worked with the `Node`'s fields being part of the container instead of its child, if that makes sense.

Kelimion (commented):
Solid documentation, thanks! I just have a few minor notes.

PR #3909, opened 2024-07-12
Add `field_name` parameter to `flags.Custom_Type_Setter`

FourteenBrush:
Motivation:
I found it quite difficult to provide an accurate error message in a type setter when the field name is not known.

### DIFF
--- a/core/flags/internal_assignment.odin
+++ b/core/flags/internal_assignment.odin
@@ -48,6 +48,7 @@ push_positional :: #force_no_inline proc (model: ^$T, parser: ^Parser, arg: stri
 	args_tag, _ := reflect.struct_tag_lookup(field.tag, TAG_ARGS)
 	field_name := get_field_name(field)
 	error = parse_and_set_pointer_by_type(ptr, arg, field.type, args_tag)
+
 	#partial switch &specific_error in error {
 	case Parse_Error:
 		specific_error.message = fmt.tprintf("Unable to set positional #%i (%s) of type %v to `%s`.%s%s",
@@ -160,6 +161,7 @@ set_option :: proc(model: ^$T, parser: ^Parser, name, option: string) -> (error:
 	ptr := cast(rawptr)(cast(uintptr)model + field.offset)
 	args_tag := reflect.struct_tag_get(field.tag, TAG_ARGS)
 	error = parse_and_set_pointer_by_type(ptr, option, field.type, args_tag)
+
 	#partial switch &specific_error in error {
 	case Parse_Error:
 		specific_error.message = fmt.tprintf("Unable to set `%s` of type %v to `%s`.%s%s",
--- a/core/flags/internal_assignment.odin
+++ b/core/flags/internal_assignment.odin
@@ -47,7 +47,7 @@ push_positional :: #force_no_inline proc (model: ^$T, parser: ^Parser, arg: stri
 	ptr := cast(rawptr)(cast(uintptr)model + field.offset)
 	args_tag, _ := reflect.struct_tag_lookup(field.tag, TAG_ARGS)
 	field_name := get_field_name(field)
-	error = parse_and_set_pointer_by_type(ptr, arg, field.type, args_tag)
+	error = parse_and_set_pointer_by_type(ptr, arg, field.type, field.name, args_tag)
 
 	#partial switch &specific_error in error {
 	case Parse_Error:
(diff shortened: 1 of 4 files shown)


Feoramund:
The type setter is not designed to know the field names, as its job is only to handle setting types, regardless of what fields it's applying to. In that sense, it's indiscriminate. The flag checker is the hook where you can act on specific fields. By default, the error handler chain provides what field name caused the error in the first place, prepended to any message provided by the type setter.

Can you provide more context as to how you're using the package?

FourteenBrush:
> The type setter is not designed to know the field names, as its job is only to handle setting types, regardless of what fields it's applying to. In that sense, it's indiscriminate. The flag checker is the hook where you can act on specific fields. By default, the error handler chain provides what field name caused the error in the first place, prepended to any message provided by the type setter.
> 
> Can you provide more context as to how you're using the package?

I might not understand the scope of those two components correctly, but regardless I had a type setter where I had to do some parsing on the raw option string in order to end up with an os.Handle. When that parsing failed, I wanted to provide an accurate error message, but due to the lack of a field name, the only reasonable thing to do was to just assing an os.INVALID_HANDLE and wait till the flag checker to provide an accurate error message.

Maybe that's intended behavior but it just felt strange to do.

Feoramund:
If your error is dependent on the field itself, then it sounds like you're doing the right thing by waiting until the flag checker to produce an error.

Do you have a minimal example of your error situation? Perhaps what your use case looks like if you err only in the type setter compared to in the flag checker.

FourteenBrush:
Sorry for the late response, had some other things going on..

This could be considered the situation that I have:
```go
Opts :: struct {
    initial_mem: MemSize `flags:"required"`,
    max_mem:     MemSize `flags:"required"`,
}

MemSize :: distinct uint

opts_type_setter :: proc(
    data: rawptr,
    datatype: typeid,
    unparsed: string,
    args_tag: string,
) -> (
    error: string,
    handled: bool,
    alloc_error: mem.Allocator_Error,
) {
    if datatype != MemSize do return

    size, ok := parse_mem_size(unparsed)
    if !ok {
        // set error, we can only provide a generic "invalid memory size"
        // instead of indicating which property was invalid
        return "Invalid memory size", true, alloc_error
    }

    (^MemSize)(data)^ = size
    handled = true
    return
}

parse_mem_size :: proc(raw: string) -> (size: MemSize, ok: bool) {
    // do some parsing here to convert e.g. "10M" to 10 * mem.Megabyte
    return
}
```

So the lack of a field name makes me provide a generic error instead of saying which specific arg was invalid.
Of course this can be overcome with deferring the error handling to the flag checker and just setting the field as -1 instead.

Now imagine MemSize would be a signed type, what could we do? Have an extra hidden boolean field that says whether the field is erroneous or not?

In my eyes all these situations of deferring the error, which can be the result of parsing, can be avoided by simply having a field name provided in order to fail with the wanted error message immediately.

I don't know what your thoughts are on this?

Feoramund:
> So the lack of a field name makes me provide a generic error instead of saying which specific arg was invalid.

This shouldn't be needed, as the error handler chain already prepends the name of the field to the error message. From the example, if you were to type `./example example -accuracy=asdf`, you would be told:
```
[Parse_Error.Bad_Value] Unable to set `accuracy` of type Fixed_Point1_1 to `asdf`. Incorrect format. Must be in the form of `i.f`.
```

> Now imagine MemSize would be a signed type, what could we do? Have an extra hidden boolean field that says whether the field is erroneous or not?

I have not tried it myself, but I can't imagine why using a `Maybe` or another sort of `union` wouldn't work.

> In my eyes all these situations of deferring the error, which can be the result of parsing, can be avoided by simply having a field name provided in order to fail with the wanted error message immediately.
> 
> I don't know what your thoughts are on this?

Based on your example code, you're still doing things correctly by deferring the error reporting if you need to know which specific flag was in error. However, there is no need to put the flag name into the type setter merely to report the name of the flag. That's already handled.

The situation in which I can imagine you would need to know the name of the flag would be if you had two fields of the same type which have very different consequences in how they are handled. I.e. the possible errors and error responses have nothing to do with each other at the type level, which is why the flag checker gets the name and not the type setter.

Say hypothetically you have two flags `a` and `b` of some type `distinct int` with generic parsing, but you want only `a` to be within a certain range: that is when the flag checker is meant to be used to verify against the name of the field. If it was common enough, you could even encode your custom requirements into the `args_tag`. However, if you go that far, it would be better to move verification back to the type setter stage, since that requirement would be encoded into the type now.

The type setter is purely for setting types and mustn't need to know about field names: if it gets to that point, you may as well have two different types (which might be another solution for you).

But as far as I can tell just based on this example code, I see no reason why you would need to know the field name at that stage if all you want to do is put the field name into the error message. That's already done for you.

FourteenBrush:
> But as far as I can tell just based on this example code, I see no reason why you would need to know the field name at that stage if all you want to do is put the field name into the error message. That's already done for you.

I mean yes, but that still requires you to get the field name out of the error message.
Apart from that, I think you've given enough alternatives so I think I'll just leave it here..

PR #3914, opened 2024-07-13
Add support of `ignore` tag for `json.marshal`

VladPavliuk:
Propose an ability to specify optional `ignore` tag for struct fields to exclude them from `json.marshal` json object output.

Example:
```
My_Struct :: struct {
	a: string `json:"-"`,
}

my_struct := My_Struct{
	a = "test",
}

marshaled_data, _ := json.marshal(my_struct)
```
will produce:
```
{}
```


laytan:
IMO it would be better to do `json:"-"` to match `cbor:"-"` and `fmt:"-"` which do the same thing in their respective packages.

gingerBill:
Please use `-` not `_`.

VladPavliuk:
@laytan @gingerBill thanks for pointing that out! I didn't know about that already existing ignore functionality in `cbor` and `fmt`.
The changes where pushed to the branch.

Issue #2657, opened 2023-07-14
Clarification of Odin's return by value calling convention behavior


JesseRMeyer:
What should the value of `game.bob^` be?

`Game` is larger than a register size.  According to Odin's calling convention, the callee should reserve the stack space, and assignments in this case should be made directly ("copy elision").  But that does not happen.  Apparently the callee and the caller reserve space, and at the function end, there is a copy.

```odin
package main

import "core:fmt"

Game :: struct {
    bob: ^u64,
    alice: u64,
}

gameInit :: proc() -> (game: Game) {//(game: ^Game) {
    game.bob = &game.alice

    game.alice = 5;

    return
}

main :: proc() {
    //game: Game
    //gameInit(&game)
    game := gameInit()

    fmt.println(game.bob^)
}
```

Sanian-Creations:
I'm not sure where you get your understanding from.

Copy elison is an _optimization_, not expected behaviour. Named returns are treated like regular local variables, only with optimizations turned on will the compiler even _consider_ copy elison. But even with optimized builds, it may or may not apply it. Don't go writing code that expects this to happen.

Here's a little snippet that returns a struct and it checks whether its address is the same in the caller as in the callee.

```odin example.odin
package main
import "core:fmt"

Thing :: struct {
    self_ptr : ^Thing,
    padding  : [16]u8 // make bigger to increase chance of optimization
}

main :: proc() {
    fmt.println("size of Thing:", size_of(Thing))
    caller()
}

caller :: proc() {
    t := callee()
    fmt.printf("addr in callee: %#p\n", &t.self_ptr)
    fmt.printf("addr in caller: %#p\n", &t)
    fmt.println(t.self_ptr == &t ? "used" : "did not use", "copy elison")
}

callee :: proc() -> (retval: Thing) {
    retval.self_ptr = &retval
    // printing retval here may also trigger the optimization, even if the struct is not that big.
    return
}
```

Run with `odin run example.odin -file -o:speed` (without `-o:speed` it never uses copy elison)

What you'll find is that optimization is unpredictable. The code in your own example, for instance, doesn't prints 0 if that is the whole file, but when I renamed the main function and called it from some other procs I was messing with, it printed 5 like you expected it to. There are effectively infinite factors that could influence how the compiler optimizes, you cannot rely on it.

JesseRMeyer:
Odin's calling convention for return values greater than register size is (essentially):

Callee reserve stack memory for return value
Call func()
[...]

PR #3916, opened 2024-07-13
Add `#by_ptr` modifier to procedures in the small array package

thetarnav:
even though Small_Array has small in its name, it's not uncommon to just use it as a global array with thousands of items.
Using procedures that do not receive the array by a pointer will cause the array to be copied, if not optimized by llvm. (which didn't happen in my case)
This was visibly slowing down my game when the `get` was proc was getting called once for each item every frame.
Adding `#by_ptr` solved the issue.
Which is a bit weird considering the odin calling convention.

### DIFF
--- a/core/container/small_array/small_array.odin
+++ b/core/container/small_array/small_array.odin
@@ -10,15 +10,15 @@ Small_Array :: struct($N: int, $T: typeid) where N >= 0 {
 }
 
 
-len :: proc "contextless" (a: $A/Small_Array) -> int {
+len :: proc "contextless" (#by_ptr a: $A/Small_Array) -> int {
 	return a.len
 }
 
-cap :: proc "contextless" (a: $A/Small_Array) -> int {
+cap :: proc "contextless" (#by_ptr a: $A/Small_Array) -> int {
 	return builtin.len(a.data)
 }
 
-space :: proc "contextless" (a: $A/Small_Array) -> int {
+space :: proc "contextless" (#by_ptr a: $A/Small_Array) -> int {
 	return builtin.len(a.data) - a.len
 }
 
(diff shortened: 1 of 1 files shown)


flysand7:
In Odin all objects larger than 16 bytes in size will be passed by pointer, and that is supposed to be documented, so in this case `#by_ptr` is not needed for those optimizations to take place.

`#by_ptr` is meant to be used in very specific circumstances (bindings to C functions with const reference parameters), and normally you never have to use it within Odin code. It is good to avoid using this tag outside of ffi-related code.

thetarnav:
> In Odin all objects larger than 16 bytes in size will be passed by pointer, and that is supposed to be documented, so in this case `#by_ptr` is not needed for those optimizations to take place.

I also heard that before, this is why I said it was weird that `#by_ptr` had such an impact, but it did.

Here is a little repro if you want to try it:

```odin
import "core:time"
import "core:fmt"


Small_Array :: struct ($N: int, $T: typeid) where N >= 0 {
	data: [N]T,
	len:  int,
}

get_odin   :: proc (a: $A/Small_Array($N, $T), index: int) -> T {
	return a.data[index]
}

get_c      :: proc "c" (a: $A/Small_Array($N, $T), index: int) -> T {
	return a.data[index]
}

get_ptr    :: proc (a: ^$A/Small_Array($N, $T), index: int) -> T {
	return a.data[index]
}

get_by_ptr :: proc (#by_ptr a: $A/Small_Array($N, $T), index: int) -> T {
	return a.data[index]
}


main :: proc () {
	numbers: Small_Array(30000, int)
	numbers.len = len(numbers.data)
	numbers.data = 1

	{
		start := time.tick_now()
		res: int
		for i in 0 ..< numbers.len {
			res += get_odin(numbers, i)
		}
		duration := time.tick_since(start)
		fmt.printfln("get_odin  :{} | {}", res, duration)
	}

	{
		start := time.tick_now()
		res: int
		for i in 0 ..< numbers.len {
			res += get_c(numbers, i)
		}
		duration := time.tick_since(start)
		fmt.printfln("get_c     :{} | {}", res, duration)
	}

	{
		start := time.tick_now()
		res: int
		for i in 0 ..< numbers.len {
			res += get_ptr(&numbers, i)
		}
		duration := time.tick_since(start)
		fmt.printfln("get_ptr   :{} | {}", res, duration)
	}

	{
		start := time.tick_now()
		res: int
		for i in 0 ..< numbers.len {
			res += get_by_ptr(numbers, i)
		}
		duration := time.tick_since(start)
		fmt.printfln("get_by_ptr:{} | {}", res, duration)
	}
}
```

result on my machine:

```
$ odin run .
get_odin  :30000 | 443.504189ms
get_c     :30000 | 444.267005ms
get_ptr   :30000 | 246.76µs
get_by_ptr:30000 | 230.612µs
```

odin report:

```
	Odin:    dev-2024-07:7a8775592
	OS:      Ubuntu 22.04.4 LTS, Linux 6.5.0-41-generic
	RAM:     7631 MiB
	Backend: LLVM 17.0.6
```

thetarnav:
this might also be a relevant issue: https://github.com/odin-lang/Odin/issues/2657

laytan:
Fixed in 8650180cccad02b39a00dcd05f696a8e3879c29e so this can be closed

thetarnav:
thats even better 🙌

PR #3917, opened 2024-07-14
[os2/file]: Fix file handle inheritance issues on Windows

flysand7:
On windows when a file was created using `os2` functions, the `.Close_On_Exec` flag was forced onto the handle, even if it wasn't set by the user, forcing all handles to be non-inheritable. This PR makes all handles inheritable by default (same behavior as on linux), and fixes the bug where `SECURITY_ATTRIBUTES` struct was being used after free. Somehow after the scope ends, it was observied in the debugger that the pointer to the struct was being reset to `nil`, making it a valid parameter to `CreateFile`.


flysand7:
Wrong branch

PR #3938, opened 2024-07-17
Fix memory leak in encoding/ini

Ronaldr1985:
A simple change that fixes a memory leak caused by not deleting all the values in the ini.map.  The type that is returned when parsing an ini file/string.

### DIFF
--- a/core/encoding/ini/ini.odin
+++ b/core/encoding/ini/ini.odin
@@ -92,7 +92,6 @@ load_map_from_string :: proc(src: string, allocator: runtime.Allocator, options 
 			}
 		}
 		return strings.clone(val)
-
 	}
 
 	context.allocator = allocator
@@ -114,7 +113,10 @@ load_map_from_string :: proc(src: string, allocator: runtime.Allocator, options 
 			new_key = strings.to_lower(key) or_return
 			delete(old_key) or_return
 		}
-		pairs[new_key] = unquote(value) or_return
+		pairs[new_key], err = unquote(value)
+		if err != nil {
+			return
+		}
 	}
 	return
 }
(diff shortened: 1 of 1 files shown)


Kelimion:
Thanks. I have two questions:
- Do you have an example of an `.ini` file and usage code that causes it to leak without this fix?
- Are you up to writing `tests/core/encoding/ini`?

The first question here is the most important. I'd like to understand what usage pattern caused it to leak.

Ronaldr1985:
Certainly, here is the example code:
ini.odin:
```odin
package main

import "core:fmt"
import "core:encoding/ini"
import "core:mem"

_main :: proc() {
	ini_map, err, ok := ini.load_map_from_path("config.ini", allocator=context.allocator)
	if err != nil {
		fmt.eprintln("Got an error")
	}
	defer ini.delete_map(ini_map)

	fmt.println("map: ", ini_map["api"])

	for key, value in ini_map {
		fmt.println("key:   ", key)
		fmt.println("value: ", value)
		fmt.println()
	}

	fmt.println("ini_map[\"API\"]:", ini_map["API"])
}

main :: proc() {
	track: mem.Tracking_Allocator

	_main()

	for _, leak in track.allocation_map {
		fmt.printf("%v leaked %v bytes\n", leak.location, leak.size)
	}
	for bad_free in track.bad_free_array {
		fmt.printf("%v allocation %p was freed badly\n", bad_free.location, bad_free.memory)
	}
}
```
config.ini:
```ini
[LOG]
LEVEL = "DEBUG"

[API]
ADDRESS = "HTTP://127.0.0.1:8000"
```

Yes I will try and write the test, should the test be included in this PR?

Kelimion:
> Yes I will try and write the test, should the test be included in this PR?

Its own PR is fine. Thank you for the bug report and fix.

PR #3951, opened 2024-07-20
Add Lua example to `vendor:lua` as well as basic tests.

Kelimion:


Kelimion:
Peculiar. Runs fine twice, then wrong argument and then it crashes `vendor.exe` (the test) and the terminal window along with it.

image

PR #3971, opened 2024-07-23
os2 process linux implementation

jasonKercher:
- Add process implementation.
- Update some incompatibilities in the file_linux.odin
- ~~Add `ctprintf` to internal util.~~


core/os/os2/process_linux.odin
```diff
@@ -71,25 +271,301 @@ _Sys_Process_Attributes :: struct {}
...
+					return process_state, _get_platform_error(errno)
+				}
+				if n == 0 {
+					_process_state_update_times(process, &process_state)
+					return
+				}
+				break
+			}
+		} else {
+			mask: bit_set[0..=63]
```
gingerBill, line 484:
```odin
mask: bit_set[0..<64; u64] // be explicit
```

core/os/os2/process_linux.odin
```diff
@@ -71,25 +271,301 @@ _Sys_Process_Attributes :: struct {}
...
+		return
+	}
+
+	// ')' will be the end of the executable name (item 2)
+	idx := strings.last_index_byte(string(stat_buf), ')')
+	stats := string(stat_buf[idx + 2:])
+
+	// utime and stime are the 14 and 15th items, respectively, and we are
+	// currently on item 3. Skip 11 items here.
+	for i := 0; i < 11; i += 1 {
```
gingerBill, line 412:
```odin
for _ in 0..<11 {
```

core/os/os2/process_linux.odin
```diff
@@ -71,25 +271,301 @@ _Sys_Process_Attributes :: struct {}
...
+	if not_exec, errno = linux.faccessat(dir_fd, executable_path, linux.F_OK | linux.X_OK); errno != .NONE || not_exec {
+		return process, errno == .NONE ? .Permission_Denied : _get_platform_error(errno)
+	}
+
+	// args and environment need to be a list of cstrings
+	// that are terminated by a nil pointer.
+	cargs := make([]cstring, len(desc.command) + 1, temp_allocator())
+	for i := 0; i < len(desc.command); i += 1{
+		cargs[i] = temp_cstring(desc.command[i]) or_return
+	}
```
gingerBill, line 326:
```odin
for command, i in desc.command {
	cargs[i] = temp_cstring(command) or_return
}
```

core/os/os2/process_linux.odin
```diff
@@ -71,25 +271,301 @@ _Sys_Process_Attributes :: struct {}
...
+	// Use current process' environment if description didn't provide it.
+	env: [^]cstring
+	if desc.env == nil {
+		// take this process's current environment
+		env = raw_data(export_cstring_environment(temp_allocator()))
+	} else {
+		cenv := make([]cstring, len(desc.env) + 1, temp_allocator())
+		for i := 0; i < len(desc.env); i += 1 {
+			cenv[i] = temp_cstring(desc.env[i]) or_return
+		}
```
gingerBill, line 337:
```odin
for env, in desc.env {
	cenv[i] = temp_cstring(env) or_return
}
```

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,268 @@
...
+_process_open :: proc(pid: int, _: Process_Open_Flags) -> (process: Process, err: Error) {
+	process.pid = pid
+	process.handle = PIDFD_UNASSIGNED
+	if !_has_pidfd_open {
+		return process, .Unsupported
+	}
+
+	pidfd, errno := linux.pidfd_open(linux.Pid(pid), {})
+	if errno == .ENOSYS {
+		_has_pidfd_open = false
```
gingerBill, line 258: Should this be atomic or `@(thread_local)`? Because it's a global variable that could be accessed from multiple threads.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,268 @@
...
+		if .PPid in selection {
+			ppid_str := stats[strings.index_byte(stats, ' '):]
+			if ppid, ok := strconv.parse_int(ppid_str); ok {
+				info.ppid = ppid
+			}
+		}
+
+		if .Priority in selection {
+			// On 4th field. Priority is field 18 and niceness is field 19.
+			for i := 4; i < 19; i += 1 {
```
gingerBill, line 177:
```odin
for _ in 4..<19 {
```

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,268 @@
...
 @(private="package")
 _get_ppid :: proc() -> int {
-	return -1
+	return int(linux.getppid())
 }
 
 @(private="package")
-_process_list :: proc(allocator: runtime.Allocator) -> (list: []int, err: Error) {
-	return
+_process_list :: proc(allocator: runtime.Allocator) -> ([]int, Error) {
```
gingerBill, line 55:
```odin
-> (list: []int, err: Error)
```
This will be needed for the following change

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,268 @@
...
+	errno:  linux.Errno
+	#partial switch dir_fd, errno = linux.open("/proc/", _OPENDIR_FLAGS); errno {
+	case .ENOTDIR:
+		return {}, .Invalid_Dir
+	case .ENOENT:
+		return {}, .Not_Exist
+	}
+	defer linux.close(dir_fd)
+
+	dynamic_list := make([dynamic]int, allocator)
```
gingerBill, line 68:
```odin
defer if err != nil {
	delete(dynamic_list)
}
```

gingerBill, line 68: Change this too:
```odin
dynamic_list := make([dynamic]int, temp_allocator())
```

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,268 @@
...
+			d_name_cstr := cstring(&d.d_name[0])
+			#no_bounds_check d_name_str := string(d.d_name[:len(d_name_cstr)])
+
+			if pid, ok := strconv.parse_int(d_name_str); ok {
+				append(&dynamic_list, pid)
+			}
+		}
+	}
+
+	return dynamic_list[:], nil
```
gingerBill, line 97:
```
return slice.clone(dynamic_list[:], allocator)
```

core/os/os2/file_linux.odin
```diff
@@ -373,17 +393,13 @@ _exists :: proc(name: string) -> bool {
...
- */
+/* For reading Linux system files that stat to size 0 */
 _read_entire_pseudo_file :: proc { _read_entire_pseudo_file_string, _read_entire_pseudo_file_cstring }
 
 _read_entire_pseudo_file_string :: proc(name: string, allocator: runtime.Allocator) -> (b: []u8, e: Error) {
-	name_cstr := clone_to_cstring(name, allocator) or_return
-	defer delete(name, allocator)
+	TEMP_ALLOCATOR_GUARD()
+	name_cstr := clone_to_cstring(name, temp_allocator()) or_return
+	defer delete(name, temp_allocator())
```
gingerBill, line 402: `defer delete` is not needed with `temp_allocator()`

jasonKercher:
I think I got everything.

core/os/os2/process_linux.odin
```diff
@@ -71,25 +307,303 @@ _Sys_Process_Attributes :: struct {}
...
+
 	return
 }
 
 @(private="package")
 _process_wait :: proc(process: Process, timeout: time.Duration) -> (process_state: Process_State, err: Error) {
+	process_state.pid = process.pid
+
+	options: linux.Wait_Options
+	big_if: if timeout == 0 {
```
flysand7, line 473: If timeout < 0, it's considered infinite wait time. Timeout = 0 should return immediately.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,304 @@
...
-	return
+	return _process_info_by_pid(get_pid(), selection, allocator)
 }
 
 @(private="package")
-_process_open :: proc(pid: int, flags: Process_Open_Flags) -> (process: Process, err: Error) {
+_process_open :: proc(pid: int, _: Process_Open_Flags) -> (process: Process, err: Error) {
+	process.pid = pid
+	process.handle = PIDFD_UNASSIGNED
+	if !_has_pidfd_open() {
```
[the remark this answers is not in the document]
jasonKercher: It's also used in `process_wait`.

core/os/os2/process_linux.odin
jasonKercher: That is how it works.  Default wait option causes infinite wait.  If the wait time is < 0, it will never enter the if statement which handles everything else.

core/os/os2/file_linux.odin
```diff
@@ -6,41 +6,32 @@ import "core:time"
 import "base:runtime"
 import "core:sys/linux"
 
+File_Impl_Kind :: enum u8 {
```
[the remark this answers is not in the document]
jasonKercher: I think the general idea for that `enum` was more for future use.  I will just take it out.  Why do we care if `stdin` is a pipe?

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,304 @@
...
+
+	info.fields = selection
+
+	// Use this so we can use bprintf to make cstrings with less copying.
+	// The path_backing is manually zero terminated as we go.
+	path_backing: [48]u8
+
+	path_slice := path_backing[:len(path_backing) - 1]
+	path_cstr  := cstring(&path_slice[0])
+	path_len   := len(fmt.bprintf(path_slice, "/proc/%d", pid))
```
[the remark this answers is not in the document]
jasonKercher: `bprintf` creates a string builder from the provided buffer (`path_slice` in this case).  No context allocations going on here.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,304 @@
...
+			}
+			passwd = passwd[eol + 1:]
+		}
+	}
+
+	cmdline_if: if selection & {.Working_Dir, .Command_Line, .Command_Args, .Executable_Path} != {} {
+		path_len = len(fmt.bprintf(path_slice, "/proc/%d/cmdline", pid))
+		path_backing[path_len] = 0
+
+		cmdline_bytes := _read_entire_pseudo_file(path_cstr, temp_allocator()) or_return
```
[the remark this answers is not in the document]
jasonKercher: I don't know about Windows, but many of these pieces of data have different permissions.  For example, I can get any process command line invocation, but I cannot always get its environment.  So I figured, we would just grab what we can.

core/os/os2/file_linux.odin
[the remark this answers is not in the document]
flysand7: It might be a file, in which case a program could attempt to read it's size. The `enum` was made is already available in the `master` branch and `os2.stat` should output the type of the handle. I'm not sure if your PR has my file_type changes, and if not i'd suggest rebasing off of master and see how I'm doing it right now (in short that information is not stored), however the change where you return a `.No_Size` error is still good.

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
flysand7: Yeah on windows it's a little bit tricky. Some of the information is retrieved from general API without needing extra permissions, other stuff requires opening the process with `QUERY_LIMITED_INFORMATION` and some other stuff required reading that process' PEB block which also required the permission to read that process' memory. I think this matches in both cases. I'll have to make a note to change the way the windows version of the function to just grab whatever I can.

core/os/os2/process_linux.odin
```diff
@@ -71,25 +307,303 @@ _Sys_Process_Attributes :: struct {}
...
+		ts: linux.Time_Spec = {
+			time_sec  = uint(timeout / time.Second),
+			time_nsec = uint(timeout % time.Second),
+		}
+
+		// pidfd_open is fairly new, so don't error out on ENOSYS
+		pid_fd: linux.Pid_FD
+		errno: linux.Errno
+		if _has_pidfd_open() {
+			if process.handle == PIDFD_UNASSIGNED {
```
flysand7, line 485: Shouldn't `process_start` open the pidfd? I feel like once you get the `Process` you automatically should get the pidfd if its available. That's sort of the point of the process handle, as long as you have it, it should refer to the same process. This can probably also simplify the caching code, if not completely remove it.

flysand7 (commented):
a

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
flysand7: I'll move the discussion there then

core/os/os2/process_linux.odin
flysand7: Misread the code, I see what's going on now.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,304 @@
...
+			d_name_cstr := cstring(&d.d_name[0])
+			#no_bounds_check d_name_str := string(d.d_name[:len(d_name_cstr)])
+
+			if pid, ok := strconv.parse_int(d_name_str); ok {
+				append(&dynamic_list, pid)
+			}
+		}
+	}
+
+	return slice.clone(dynamic_list[:], allocator), nil
```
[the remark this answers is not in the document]
jasonKercher: This is done except for a `filepath.join`. Maybe those procs should return `Allocator_Error`.

core/os/os2/process_linux.odin
jasonKercher: Agreed. I think I was worried about waiting on a process that isn't my child, but that should never happen. This removes the need for caching.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,304 @@
...
+			continue loop
+		case .NONE:
+			if buflen == 0 { break loop }
+		case:
+			return {}, _get_platform_error(errno)
+		}
+
+		d: ^dirent64
+
+		for i := 0; i < buflen; i += int(d.d_reclen) {
```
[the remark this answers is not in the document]
jasonKercher: Got this one as well as the others if you want to double check it.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,304 @@
...
+			d = (^dirent64)(rawptr(&buf[i]))
+			d_name_cstr := cstring(&d.d_name[0])
+			#no_bounds_check d_name_str := string(d.d_name[:len(d_name_cstr)])
+
+			if pid, ok := strconv.parse_int(d_name_str); ok {
```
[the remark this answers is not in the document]
flysand7: =]

core/os/os2/file_linux.odin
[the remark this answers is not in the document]
jasonKercher: Oh, now I remember, why I added that.  I forgot to actually implement the other side of it.  It was to prevent non-blocking reads from returning EOF all the time.

core/os/os2/process_linux.odin
```diff
@@ -71,25 +307,303 @@ _Sys_Process_Attributes :: struct {}
...
+		if errno = linux.execveat(dir_fd, executable_path, &cargs[0], env); errno != .NONE {
+			print_error(stderr, _get_platform_error(errno), string(executable_path))
+			panic("execve failed to replace process")
+		}
+		unreachable()
+	}
+
+	process.pid = int(pid)
+	process.handle = PIDFD_UNASSIGNED
+	return
```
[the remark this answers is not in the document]
jasonKercher: I changed this to `exit`, but the more I look at it the more the logic is wrong.   It should be aborting with something like an illegal instruction.  This gives us 2 options:

 - The kernel will get the message out for us no matter what we did to the pipes.  It probably just writes directly to `/dev/stderr` or something.  It is up to the user to handle the state they get back from `process_wait` and `process_start` still returns success.
 - Catch a signal from the child sent via `kill(get_ppid(), ...)`

For the signal path, either we lock and take over a signal temporarily, or we have to take full ownership of an arbitrary signal number that the child can report to.  Both have their complications, so maybe best to just leave it to the user.

What's the best way to hard fault here?  I cannot find an `abort` equivalent that doesn't try to print something.

core/os/os2/file_linux.odin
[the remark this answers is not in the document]
flysand7: I'll have to do some research to know for sure whether there are any better solutions to this problem, but can you provide the setup where returning EOF is a problem?

Something like setting up the pipe and calling `read_entire_file`? You say non-blocking and I'm a little bit confused by that, clearly you mean `O_NONBLOCK` ops, because for now we don't have a way to queue up asynchronous operations, but why are pipes non-blocking? My perception is that this should apply to all kinds of files, potentially, be it a regular file on a network drive, or a named pipe in the filesystem.

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
flysand7: `intrinsics.ud2`. I think it's called something else though, but basically on x86 it should produce a UD2 instruction which will be interpreted in a debugger as an invalid instruction, pretty telling for something that goes "expectedly wrong". It might be `intrinsics.trap()`, or `debug_trap()`, but honestly I don't remember, it's one of those low-level procedures.

As for the error handling, I do have a couple of ideas, but none of them are good.

I think one of the ways to resolve it is to use user-defined signals `SIG_USR+n`, for custom error codes. The child would send one of these to its parent and abort. Then it would send a cross-process message saying that it's about to be started. The parent would wait on that message after calling `fork`, and after the wait it would examine the status of the child process and return the according status. We can guarantee to be able to do this upto the point of `execve`, but we won't be able to handle `execve` errors (because we cant synchronize the message in the case of success, at that point the code is out of our control).

The issues with this approach are:

1. It's intrusive. We're taking up user signals for error handling purposes
2. We can't handle errors returned by execve

The first issue is probably not an issue, because `os2` doesn't provide signals to users, but it can be a problem in an application that uses both `os2` and system-specific functions. Nevertheless, it can be resolved by using literally any other IPC mechanism. I'm not sure how to communicate a child process an inherited pipe, but I think communicating error codes via an extra pipe that will be closed right before execve is not a bad choice.

The second issue is mainly an issue because normally execve would return an error that says that a file can't be found. This is trivially fixed by using `fexecve` and opening file descriptor as a separate operation.

`fexecve` could still return us an error, and here we give up and just debug trap our process. I think that's the best error handling that we can afford.

I'd need to look into how other similar programming languages handle these errors. Particularly curious about zig, because it's also a type of language that uses raw syscalls in their standard library.

core/os/os2/file_linux.odin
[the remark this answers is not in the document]
jasonKercher: Indeed, I was thinking about `read_entire_file`.  I have personally used it after modifying termios, so that is sort of outside the scope of core anyway.  Is it really a problem?  No, in practice, it can always be mitigated.

core/os/os2/process_linux.odin
```diff
@@ -148,10 +136,15 @@ _process_info_by_pid :: proc(pid: int, selection: Process_Info_Fields, allocator
...
 			n = strings.index_byte(passwd, ':')
-			if uid, ok := strconv.parse_int(passwd[:n]); ok && uid == int(s.uid) {
-				info.username = strings.clone(username, allocator)
+			uid: int
+			ok: bool
+			if uid, ok = strconv.parse_int(passwd[:n]); ok && uid == int(s.uid) {
+				info.username = strings.clone(username, allocator) or_return
 				break
 			}
+			if !ok {
```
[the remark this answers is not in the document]
jasonKercher: Good point. Fixed.

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
jasonKercher: That's simple enough.  I like the idea of a temporary pipe instead of playing with signals.   `fexecve` is libc.  I'm curious how they implement that..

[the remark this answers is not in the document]
jasonKercher: WHAT!?  `execveat` can act like an `fexecve` if the pathname is blank an is provided `AT_EMPTY_PATH`.  I didin't know that...

core/os/os2/process_linux.odin
```diff
@@ -71,25 +300,295 @@ _Sys_Process_Attributes :: struct {}
...
+		return process, _get_platform_error(errno)
+	}
+
+	STDIN  :: linux.Fd(0)
+	STDOUT :: linux.Fd(1)
+	STDERR :: linux.Fd(2)
+
+	if pid == 0 {
+		// in child process now
+		if desc.stdin != nil {
```
laytan, line 393: > 	// The `stderr` handle to give to the child process. It can be either a file
	// or a writeable end of a pipe. Passing `nil` will shut down the process'
	// stderr output.

According to that doc comment, passing nil should not inherit, but close.

Probably want to either dup2 to the given file or to /dev/null if it is nil.

flysand7: Oh yeah this might have been due to copying windows implementation. On windows if the handle isn't valid that would shut down process' output (equivalent of piping to /dev/null). That said, I haven't figured out what to do with /dev/stdin though. I probably copied the docs and said something along the lines of "it shuts down the input", but what exactly that means I haven't thought about. Piping /dev/zero to it sounds silly.

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
flysand7: Here's how zig does it: https://github.com/ziglang/zig/blob/master/lib/std/process/Child.zig#L646

My intuition about setting up a separate pipe was correct

core/os/os2/process_linux.odin
laytan: Afaik you can also use dev/null for stdin and there will just not be anything to read which seems correct.

flysand7: Oh makes sense!

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
jasonKercher: That's interesting.  They hold onto it and read it during the wait routine.  In my head, I thought pipe on the stack, child sends a single byte status, and parent spins waiting for it before leaving `process_start`.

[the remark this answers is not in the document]
flysand7: There are still these funny little issues in sys/linux that are going to take a few months to fix all of them. That's kind of how manual binding works :)

jasonKercher:
Okay @laytan @flysand, *now* it's ready.  Please take a look when you have time.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,296 @@
...
+	TEMP_ALLOCATOR_GUARD()
+
+	dir_fd: linux.Fd
+	errno:  linux.Errno
+	#partial switch dir_fd, errno = linux.open("/proc/", _OPENDIR_FLAGS); errno {
+	case .ENOTDIR:
+		return {}, .Invalid_Dir
+	case .ENOENT:
+		return {}, .Not_Exist
+	}
```
flysand7, line 64: Other errors?

core/os/os2/process_linux.odin
```diff
@@ -71,25 +299,439 @@ _Sys_Process_Attributes :: struct {}
...
+		}
+		if _, errno = linux.dup2(stderr_fd, STDERR); errno != .NONE {
+			write_errno_to_parent_and_abort(child_pipe_fds[WRITE], errno)
+		}
+
+		success_byte: [1]u8
+		linux.write(child_pipe_fds[WRITE], success_byte[:])
+
+		if errno = linux.execveat(exe_fd, "", &cargs[0], env, {.AT_EMPTY_PATH}); errno != .NONE {
+			write_errno_to_parent_and_abort(child_pipe_fds[WRITE], errno)
```
flysand7, line 479: OMG i literally just realised that's what `CLOEXEC` is for! I thought there would be no way for a parent to know that the child has successfully summoned the child process, but turns out there is! They closed the pipe without writing anything to it.

Thank you for figuring this out, it took me some time to finally understand this.

core/os/os2/process_linux.odin
```diff
@@ -71,25 +299,439 @@ _Sys_Process_Attributes :: struct {}
...
+	}
+
+	linux.close(child_pipe_fds[WRITE])
+	defer linux.close(child_pipe_fds[WRITE])
+
+	n: int
+	child_byte: [1]u8
+	n, errno = linux.read(child_pipe_fds[READ], child_byte[:])
+	if errno != .NONE {
+		return process, _get_platform_error(errno)
```
flysand7, line 491: Can't this throw BROKEN_PIPE?

I thought that if the child is started successfully, the pipe will be closed and errno set to broken pipe.

flysand7 (commented):
Other than a question and a small issue with error handling, everything else looks good to merge.

core/os/os2/process_linux.odin
jasonKercher: Ha, that was your idea.  I just wrote it.

core/os/os2/process_linux.odin
jasonKercher: Good catch

core/os/os2/process_linux.odin
jasonKercher: I don't think it can without it being terminated outside of our control.  Just to verify, I added a 1 second busy loop right before this read and ran it through a set of tests that mostly run programs that complete near instantly and it still reads just fine.  The thinking here was that everything is still valid until the process is reaped.  Now... for the life of me.. I cannot find documentation that says that anywhere.

But, I'm glad you pointed this out.  The error handling doesn't wait on the process and leaks the PID.

jasonKercher: Check out the update then.. I just stopped trying to handle the read error all together.  If the read fails, something extraordinary happened, but the user now has a process to wait on.  They can figure it out on that end.

flysand7 (commented):
I guess this looks good now

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,299 @@
...
+
+		// We are now on the 3rd field (skip)
+		stats = stats[strings.index_byte(stats, ' ') + 1:]
+
+		if .PPid in selection {
+			ppid_str := stats[:strings.index_byte(stats, ' ')]
+			if ppid, ok := strconv.parse_int(ppid_str); ok {
+				info.ppid = ppid
+			} else {
+				return info, .Invalid_File
```
laytan, line 232: I assume this only happens when it is an invalid thing? If it isn't then you might want to set the error and break to allow retrieval of other fields.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,299 @@
...
+		if .Priority in selection {
+			// On 4th field. Priority is field 18 and niceness is field 19.
+			for _ in 4..<19 {
+				stats = stats[strings.index_byte(stats, ' ') + 1:]
+			}
+			nice_str := stats[:strings.index_byte(stats, ' ')]
+			if nice, ok := strconv.parse_int(nice_str); ok {
+				info.priority = nice
+			} else {
+				return info, .Invalid_File
```
laytan, line 245: I assume this only happens when it is an invalid thing? If it isn't then you might want to set the error and break to allow retrieval of other fields.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,299 @@
...
+			}
+		}
+	}
+
+	if .Environment in selection {
+		path_len = len(fmt.bprintf(path_slice, "/proc/%d/environ", pid))
+		path_backing[path_len] = 0
+
+		if env_bytes, env_err := _read_entire_pseudo_file(path_cstr, temp_allocator()); env_err == nil {
```
laytan, line 254: If there is an error it should still be set and given to the user imo

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+	has_executable_permissions :: proc(fd: linux.Fd) -> bool {
+		backing: [48]u8
+		_ = fmt.bprintf(backing[:], "/proc/self/fd/%d", fd)
+		return linux.access(cstring(&backing[0]), linux.X_OK) == .NONE
+	}
+
+	TEMP_ALLOCATOR_GUARD()
+
+	if len(desc.command) == 0 {
+		return process, .Invalid_File
```
laytan, line 314: `.Invalid_File` doesn't really fit this condition, possibly need a new error here like `.Invalid_Command`

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+	// search PATH if just a plain name is provided
+	exe_fd: linux.Fd
+	executable_name := desc.command[0]
+	if strings.index_byte(executable_name, '/') == -1 {
+		path_env := get_env("PATH", temp_allocator())
+		path_dirs := filepath.split_list(path_env, temp_allocator())
+
+		found: bool
+		for dir in path_dirs {
+			exe_path := fmt.caprintf("%s/%s", dir, executable_name, allocator=temp_allocator())
```
laytan, line 338: Can't use fmt in this package

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+			if !has_executable_permissions(exe_fd) {
+				linux.close(exe_fd)
+				continue
+			}
+			found = true
+			break
+		}
+		if !found {
+			// check in cwd to match windows behavior
+			exe_path := fmt.caprintf("./%s", executable_name, allocator=temp_allocator())
```
laytan, line 351: Can't use fmt in this package

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+	}
+
+	STDIN  :: linux.Fd(0)
+	STDOUT :: linux.Fd(1)
+	STDERR :: linux.Fd(2)
+
+	READ :: 0
+	WRITE :: 1
+
+	if pid == 0 {
```
laytan, line 423: Should the child not close the read side of the pipe here? `linux.close(child_pipe_fds[READ])`

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
 }
 
-@(private="package")
-_process_wait :: proc(process: Process, timeout: time.Duration) -> (process_state: Process_State, err: Error) {
+_process_state_update_times :: proc(state: ^Process_State) -> (err: Error) {
+	TEMP_ALLOCATOR_GUARD()
+
+	stat_path_buf: [32]u8
+	_ = fmt.bprintf(stat_path_buf[:], "/proc/%d/stat", state.pid)
+	stat_buf: []u8
```
laytan, line 520: Any reason this isn't using `or_return`?

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+	stat_buf, err = _read_entire_pseudo_file(cstring(&stat_path_buf[0]), temp_allocator())
+	if err != nil {
+		return
+	}
+
+	// ')' will be the end of the executable name (item 2)
+	idx := strings.last_index_byte(string(stat_buf), ')')
```
laytan, line 527: Similar comment to the one before, but you are assuming a lot of things about the buffer here, I wonder if you should do some checks.

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+	utime, stime: int
+	ok: bool
+	if utime, ok = strconv.parse_int(utime_str, 10); !ok {
+		return .Invalid_File
+	}
+	if stime, ok = strconv.parse_int(stime_str, 10); !ok {
+		return .Invalid_File
+	}
+
+	// NOTE: Assuming HZ of 100, 1 jiffy == 10 ms
```
laytan, line 551: Is there no syscall like `sysconf` (which is libc I know) to get the HZ? Or is assuming 100 basically always correct?

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+			}
+			return process_state, _get_platform_error(errno)
+		}
+
+		if n == 0 {  // timeout with no events
+			_process_state_update_times(&process_state)
+			return
+		}
+
+		// This throws EBADF with pidfd
```
laytan, line 630: > // This throws EBADF with pidfd

Really, `linux.waitid(.PIDFD, linux.Id(pidfd), ...)` does that? Did you not forget to change `.PID` to `.PIDFD`?

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+		if errno = linux.waitid(.PID, linux.Id(process.pid), &info, {.WEXITED, .WNOHANG, .WNOWAIT}, nil); errno != .NONE {
```
laytan, line 631: Any particular reason for `.WNOWAIT`, are you expecting the user to wait on it again later?

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+			return process_state, _get_platform_error(errno)
+		}
+
+		if info.signo == .SIGCHLD {
+			break
+		}
+
+		new_tick := time.tick_now()
+		timeout -= time.tick_diff(start_tick, new_tick)
```
laytan, line 640: Can use `time.tick_since(start_tick)` here

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+		ts: linux.Time_Spec = {
+			time_sec  = uint(timeout / time.Second),
+			time_nsec = uint(timeout % time.Second),
+		}
+
+		n, errno := linux.ppoll(pollfd[:], &ts, &sigchld_set)
+		if errno != .NONE {
+			if errno == .EINTR {
+				new_tick := time.tick_now()
+				timeout -= time.tick_diff(start_tick, new_tick)
```
laytan, line 618: Can use `time.tick_since(start_tick)` here

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+		}
+
+		_, errno = linux.rt_sigtimedwait(&sigchld_set, &info, &ts)
+		#partial switch errno {
+		case .EAGAIN:   // timeout
+			_process_state_update_times(&process_state)
+			return
+		case .EINTR:
+			new_tick := time.tick_now()
+			timeout -= time.tick_diff(start_tick, new_tick)
```
laytan, line 687: Can use time.tick_since(start_tick) here

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+	options: linux.Wait_Options = {.WEXITED}
+	if timeout == 0 {
+		options += {.WNOHANG}
+	}
+
+	info: linux.Sig_Info
+
+	errno = .EINTR
+	for errno == .EINTR {
+		errno = linux.waitid(.PID, linux.Id(process.pid), &info, options + {.WNOWAIT}, nil)
```
laytan, line 721: Same comment as above about `.WNOWAIT`, and if `.PIDFD` does work (like I asked above) I guess it makes sense to use that if it isn't `PIDFD_UNASSIGNED`

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+	}
+	if errno == .EAGAIN || (errno == .NONE && info.signo != .SIGCHLD) {
+		_process_state_update_times(&process_state)
+		return process_state, nil
```
laytan, line 725: Shouldn't this return the `.Timeout` if `errno == .EAGAIN`? I don't see any specific handling for 0 second timeouts in the windows implementation.

core/os/os2/process_linux.odin
```diff
@@ -71,25 +302,446 @@ _Sys_Process_Attributes :: struct {}
...
+	}
+
+	return _reap_terminated(process)
+}
+
 @(private="package")
 _process_close :: proc(process: Process) -> Error {
-	return nil
+	pidfd := linux.Fd(process.handle)
+	if pidfd < 0 {
```
laytan, line 737: Should this also check for `PIDFD_UNASSIGNED`? Or does casting that to `Fd` make it negative?

core/os/os2/file_linux.odin
```diff
@@ -54,10 +39,33 @@ _stderr := File{
 
 @init
 _standard_stream_init :: proc() {
```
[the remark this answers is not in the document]
jasonKercher: Let me see if I can get it to work. This is indeed better, but it was hitting an assert in the compiler for me.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,299 @@
...
+				info.working_dir = strings.clone(cwd, allocator) or_return
+			}
+		}
+
+		if .Executable_Path in selection {
+			if cmdline[0] == '/' {
+				info.executable_path = strings.clone(cmdline[:terminator], allocator) or_return
+			} else if cwd_err == nil {
+				join_paths: [2]string = { cwd, cmdline[:terminator] }
+				info.executable_path = filepath.join(join_paths[:], allocator)
```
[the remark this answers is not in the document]
jasonKercher: core/path/filepath stuff doesn't return `Allocator_Error`.  I can add it in and make it optional if you want.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,299 @@
...
+			}
+		}
+		if .Command_Line in selection {
+			info.command_line = strings.clone(cmdline[:terminator], allocator) or_return
```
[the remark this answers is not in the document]
jasonKercher: It is indeed the former.  So, I will make it `command_args` is `args[1:]`.  `command_line` is `args[:]`.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,299 @@
...
+		for d in linux.dirent_iterate_buf(buf[:buflen], &offset) {
+			d_name_str := linux.dirent_name(d)
+
+			if pid, ok := strconv.parse_int(d_name_str); ok {
+				append(&dynamic_list, pid)
+			}
+		}
+	}
+
+	list, err = slice.clone(dynamic_list[:], allocator)
```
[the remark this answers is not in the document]
jasonKercher: I believe this was Bill's suggestion unless I read it wrong =]

flysand7:
Oh also github CI doesn't do any checks for `os2`, make sure you run `odin check core/os/os2 -vet -strict-style` locally.

Just making sure

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,299 @@
...
+				info.executable_path = filepath.join(join_paths[:], allocator)
+			}
+		}
+		if .Command_Line in selection {
+			info.command_line = strings.clone(cmdline[:terminator], allocator) or_return
+		}
+
+		if .Command_Args in selection {
+			// skip to first arg
+			cmdline = cmdline[terminator + 1:]
```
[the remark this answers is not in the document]
jasonKercher: This is a good catch.

core/os/os2/process_linux.odin
```diff
@@ -1,68 +1,299 @@
...
+	stat_if: if selection & {.PPid, .Priority} != {} {
+		path_len = len(fmt.bprintf(path_slice, "/proc/%d/stat", pid))
+		path_backing[path_len] = 0
+
+		proc_stat_bytes := _read_entire_pseudo_file(path_cstr, temp_allocator()) or_return
+		if len(proc_stat_bytes) <= 0 {
+			break stat_if
+		}
+
+		start := strings.last_index_byte(string(proc_stat_bytes), ')')
```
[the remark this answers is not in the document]
jasonKercher: Generally, this is all very well defined and can be assumed.  However, that doesn't rule out extraordinary things happening.  I will add checks.

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
laytan: Oh, alright, was a nitpick anyway

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
laytan: Oh, adding an optional error is a good idea

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
laytan: I figured it was pretty much guaranteed to be valid, just some assertions that make readers see that's the case would already be great imo

core/os/os2/file_linux.odin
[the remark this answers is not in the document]
laytan: Ah I think I only checked if it got through `odin check` and that crash happens in the backend which isn't called with just `odin check`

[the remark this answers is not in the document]
laytan: Also, I just realised that this is taking a bunch of pointers to the stack and assigning it, not sure how @(init) /globals interact with that but I don't think it is right, let's just keep it how you got it, sorry about this.

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
jasonKercher: Maybe this is the wrong approach, but to get the list of PIDs, I'm just going through the list of files in `/proc/`.  That includes non-numeric files with names like `cpuinfo`.  I'm using the failure of `parse_int` to say, "this isn't a process."

core/os/os2/process_linux.odin
jasonKercher: Your assumption is correct, but that was certainly cryptic to write it like that.  I'll just make it `if process.handle == PIDFD_UNASSIGNED`.

jasonKercher:
> Oh also github CI doesn't do any checks for `os2`, make sure you run `odin check core/os/os2 -vet -strict-style` locally.
> 
> Just making sure

+ `-vet-tabs`

core/os/os2/process_linux.odin
jasonKercher: That pipe is opened with `CLOEXEC`.  In the child, they will not exist passed the `execve`.

core/os/os2/process_linux.odin
jasonKercher: `bprintf :: proc(...) -> string`

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
laytan: Alright, that's a fine approach I think

core/os/os2/process_linux.odin
jasonKercher: So, I tried this again, and it just worked.  I must have been doing something incredibly stupid before.

core/os/os2/process_linux.odin
laytan: Alright let's just assume 100hz then

core/os/os2/process_linux.odin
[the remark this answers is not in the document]
jasonKercher: To make sure I understand correctly, you want me to add optional `Allocator_Error` to the `filepath` procs?  Adding `Allocator_Error` has a sort of contagious nature to it.  For example, adding it to `join` means adding it to `clean` -> `from_slash` -> `strings.replace_all` -> `strings.replace`.  None of which currently returns `Allocator_Error`.  I don't mind doing that by the way, but I wanted to make sure we are on the same page, first.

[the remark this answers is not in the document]
laytan: I think I'm fine with the contagious nature of adding them, although if you want to keep this pr a bit more local I also understand. Maybe as a compromise you can check if the returned string is empty and assume allocation failure.

PR #3972, opened 2024-07-23
testing: add json reporting

laytan:
The fancy and non-fancy output of the new test runner is hard/impossible to parse for automation, this adds a `-define:ODIN_TEST_JSON_REPORT=<path>` that outputs an easy to parse report of all the tests and if they succeeded.

@Feoramund please also review / give your thoughts :)

### DIFF
--- a/core/testing/runner.odin
+++ b/core/testing/runner.odin
@@ -6,6 +6,7 @@ import "base:runtime"
 import "core:bytes"
 import "core:encoding/ansi"
 @require import "core:encoding/base64"
+@require import "core:encoding/json"
 import "core:fmt"
 import "core:io"
 @require import pkg_log "core:log"
@@ -44,7 +45,8 @@ SHARED_RANDOM_SEED    : u64    : #config(ODIN_TEST_RANDOM_SEED, 0)
 LOG_LEVEL             : string : #config(ODIN_TEST_LOG_LEVEL, "info")
 // Show only the most necessary logging information.
 USING_SHORT_LOGS      : bool   : #config(ODIN_TEST_SHORT_LOGS, false)
-
+// Output a report of the tests to the given path.
+JSON_REPORT           : string : #config(ODIN_TEST_JSON_REPORT, "")
 
 get_log_level :: #force_inline proc() -> runtime.Logger_Level {
 	when ODIN_DEBUG {
@@ -59,6 +61,18 @@ get_log_level :: #force_inline proc() -> runtime.Logger_Level {
 			#panic("Unknown `ODIN_TEST_LOG_LEVEL`: \"" + LOG_LEVEL + "\", possible levels are: \"debug\", \"info\", \"warning\", \"error\", or \"fatal\".")
 		}
 	}
+}
+
+JSON :: struct {
+	total:    int,
+	success:  int,
+	duration: time.Duration,
+	packages: map[string][dynamic]JSON_Test,
+}
+
+JSON_Test :: struct {
+	success: bool,
+	name:    string,
 }
 
 end_t :: proc(t: ^T) {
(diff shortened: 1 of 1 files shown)


core/testing/runner.odin
```diff
@@ -61,6 +63,18 @@ get_log_level :: #force_inline proc() -> runtime.Logger_Level {
 	}
 }
 
+JSON :: struct {
+	total:    int,
```
Feoramund, line 67: I would have given this a more descriptive name, like `JSON_Report` and changed the config to `EXPORTING_JSON`.

core/testing/runner.odin
```diff
@@ -61,6 +63,18 @@ get_log_level :: #force_inline proc() -> runtime.Logger_Level {
...
+	success:  int,
+	duration: time.Duration,
+	packages: map[string][dynamic]JSON_Test,
+}
+
+JSON_Test :: struct {
+	success: bool,
+	name:    string,
+}
+
```
Feoramund, line 77: Because these two structs are never used outside of their particular `when` statement below, I would've also put them inside the `when` block itself. They're never used anywhere else in the program, and it's one less thing the compiler has to be concerned about. More importantly, it fences off all the possible places they could be used and makes understanding the scope a little easier.

core/testing/runner.odin
```diff
@@ -847,5 +861,35 @@ To partly mitigate this, redirect STDERR to a file or use the -define:ODIN_TEST_
...
 	fmt.wprintln(stderr, bytes.buffer_to_string(&batch_buffer))
 
+	when JSON_REPORT != "" {
+		json_report: JSON
+
+		mode: int
+		when ODIN_OS != .Windows {
+			mode = os.S_IRUSR|os.S_IWUSR|os.S_IRGRP|os.S_IROTH
+		}
+		json_fd, errno := os.open(JSON_REPORT, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, mode)
```
Feoramund, line 871: If I recall correctly, Windows already ignores the permissions argument for `open`.

core/testing/runner.odin
```diff
@@ -44,7 +45,8 @@ SHARED_RANDOM_SEED    : u64    : #config(ODIN_TEST_RANDOM_SEED, 0)
 LOG_LEVEL             : string : #config(ODIN_TEST_LOG_LEVEL, "info")
 // Show only the most necessary logging information.
 USING_SHORT_LOGS      : bool   : #config(ODIN_TEST_SHORT_LOGS, false)
-
+// Output a report of the tests to the given path.
+JSON_REPORT           : string : #config(ODIN_TEST_JSON_REPORT, "")
```
Feoramund, line 49: When trying this, my first instinct was to write `-define:ODIN_TEST_JSON_REPORT=true`. I'm trying to think of what might be a less ambiguous config name without being too long. `ODIN_TEST_SAVE_JSON_REPORT_TO`? The open preposition hints at what sort of argument it takes.

Feoramund (commented):
You're correct in that neither of the output modes are good for machine parsing; they're not intended to. One is just animated, and the other is not.

I tested the code, and both it and the output look fine.

The following is more of a higher-level criticism and my only major issue: if the intent is for machine parsing, I would've suggested a non-human readable format that benefits from the lack of that quality. JSON has a high size-to-information ratio, and the human-readable aspect of it becomes worthless outside of debugging in a plain text editor, except for in the domains where it is used as a markup language (as opposed to `ini`, `toml`, `yaml`, et cetera). I know JSON is ubiquitous, like JavaScript, but I'm generally opposed to its further adoption where there are other formats that better solve the issue.

That aside, I do like the idea of having some way of outputting the results in a way that can be easily parsed by machine, and I think this is a good, non-invasive way to do it.

Issue #3970, opened 2024-07-23
Compiler crashes when building with spall


laughsuggestion:
## Context

```
package main

import "base:runtime"
import spall "core:prof/spall"

import "core:fmt"
spall_ctx: spall.Context
spall_buffer: spall.Buffer

@(instrumentation_enter)
spall_enter :: proc "contextless" (
    proc_address, call_site_return_address: rawptr,
    loc: runtime.Source_Code_Location,
) {
    spall._buffer_begin(&spall_ctx, &spall_buffer, "", "", loc)
}

@(instrumentation_exit)
spall_exit :: proc "contextless" (
    proc_address, call_site_return_address: rawptr,
    loc: runtime.Source_Code_Location,
) {
    spall._buffer_end(&spall_ctx, &spall_buffer)
}

main :: proc() {
    fmt.printfln("hello, world!")
}
```

```
Odin:    dev-2024-07:527c0b320
	OS:      macOS Sonoma 14.5 (build: 23F79, kernel: 23.5.0)
	CPU:     Apple M2 Pro
	RAM:     16384 MiB
	Backend: LLVM 14.0.6
```

## Expected Behavior

Running "odin build" should build successfully.

## Current Behavior

The compiler crashes.

## Failure Information (for bugs)

The compiler crashes and outputs an error.

### Steps to Reproduce

1. Run: odin build src/

### Failure Logs

`35747 bus error  odin build src/`

laytan:
Unfortunately, I can't reproduce this on:
```
Where to find more information and get into contact when you encounter a bug:

	Website: https://odin-lang.org
	GitHub:  https://github.com/odin-lang/Odin/issues


Useful information to add to a bug report:

	Odin:    dev-2024-07:f9ef951b2
	OS:      macOS Sonoma 14.5 (build: 23F79, kernel: 23.5.0)
	CPU:     Apple M1
	RAM:     8192 MiB
	Backend: LLVM 18.1.4
```

Maybe it is the LLVM version difference, let me check.

laytan:
Ah yep, that's it, I reproduce this on LLVM 14.

laytan:
Ok the reason for the bus error was an error message over 4kb not fitting on the stack buffer used for printing it, here is the first snippet of that error message:
```
LLVM Error:
Call parameter type does not match function signature!
i8 (<{ i64, i64 }>, <{ i64, i64 }>)* @runtime.string_eq
[...]

PR #3974, opened 2024-07-24
compiler: fix formatting of large messages

laytan:
Was looking at #3970 which crashes because it tries to print a big error message, looks like that code was just extremely busted? Here is an attempt at fixing, after this the actual error for that issue is showing (although truncated).


laytan:
I realise now this might need to return -1 when it went out of bounds to let callers deal with the fact it did (instead of truncating). Will look at that later.

laytan:
Kind off fixed in https://github.com/odin-lang/Odin/commit/a219da14ce8bbd15fec4ea962fd0ec175da0f0aa

PR #3985, opened 2024-07-27
[sync]: Document all procedures

flysand7:
This PR implements documentation for every single one of the synchronization primitives in the `core:sync` package, and all atomic functions. Tried to be as specific as possible and introduce general terminology, words like "critical section" in the package documentation.

Some of the terminology was changed to make the flow of text easier. For example a lock on mutex is said to be "acquired" and "released", when lock and unlock operations are performed. In RW mutex, the two kinds of locks are an "exclusive" lock and a "shared" lock (not read and write). I tried using the read/write terminology but that made the text needlessly complicated and I wasn't sure how to properly form sentences with it in a way that's not confusing upon skimming. "write lock of read-write mutex" can be a bit confusing and can require a re-read of the sentence to understand.

This PR also attempts to fix an issue with `park_with_timeout` procedure, where futex isn't checked for spurious wakeups. The futex_wait call was put into a loop and the remaining calculation is calculated by subtracting the difference in monotonic time between calls to `futex_wait()`.

The documentation in the `extended.odin` is a bit on the weaker side. Some of the synchronization primitives implemented there are such that I'm not personally familiar with so describing them in detail was hard.

### DIFF
--- a/core/sync/atomic.odin
+++ b/core/sync/atomic.odin
@@ -2,44 +2,452 @@ package sync
 
 import "base:intrinsics"
 
+/*
+This procedure may lower CPU consumption or yield to a hyperthreaded twin
+processor. It's exact function is architecture specific, but the intent is to
+say that you're not doing much on a CPU.
+*/
 cpu_relax :: intrinsics.cpu_relax
 
 /*
-Atomic_Memory_Order :: enum {
-	Relaxed = 0, // Unordered
-	Consume = 1, // Monotonic
-	Acquire = 2,
-	Release = 3,
-	Acq_Rel = 4,
-	Seq_Cst = 5,
-}
+Describes memory ordering for an atomic operation.
+
+Modern CPU's contain multiple cores and caches specific to those cores. When a
+core performs a write to memory, the value is written to cache first. The issue
+is that a core doesn't typically see what's inside the caches of other cores.
+In order to make operations consistent CPU's implement mechanisms that
+synchronize memory operations across cores by asking other cores or by
+pushing data about writes to other cores.
+
+Due to how these algorithms are implemented, the stores and loads performed by
+one core may seem to happen in a different order to another core. It also may
+happen that a core reorders stores and loads (independent of how compiler put
+them into the machine code). This can cause issues when trying to synchronize
+multiple memory locations between two cores. Which is why CPU's allow for
+stronger memory ordering guarantees if certain instructions or instruction
+variants are used.
+
+In Odin there are 5 different memory ordering guaranties that can be provided
+to an atomic operation:
+
+- `Relaxed`: The memory access (load or store) is unordered with respect to
+  other memory accesses. This can be used to implement an atomic counter.
+  Multiple threads access a single variable, but it doesn't matter when
+  exactly it gets incremented, because it will become eventually consistent.
+- `Consume`: No loads or stores dependent on a memory location can be
+  reordered before a load with consume memory order. If other threads released
+  the same memory, it becomes visible.
+- `Acquire`: No loads or stores on a memory location can be reordered before a
+  load of that memory location with acquire memory ordering. If other threads
+  release the same memory, it becomes visible.
+- `Release`: No loads or stores on a memory location can be reordered after a
+  store of that memory location with release memory ordering. All threads that
+  acquire the same memory location will see all writes done by the current
+  thread.
+- `Acq_Rel`: Acquire-release memory ordering: combines acquire and release
+  memory orderings in the same operation.
+- `Seq_Cst`: Sequential consistency. The strongest memory ordering. A load will
+  always be an acquire operation, a store will always be a release operation,
+  and in addition to that all threads observe the same order of writes.
+
+Non-explicit atomics will always be sequentially consistent.
+
+	Atomic_Memory_Order :: enum {
+		Relaxed = 0, // Unordered
+		Consume = 1, // Monotonic
+		Acquire = 2,
+		Release = 3,
+		Acq_Rel = 4,
+		Seq_Cst = 5,
+	}
+
+**Note(i386, x64)**: x86 has a very strong memory model by default. It
+guarantees that all writes are ordered, stores and loads aren't reordered. In
+a sense, all operations are at least acquire and release operations. If `lock`
+prefix is used, all operations are sequentially consistent. If you use explicit
+atomics, make sure you have the correct atomic memory order, because bugs likely
+will not show up in x86, but may show up on e.g. arm. More on x86 memory
+ordering can be found
+[[here; https://www.cs.cmu.edu/~410-f10/doc/Intel_Reordering_318147.pdf]]
 */
 Atomic_Memory_Order :: intrinsics.Atomic_Memory_Order
 
+/*
+Establish memory ordering.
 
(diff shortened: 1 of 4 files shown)


core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
+multiple memory locations between two cores. Which is why CPU's allow for
+stronger memory ordering guarantees if certain instructions or instruction
+variants are used.
+
+In Odin there are 5 different memory ordering guaranties that can be provided
+to an atomic operation:
+
+- `Relaxed`: The memory access (load or store) is unordered with respect to
+  other memory accesses. This can be used to implement an atomic counter.
+  Multiple threads access a single variable, but it doesn't maatter when
```
Kelimion, line 35: maatter -> matter

flysand7:
Im planning to go over the docstings tomorrow, so for now maybe don't merge until sunday (approx. US time)

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
 
 import "base:intrinsics"
 
+/*
+This procedure may lower CPU consumption or yield to a hyperthreaded twin
+processor. It's exact function is architecture specific, but the intent is to
+say that you're not doing much on a CPU.
+*/
 cpu_relax :: intrinsics.cpu_relax
```
Feoramund, line 10: Doesn't it make more sense for documentation to live at the origin point (`base:intrinsics`), rather than an alias? Does that mean there will be no documentation in `intrinsics` or will this be duplicated (and thus need to both be maintained in sync)?

gingerBill: I'd argue both need to be documented. The alias and the intrinsics in pkg.odin-lang.org generator.

flysand7: > or will this be duplicated

Unfortunately the state of documentation is that it needs to be duplicated if as the end user you don't want to follow constant chain of references to find the relevant information. You'll see it a lot in this PR, I duplicate information as much as I can to make sure reader gets the relevant pieces without following chains of references

flysand7:
Ok I think it's pretty much final and ready for review

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
-	Relaxed = 0, // Unordered
-	Consume = 1, // Monotonic
-	Acquire = 2,
-	Release = 3,
-	Acq_Rel = 4,
-	Seq_Cst = 5,
-}
+Describes memory ordering for an atomic operation.
+
+Moden CPU's contain multiple cores and caches specific to those cores. When a
```
gingerBill, line 15: Moden -> Modern

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
+/*
+Atomically replace the memory location with the result of AND operation with
+the specified value.
+
+This procedure loads a value from memory, calculates the result of AND operation
+between the loaded value and the specified value, and stores it back into the
+same memory location as an atomic operation. This operation is an atomic
+equivalent of the following:
+
+	dst^ &&= val
```
gingerBill, line 189:
```
dst^ &= val
```

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
```
gingerBill, line 204:
```
dst^ &= val
```

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
+/*
+Atomically replace the memory location with the result of NAND operation with
+the specified value.
+
+This procedure loads a value from memory, calculates the result of NAND operation
+between the loaded value and the specified value, and stores it back into the
+same memory location as an atomic operation. This operation is an atomic
+equivalent of the following:
+
+	dst^ = !(dst^ && val)
```
gingerBill, line 219:
```
dst^ = ~(dst^ & val)
```

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
```
gingerBill, line 234: && -> &

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
+/*
+Atomically replace the memory location with the result of OR operation with
+the specified value.
+
+This procedure loads a value from memory, calculates the result of OR operation
+between the loaded value and the specified value, and stores it back into the
+same memory location as an atomic operation. This operation is an atomic
+equivalent of the following:
+
+	dst^ ||= val
```
gingerBill, line 249: || -> |

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
```
gingerBill, line 264: || -> |

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
+/*
+Atomically replace the memory location with the result of XOR operation with
+the specified value.
+
+This procedure loads a value from memory, calculates the result of XOR operation
+between the loaded value and the specified value, and stores it back into the
+same memory location as an atomic operation. This operation is an atomic
+equivalent of the following:
+
+	dst^ ~~= val
```
gingerBill, line 279: ~~ -> ~

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
```
gingerBill, line 294: ~~ -> ~

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
+This procedure checks if the value pointed to by the `dst` parameter is equal
+to `old`, and if they are, it stores the value `new` into the memory location,
+all done in a single atomic operation. This procedure returns the old value
+stored in a memory location and a boolean value signifying whether `old` was
+equal to `new`.
+
+This procedure is an atomic equivalent of the following operation:
+
+	old_dst := dst^
+	if old_dst = old {
```
gingerBill, line 344: = -> ==

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
```
gingerBill, line 374: = -> ==

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
```
gingerBill, line 404: = -> ==

core/sync/atomic.odin
```diff
@@ -2,44 +2,452 @@ package sync
...
+		// may return false here
+		dst^ = new
+		return old_dst, true
+	} else {
+		return old_dst, false
+	}
+
+The weak version of compare exchange may return false, even if `dst^` = `old`.
```
gingerBill, line 412: = -> ==

core/sync/atomic.odin
flysand7: Oh so these are bitwise.. Didn't guess from the name, actually never used these

core/sync/atomic.odin
flysand7: Omg having a job where I write shitton of SQL really rotted my brain

flysand7:
Fixed now.

PR #4007, opened 2024-08-02
Fix: compilation error when using BufferSubDataSlice in webgl package

marcosantos98:
## Problem:

`BufferSubDataSlice` calls `BufferSubData` with an extra parameter.

image

It appears to be a copy-pasta left over from `BufferDataSlice` that calls [`BufferData`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/bufferData) which requires `usage`.

## Done:
- Removed the extra parameter.


Kelimion:
Good find.

marcosantos98:
> Good find.

Yeah, since the compiler doesn't have tests for it and only appears if the specific procedure is called it lived for 3 years.
image

Kelimion:
> > Good find.
> 
> Yeah, since the compiler doesn't have tests for it and only appears if the specific procedure is called it lived for 3 years.

I'm not surprised at this. Automatically testing anything that requires a GPU and/or headless browser to run is going to be non-trivial, and as GitHub's CI runners don't come with a GPU, also potentially costly.

PR #4012, opened 2024-08-02
core:sys/posix and core:os/os2 based on it (for darwin, netbsd, freebsd and openbsd)

laytan:
## In brief

- new package `core:sys/posix`
- implementation of os2's current APIs for Darwin, NetBSD, FreeBSD and OpenBSD based on the posix package (excluding the process.odin API)
- implement the process_info API on Darwin by adding incomplete bindings to their standard process API to `core:sys/darwin`
- support `core:mem/virtual` on NetBSD, FreeBSD and OpenBSD
- make `core:time` not use libc anymore on Linux
- some fixes to other parts of os2 found while implementing the API
- some fixes in `core:c/libc`
- default to 0o755 permissions when making directories in os2
- add `intrinsics.constant_log2` to ease binding with bitsets

## `core:sys/posix`

A new package that binds to the POSIX standard.

This was added so that we can provide a posix based implementation of os2, pending OS specific implementations that may come later.
Darwin/MacOS will at least use this as it's final os2 implementation due to restrictions Apple puts on using direct syscalls.

Additionally, we get regular questions about termios, ipc, unix sockets, etc. which this will hopefully cover.

Currently only supported for Darwin, NetBSD, OpenBSD and FreeBSD.

Linux is something I want to add but it wasn't needed for the main goal of the PR (os2).
And Linux also has many different libc implementations which make it hard.

During the making of these bindings I added `intrinsics.constant_log2` this is extremly
handy because instead of manually doing the log2 of whatever the C headers specify,
we can directly copy the C constants and log2 them automatically.
There have been many bugs in the past because this manual conversion was misunderstood or done wrong.

## os2

The current os2 API has been implemented on Darwin and the BSDs using the posix package.
I left out the process API because @flysand is still working on the design of it and I had already written too much code without a PR.
I did implement the process_info stuff on Darwin (this is not a thing you can do with posix alone).

### Some os2 design notes

Some notes on the design and possible improvements, I have not changed any API here
in order for the PR to not do too many things, but still wrote these down:

- could we make `Platform_Error :: enum { None = 0 }` be `Platform_Error :: _Platform_Error` instead,
  posix would then do `_Platform_Error :: distinct posix.Errno`, same with Linux, this would make
  printing the error actually print a meaningful thing instead of `%!BAD_ENUM_VALUE=x`
- for the environment APIs specifically, they are returning bools instead of errors, I think this is a mistake
  carried over from `os`, I left some notes in the posix implementation about different conditions that a user
  should be notified about instead of just getting `false` and needing to get a debugger out for the specific errors
- the read directory iterator should probably have a `error` field on it, so implementations can set it when something goes wrong, users now can't determine if something went wrong or if all files were read
- what the heck is `File_Flag.Sparse`?
- I was thinking it may be better to abstract away the permissions octal number, I don't think we should assume
  knowledge about this
- we should probably have the file type be a field on the `File`, there are a few apis that constantly `stat`
  when all they need is the file type which is wasteful, it's also just nice to have exposed, this field could be zero initialized to Unknown and filled in/cached the first time there is a need for the filetype
- `free_process_info` would fit the core packages better as `destroy_process_info` IMO
- `clear_env` is a weird thing to have/do
- a `child_process_list` to accompany `process_list` would be good imo
- adding memory/cpu usage to process info seems great

## In the future

Some things that can/should be done in the future:

- support linux in `core:sys/posix`
  - nuke `core:sys/unix`
  - use the pthread stuff from `core:sys/posix` for `core:threads`
- process API of os2 for the posix implementation
- haiku (does anybody use this?), their posix implementation also looked weird
- improve lstat's fullpath resolving, can't use open() or realpath() because they try to resolve the symlinks


core/sys/posix/sys_socket.odin
```diff
@@ -0,0 +1,491 @@
...
+	SO_OOBINLINE  :: 0x0100
+	SO_RCVBUF     :: 0x1002
+	SO_RCVLOWAT   :: 0x1004
+	SO_REUSEADDR  :: 0x0004
+	SO_SNDBUF     :: 0x1001
+	SO_SNDLOWAT   :: 0x1003
+	SO_TYPE       :: 0x1008
+
+	when ODIN_OS == .Darwin || ODIN_OS == .FreeBSD {
+		SO_LINGER   :: 0x1080
```
Feoramund, line 441: This is `0x80` on FreeBSD.

core/sys/posix/fcntl.odin
```diff
@@ -0,0 +1,415 @@
...
+	_O_RSYNC   :: 0
+	O_RSYNC    :: O_Flags{}
+
+	O_EXEC    :: 0x00040000
+	O_RDONLY  :: 0
+	O_RDWR    :: 0x0002
+	O_WRONLY  :: 0x0001
+
+	_O_SEARCH :: O_EXEC|O_DIRECTORY
+	O_SEARCH  :: O_Flags{ .EXEC, .DIRECTORY }
```
Feoramund, line 265: `O_SEARCH` is the same value as `O_EXEC` on FreeBSD.

core/sys/posix/fcntl.odin
```diff
@@ -0,0 +1,415 @@
...
+
+	_O_TTY_INIT :: 0x00080000
+	O_TTY_INIT  :: O_Flags{O_Flag_Bits(log2(_O_TTY_INIT))}
+
+	O_APPEND   :: 0x0008
+	O_DSYNC    :: 0x01000000
+	O_NONBLOCK :: 0x0004
+	O_SYNC     :: 0x0080
+	_O_RSYNC   :: 0
+	O_RSYNC    :: O_Flags{}
```
Feoramund, line 257: FreeBSD does not implement `O_RSYNC` as of 14.0-RELEASE.

https://github.com/freebsd/freebsd-src/blob/2edac5a60a57a2a6ef80c61079d427c4ee31b7a9/sys/sys/fcntl.h#L145-L147

laytan: I know, that's why it is `0`, adding it does nothing. I can add a comment though

core/sys/posix/fcntl.odin
laytan: It is technically fine like this because O_SEARCH can only be used in combination with O_DIRECTORY, supplying O_SEARCH without O_DIRECTORY would return EINVAL afaik.

But, I see no reason to not make it the exact same so I will make that change.

Feoramund:
Attempts at OS unification have lent themselves to untested copy-paste code in the past with `core:sys/unix`, `core:thread`, and others under the assumption that everyone is more or less the same when that is not the case. It's also easier to compare and `diff` two OS-specific files defining their constants side-by-side rather than one long chain of `when`/`else` statements.

Do you have any plans to guarantee interoperability and correctness?

core/sys/posix/sys_socket.odin
laytan: Thanks!

core/sys/posix/fcntl.odin
Feoramund: Right, I noticed you commented the other cases where FreeBSD did not implement something, so I pointed this out.

laytan:
> Attempts at OS unification have lent themselves to untested copy-paste code in the past with `core:sys/unix`, `core:thread`, and others under the assumption that everyone is more or less the same when that is not the case. It's also easier to compare and `diff` two OS-specific files defining their constants side-by-side rather than one long chain of `when`/`else` statements.
> 
> Do you have any plans to guarantee interoperability and correctness?

OS unification is not what caused untested copy-paste code, see core:os for example which also has that and is not unified. I don't think it is avoidable other than being thorough with reviews and writing tests.
Since not too long ago we have CI for some of the bsds that should help a bunch, the language growing will also provide more coverage.

I can do multiple files but it would quadruple the amount of files in the package (not necessarily a bad thing) and it would make people be able to add another OS without implementing everything which I wanted to avoid.

core/sys/posix/fcntl.odin
laytan: Did I? That should be corrected, I am trying to make usage of non-private (/ underscored) symbols work on each platform, so if that's not the case please let me know.

laytan:
Another reason I did a single file with all platforms is because that's also done in core:c/libc, which this originated as a sub-package of

core/sys/posix/fcntl.odin
Feoramund: https://github.com/odin-lang/Odin/blob/b40a32b22d9aeef995f9277cdceb9f5ca1f90484/core/sys/posix/errno.odin#L266-L270

laytan: Oh I misunderstood, I thought you meant "commented out" as in not defined, I get it now :)

flysand7:
Unrelated to the contents of the PR, but do you want to fix the 'posis' typo in the title of the PR?

:3

laytan:
> Unrelated to the contents of the PR, but do you want to fix the 'posis' typo in the title of the PR?
> 
> :3

Lol, didn't even notice 😆

Feoramund:
> I don't think it is avoidable other than being thorough with reviews and writing tests.

This assumes a preference to spend time reviewing every reference to a `define` or `struct` in an external operating system's header files and keep it up to date. More tests would be good, but it would require many, many permutations of the various enumerated options of any one procedure for full coverage.

> [...] and it would make people be able to add another OS without implementing everything which I wanted to avoid.

This is a good point.

---

I do believe the issue here is avoidable, with better tooling. Recently, I've been thinking about this more. What are we fundamentally doing when we write `core:sys` packages or OS-interfacing code? Copying definitions from system header files: constant `#define` definitions, typedefs, structs, C functions, re-interpreting macros. Sometimes they have to be re-worked. What are the issues that come up with this? Incorrect copying (not necessarily from you the original author, but easily someone later down the line could innocently copy some mostly-correct structures), changes to the underlying OS over time.

Why couldn't we have a program to automatically do this for us? Better yet, a program that does this and also verifies correctness over time as OS versions change. I believe this would be the more maintainable option in the long run.

You could structure a translation language for exporting only what you want from headers, like an export manifest, or just a bunch of explicit procedures from an Odin package - no need to get too fancy about it, grouping them into enums or leaving them as constants.

Here's a tiny, hypothetical example to better illustrate what I'm dreaming of:
```c
// sys/errno.h
#define EPERM 1
#define ENOENT 2
```

```odin
import "core:c/tooling"

output, err := tooling.parse_as_enum(
	file = "netbsd/sys/errno.h",
	default = 'ENONE',
	output_name = 'Errno',
	defines = {
		'EPERM',
		'ENOENT',
	},
)
```

Run that, and you get:
```odin
Errno :: enum {
	ENONE,
	EPERM  = 1, //+refer netbsd/sys/errno.h(123:4) EPERM
	ENOENT = 2, //+refer netbsd/sys/errno.h(124:4) ENOENT
}
```

With this, you could have a test suite that re-generates all of the generated headers and runs a diff against them to verify that everything is still as expected. This sort of tooling could be used for any C library, not just system header files, and it would greatly reduce the workload, the many hours it takes to write, review, and maintain such an import, and the potential for human error in review. It could be used for the `vendor` libraries.

There's no reason to write all of this by hand when we can automate it.

The API example given above is all made up and can be whatever, but the underlying idea I'm putting forward here is sound.

laytan:
I am on board with that in general, but besides mindless copying it also requires quite a bit of reasoning about the API, types and intended usage which are not gotten from looking at the headers most of the time, not even by humans, let alone a tool.

No tool will be able to output idiomatic bindings like I wrote here, which is one of the main reasons we don't have a standard tool like it. Look at zig's @ cimport as an example, it quickly falls apart and most of the bindings are still handwritten wrappers over it.


I do see your snippet is not fully automatic which is certainly the way to approach this, that could actually output some nice things if done well.

There is also a reason we mainly do vendor bindings for stable things that don't change often anymore, once bindings are written, updates are almost a non-issue because they are rare and small.

core/os/os2/allocators.odin
```diff
@@ -46,3 +46,6 @@ TEMP_ALLOCATOR_GUARD :: #force_inline proc(ignore := false, loc := #caller_locat
 	}
 }
 
+is_temp :: proc(allocator: runtime.Allocator) -> bool {
+	return allocator.data == &global_default_temp_allocator_arena
+}
```
gingerBill, line 51: Why is this even needed in the first place?

gingerBill: Especially since the user can NEVER even call `temp_allocator()`

core/os/os2/env_posix.odin
```diff
@@ -0,0 +1,77 @@
...
+
+import "core:strings"
+import "core:sys/posix"
+
+_lookup_env :: proc(key: string, allocator: runtime.Allocator) -> (value: string, found: bool) {
+	if key == "" {
+		return
+	}
+
+	TEMP_ALLOCATOR_GUARD(ignore=is_temp(allocator))
```
gingerBill, line 15: Remove the `ignore` check. That is useless in this context because `temp_allocator()` is private the package anyway.

Don't try to repeat the bodge that was in `core:os`.

laytan:
The is_temp was actually needed, you had a few use-after-frees because of something like `stat(fd, temp_allocator())` calls (internally in the package), and then in stat you had a TEMP_ALLOCATOR_GUARD, stat would allocate the results on the arena and return it, but the guard had already freed it at that point.

laytan:
https://github.com/odin-lang/Odin/commit/d0709a7de21efded4625167dbff4a7dd13d561b4 will allow me to remove the is_temp stuff because it fixes the use-after-free bugs another way.

laytan:
> [d0709a7](https://github.com/odin-lang/Odin/commit/d0709a7de21efded4625167dbff4a7dd13d561b4) will allow me to remove the is_temp stuff because it fixes the use-after-free bugs another way.

done

PR #4035, opened 2024-08-08
Added a hint at end of build.bat about release mode when building in debug mode

karl-zylinski:
The motivation is that a bunch of people seem to not realize they use a slow debug compiler, sometimes for several years.

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -116,6 +116,9 @@ if %errorlevel% neq 0 goto end_of_build
 rem If the demo doesn't run for you and your CPU is more than a decade old, try -microarch:native
 if %release_mode% EQU 0 odin run examples/demo -vet -strict-style -- Hellope World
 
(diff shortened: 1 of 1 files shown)


JesseRMeyer:
Alternatively, we could just emit the build mode when building, like `Build: [DEBUG]` or `Build: [RELEASE]`, but I agree that a change here is warranted.

Kelimion:
And for consistency, we may also want to add `build.bat release_native` (to go with `make release_native`), but that could be its own pull request.

laytan:
Should've been added to build.sh too, no?

karl-zylinski:
> Should've been added to build.sh too, no?

Can be done separately, I don't have quick access to linux right now.

laytan:
Hmm I don't like adding things for one OS. It's fine I will just add it.

PR #4043, opened 2024-08-08
Allow testing for intentional leaks in test runner

Kelimion:
Adds `expect_leak_or_bad_free :: proc(t: ^T, client_test: proc(t: ^T), verifier: Memory_Verifier_Proc)`.

It sets up its own `Tracking_Allocator`, runs the `client_test`, and then calls the `verifier` procedure.
The verifier can then inspect the contents of the tracking allocator and call `testing.expect*` as sensible for the test in question.

Any allocations are then cleared so that the test runner doesn't itself complain about leaks.

Additionally, `ODIN_TEST_LOG_LEVEL_MEMORY` has been added as a define to set the severity of the test runner's memory tracker. You can use `-define:ODIN_TEST_LOG_LEVEL_MEMORY=error` to make tests fail rather than warn if leaks or bad frees have been found.

### DIFF
--- a/core/testing/runner.odin
+++ b/core/testing/runner.odin
@@ -25,6 +25,8 @@ TEST_THREADS          : int    : #config(ODIN_TEST_THREADS, 0)
 TRACKING_MEMORY       : bool   : #config(ODIN_TEST_TRACK_MEMORY, true)
 // Always report how much memory is used, even when there are no leaks or bad frees.
 ALWAYS_REPORT_MEMORY  : bool   : #config(ODIN_TEST_ALWAYS_REPORT_MEMORY, false)
+// Log level for memory leaks and bad frees: debug, info, warning, error, fatal
+LOG_LEVEL_MEMORY      : string : #config(ODIN_TEST_LOG_LEVEL_MEMORY, "warning")
 // Specify how much memory each thread allocator starts with.
 PER_THREAD_MEMORY     : int    : #config(ODIN_TEST_THREAD_MEMORY, mem.ROLLBACK_STACK_DEFAULT_BLOCK_SIZE)
 // Select a specific set of tests to run by name.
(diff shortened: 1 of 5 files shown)


Feoramund:
`-define:ODIN_TEST_LOG_LEVEL_MEMORY=error` does not actually cause a test to fail if it leaks though. Have I missed something here?

```odin
package main

import "core:testing"

@test
a :: proc(t: ^testing.T) {
	b := new(int)
}
```

`odin test . -define:ODIN_TEST_LOG_LEVEL_MEMORY=error -define:ODIN_TEST_SHORT_LOGS=true`

```
[INFO ] --- Starting test runner with 1 thread. Set with -define:ODIN_TEST_THREADS=n.
[INFO ] --- The random seed sent to every test is: 148072560220716. Set with -define:ODIN_TEST_RANDOM_SEED=n.
[INFO ] --- Memory tracking is enabled. Tests will log their memory usage if there's an issue.
[INFO ] --- < Final Mem/ Total Mem> <  Peak Mem> (#Free/Alloc) :: [package.test_name]
[ERROR] --- <        8B/        8B> <        8B> (    0/    1) :: main.a
        +++ leak         8B @ 0x750BCF000048 [memtest.odin:7:a()]
main  [|                       ]         1 :: [package done]

Finished 1 test in 193.756µs. The test was successful.
```

Kelimion:
> `-define:ODIN_TEST_LOG_LEVEL_MEMORY=error` does not actually cause a test to fail if it leaks though. Have I missed something here?
> 
> ```odin
> package main
> 
> import "core:testing"
> 
> @test
> a :: proc(t: ^testing.T) {
> 	b := new(int)
> }
> ```
> 
> `odin test . -define:ODIN_TEST_LOG_LEVEL_MEMORY=error -define:ODIN_TEST_SHORT_LOGS=true`
> 
> ```
> [INFO ] --- Starting test runner with 1 thread. Set with -define:ODIN_TEST_THREADS=n.
> [INFO ] --- The random seed sent to every test is: 148072560220716. Set with -define:ODIN_TEST_RANDOM_SEED=n.
> [INFO ] --- Memory tracking is enabled. Tests will log their memory usage if there's an issue.
> [INFO ] --- < Final Mem/ Total Mem> <  Peak Mem> (#Free/Alloc) :: [package.test_name]
> [ERROR] --- <        8B/        8B> <        8B> (    0/    1) :: main.a
>         +++ leak         8B @ 0x750BCF000048 [memtest.odin:7:a()]
> main  [|                       ]         1 :: [package done]
> 
> Finished 1 test in 193.756µs. The test was successful.
> ```

Look at the return value. It now checks for bad frees + leaks if mem tracking is active.

core/testing/runner.odin
```diff
@@ -63,6 +65,21 @@ get_log_level :: #force_inline proc() -> runtime.Logger_Level {
 	}
 }
 
+get_memory_log_level :: #force_inline proc() -> runtime.Logger_Level {
+	when ODIN_DEBUG {
+		// Always use .Debug in `-debug` mode.
+		return .Debug
+	} else {
```
Feoramund, line 72: Is there a reason for this to default to Debug in `-debug` mode? The reason why general logging forces it to Debug is as a fallback to catch every possible log message, but I'm not sure why memory logging would do this, especially if there's an intent to catch leaks and errors by virtue of setting a higher log level.

Kelimion: Not really, no. I just copy/pasted it with minimal edits.

PR #4049, opened 2024-08-10
vendor/x11: added a few procedures for xrandr, xinput, and xfixes

IllusionMan1212:


flysand7:
xinput and xrandr are part of xlib and should be inside the folder. The `x11/xlib` package is named after the x11 library `XLib`, not just `xlib.h`. It is opposed to XCB, which may be implemented by someone in the future.

IllusionMan1212:
Merged them into the xlib package.

PR #4053, opened 2024-08-10
Fix `.gitignore`

Feoramund:
I'm not familiar enough with non-UNIX build systems to suggest what to do about the lines around the all-encompassing `x86/` (or if they're even relevant anymore), but this change at least lets me make modifications to `core:simd/x86` and `tests/benchmark` without complaints from `git`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,8 +23,6 @@ bld/
 [Oo]bj/
 [Ll]og/
 ![Cc]ore/[Ll]og/
-tests/documentation/verify/
-tests/documentation/all.odin-doc
 # Visual Studio 2015 cache/options directory
 .vs/
 # Visual Studio Code options directory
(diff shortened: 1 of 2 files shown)


Kelimion:
That gitignore line was about the benchmark executable accidentally ending up in the repository.

Feoramund:
Would a better line be `/tests/benchmark/benchmark*`? Come to think of it, we don't have lines for tests ending up in the repository.

Kelimion:
> Would a better line be `/tests/benchmark/benchmark*`?

Maybe, yes. But then we'd probably want a `tests/bench.sh` or `tests/benchmark/bench.sh`, to make it easy for people to build and run the benchmark locally and have it clean itself up. The CI doesn't need to care about cleaning up.

> Come to think of it, we don't have lines for tests ending up in the repository.

I mean we do for Windows, it's mostly just an issue on *nix because of the lack of extension.

Feoramund:
I added a `.gitignore` specifically for the `tests` directory. Do you think this might adequately solve the problem now?

PR #4069, opened 2024-08-12
Fix alias handling of procedures

### LINKED ISSUE #1738: Assertion bug when aliasing procedure multiple times in a specific way
## Context

        Odin: dev-2022-04:f92ffe60
        OS:   Windows 10 Unknown Edition (0000007d) (version: 21H2), build 19044.1645
        CPU:  Intel(R) Core(TM)2 Duo CPU     P8600  @ 2.40GHz
        RAM:  8086 MiB

## Bug
In this directory structure:
image

`alias.odin`
```odin
package alias

import a "alias_pkg"

main :: proc() {
	a.wait_vblank()
}
```

`alias_pkg.odin`
```odin
package alias_pkg

wait_vblank :: _wait_vblank
```

`alias_pkg_windows.odin`
```odin
//+build windows
package alias_pkg

_wait_vblank :: win32.DwmFlush
```

When compiling this error happens:
```
Odin\src\llvm_backend_proc.cpp(65): Assertion Failure: `entity->flags & EntityFlag_ProcBodyChecked` DwmFlush :: proc "stdcall" () -> HRESULT
```

zen3ger:
An incorrect memmove when overriding entities caused multiple ones to point to the same procedure with incomplete variant data, resulting in later hiting a compiler assertion.

Introduced delayed type checking for procedure aliases, as it was masked by the previous error in the override logic.

Fixes #1738

### DIFF
--- a/src/check_decl.cpp
+++ b/src/check_decl.cpp
@@ -180,6 +180,8 @@ gb_internal void override_entity_in_scope(Entity *original_entity, Entity *new_e
 
 	original_entity->flags |= EntityFlag_Overridden;
 	original_entity->type = new_entity->type;
+	original_entity->kind = new_entity->kind;
+	original_entity->decl_info = new_entity->decl_info;
 	original_entity->aliased_of = new_entity;
 
 	original_entity->identifier.store(new_entity->identifier);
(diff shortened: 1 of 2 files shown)


src/check_decl.cpp
```diff
@@ -180,6 +180,8 @@ gb_internal void override_entity_in_scope(Entity *original_entity, Entity *new_e
 
 	original_entity->flags |= EntityFlag_Overridden;
 	original_entity->type = new_entity->type;
+	original_entity->kind = new_entity->kind;
+	original_entity->decl_info = new_entity->decl_info;
```
zen3ger, line 184: I'm not sure if this is entirely correct, assuming the comment above holds...
I've found that changing the `kind` is a must otherwise a chain of aliases result in e.g.: `Constant -> Constant -> Procedure` entity types.

The change on `decl_info` is due to other parts of the code asserting that an entity of `kind` holds matching `decl_info` which does not hold with aliases.

PR #4070, opened 2024-08-12
Let memory failures be errors in the test runner

Feoramund:
This PR implements the secondary idea in #4043 as a first-class feature.

Enable with `-define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true`.


Kelimion:
Thanks once again.

PR #4071, opened 2024-08-12
Use another URL for pkgsrc on CI

andreas-jonsson:
I'm not entirely sure what the issue here is? But on my system I use "amd64" and not "x86_64".
Perhaps there is an issue using this architecture name on some mirrors?

Anyway, amd64 is what NetBSD calls it these days. :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,10 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build_netbsd:
-    if: false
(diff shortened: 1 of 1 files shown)


Kelimion:
Let's hope that keeps it running. Thanks for looking into this.

andreas-jonsson:
The PKGSRC_BRANCH needs to be updated at some point. But they keep old branches around for quite some time.
Updating it will likely result in a newer LLVM version so that would have to be coordinated so it is supported by Odin at that point.

PR #4077, opened 2024-08-13
fix vararg default values

laytan:
Default values for varargs didn't work, no error but also no default value (always got `nil`). This simple change makes them work properly.
I could also see not allowing it at all, in which case #2977 would be a good merge.


gingerBill:
This might fix it, but should we allow this in the first place?

Default values on varargs seems like a bad idea.

laytan:
> This might fix it, but should we allow this in the first place?
> 
> Default values on varargs seems like a bad idea.

I'm fine with making it an error too, that's why I wrote this in the description:
> I could also see not allowing it at all, in which case #2977 would be a good merge.

gingerBill:
I think we should just disallow it entirely.

PR #4080, opened 2024-08-14
fix os2.read_entire_file_from_file for files with known size

DerTee:
`os2.read_entire_file_from_file` has a couple of bugs when using a `os2.File` with a known size.
The size that is determined gets reset to 0, in case there is a problem with the OS's int type being too small then inconsistent is passed that leads to more problems.

In the next step, the code in the "known size" branch of the logic creates a loop that throws away useful data until an error occurs, so the "known size" branch can never succeed.

I haven't tested the "unknown size" branch yet, but so far, that one looks good to me, so there's probably still a bug.

### DIFF
--- a/core/os/os2/file_util.odin
+++ b/core/os/os2/file_util.odin
@@ -133,22 +133,14 @@ read_entire_file_from_file :: proc(f: ^File, allocator: runtime.Allocator) -> (d
 		err = serr
 		return
 	}
-	size += 1 // for EOF
 
-	// TODO(bill): Is this correct logic?
 	if has_size {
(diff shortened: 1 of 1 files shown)


laytan:
I think I fixed these problems here too: https://github.com/odin-lang/Odin/pull/4012

DerTee:
@laytan Awesome, then I'll try your changes out and check in with you if there are any problems. Thanks for the heads up!

-> I'm closing this PR because this should be solved and even if not, then the code underneath has changed, so it would have to be reworked anyway

PR #4082, opened 2024-08-14
Fix 'odin [run|test] --' giving a compiler out of bounds

FourteenBrush:
`src/array.cpp(13): Assertion Failure: ``cast(usize)index < cast(usize)count`` Index 2 is out of bounds ranges 0..<2`

Fixes the following invocations (missing source path):
- `odin [run|test] --`
- `odin [run|test] -- [..ARGS]`

Replaces them both with a usage message

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2973,13 +2973,19 @@ int main(int arg_count, char const **arg_ptr) {
 				break;
 			}
 		}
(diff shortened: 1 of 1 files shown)


flysand7:
I can tell by the missing space after the `if` keyword -- that was my code :)

PR #4089, opened 2024-08-16
add support for linux_riscv64 and freestanding_riscv64

laytan:
Adds support for riscv64 through a Linux and freestanding target.

We require the float, doubles and integer multiplication extensions to be enabled. We default to the same features as clang with `-march=rv64gc` which stands for "general computing". Users can tweak extensions with -microarch and -target-features. These default features are: `64bit,a,c,d,f,m,relax,zicsr,zifencei` you can find more info on them by compiling with `-target-features:"help"`.

You can either build with no libc through `odin build . -target:linux_riscv64 -no-crt -default-to-nil-allocator` and in my testing I needed `-reloc-mode:static` or it would complain about thread local storage.

Or, install the gnu toolchain `sudo apt install gcc-12-riscv64-linux-gnu libc6-dev-riscv64-cross` to get the cross-compilation toolchain and use `odin build . -target:linux_riscv64 -extra-linker-flags:"-fuse-ld=/usr/bin/riscv64-linux-gnu-gcc-12"` to use the riscv enabled linker.


src/llvm_abi.cpp
```diff
@@ -332,7 +332,8 @@ gb_internal i64 lb_alignof(LLVMTypeRef type) {
 }
 
 
-#define LB_ABI_INFO(name) lbFunctionType *name(LLVMContextRef c, LLVMTypeRef *arg_types, unsigned arg_count, LLVMTypeRef return_type, bool return_is_defined, bool return_is_tuple, ProcCallingConvention calling_convention, Type *original_type)
+#define LB_ABI_INFO(name) lbFunctionType *name(lbModule *m, LLVMTypeRef *arg_types, unsigned arg_count, LLVMTypeRef return_type, bool return_is_defined, bool return_is_tuple, ProcCallingConvention calling_convention, Type *original_type)
+#define LB_ABI_INFO_CTX(name) lbFunctionType *name(LLVMContextRef c, LLVMTypeRef *arg_types, unsigned arg_count, LLVMTypeRef return_type, bool return_is_defined, bool return_is_tuple, ProcCallingConvention calling_convention, Type *original_type)
```
gingerBill, line 336: My not just unify these both?

laytan: Done!

core/sys/linux/types.odin
```diff
@@ -120,6 +120,25 @@ when ODIN_ARCH == .amd64 {
...
+		size:       int,
+		blksize:    i32,
+		_:          i32,
+		blocks:     int,
+		atime:      Time_Spec,
+		mtime:      Time_Spec,
+		ctime:      Time_Spec,
+		_:          [2]u32,
+	}
+} else when ODIN_ARCH == .riscv64 {
```
flysand7, line 141: Are you use this is correct...? You added these lines above where you check for arm64, not for riscv

laytan: Looks like arm64 and riscv64 have the same struct for stat actually, don't know where I got the wrong idea from, good spot

PR #4091, opened 2024-08-16
Check if file open-mode is `O_CREATE` on Darwin before forcing chmod

### LINKED ISSUE #4087: core:flags package tries to change permissions of input files defined via os.Handle type
## Context

Running the example Odin program provided by the flags package:

https://github.com/odin-lang/Odin/blob/master/core/flags/example/example.odin

        Odin:    dev-2024-08:1a16585b1
        OS:      macOS Sonoma 14.5 (build: 23F79, kernel: 23.5.0)
        CPU:     Apple M3 Max
        RAM:     36864 MiB
        Backend: LLVM 18.1.8

## Expected Behavior

A file marked as readonly should not have other permissions changed

For example if I want to _read_  a file (e.g. `/etc/passwd`) the flags package should NOT try to change the permissions of the file

## Current Behavior

```
./argparse_test /etc/passwd -hub 4 -accuracy 6.3
[Open_File_Error#-1] Unable to open file with perms 0o444 in mode 0x0: /etc/passwd

```

More insidious is if you _do_ have permissions to change the file then something that is marked `rwx` silently gets the permissions changed:

Before:
```
$  ls -lrt foo  
-rwxr-xr-x  1 acochrane  staff  0 Aug 16 09:32 foo
$  ./argparse_test foo -hub 4 -accuracy 6.3
[Parse_Error.Bad_Value] Unable to set `hub` of type Host_Or_Endpoint to true.
$  ls -lrt foo                             
-r--r--r--  1 acochrane  staff  0 Aug 16 09:32 foo
```
 

### Steps to Reproduce

Copy and paste the example flags code from https://github.com/odin-lang/Odin/blob/master/core/flags/example/example.odin then run it with a file that is readable, but not writable and the error above will be shown.

Feoramund:
I do not have a Darwin machine to test this with, but this should address #4087.

Fixes #4087

### DIFF
--- a/core/os/os_darwin.odin
+++ b/core/os/os_darwin.odin
@@ -702,7 +702,7 @@ open :: proc(path: string, flags: int = O_RDWR, mode: int = 0) -> (handle: Handl
 		@INFO(Platin): this is only done because O_CREATE for some reason fails to apply mode
 		               should not happen if the handle is a directory
 	*/
-	if mode != 0 && !isDir {
+	if flags & O_CREATE != 0 && mode != 0 && !isDir {
 		err = fchmod(handle, cast(u16)mode)
 		if err != nil {
 			_unix_close(handle)


laytan:
> @INFO(Platin): this is only done because O_CREATE for some reason fails to apply mode

I wonder if this is outdated and it can be removed completely, I know we had some c_vararg bugs at some point, I can check on my mac in a bit.

PR #4101, opened 2024-08-18
core/bytes: Tweak `index_byte`/`last_index_byte`

Yawning:
- Assume unaligned loads are cheap
- Explicilty use 256-bit or 128-bit SIMD to avoid AVX512
- Limit "vectorized" scanning to 128-bits if SIMD is emulated via SWAR
- Add a few more benchmark cases

I also think `last_index_byte` would crash if `s == nil` (at least it looks suspicious).  This is objectively worse on huge `s` with the character at the tail, if the starting address is not 16 or 32-byte aligned (depending on microarch), because the penalties will add up.  I don't particularly think that case is worth optimizing for (and this has a code-path for < 4*SIMD_REG_SIZE), so it should be better on reasonable inputs.


core/bytes/bytes.odin
```diff
@@ -334,52 +334,126 @@ Inputs:
...
 - index: The index of the byte `c`, or -1 if it was not found.
 */
-index_byte :: proc(s: []byte, c: byte) -> (index: int) #no_bounds_check {
-	length := len(s)
-	i := 0
-
-	// Guard against small strings.
-	if length < SIMD_SCAN_WIDTH {
-		for /**/; i < length; i += 1 {
+index_byte :: proc "contextless" (s: []byte, c: byte) -> (index: int) #no_bounds_check {
```
Feoramund, line 337: These procs have been made contextless but not the ones in `core:strings`. Why is that?

core/bytes/bytes.odin
```diff
@@ -334,52 +334,126 @@ Inputs:
...
+							off := simd.reduce_min(sel)
+							return i + j * SIMD_REG_SIZE_256 + int(off)
+						}
+					}
+				}
+
+				i += 2 * SIMD_REG_SIZE_256
+			}
+		} else {
+			// Note: ARM can probaby do 128 or 256-bytes at a time.
```
Feoramund, line 409:
```suggestion
			// Note: ARM can probably do 128 or 256-bytes at a time.
```

core/bytes/bytes.odin
```diff
@@ -401,56 +475,129 @@ Inputs:
...
+						if m_vec[1-j] > 0 {
+							sel := simd.select(c_vecs[1-j], SCANNER_INDICES_256, SCANNER_SENTINEL_MAX_256)
+							off := simd.reduce_max(sel)
+							return i + (1-j) * SIMD_REG_SIZE_256 + int(off)
+						}
+					}
+				}
+			}
+		} else {
+			// Note: ARM can probaby do 128 or 256-bytes at a time.
```
Feoramund, line 552:
```suggestion
			// Note: ARM can probably do 128 or 256-bytes at a time.
```

core/bytes/bytes.odin
```diff
@@ -334,52 +334,126 @@ Inputs:
...
+		// Note: While this is something that could also logically take
+		// advantage of AVX512, the various downclocking and power
+		// consumption related woes make premature to have a dedicated
+		// code path.
+		when ODIN_ARCH == .amd64 && intrinsics.has_target_feature("avx2") {
+			c_vec_256 := simd.swizzle(
+				c_vec,
+				0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+				0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+			)
```
Feoramund, line 363: How does this differ from `c_vec_256: simd.u8x32 = c`?

core/bytes/bytes.odin
```diff
@@ -334,52 +334,126 @@ Inputs:
...
+			// Note: ARM can probaby do 128 or 256-bytes at a time.
+
+			s_vecs: [4]simd.u8x16 = ---
+			c_vecs: [4]simd.u8x16 = ---
+			m_vecs: [4]u8 = ---
+
+			// Scan 64-byte chunks, using 128-bit SIMD.
+			for nr_blocks := l / (4 * SIMD_REG_SIZE_128); nr_blocks > 0; nr_blocks -= 1 {
+				#unroll for j in 0..<4 {
+					s_vecs[j ]= intrinsics.unaligned_load(cast(^simd.u8x16)raw_data(s[i+j*SIMD_REG_SIZE_128:]))
```
Feoramund, line 418:
```suggestion
					s_vecs[j]= intrinsics.unaligned_load(cast(^simd.u8x16)raw_data(s[i+j*SIMD_REG_SIZE_128:]))
```

core/bytes/bytes.odin
Yawning: More things in core can be made contextless, but I reverted this for now to delay it for a later day when a pass can be done over all the things.

core/bytes/bytes.odin
Yawning: Looks like it's a wash between VPBROADCASTB and VPERM2I128 (and a gain over VINSERTI128), changed (since it's also more readable).

PR #4104, opened 2024-08-18
query brew for llvm-config location if possible

laytan:
Rewriting the install docs, this allows us to go with brew's recommendation of not adding the llvm directory to the user's $PATH.

### DIFF
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -23,6 +23,14 @@ error() {
 	exit 1
 }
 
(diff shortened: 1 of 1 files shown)


build_odin.sh
```diff
@@ -23,6 +23,19 @@ error() {
 	exit 1
 }
 
+# Brew advices people not to add llvm to their $PATH, so try and use brew to find it.
```
Kelimion, line 26: advises

PR #4110, opened 2024-08-19
Add support for Linux on os/sys/posix termios.

andradei:
- Add support for [POSIX compliant](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/termios.h.html) linux flag values for termios
- Add enum for [tcsetattr](https://pubs.opengroup.org/onlinepubs/9699919799/functions/tcsetattr.html)'s `optional_actions` paremeter
- Minor styling changes on Darwin and BSDs `when` blocks

### DIFF
--- a/core/sys/posix/termios.odin
+++ b/core/sys/posix/termios.odin
@@ -88,7 +88,7 @@ foreign lib {
 
 	[[ More; https://pubs.opengroup.org/onlinepubs/9699919799/functions/tcsetattr.html ]]
 	*/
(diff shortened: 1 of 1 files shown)


laytan:
The PR looks good, verified it is correct too (glibc and musl).

I could just merge it, but it won't be usable until the rest of the package is implemented.

andradei:
@laytan I pulled just this part and could already a lot with it. Can the rest of the `posix` linux support be implemented on separate PRs so it's easier to review? Either way I'm willing to keep working on it until it's completed.

laytan:
Sure, that's alright, if you mark it as ready I can merge it.

PR #4111, opened 2024-08-19
fix simd var typo

RoryO:
Found when building OLS. I have no idea how this compiles or how to get the compiler in the path to test it.


Kelimion:
Good find. Thanks!

PR #4113, opened 2024-08-20
Corrects ColorToInt signature to return int instead of uint

jolson88:
I think this could definitely use a review from somebody with more experience than me. I was working through raygui examples when I discovered that I couldn't port one. The issue came down to being unable to use `GuiSetStyle` in raygui with a raylib color. This is because GuiSetStyle's value param is a `c.int` where as `ColorToInt` in raylib is a `c.uint.`

At first glance, it would make sense for Color to be an unsigned integer since it's basically 4 8-bit integers (R/G/B/A). But as shown below, the actual raylib ABI is an `int`, not an `unsigned int`. The implementation also explicitly casts to (int).

I couldn't find a way for the compiler to let me to force the color white (`0xffffffff`) into the `c.int` parameter in `GuiSetStyle` of raygui as it always reports it as (validly) not being able to fit into the max value of a i32. For example, to be able to set the color of the lines drawn by `GuiGrid()`, you have to set the default LINE_COLOR property:

```
    // Not possible before this change
    rl.GuiSetStyle(.DEFAULT, i32(rl.GuiDefaultProperty.LINE_COLOR), rl.ColorToInt(rl.WHITE))
```

If folks feel strongly about this having the "correct" type in Odin, even though it appears to conflict with raylib ABI, and there's a workaround, I could use that instead.

# Description

The current C definition of ColorToInt in the raylib.h header itself is: `RLAPI int ColorToInt(Color color);`

Though it feels improper (and perhaps relies on undefined behavior?), the raylib implementation does indeed cast an unsigned integer to a signed one:

https://github.com/raysan5/raylib/blob/7fab03c0b4f079a6450f03d82ca17f2e18be3477/src/rtextures.c#L4893
```
int ColorToInt(Color color)
{
    int result = 0;

    result = (int)(((unsigned int)color.r << 24) |
                   ((unsigned int)color.g << 16) |
                   ((unsigned int)color.b << 8) |
                    (unsigned int)color.a);

    return result;
}
```

This is unfortunately a breaking change for users of this package. I'm unsure of whether a different approach would be better. This incorrect signature made it impossible, from what I could tell, to be able to set default control colors in raygui.

```
    rl.GuiSetStyle(.DEFAULT, i32(rl.GuiDefaultProperty.LINE_COLOR), rl.ColorToInt(rl.WHITE))
```

With the previously incorrect signature, I couldn't find a way to cast a raylib color to i32 in a forced way. Perhaps my own Odin knowledge shortcoming.

### DIFF
--- a/vendor/raylib/raylib.odin
+++ b/vendor/raylib/raylib.odin
@@ -1366,7 +1366,7 @@ foreign lib {
 	// Color/pixel related functions
 
 	Fade                :: proc(color: Color, alpha: f32) -> Color ---                  // Get color with alpha applied, alpha goes from 0.0f to 1.0f
-	ColorToInt          :: proc(color: Color) -> c.uint ---                             // Get hexadecimal value for a Color
+	ColorToInt          :: proc(color: Color) -> c.int ---                              // Get hexadecimal value for a Color
 	ColorNormalize      :: proc(color: Color) -> Vector4 ---                            // Get Color normalized as float [0..1]
 	ColorFromNormalized :: proc(normalized: Vector4) -> Color ---                       // Get Color from normalized values [0..1]
 	ColorToHSV          :: proc(color: Color) -> Vector3 ---                            // Get HSV values for a Color, hue [0..360], saturation/value [0..1]


flysand7:
`i32` and `u32` have the same ABI. Odin assumes the two's complement representation for signed integer, as do the ABIs for all the platforms that Odin ships on. The reason raylib went with signed integer is probably to reduce the warnings from C compilers thinking that since integer literals are signed integers, they trivially overflow for colors like white (`0xffffffff`), but that's my speculation.

So yeah my take on this is to use `u32` for color. This does not violate the ABI and works better with Odin's type system. I did notice the issue with `GuiSetStyle` and excessive casts, and kinda forgot about it, I forgot why I didn't do a PR...

Now thinking about it it was right before raylib 5.0, and someone else wanted to update the bindings, which I just said them to do in hopes they would do the proper adjustments to types... guess that part didn't happen

jolson88:
> `i32` and `u32` have the same ABI. Odin assumes the two's complement representation for signed integer, as do the ABIs for all the platforms that Odin ships on. The reason raylib went with signed integer is probably to reduce the warnings from C compilers thinking that since integer literals are signed integers, they trivially overflow for colors like white (`0xffffffff`), but that's my speculation.
> 
> So yeah my take on this is to use `u32` for color. This does not violate the ABI and works better with Odin's type system. I did notice the issue with `GuiSetStyle` and excessive casts, and kinda forgot about it, I forgot why I didn't do a PR...
> 
> Now thinking about it it was right before raylib 5.0, and someone else wanted to update the bindings, which I just said them to do in hopes they would do the proper adjustments to types... guess that part didn't happen

Thanks for sharing! That's great to know and exactly the type of feedback I was hoping for :D.

Do you have an idea on how I can get around the compilation failure of not allowing the full color value to be cast as an i32? In order to use `GuiSetStyle` with existing conditions, I have to force that conversion for raylib to work properly. On my machine at least, the other problem is that I think the underlying graphics format is a format where I don't believe Alpha is in the most significant digit place, so I have been unable to get the color white working via GuiSetStyle.

If there's a good workaround, I can then just submit an example to the examples repos that shows how to use GuiSetStyle given the constraint.

jolson88:
Nevermind, my failure was because I was trying to get "clever" and hard-code white as 0xffffffff (which resulted in the code `i32(0xffffffff)` which was failing.

I was able to get it working using ColorToInt with the `rl.WHITE`. I must have missed it with it being so late at night :P. 

```
rl.GuiSetStyle(.DEFAULT, i32(rl.GuiDefaultProperty.LINE_COLOR), i32(rl.ColorToInt(rl.WHITE)))
```

Hence, I'm closing this PR :). I'll look at introducing a separate PR for `examples` repo that shows GuiSetStyle in action to change color.

PR #4116, opened 2024-08-20
Add table-driven (in)definite article to some errors.

Kelimion:
And a handful of em-dashes added where applicable, e.g.
```
Error: Cannot assign 'f32' which is a type in procedure argument
```
Becomes:
```
Error: Cannot assign 'f32' — a type — to a procedure argument 
```

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -1071,16 +1071,19 @@ gb_internal void check_assignment(CheckerContext *c, Operand *operand, Type *typ
 		return;
 	}
 
+	// Grab definite or indefinite article matching `context_name`, or "" if not found.
+	String article = error_article(context_name);
+
 	if (is_type_untyped(operand->type)) {
 		Type *target_type = type;
 		if (type == nullptr || is_type_any(type)) {
 			if (type == nullptr && is_type_untyped_uninit(operand->type)) {
-				error(operand->expr, "Use of --- in %.*s", LIT(context_name));
+				error(operand->expr, "Use of --- in %.*s%.*s", LIT(article), LIT(context_name));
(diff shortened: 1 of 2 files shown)


flysand7:
For terminals that don't support UTF-8 (cough cough windows cmd with default settings on foreign computers cough cough) -- maybe we should just stick with ASCII?

Feoramund:
Two dashes (` -- `) should do the same job.

src/check_expr.cpp
```diff
@@ -1163,20 +1167,23 @@ gb_internal void check_assignment(CheckerContext *c, Operand *operand, Type *typ
...
+			      "Cannot assign built-in procedure '%s' to %.*s%.*s",
 			      expr_str,
+			      LIT(article),
 			      LIT(context_name));
 			break;
 		case Addressing_Type:
 			if (is_type_polymorphic(operand->type)) {
 				error(operand->expr,
-				      "Cannot assign '%s' which is a polymorphic type in %.*s",
+				      "Cannot assign '%s' — a polymorphic type — to %.*s%.*s",
```
gingerBill, line 1178: I recommend using a hyphen `-`, even if an em dash is more correct. And if you were going to be very correct, you shouldn't have any space around the em dash either.

Kelimion: > And if you were going to be very correct, you shouldn't have any space around the em dash either.

I'm aware and tried it. It looks worse.

gingerBill: Just use commas instead? e.g.
```
Cannot assign '%s', a polymorphic type, to %.*s%.*s
```

Kelimion: I just changed it to `-` at your recommendation?

PR #4118, opened 2024-08-21
Linux POSIX support

andradei:
Add the missing Linux-specific support for many POSIX headers.


core/sys/posix/dirent.odin
```diff
@@ -193,11 +193,21 @@ when ODIN_OS == .Darwin {
...
+	}
+
+} else when ODIN_OS == .Linux {
+
+	dirent :: struct {
+		d_ino: ino_t,                   /* [PSX] file number of entry */
+		d_off: off_t,                   /* directory offset of the next entry */
+		d_reclen: c.uint16_t,           /* length of this record */
+		d_type: D_Type,                 /* file type  */
+		d_name: [256]c.char `fmt:"s,0"` /* [PSX] entry name */
```
Feoramund, line 210: This entry needs to end with a comma.

core/sys/posix/errno.odin
```diff
@@ -367,6 +367,97 @@ when ODIN_OS == .Darwin {
...
+	ESTALE          :: 116
+
+	EDQUOT          :: 122
+	ECANCELED       :: 125
+
+	EOWNERDEAD      :: 130
+	ENOTRECOVERABLE :: 131
+
+	// NOTE: Note defined for linux
+	ENOTSUP         :: -1
```
Feoramund, line 460:
```suggestion
	// NOTE: Not defined for Linux
	ENOTSUP         :: -1
```

Moreover:
```
EOPNOTSUPP      Operation not supported on socket (POSIX.1-2001).

                (ENOTSUP and EOPNOTSUPP have the same  value  on  Linux,
                but  according  to  POSIX.1 these error values should be
                distinct.)
```
Source: `man 3 errno`.

Looking through the rest of the `errno.odin` file, it looks like the other platforms alias `ENOTSUP` and `EOPNOTSUPP` together. Might be best to do the same for Linux to be consistent.

andradei: Thanks you for this. Very helpful.

core/sys/posix/sys_stat.odin
```diff
@@ -427,6 +427,9 @@ when ODIN_OS == .Darwin {
 	UTIME_NOW  :: -2
 	UTIME_OMIT :: -1
 
+} else when ODIN_OS == .Linux {
+	ino_t   :: distinct c.unit64_t
+	ino_t32 :: distinct c.unit32_t
```
gingerBill, line 432: Typos here. `unit` -> `uint`, and I'd also prefer `u64` and `u32` instead too in general

andradei: @gingerBill Thanks for spotting this.
I was following the style of the package with the types but would love to use the `builtin` types directly.
How about making this same change on the entire file and places of the package where it makes sense?

gingerBill: Honestly, it doesn't really matter which you use. It's more of an aesthetic judgment as to which is better. As long as you are consistent. with the `core` packages, I'd probably prefer `u64` over `c.uint64_t`, but if you have one of the more exotic types, then sure use the `c.*_t` equivalent.

core/sys/posix/fcntl.odin
```diff
@@ -410,6 +410,9 @@ when ODIN_OS == .Darwin {
 		l_whence: c.short,   /* [PSX] flag (Whence) of starting offset */
 	}
 
+} else when ODIN_OS == .Linux {
+	off_t :: distinct c.uint64_t
```
laytan, line 414: no need for off_t32 and afaict off_t is not unsigned

core/sys/posix/sys_stat.odin
```diff
@@ -427,6 +427,9 @@ when ODIN_OS == .Darwin {
 	UTIME_NOW  :: -2
 	UTIME_OMIT :: -1
 
+} else when ODIN_OS == .Linux {
+	ino_t   :: distinct u64
+	ino_t32 :: distinct u32
```
laytan, line 432: no need for ino_t32

core/sys/posix/dirent.odin
```diff
@@ -193,11 +193,41 @@ when ODIN_OS == .Darwin {
...
+		d_type:    D_Type,                  /* file type  */
+		d_name:    [512]c.char `fmt:"s,0"`, /* [PSX] entry name */
+	}
+
+} else when ODIN_OS == .Linux {
+
+	when ODIN_ARCH == .i386 || ODIN_ARCH == .wasm32 || ODIN_ARCH == .arm32 {
+
+		dirent :: struct {
+			d_ino: ino_t32,                  /* [PSX] file number of entry */
```
laytan, line 208: afaict even on 32 bit these fields are 64 bit (the normal ino_t and off_t types), so I think the struct you are using below for 64 bit should be used for all linux

core/sys/posix/dirent.odin
```diff
@@ -193,11 +193,41 @@ when ODIN_OS == .Darwin {
...
+		d_ino:     ino_t,                   /* [PSX] file number of entry */
+		d_reclen:  c.uint16_t,              /* length of this record */
+		d_namelen: c.uint16_t,              /* length of string in d_name */
+		d_type:    D_Type,                  /* file type  */
+		d_name:    [512]c.char `fmt:"s,0"`, /* [PSX] entry name */
+	}
+
+} else when ODIN_OS == .Linux {
+
+	when ODIN_ARCH == .i386 || ODIN_ARCH == .wasm32 || ODIN_ARCH == .arm32 {
```
laytan, line 205: wasm is not an architecture for linux

core/sys/posix/dirent.odin
```diff
@@ -193,11 +193,41 @@ when ODIN_OS == .Darwin {
...
+
+		dirent :: struct {
+			d_ino: ino_t32,                  /* [PSX] file number of entry */
+			d_off: off_t32,                  /* directory offset of the next entry */
+			d_reclen: c.uint16_t,            /* length of this record */
+			d_type: D_Type,                  /* file type  */
+			d_name: [256]c.char `fmt:"s,0"`, /* [PSX] entry name */
+		}
+	} else when ODIN_ARCH == .amd64 || ODIN_ARCH == .wasm64p32 || ODIN_ARCH == .arm64 {
```
laytan, line 214: wasm is not an architecture for linux

core/sys/posix/dirent.odin
```diff
@@ -193,11 +193,41 @@ when ODIN_OS == .Darwin {
...
+		dirent :: struct {
+			d_ino: ino_t,                    /* [PSX] file number of entry */
+			d_off: off_t,                    /* directory offset of the next entry */
+			d_reclen: c.uint16_t,            /* length of this record */
+			d_type: D_Type,                  /* file type  */
+			d_name: [256]c.char `fmt:"s,0"`, /* [PSX] entry name */
+		}
+	}
+
+	dirent64 :: struct {
```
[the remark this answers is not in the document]
andradei: I was following the `dirent.h` definitions on Linux, on my system this options exists (see below). But if it's still not needed because it's actually not POSIX-specified, I'll remove it.

image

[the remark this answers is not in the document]
flysand7: Yeah the `*64` types are meant for 32-bit platforms that want to use 64-bit functions and types, if it's not posix specified, Odin doesn't really need that because all Odin platforms support 64-bit types and we don't need to provide it as an option, we can just force 64-bit for all of these structs.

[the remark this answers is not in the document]
laytan: Right, we don't do anything in this package that isn't posix specified.

core/sys/posix/dirent.odin
```diff
@@ -193,13 +193,23 @@ when ODIN_OS == .Darwin {
...
+		d_ino:     ino_t,                   /* [PSX] file number of entry */
+		d_reclen:  c.uint16_t,              /* length of this record */
+		d_namelen: c.uint16_t,              /* length of string in d_name */
+		d_type:    D_Type,                  /* file type  */
+		d_name:    [512]c.char `fmt:"s,0"`, /* [PSX] entry name */
 	}
 
+} else when ODIN_OS == .Linux {
+
+		dirent :: struct {
```
laytan, line 205: Could you align the types and comments like on the others?

laytan:
Also, are you planning on keeping this PR going or should we merge this and do other PRs for the rest?

core/sys/posix/dirent.odin
andradei: Done.

andradei:
@laytan Now that I have a better idea of the work to be done, 'll keep this PR going (unless I find a change that's so massive it would be saner if it had its own PR).

core/sys/posix/sys_sem.odin
```diff
@@ -127,6 +127,36 @@ when ODIN_OS == .Darwin || ODIN_OS == .FreeBSD || ODIN_OS == .NetBSD || ODIN_OS
...
+	GETALL  :: 13
+	GETNCNT :: 14
+	GETZCNT :: 15
+	SETVAL  :: 16
+	SETALL  :: 17
+
+	semid_ds :: struct {
+		sem_perm:  ipc_perm,  // [PSX] operation permission structure
+		sem_otime: time_t,    // [PSX] last semop()
+    __sem_otime_high: c.ulong,
```
flysand7, line 146: Indentation issue, this should be a tab. Also check some other lines in this struct

core/sys/posix/signal.odin
```diff
@@ -1126,6 +1136,141 @@ when ODIN_OS == .Darwin {
...
+	MINSIGSTKSZ :: 2048
+	SIGSTKSZ    :: 8192
+
+	stack_t :: struct {
+		ss_sp:    rawptr,   /* [PSX] stack base or pointer */
+		ss_flags: SS_Flags, /* [PSX] flags */
+		ss_size:  c.size_t, /* [PSX] stack size */
+	}
+
+	// WARNING: This implementaion might be completely wrong and need to be reviewed and corrected.
```
andradei, line 1205: Requesting help figuring out how to translate this from C's implementation to Odin as I doubt this implementation is it.

andradei:
@laytan As I get experience writing these POSIX implementations, I realize there is a lot of work still to be done. This PR became very, very big.

I'd like to consider it ready for review for what it implements. I'll create smaller PRs with another chunk of POSIX implementations after this is fully reviewed, approved, and merged. It'll be smaller too, with 1-3 files in `core/sys/posix` edited to add Linux support.

This way we can keep this process saner for everyone.

laytan:
@andradei instead of putting review comments everywhere and taking the time to explain every little thing and having you then make the changes I opted to just commit my reviews/corrections/cleanups here directly. Let me know about any questions regarding the changes. I will merge this once those questions have been answered/if you got no questions.

Still need to do a couple headers I see and then we can also enable the tests etc. I see from your last comment that you are picking those up in separate PRs which is fine by me.

andradei:
@laytan I left questions on your commit. Thank you for reviewing this.

core/sys/posix/netinet_in.odin
```diff
@@ -45,20 +45,18 @@ when ODIN_OS == .Darwin || ODIN_OS == .FreeBSD || ODIN_OS == .NetBSD || ODIN_OS
...
 		sockaddr_in :: struct {
 			sin_family: sa_family_t, /* [PSX] AF_INET (but a smaller size) */
 			sin_port:   in_port_t,   /* [PSX] port number */
 			sin_addr:   in_addr,     /* [PSX] IP address */
-			sin_zero: [size_of(sockaddr) -
-				u16 -
-				size_of(in_port_t) -
-				size_of(in_addr)]c.char,
+			sin_zero:   [8]c.char,
 		}
```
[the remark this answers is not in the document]
laytan: IIRC (from when I did it) that whole expression comes out as `8` so they should be equivalent.

core/sys/posix/pthread.odin
```diff
@@ -521,7 +521,7 @@ when ODIN_OS == .Darwin {
 	PTHREAD_CANCEL_ENABLE       :: 0
 	PTHREAD_CANCEL_DISABLE      :: 1
 
-	PTHREAD_CANCELED :: rawptr(uintptr(-1))
```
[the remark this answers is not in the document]
laytan: It should've shown on a local Odin check, ci is not running on the linux part of this package though.

core/sys/posix/netinet_in.odin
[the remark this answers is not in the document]
andradei: I thought having that be a 1:1 formula conversion from C would make it automatically compatible in case this ever changes in glibc. But I don't know how often, if ever, that changes due to backwards compatibility.
So I'll mark this as resolved.

core/sys/posix/pthread.odin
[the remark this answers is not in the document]
andradei: It didn't show even a single warning on `odin check`.

PR #4123, opened 2024-08-23
orca: implement core:time and core:log

laytan:
Turns out importing `fmt` from `sys/orca` was a bad idea due to cyclic imports (which made it impossible to implement `time`). What I did now is make the log package plug into the orca log nicely which should still allow nice usage without the imports.

```odin
package main

import    "core:log"
import    "core:time"
import oc "core:sys/orca"

main :: proc() {
	context.logger = oc.create_odin_logger()	

	log.infof("The current time is: %v", time.now())

	start := time.tick_now()
	sum := 0
	for i in 0..<100 {
		sum += i
	}

	log.infof("Calculated %v in %v", sum, time.tick_since(start))
}
```

### DIFF
--- a/core/log/file_console_logger.odin
+++ b/core/log/file_console_logger.odin
@@ -1,4 +1,5 @@
 //+build !freestanding
+//+build !orca
 package log
 
 import "core:encoding/ansi"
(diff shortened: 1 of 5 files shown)


Skytrias (approved):
Looking good, great job!

PR #4134, opened 2024-08-24
Fix binding: Make rl.LoadImageAnim take a pointer instead of a multi-pointer. 

karl-zylinski:
The param `frames` is just for returning an int, it's not for passing a multi-pointer into the proc.


bersK:
Can confirm 👍:
v5.0: https://github.com/raysan5/raylib/blob/5.0/src/rtextures.c#L401
v4.5: https://github.com/raysan5/raylib/blob/4.5.0/src/rtextures.c#L293

PR #4136, opened 2024-08-25
[os2/process]: Don't free process info fields in partial success scenarios

flysand7:
This PR makes the behavior of `get_process_info` procedure consistent with linux and other platforms' implementations by not freeing all the fields if an error has occurred during retrieval of requested information, after other information has been successfully retrieved.

It also allows for the procedure to complete with "partial success" state, where the function will complete successfully, even if not all of the requested information has been obtained.

The error handling has been altered according to the following logic. Depending on which errors has happened internally, the handling of it will differ, according to the following rules:

- Allocator errors: `process_info` returns immediately with the error
- Process isn't running: `process_info` returns immediately with `General_Error.Not_Found`
- Permissions and other similar issues: `process_info` records the error, but continues to fetch other data
- Unexpected situations (address in PEB is `nil`, length of command line is 0, etc): Panic

Which is the behavior currently implemented for other platforms (more or less)

### DIFF
--- a/core/os/os2/process.odin
+++ b/core/os/os2/process.odin
@@ -169,7 +169,7 @@ Process_Info :: struct {
 
 	Use `free_process_info` to free the memory allocated by this procedure. The
 	`free_process_info` procedure needs to be called, even if this procedure
-	returned an error, as some of the fields have been allocated.
+	returned an error, as some of the fields may have been allocated.
 
 	**Note**: The resulting information may or may contain the fields specified
 	by the `selection` parameter. Always check whether the returned
@@ -190,7 +190,7 @@ process_info_by_pid :: proc(pid: int, selection: Process_Info_Fields, allocator:
 
 	Use `free_process_info` to free the memory allocated by this procedure. The
 	`free_process_info` procedure needs to be called, even if this procedure
-	returned an error, as some of the fields have been allocated.
+	returned an error, as some of the fields may have been allocated.
 
 	**Note**: The resulting information may or may contain the fields specified
 	by the `selection` parameter. Always check whether the returned
@@ -210,7 +210,7 @@ process_info_by_handle :: proc(process: Process, selection: Process_Info_Fields,
 
 	Use `free_process_info` to free the memory allocated by this procedure. The
 	`free_process_info` procedure needs to be called, even if this procedure
-	returned an error, as some of the fields have been allocated.
+	returned an error, as some of the fields may have been allocated.
 
 	**Note**: The resulting information may or may contain the fields specified
 	by the `selection` parameter. Always check whether the returned
--- a/core/os/os2/process.odin
+++ b/core/os/os2/process.odin
@@ -166,15 +166,15 @@ Process_Info :: struct {
 
 	This procedure obtains an information, specified by `selection` parameter of
 	a process given by `pid`.
-	
-	Use `free_process_info` to free the memory allocated by this procedure. In
-	case the function returns an error it may only have been an error for one part
-	of the information and you would still need to call it to free the other parts.
+
+	Use `free_process_info` to free the memory allocated by this procedure. The
+	`free_process_info` procedure needs to be called, even if this procedure
+	returned an error, as some of the fields have been allocated.
 
 	**Note**: The resulting information may or may contain the fields specified
 	by the `selection` parameter. Always check whether the returned
 	`Process_Info` struct has the required fields before checking the error code
-	returned by this function.
+	returned by this procedure.
 */
 @(require_results)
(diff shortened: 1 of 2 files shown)


flysand7:
I'm not entirely sure how to correctly handle the errors that occur when the process information can not be retrieved. The user requests what information they want to retrieve from the process, and on most platforms different information is obtained differently. For example, on windows information about parent process can only be obtained through a snapshot, and process' command line is obtained by reading PEB. So it is possible that we may be able to obtain only partial information about some processes, for example it's parent process but not the cmdline, in case of `ntoskrnl.exe`, which doesn't allow userspace processes to read its memory.

I generally view the three possible ways in which obtaining the information can result in an error:

- User allocator failure
- Process is not running
- Permissions issue (trying to read info or memory of privileged processes)
- Anything else (panic)

In case an allocation has failed, the function should return immediately, returning the information that has been retrieved successfully so far, while also returning the error code for an allocation failure.

In case the target process is not running, this could mean a race condition, i.e. we obtained 50% of the requested information, process stopped, and when trying to get the other information, by e.g. opening a process handle we find out that it isn't running anymore. I think that race conditions in this function should be eliminated at a root, by opening the process handle right away and obtaining as much information as possible via that handle.

If it is a permissions issue, this error can be ignored and we should continue fetching the other information.

If any other error occurs (such as pointers in PEB being invalid, etc) this should be a panic, as it's probably a bug in the implementation and not accounting for certain scenarios.

flysand7:
It's interesting that you have the same thoughts on this. Guess it must be correct then haha, since I haven't looked at process code for darwin/bsd.

core/os/os2/process.odin
```diff
@@ -166,15 +166,15 @@ Process_Info :: struct {
...
 	This procedure obtains an information, specified by `selection` parameter of
 	a process given by `pid`.
-	
-	Use `free_process_info` to free the memory allocated by this procedure. In
-	case the function returns an error it may only have been an error for one part
-	of the information and you would still need to call it to free the other parts.
+
+	Use `free_process_info` to free the memory allocated by this procedure. The
+	`free_process_info` procedure needs to be called, even if this procedure
+	returned an error, as some of the fields have been allocated.
```
laytan, line 172: I think this should be "the fields may have been allocated."

I also think the docs should say something about how even if an error is returned, the `.fields` can be checked for things that have been successfully retrieved, and that the error is for the other fields that couldn't be retrieved.

It does get pretty lengthy but it is a weird procedure.

flysand7: I am contemplating whether it was a good choice for the API, but still not entirely sure that's the best solution. Thinking about windows, you get process information from different places, depending on which information you want to retrieve.

For example, if you want to know the command line of the process, on windows you'd read the other process' PEB. If you want to know parent pid of a process, the only way I found without the NTDLL weirdness is to make a toolhelp32 process snapshot. And on linux all of these things are done by reading from different files of the `/proc` filesystem (`/proc/cmdline`, `/proc/stat`, etc).

With this API I didn't want it to be too slow by doing every single thing a user "might" want to have and filling out the process info struct completely. So I added a mechanism that would allow the user to "select" what info they want, so that the inner bodies of the `process_info` procedures can be optimized to only retrieve the things the user has requested.

...

Got distracted, lost my train of thought.

So if the procedure, instead of how it works currently, just retrieved all the information, that would be a waste of resources, i.e. "why is this reading process memory and failing when I just asked for parent process id", and if it was made in separate calls, it would lead to a question of "why is the process being opened and closed 8 times for each process im asking".

So I decided on an API that would allow the user to chose what they want and what they don't want, to provide the best speed and cleanliness for whatever set of information you may want to retrieve.

This was in response to "this is a weird procedure", because I do think it is... unique, and I'm wondering if I'm overthinking things. It's not exactly the type of thing you do millions times per second, so perhaps a better way to think about it is in terms of how easy it would be to handle errors instead? In whichcase maybe separating API out into separate procedures is a better idea. What are your thoughts?

laytan: IMO the API is pretty nice as is. I like the idea of requesting fields and getting whatever was available back, instead of an error because one of the things wasn't available. Splitting it up comes with it's own downsides as you also wrote.

Thinking of a `top` like program, you would probably request every field and display whatever was available. Other uses could request whatever they really need and take the returned error as a "fact" and bubble it up.

In terms of handling errors, the only information you might lose in this API is if multiple fields were erroneous in different ways, just one of the errors would be returned back, IMO that is acceptable. The general flow I see would be: request x, y, z; get x back (in the .fields bit_set); see the error for why y and z weren't retrieved. 

Some functionality is just a bit unique and requires more explaining / documentation.

flysand7: I won't split it up, at least not now. I think this PR is ready for merging

PR #4140, opened 2024-08-25
Mach Process control

colrdavidson:


flysand7:
eeeeeeeeeeeeeeeeeh, macharena (kyaah)!

base/intrinsics/intrinsics.odin
```diff
@@ -361,6 +361,9 @@ wasm_memory_atomic_notify32 :: proc(ptr: ^u32, waiters: u32) -> (waiters_woken_u
 x86_cpuid  :: proc(ax, cx: u32) -> (eax, ebx, ecx, edx: u32) ---
 x86_xgetbv :: proc(cx: u32) -> (eax, edx: u32) ---
 
+// Arm64 targets
+arm64_read_cycle_counter_frequency :: proc() -> i64 ---
```
laytan: Is this never something any other architecture will have? If it is, I don't like the arm64 prefix, could make it return 0 on targets that don't have it. I was thinking of riscv that might have this.

I would also put this with `read_cycle_counter` in this file.

src/check_builtin.cpp
```diff
@@ -4713,6 +4713,7 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 		}
 		break;
 
+	case BuiltinProc_arm64_read_cycle_counter_frequency:
```
laytan: If this stays arm64 only, you are missing a check and error for other targets.

core/os/os_darwin.odin
```diff
@@ -677,6 +679,39 @@ get_last_error_string :: proc() -> string {
 	return string(_darwin_string_error(__error()^))
 }
 
+_spawn :: #force_inline proc(path: string, args: []string, envs: []string, file_actions: rawptr, attributes: rawptr, is_spawnp: bool) -> (posix.pid_t, Error) {
```
laytan: We shouldnt really add anything to `core:os` if we can help it, since it will be replaced in the future.

core/sys/darwin/mach_darwin.odin
```diff
@@ -5,15 +5,55 @@ foreign import mach "system:System"
...
+thread_list_t  :: [^]thread_act_t
+vm_region_recurse_info_t :: distinct ^i32
+task_info_t :: distinct ^i32
+
+MACH_PORT_NULL :: 0
+MACH_PORT_DEAD :: ~mach_port_t(0)
+
+MACH_MSG_PORT_DESCRIPTOR :: 0
+
+MACH_SEND_MSG     :: 0x00000001
```
laytan: Can this be a bit_set?

core/sys/darwin/mach_darwin.odin
```diff
@@ -5,15 +5,55 @@ foreign import mach "system:System"
...
+MACH_RCV_MSG      :: 0x00000002
+MACH_SEND_TIMEOUT :: 0x00000010
+MACH_RCV_TIMEOUT  :: 0x00000100
+
+MACH_MSG_TYPE_COPY_SEND :: 19
+MACH_MSG_TYPE_MAKE_SEND :: 20
+MACH_MSGH_BITS_COMPLEX :: 0x80000000
+
+MACH_PORT_RIGHT_SEND    :: 0
```
laytan: Can this be an enum?

core/sys/darwin/mach_darwin.odin
```diff
@@ -5,15 +5,55 @@ foreign import mach "system:System"
...
+MACH_PORT_RIGHT_RECEIVE :: 1
+
+VM_INHERIT_SHARE        :: 0
```
laytan: There is an enum for this already, a bit lower down in this file.

core/sys/darwin/mach_darwin.odin
laytan: In general, if an enum or bit set makes sense I highly prefer it over bare constants you have to figure out how/where they are valid.

core/sys/darwin/mach_darwin.odin
```diff
@@ -31,9 +71,204 @@ mach_port_name_t :: distinct c.uint
...
+	aot_info_array_change_timestamp:     u64,
+	aot_shared_cache_base_address:       u64,
+	aot_shared_cache_uuid:            [16]u8,
+}
+
+
 @(default_calling_convention="c")
 foreign mach {
-	mach_task_self :: proc() -> mach_port_t ---
+	mach_task_self     :: proc() -> task_t ---
```
laytan: This is not correct right? I see it returning `mach_port_t` on my end.

core/sys/darwin/mach_darwin.odin
```diff
@@ -44,9 +279,11 @@ foreign mach {
 
 	semaphore_wait :: proc(semaphore: semaphore_t) -> Kern_Return ---
 
-	vm_allocate :: proc (target_task : vm_map_t, address: ^vm_address_t, size: vm_size_t, flags: VM_Flags) -> Kern_Return ---
```
laytan, line 47: Why remove this?

core/sys/darwin/mach_darwin.odin
```diff
@@ -44,9 +279,11 @@ foreign mach {
...
+	thread_get_state :: proc(thread: thread_act_t, flavor: i32, thread_state: thread_state_t, old_state_count: ^u32) -> kern_return_t ---
+	thread_info :: proc(thread: thread_act_t, flavor: u32, thread_info: ^thread_identifier_info, info_count: ^u32) -> kern_return_t ---
 
-	vm_deallocate :: proc(target_task: vm_map_t, address: vm_address_t, size: vm_size_t) -> Kern_Return ---
```
laytan, line 49: Why remove this?

core/sys/darwin/mach_darwin.odin
```diff
@@ -31,9 +71,204 @@ mach_port_name_t :: distinct c.uint
...
+	aot_shared_cache_base_address:       u64,
+	aot_shared_cache_uuid:            [16]u8,
+}
+
+
 @(default_calling_convention="c")
 foreign mach {
-	mach_task_self :: proc() -> mach_port_t ---
+	mach_task_self     :: proc() -> task_t ---
+	mach_msg           :: proc(header: rawptr, option: mach_msg_option_t, send_size: u32, receive_limit: u32, receive_name: mach_port_t, timeout: u32, notify: mach_port_t) -> kern_return_t ---
```
laytan: Everything returning `kern_return_t` can return `Kern_Return` instead

core/sys/posix/spawn.odin
```diff
@@ -0,0 +1,14 @@
...
+import "core:c"
+
+when ODIN_OS == .Darwin {
+	foreign import lib "system:System.framework"
+} else {
+	foreign import lib "system:c"
+}
+
+foreign lib {
+	posix_spawn :: proc(pid: ^pid_t, path: cstring, file_actions: rawptr, attrp: rawptr, argv: [^]cstring, envp: [^]cstring) -> c.int ---
```
laytan: Can you add some docs like everything else in the package?

core/sys/posix/spawn.odin
```diff
@@ -0,0 +1,14 @@
...
+
+import "core:c"
+
+when ODIN_OS == .Darwin {
+	foreign import lib "system:System.framework"
+} else {
+	foreign import lib "system:c"
+}
+
+foreign lib {
```
laytan, line 9: These seems to return an error number, is `Errno` not better?

gingerBill:
There are a few things failing because of missing `import`s.

core/sys/darwin/mach_darwin.odin
colrdavidson: This is a rename to make it consistent with the rest of the binds.
vm_deallocate -> mach_vm_deallocate

core/sys/darwin/mach_darwin.odin
colrdavidson: This is a rename to make it consistent with the rest of the binds.
vm_allocate -> mach_vm_allocate

laytan: Is it, because `vm_allocate` and `mach_vm_allocate` are two different symbols on my end?

My googling tell me that `mach_vm_allocate` is a newer version that everyone should pretty much be using so I am fine with the patch in that regard.

core/sys/darwin/mach_darwin.odin
```diff
@@ -5,15 +5,55 @@ foreign import mach "system:System"
...
+MACH_PORT_DEAD :: ~mach_port_t(0)
+
+MACH_MSG_PORT_DESCRIPTOR :: 0
+
+MACH_SEND_MSG     :: 0x00000001
+MACH_RCV_MSG      :: 0x00000002
+MACH_SEND_TIMEOUT :: 0x00000010
+MACH_RCV_TIMEOUT  :: 0x00000100
+
+MACH_MSG_TYPE_COPY_SEND :: 19
```
colrdavidson: Not really sure how to handle msg types.
These have a lot of overlaps, and get used in an "anyone's guess" flexible way.
https://www.cs.cmu.edu/afs/cs/project/mach/public/src/mkernel/src/kernel/mach/message.h

The field is literally just a pile of bits
`send_msg.header.msgh_bits = darwin.MACH_MSG_TYPE_COPY_SEND | darwin.MACH_MSGH_BITS_COMPLEX`

colrdavidson: I've just made two enums that I cast to u32s locally for now.
This one is pretty gnarly, don't think there's really any great solutions here

src/llvm_backend_proc.cpp
```diff
@@ -2809,6 +2809,23 @@ gb_internal lbValue lb_build_builtin_proc(lbProcedure *p, Ast *expr, TypeAndValu
...
+			res.type = tv.type;
+
+			if (build_context.metrics.arch == TargetArch_arm64) {
+				LLVMTypeRef func_type = LLVMFunctionType(LLVMInt64TypeInContext(p->module->ctx), nullptr, 0, false);
+				bool has_side_effects = false;
+				LLVMValueRef the_asm = llvm_get_inline_asm(func_type, str_lit("mrs $0, cntfrq_el0"), str_lit("=r"), has_side_effects);
+				GB_ASSERT(the_asm != nullptr);
+				res.value = LLVMBuildCall2(p->builder, func_type, the_asm, nullptr, 0, "");
+			} else {
+				GB_PANIC("Unsupported architecture: %.*s", LIT(target_arch_names[build_context.metrics.arch]));
```
colrdavidson: I'm not sure this is the right move, but it's probably better than nothing?
Might be worth looking at our other arch-specific calls and adding unsupport arch there too, I don't think x86_cpuid / x86_xgetvb or friends check either, I used them for reference.

laytan: A panic isn't the best idea though, we class compiler panics as compiler bugs (ie if a user hits a cpp panic/assert it's a compiler bug) and this seems like something you can hit quite easily. I think you'd want a check in the checker and emit a proper error there instead. `wasm_memory_grow` has such a check and error message for example.

core/sys/darwin/mach_darwin.odin
laytan: Whatever you feel is best, with some apis it's just a mess and this seems like one of those.

PR #4143, opened 2024-08-25
Float emulation start for more embedded support

laytan:
Had a quick look at float emulation today so we can support use cases where targets don't support floats in the future. With the current progress in this PR a simple hello world compiles for riscv64 without floats:

```odin
package main

import "base:runtime"

main :: proc() {
	runtime.print_string("Hello")
}
```
`odin run test.odin -file -target:linux_riscv64 -no-crt -default-to-nil-allocator -reloc-mode:static -microarch:generic -target-features:"m"` (you want to have qemu user mode emulation if you are on another architecture for your host `sudo apt install qemu-user qemu-user-static`)

When you start importing more packages and using more float features you will keep getting more linker errors about functions that are needed. The basic process is adding them to the runtime package and to the checker (in order to add it to the final output).

We think the implementations can just be ported from LLVM's compiler-rt https://github.com/llvm/llvm-project/tree/main/compiler-rt which uses the same license as llvm as a backend itself, so there shouldn't be any licensing issues.

Once we have float emulation for one target, adding it for other targets should be fairly simple.

I am making this a draft so somebody else can pick it up if they want to, because I don't have much time to spend on this in the near future.

So if anybody wants to pick this up further, you can just take the changes here as a base and add to it.

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -1803,6 +1803,22 @@ gb_internal bool check_target_feature_is_superset_of(String const &superset, Str
 	return true;
 }
 
+gb_internal bool target_supports_f32() {
+	switch (build_context.metrics.arch) {
+	case TargetArch_riscv64:
+		return check_target_feature_is_enabled(str_lit("f"), nullptr);
+	}
+	return false;
+}
+
+gb_internal bool target_supports_f64() {
+	switch (build_context.metrics.arch) {
+	case TargetArch_riscv64:
+		return check_target_feature_is_enabled(str_lit("d"), nullptr);
+	}
+	return false;
+}
+
 // NOTE(Jeroen): Set/create the output and other paths and report an error as appropriate.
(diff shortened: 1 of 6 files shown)


src/build_settings.cpp
```diff
@@ -1803,6 +1803,22 @@ gb_internal bool check_target_feature_is_superset_of(String const &superset, Str
 	return true;
 }
 
+gb_internal bool target_supports_f32() {
+	switch (build_context.metrics.arch) {
+	case TargetArch_riscv64:
+		return check_target_feature_is_enabled(str_lit("f"), nullptr);
+	}
+	return false;
```
gingerBill, line 1811: Shouldn't this be `true`?

laytan: Yes!

Platin21:
Did already start with porting stuff from compiler-rt  the code is quite generic as in does implement f32 and f64 and f128 in one go and changes the types with defines.

PR #4152, opened 2024-08-26
Fix error for uuid to require 36 bytes instead 32 (4 bytes for dashes)

edyu:
uuid write/to_string functions actually need 36 bytes not 32 as 8-4-4-4-12 is 8 + 4 + 4 + 4 + 12 + (4) = 36 not 32 because you need the extra (4) for the dashes.


Feoramund:
Looks like the only thing that needed to be changed was the documentation and the assertion string. This otherwise looks fine.

PR #4158, opened 2024-08-28
Win32 clipboard procedures

sqrt-negativeone:


tests/core/sys/windows/test_clipboard.odin
```diff
@@ -0,0 +1,52 @@
...
+
+import "core:testing"
+import win32 "core:sys/windows"
+import "base:runtime"
+import "core:strings"
+import "core:mem"
+
+read_from_clipboard :: proc(wnd_handle: win32.HWND, allocator: runtime.Allocator) -> (result: string)
+{
+	if !win32.IsClipboardFormatAvailable(win32.CF_TEXT) do return;
```
Kelimion, line 12: Please avoid using `do` in the Odin codebase. Prefer `{ return }` if you want to keep it a one-liner.

tests/core/sys/windows/test_clipboard.odin
```diff
@@ -0,0 +1,52 @@
...
+		if clipboard_data != nil {
+			if cstr := cstring(win32.GlobalLock(win32.HGLOBAL(clipboard_data))); cstr != nil {
+				result = strings.clone_from_cstring(cstr, allocator);
+			}
+			win32.GlobalUnlock(win32.HGLOBAL(clipboard_data));
+		}
+	return;
+}
+
+write_to_clipboard :: proc(wnd_handle: win32.HWND, text: string)
```
Kelimion, line 27: You could possibly also make these procedures return `win32.BOOL` and go with win32.OpenClipboard(wnd_handle) or_return`, and so on.

tests/core/sys/windows/test_clipboard.odin
sqrt-negativeone: Oh I didn't know that, okay I'll fix it!

Kelimion: No problem. One last thing, can you remove the end of line semicolons? They're unnecessary.

The main codebase including tests is written to pass `-vet -strict-style -disallow-do`.

We have overlooked adding those switches to some of the CI tests, and so I'm adding and addressing them [here](https://github.com/odin-lang/Odin/pull/4160). But once that's finished and merged, your clipboard test will complain about the optional semicolons.

sqrt-negativeone: okay done!

Kelimion: Thanks. We've been quite good about it, it appears. When I added those flags to the remainder of the CI tests, it only flagged up one unused import.

sqrt-negativeone: That's great! 

Btw is there a reason why `do` is not allowed? My guess is to allow people who set the `disallow-do` flag to build their code without errors, is this correct or is there some other reason?

Kelimion: That is one of the main reasons, yes. The other is that it's part of the official code style.

Kelimion: The problem is that `do` is occasionally very useful in constructs like `for foo in bar do if foo & 1 == 1 { ... }`.

PR #4180, opened 2024-09-02
fix(os2): mkdir_all on path_posix

dvrd:
## Changes

* Fixes `mkdir_all` logic on `os2.path_posix`. Currently it does not execute correctly, since the `internal_mkdir_all` function checks the `dir` against the `path` instead of the `file` this means that it eventually tries to run `_mkdir` on empty strings and doesn't create any directory and returns a `.Not_exist` error


laytan:
It's usually nice to also explain what was wrong with it / on what inputs it failed, instead of just saying you fixed it.

dvrd:
> It's usually nice to also explain what was wrong with it / on what inputs it failed, instead of just saying you fixed it.



> It's usually nice to also explain what was wrong with it / on what inputs it failed, instead of just saying you fixed it.

Sure, I have just updated the description ✌🏼

PR #4188, opened 2024-09-03
Added `#any_int` to some `soa` procs to mirror the behaviour of their non `soa` counterparts

FrancisTheCat:
I also noticed that the `remove_*` procedures also take an index as a paramter that does not have `#any_int` applied to it, tho this is the case for both the `soa` and non `soa` versions so it might be intentional?


Feoramund:
I can't think of any reason off the top of my head why the `remove_*` procs couldn't also be `#any_int`, if the others are as well.

It's worth keeping in mind that if you pass a `u64` to an `#any_int int`, or an `i64` on a platform with 32-bit `int`s, it will be silently downgraded with a loss of precision. I don't believe that should be an issue here.

PR #4194, opened 2024-09-03
Update `bit_array`

Feoramund:
- Allow creation of zero-length `Bit_Array`s; it's harmless.
- Add `bit_array.shrink` to safely recover memory unused by unset bits.
- Change `max_index` to `length` to be more consistent with a dynamic array and document explicitly what range a `create`d array will cover.
- Fix iteration of empty `Bit_Array`s; previously there would always be at least one unset bit returned.
- Fix iteration of biased `Bit_Array`s.
- Add a test suite for all of this.

The name of `max_index` was explicitly changed to both represent the underlying value better after the change and to break the API of anyone who might've been using it, to give them a chance to catch the failure at compile-time and adjust.

### DIFF
--- a/core/container/bit_array/bit_array.odin
+++ b/core/container/bit_array/bit_array.odin
@@ -19,7 +19,7 @@ NUM_BITS :: 64
 Bit_Array :: struct {
 	bits:         [dynamic]u64,
 	bias:         int,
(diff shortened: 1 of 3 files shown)


Kelimion:
Great work, as usual. :-)

PR #4196, opened 2024-09-04
Support unmarshalling bit_sets in `core:encoding/json`

jakubtomsu:
For whatever reason unmarshalling bit sets wasn't supported in the `json` package, it always returned an `Unsupported_Type_Error`.

So I added support for it, I just copied the implementation from `core:encoding/cbor/unmarshal.odin` by Laytan. It seems to work fine.

### DIFF
--- a/core/encoding/json/unmarshal.odin
+++ b/core/encoding/json/unmarshal.odin
@@ -116,7 +116,30 @@ assign_int :: proc(val: any, i: $T) -> bool {
 	case int:     dst = int    (i)
 	case uint:    dst = uint   (i)
 	case uintptr: dst = uintptr(i)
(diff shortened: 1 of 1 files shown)


gingerBill:
So the reason why they are not supported is because how should you serialize them? As a number or as a set with the individual values?

It's not as simple as you think.

jakubtomsu:
well the thing is... the marshaller already serializes them, as a raw integer value. So I simply added a way to deserialize that.
But I agree  it's not as simple, and it might be worth adding an option to generate an array of values instead.

gingerBill:
I might just merge this because then it'll at least do something, even if it is not exactly as expected.

PR #4204, opened 2024-09-06
Changed Linux CI builds to static linking with Musl for better compatibility

A1029384756:
Updates CI so that Linux builds will be done with Alpine and statically link Musl. This will fix compatibility with non-Debian systems at the cost of some compilation speed.


A1029384756:
I have also experimented with linking `glibc` in the past but decided against it as it is heavily discouraged and could lead to issues. Another option would be to statically link everything but `glibc` but that runs into issues for people using Musl and exotic setups. This also would introduce issues with versioning where we could find an old version of `glibc` to link against but that introduces another set of issues around what versions are "officially supported" and harms portability.

A1029384756:
I can't test the second part of the workflow revolving around artifact uploading to Backblaze, if someone who has access would be able to test that portion, that would be great (willing to help however I can).

laytan:
This is looking like a good improvement to the dire state of our Linux releases 👍 

Best way forward would probably be merging this and testing out the next nightly build thoroughly.

laytan:
We may get some complaints from people relying on the release being a zip though, I don't mind but something to keep in mind.

Kelimion:
Because of the updated `{download,upload}-artifact` actions, we may still need to fix this in practice. But it's merged so we'll get some CI feedback of the nightly action.

PR #4206, opened 2024-09-06
Improve linking shared libraries

laytan:
Goal: Make the behaviour of shared libraries the same on all platforms.

Windows seems to have the behaviour figured out and seems to be what we want, which is: require the user to copy the library files next to the output executable.

In order to get this same behaviour on other platforms I set the rpath to $ORIGIN, this is afaict a special linker variable that points to the executable, setting the rpath to it makes the linker look in the executable directory for the libraries, a similar thing is `@loader_path` on macos.

Before, defining `-define:RAYLIB_SHARED=true` would cause it to expect the shared library at `$(pwd)/linux/libraylib.so` that would also be `$(pwd)` at compile time, which is problematic.
After: The linker looks for the library in general, but because we added the rpath of $ORIGIN it will find the library if it is at `$executable_directory/libraylib.so`

I added a way to opt-out of setting this rpath for advanced use cases with `-no-rpath`.

---

Shared libraries are still not going to be in an ideal state in part because it requires users to copy things, in the future it would be nice to have the compiler figure out relative foreign imports to shared libraries and automatically execute the copies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,16 +257,16 @@ jobs:
         run: sudo apt-get install -y qemu-user qemu-user-static gcc-12-riscv64-linux-gnu libc6-riscv64-cross
 
       - name: Odin run
-        run: ./odin run examples/demo -vet -strict-style -disallow-do -target:linux_riscv64 -extra-linker-flags:"-fuse-ld=/usr/bin/riscv64-linux-gnu-gcc-12 -static -Wl,-static"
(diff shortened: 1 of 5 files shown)


Yawning:
I am for this change, I haven't looked at the implementation in-depth but:
- It is a general quality-of-life improvement for the common case
- There is an opt-out for cases where this is undesirable
- patchelf can be used to alter the rpath if needed later

PR #4208, opened 2024-09-06
wasm: support more vendor libraries

laytan:
Adds support for:
- box2d
- cgltf
- stb image
- stb rect pack

The PR is arguably a little gross, but it enables a lot of things in WASM because you can now use C libraries that just use a small bit of libc.


core/mem/allocators.odin
```diff
@@ -1137,3 +1137,81 @@ buddy_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
...
+
+	rra := (^Compat_Allocator)(allocator_data)
+	switch mode {
+	case .Alloc, .Alloc_Non_Zeroed:
+		a    := max(alignment, size_of(Header))
+		size += a
+		assert(size >= 0, "overflow")
+
+		allocation := rra.parent.procedure(rra.parent.data, mode, size, alignment, old_memory, old_size, location) or_return
+		#no_bounds_check data = allocation[a:]
```
gingerBill, line 1184: That might not be aligned properly still.

core/mem/allocators.odin
```diff
@@ -1137,3 +1137,81 @@ buddy_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
...
+
+		([^]Header)(raw_data(data))[-1] = {
```
flysand7, line 1186: Wouldn't the `-1` index trip a bounds check?

laytan: Multipointers don't have bounds checks

flysand7: I see now

PR #4212, opened 2024-09-07
Fix build tags that use ! multiple times on one line.

karl-zylinski:
These build tags don't actually do anything since build tags use OR within the line. So something like `//+build !windows, !linux` would actually build on both linux and windows. What was intended in all these cases was probably AND, which you get by splitting them into separate lines.

### DIFF
--- a/core/debug/trace/trace_nil.odin
+++ b/core/debug/trace/trace_nil.odin
@@ -1,4 +1,6 @@
-//+build !windows !linux !darwin
+//+build !windows
+//+build !linux
+//+build !darwin
 package debug_trace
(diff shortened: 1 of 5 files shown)


karl-zylinski:
Actually I was wrong, all these had no commas. No comma means AND. I did not know this. But the commited changes do the same thing so it doesn't really do any harm.

laytan:
> Actually I was wrong, all these had no commas. No comma means AND. I did not know this. But the commited changes do the same thing so it doesn't really do any harm.

Huh, I didn't even know that...

PR #4226, opened 2024-09-10
Fix #4225

### LINKED ISSUE #4225: (BUG) Compiler crash (.cpp assertion) due to non-exact integer value from casting
### Hardware and Software
2024-09 Odin Master Repo (release build)
Windows 11 Intel x64 CPU
Fails on build and/or run
### Symptoms
Compiler crash, best seen by running code below
```odin
package main
import "core:fmt"
main :: proc() {
	_ := transmute(i64)(u64(0.1))
}
```
### Error
\odin\src\check_expr.cpp(3618): Assertion Failure: `src_v.kind == ExactValue_Integer`
Cpp Source Code link and specific code line in error:
Line 3618 of .cpp source (https://github.com/odin-lang/Odin/blob/d783bca2979db5955885ee6b2a06f20f8a847582/src/check_expr.cpp#L3618) 
### Expected Behavior
No compiler crash and appropriate compiler warning/error

Kelimion:
Fixes #4225. Turns compiler assert into:
```
W:/Odin/bug/bug.odin(6:26) Error: Cannot cast '42.1' as 'u64' from 'untyped float' 
	v := transmute(i64)(u64(42.1)) 
	                        ^~~^ 
	42.100000 cannot be represented without truncation/rounding as the type 'u64' 
```


Kelimion:
Just commenting out the assert also works, so it could be that it's a remnant from before the `inexact` error message was added. I leave this to @gingerBill to decide which approach is better here.

src/check_expr.cpp
```diff
@@ -3615,7 +3615,7 @@ gb_internal bool check_transmute(CheckerContext *c, Ast *node, Operand *o, Type
 		if (is_type_integer(src_t) && is_type_integer(dst_t)) {
 			if (types_have_same_internal_endian(src_t, dst_t)) {
 				ExactValue src_v = exact_value_to_integer(o->value);
-				GB_ASSERT(src_v.kind == ExactValue_Integer);
+				// GB_ASSERT(src_v.kind == ExactValue_Integer);
 				BigInt v = src_v.value_integer;
```
gingerBill, line 3619: The assert is there to make sure this access is safe. Unless you are relying on the fact that the kind is Invalid. 

So why not make the assert check for either of those?

PR #4242, opened 2024-09-14
add '#caller_expression'

laytan:
Adds a new directive `#caller_expression` that works similarly to the `#caller_location` directive.

This one provides the function with the entire call expression, or with the expression used to create one of the parameters. This is very useful for `assert`-like procedures:

```odin
// A single parameter (added to the builtin assert function):
assert(err == nil) // test.odin(1:2) runtime assertion: err == nil
// A custom message is still possible:
assert(err == nil, "Hellope!") // test.odin(1:2) runtime assertion: Hellope!

// Or the entire expression:
foo :: proc(hello: string, ok: bool, expr := #caller_expression) {
    fmt.println(expr)
}
foo("Hellope!", err == nil) // foo("Hellope!", err == nil)
```

WIth this PR I've added it to the builtin assert as the default message and to some of the testing expectation procedures.

### DIFF
--- a/base/runtime/core_builtin.odin
+++ b/base/runtime/core_builtin.odin
@@ -913,7 +913,7 @@ card :: proc "contextless" (s: $S/bit_set[$E; $U]) -> int {
 
 @builtin
 @(disabled=ODIN_DISABLE_ASSERT)
-assert :: proc(condition: bool, message := "", loc := #caller_location) {
+assert :: proc(condition: bool, message := #caller_expression(condition), loc := #caller_location) {
 	if !condition {
 		// NOTE(bill): This is wrapped in a procedure call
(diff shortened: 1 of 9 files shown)


FourteenBrush:
I keep wondering, why `#caller_location` and `#caller_expression` can only be default parameters..

Kelimion:
> I keep wondering, why `#caller_location` and `#caller_expression` can only be default parameters..

Because they're input parameters coming from outside of the procedure. If you put them in a callee's body, you'd then either have to specialize that procedure body, or optimize around that, lift them out to the API boundary or the caller's body. And that would make them a parameter again, so we may as well be explicit about it.

gingerBill (approved):

PR #4253, opened 2024-09-16
Fix core sync test deadlock on darwin

pkova:
The deadlock was discussed in #4232.

I'm running macos 14.6.1 and `__ulock_wait` randomly deadlocks in the core sync tests. I guess it shouldn't be too surprising since this isn't really a public API and as [the zig guys discovered](https://github.com/ziglang/zig/blob/812557bfde3c577b5f00cb556201c71ad5ed6fa4/lib/std/Thread/Futex.zig#L175-L179) Apple has been migrating to `__ulock_wait2`.

In any case using `__ulock_wait2` seems to fix the problem. The corresponding change should probably be made to the C++ code as well.

### DIFF
--- a/core/sync/futex_darwin.odin
+++ b/core/sync/futex_darwin.odin
@@ -12,6 +12,7 @@ foreign System {
 	// __ulock_wait is not available on 10.15
 	// See https://github.com/odin-lang/Odin/issues/1959
 	__ulock_wait  :: proc "c" (operation: u32, addr: rawptr, value: u64, timeout_us: u32) -> c.int ---
+	__ulock_wait2  :: proc "c" (operation: u32, addr: rawptr, value: u64, timeout_ns: u64, value2: u64) -> c.int ---
 	__ulock_wake  :: proc "c" (operation: u32, addr: rawptr, wake_value: u64) -> c.int ---
 }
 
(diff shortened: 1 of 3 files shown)


core/sync/futex_darwin.odin
```diff
@@ -12,6 +12,7 @@ foreign System {
 	// __ulock_wait is not available on 10.15
 	// See https://github.com/odin-lang/Odin/issues/1959
 	__ulock_wait  :: proc "c" (operation: u32, addr: rawptr, value: u64, timeout_us: u32) -> c.int ---
+	__ulock_wait2  :: proc "c" (operation: u32, addr: rawptr, value: u64, timeout_ns: u64, value2: u64) -> c.int ---
```
Feoramund, line 15:
```suggestion
	__ulock_wait  :: proc "c" (operation: u32, addr: rawptr, value: u64, timeout_us: u32) -> c.int ---
	__ulock_wait2 :: proc "c" (operation: u32, addr: rawptr, value: u64, timeout_ns: u64, value2: u64) -> c.int ---
```

Feoramund (commented):
I was going to mention the comment, but it looks like you got it already. Great job and thank you for looking into this.

laytan:
Great! I wonder, if you do -minimum-os-version:14.4.0 (so it uses the new public os_sync api) does that also not deadlock? I assume it doesn't?

laytan:
And indeed, would want an improvement like this in the compiler too, but if you aren't comfortable there I can pick that up too.

pkova:
Can confirm no deadlocks with `./odin test tests/core/sync -minimum-os-version:14.4.0`.

laytan:
Another thing catching my eye (clicking through that link) is that ulock_wait is microseconds and ulock_wait2 is nanoseconds. Looks like we handle that wrong here too, passing in nanoseconds where the api expects microseconds.

PR #4254, opened 2024-09-17
Allow import inside `when` statement

jeft5797:
Hello,

I am working on a platform that makes it easy to test snippets of code to test portability.

I'd like to be able to stick to a simple single file format, being able to import within `when` makes it easier and less messy.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -4948,10 +4948,6 @@ gb_internal Ast *parse_import_decl(AstFile *f, ImportDeclKind kind) {
 		array_add(&f->imports, s);
 	}
 
-	if (f->in_when_statement) {
-		syntax_error(import_name, "Cannot use 'import' within a 'when' statement. Prefer using the file suffixes (e.g. foo_windows.odin) or '//+build' tags");
-	}
-
 	if (kind != ImportDecl_Standard) {
 		syntax_error(import_name, "'using import' is not allowed, please use the import name explicitly");
 	}


flysand7:
Imports within `when` statements has been disabled on purpose. This is probably a topic worth some discussion, can you elaborate on what exactly you're trying to do with your platform for testing snippets and what you're trying to achieve by using imports within a `when` statement?

jeft5797:
That's what i thought, I had to change my old code to support the latest version of the compiler, which prompted me to just submit this PR.

It comes down to simplicity, really, why have to manage separate files, when you could just wrap imports inside a `when` block to check for OS/ARCH.

Multiple files involve making the infrastructure more complex, since users now need a way to input multiple files.

gingerBill:
This is not a bug and literally by design. 

Did you not consider that the error was there for a reason?

jeft5797:
> This is not a bug and literally by design.

Who said it was a bug? You didn't read what I wrote.

> Did you not consider that the error was there for a reason?

It's hard to guess without documentation/spec.

Let's take this opportunity to finally document it?

Why is it not allowed?

flysand7:
The documentation is in a rough state right now and I wouldn't expect 100% of the language to be documented at this point. That said, when the time comes, this behavior will be documented, along with recommendations for organizing cross-platform code.

> It comes down to simplicity, really, why have to manage separate files, when you could just wrap imports inside a when block to check for OS/ARCH.

It is certainly an easier thing to do, because depending on how you use this feature you can avoid declaring an extra set of "platform-independent" declarations, and maybe avoid thinking about interfaces for platform-independent packages. But there may be some things that you want to do that you don't *need* to do. I'll explain.

It is understandable that, coming from a different programming language you might want to use Odin in a way that matches the way you wrote code previously. Maybe you used C or C++, where you guard `#include` statements behind `#ifdef` to make multiplatform behavior. But we don't support that paradigm. We also don't support many other ways in which other programming languages handle certain things, which is why it's important to learn what the preferred way of handling things here is, because the way the language is made is to make life easy when things are used in a specific way and when you're not trying to grind against the rules to make a specific thing work. If you do, well... you're going to be met with resistance, and for Odin's design we generally don't try to reduce this resistance by allowing more features, but rather we embrace it because if it feels weird, you're probably doing something weird.

To be more specific, when you want to make something "cross-platform", you can just put the code into multiple files in the same directory, each marked with `#+build` tag (I don't remember if karl's PR is marged and if its still `//+build` or not). That does the bare minimum of the job you want. If you want each job in a separate package, then I guess the question is why is that so important? You're not going to use a cross-platform library to then use one package that's just for windows and completely ignore everything else. I'm assuming that should be along the lines of the issue you've had with this issue being missing, and the response to that is that well, you might have done something weird. You didn't exactly clarify why you didn't do that instead in your response to my question, so my guess is that that's not something you've considered.

TL;DR Allowing this feature may allow some thoughlessness and sillyness in the API, so we disallow that, because a lot of the time you don't need it.

Hopefully my response didn't come out as cold or unwelcoming, I hope for your understanding on this matter. It's just that... sometimes the same questions are getting asked over and over again and writing a detailed answer every time may get a little bit tiring

karl-zylinski:
Just for the record, I've come across cases where what this message suggests:
```
Cannot use 'import' within a 'when' statement. Prefer using the file suffixes (e.g. foo_windows.odin) or '//+build' tags
```
is impossible to do.

Sometimes you want to skip imports based on a config flag, not based on a platform. The only way I found to do that was to go into the package and wrap everything within that package in `when` blocks that checked the config flag.

As an example I wanted to only import a package that contains the steamworks API based on a config flag, because I build my game both with and without steam support depending on where I'm gonna upload it.

DaveTheGameDev:
I too have come across needing a when config do import situation. I have to do the following to keep vet happy. 
The goal is to have a single main file with configs rather than two or more packages.

```odin
when ENABLE_HOT_RELOAD {
	_ :: game
}
```

PR #4266, opened 2024-09-18
[core/os] get_current_directory: Add allocator arg to targets where i…

greenya:
At the moment, `os.get_current_directory` has different signature on different targets. Only Windows and JS targets has `allocator := context.allocator` argument; all other targets missing it, leading to compilation error of user code if it uses that argument on other targets.

**This PR adds `allocator` argument to implementation of `os.get_current_directory` on targets where it is missing.**

### DIFF
--- a/core/os/os_darwin.odin
+++ b/core/os/os_darwin.odin
@@ -1095,7 +1095,7 @@ unset_env :: proc(key: string) -> Error {
 }
 
 @(require_results)
-get_current_directory :: proc() -> string {
+get_current_directory :: proc(allocator := context.allocator) -> string {
 	page_size := get_page_size() // NOTE(tetra): See note in os_linux.odin/get_current_directory.
 	buf := make([dynamic]u8, page_size)
 	for {
--- a/core/os/os_darwin.odin
+++ b/core/os/os_darwin.odin
@@ -1096,6 +1096,7 @@ unset_env :: proc(key: string) -> Error {
 
 @(require_results)
 get_current_directory :: proc(allocator := context.allocator) -> string {
+	context.allocator = allocator
 	page_size := get_page_size() // NOTE(tetra): See note in os_linux.odin/get_current_directory.
 	buf := make([dynamic]u8, page_size)
 	for {
--- a/core/os/os_freebsd.odin
+++ b/core/os/os_freebsd.odin
@@ -841,6 +841,7 @@ get_env :: proc(key: string, allocator := context.allocator) -> (value: string) 
 
 @(require_results)
 get_current_directory :: proc(allocator := context.allocator) -> string {
+	context.allocator = allocator
 	// NOTE(tetra): I would use PATH_MAX here, but I was not able to find
 	// an authoritative value for it across all systems.
 	// The largest value I could find was 4096, so might as well use the page size.
--- a/core/os/os_freebsd.odin
+++ b/core/os/os_freebsd.odin
@@ -840,7 +840,7 @@ get_env :: proc(key: string, allocator := context.allocator) -> (value: string) 
 }
 
 @(require_results)
-get_current_directory :: proc() -> string {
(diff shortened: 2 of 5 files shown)


Feoramund:
`context.allocator` must be assigned to with the allocator, or the allocator must be used in relevant allocating procedures for this to work as intended; merely adding the argument does nothing.

greenya:
The PR fixes compilation errors, which are due to different proc signatures. If user code has line like this:
```odin
log.debugf("current dir is %v", os.get_current_directory(context.temp_allocator))
```
It will compile on Windows and not on Mac (and other targets). To make it compilable, user have to check target with construction like `when ODIN_OS == .Windows {`. I don't think its intended.

jasonKercher:
What Feormund is saying that just because the parameter exists doesn't mean it is being used.  In your example, you send `context.temp_allocator` to `os.get_current_directory`, but that is not going to cause `get_current_directory` to use it.

core/os/os_darwin.odin
```diff
@@ -1095,7 +1095,7 @@ unset_env :: proc(key: string) -> Error {
 }
 
 @(require_results)
-get_current_directory :: proc() -> string {
+get_current_directory :: proc(allocator := context.allocator) -> string {
 	page_size := get_page_size() // NOTE(tetra): See note in os_linux.odin/get_current_directory.
```
laytan, line 1099:
```suggestion
	context.allocator = allocator
	page_size := get_page_size() // NOTE(tetra): See note in os_linux.odin/get_current_directory.
```

core/os/os_freebsd.odin
```diff
@@ -840,7 +840,7 @@ get_env :: proc(key: string, allocator := context.allocator) -> (value: string)
 }
 
 @(require_results)
-get_current_directory :: proc() -> string {
+get_current_directory :: proc(allocator := context.allocator) -> string {
 	// NOTE(tetra): I would use PATH_MAX here, but I was not able to find
```
laytan, line 844:
```suggestion
	context.allocator = allocator
	// NOTE(tetra): I would use PATH_MAX here, but I was not able to find
```

core/os/os_linux.odin
```diff
@@ -985,7 +985,7 @@ unset_env :: proc(key: string) -> Error {
 }
 
 @(require_results)
-get_current_directory :: proc() -> string {
+get_current_directory :: proc(allocator := context.allocator) -> string {
 	// NOTE(tetra): I would use PATH_MAX here, but I was not able to find
```
laytan, line 989:
```suggestion
	context.allocator = allocator
	// NOTE(tetra): I would use PATH_MAX here, but I was not able to find
```

core/os/os_netbsd.odin
```diff
@@ -894,7 +894,7 @@ get_env :: proc(key: string, allocator := context.allocator) -> (value: string)
 }
 
 @(require_results)
-get_current_directory :: proc() -> string {
+get_current_directory :: proc(allocator := context.allocator) -> string {
 	// NOTE(tetra): I would use PATH_MAX here, but I was not able to find
```
laytan, line 898:
```suggestion
	context.allocator = allocator
	// NOTE(tetra): I would use PATH_MAX here, but I was not able to find
```

core/os/os_openbsd.odin
```diff
@@ -806,7 +806,7 @@ get_env :: proc(key: string, allocator := context.allocator) -> (value: string)
 }
 
 @(require_results)
-get_current_directory :: proc() -> string {
+get_current_directory :: proc(allocator := context.allocator) -> string {
 	buf := make([dynamic]u8, MAX_PATH)
```
laytan, line 810:
```suggestion
	context.allocator = allocator
	buf := make([dynamic]u8, MAX_PATH)
```

PR #4269, opened 2024-09-18
Specify integer-like only for some `atomic_*` intrinsics

### LINKED ISSUE #4256: `sync.atomic_add()` causes LLVM code gen failure when using non-integer types
## Context

```
Odin:    dev-2024-09-nightly:dd1f151
OS:      Windows 10 Home Basic (version: 22H2), build 19045.4780
CPU:     AMD Ryzen 9 4900HS with Radeon Graphics
RAM:     15789 MiB
Backend: LLVM 18.1.8
```

## Expected Behavior
Expected this code to either successfully compile or throw an "invalid type" compile time error.

```
import "core::sync"

main :: proc() {
  x := f64(0)
  sync.atomic_add(&x, 1.0)
  _ = x
}
```

## Current Behavior
Throws this monster:
```
LLVM CODE GEN FAILED FOR PROCEDURE: main.main
define internal void @main.main(ptr noalias nocapture nonnull %__.context_ptr) {
decls:
  %x = alloca double, align 8
  br label %entry

  store double 0.000000e+00, ptr %x, align 8
  %0 = atomicrmw add ptr %x, double 1.000000e+00 seq_cst, align 8
  %1 = load double, ptr %x, align 8
  ret void
}

atomicrmw add operand must have integer type!
  %0 = atomicrmw add ptr %x, double 1.000000e+00 seq_cst, align 8
 double
```

Possibly missing `where type_is_integer(T)` clauses from `intrinsics.atomic_()` functions.

### Steps to Reproduce
Run above code.

Feoramund:
Fixes #4256

I left out adding the `where` clauses in `intrinsics.odin` since it's an unwieldy construct (and that file is only for reference, of course), but if it needs to be added, it is something like: `where (intrinsics.type_is_integer(T) || intrinsics.type_is_bit_set(T) || intrinsics.type_is_boolean(T)) && intrinsics.type_is_endian_platform(T)`

I only did cursory testing, but it seems the requirement for same-platform endianness may not be required. At the very least, LLVM did not panic and the output made sense with `atomic_add`. I left it as-is, however. Just noting that here.

Referenced this part of the documentation to check integer requirements: https://llvm.org/docs/LangRef.html#id231

### DIFF
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -4969,16 +4969,14 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 			check_assignment(c, &x, elem, builtin_name);
 
 			Type *t = type_deref(operand->type);
-			switch (id) {
-			case BuiltinProc_atomic_add:
-			case BuiltinProc_atomic_sub:
-				if (!is_type_numeric(t)) {
+			if (id != BuiltinProc_atomic_exchange) {
+				if (!is_type_integer_like(t)) {
 					gbString str = type_to_string(t);
-					error(operand->expr, "Expected a numeric type for '%.*s', got %s", LIT(builtin_name), str);
+					error(operand->expr, "Expected an integer type for '%.*s', got %s", LIT(builtin_name), str);
 					gb_string_free(str);
 				} else if (is_type_different_to_arch_endianness(t)) {
 					gbString str = type_to_string(t);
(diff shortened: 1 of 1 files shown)


flysand7:
Hmm.. do atomics work with 128 bit integers?

Feoramund:
> Hmm.. do atomics work with 128 bit integers?

I tried adding and subtracting 1 to `max(u128)`, and there was no LLVM panic. The results were sane, too.

flysand7:
Good

PR #4274, opened 2024-09-19
Add #+build darwin to everything in core/sys/darwin and vendor/darwin

karl-zylinski:
I tried to compile the package `examples/all` and hit compilation errors because of imports of packages in `vendor/darwin` and `core/sys/darwin`. So I added in `#+build darwin` in those files.

### DIFF
--- a/core/sys/darwin/CoreFoundation/CFBase.odin
+++ b/core/sys/darwin/CoreFoundation/CFBase.odin
@@ -1,3 +1,4 @@
+#+build darwin
 package CoreFoundation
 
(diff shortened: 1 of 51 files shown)


gingerBill:
So be very careful. A lot of that the `examples/all` stuff is there to generate the pkg.odin-lang.org docs.

It is also never meant to be "compiled", just `odin check`.

karl-zylinski:
Closing this, someone else can add the buidlsback in if they run into an actual use case.

PR #4279, opened 2024-09-20
Fix compile errors

obiwan87:
Fix compile errors in map.odin and stdatomic.odin.


obiwan87:
Fixed the problem map.odin, by returning an allocator error in map_entry_infos. Also found some other compile errors, mostly missing package qualifiers and typos

Issue #3474, opened 2024-04-23
encoding/json: unmarshal uses the first struct in a union of structs


Feoramund:
## Context

`json.unmarshal` seems to be ignoring alternative structs in a union, because if it doesn't find the fields it's looking for on the first struct variant it tries, it parses it as any value, skipping the other possibilities.

https://github.com/odin-lang/Odin/blob/5b6c96cd18af9818c37ecdef5130d79ca7149b87/core/encoding/json/unmarshal.odin#L461-L474

	Odin:    dev-2024-04:2416380f3
	OS:      Arch Linux, Linux 6.8.7-arch1-1
	CPU:     12th Gen Intel(R) Core(TM) i7-12700K
	RAM:     31916 MiB
	Backend: LLVM 14.0.6

## Expected Behavior

JSON unmarshalling to `B` instead of `A`, or some sort of error/advisory/workaround.

## Current Behavior

`Invalid type assertion from U to B, actual type: A`

### Steps to Reproduce

```odin
package main

import "core:encoding/json"

A :: struct {
    a: int,
}

B :: struct {
    b: string,
}

U :: union {
    rawptr,
    A,
    B,
}

main :: proc() {
    u: U

    error := json.unmarshal_string(`{"b": "Hellope"}`, &u)
    assert(error == nil)
    assert(u.(B).b == "Hellope")
}
```

Spooky309:
Other JSON implementations I've seen use a "$Variant" key in the JSON, usually with a string name of the variant (to prevent changes in the union declaration from causing the IDs to shift and break the JSON), using the $ specifically because it's illegal in the language to begin a field name with $ so there won't be any conflicts with actual field names. There's a [closed PR](https://github.com/odin-lang/Odin/pull/2617) I saw a couple of months ago that did something similar, but it wasn't quite right.

Would something like this be at all appropriate? I feel it would beat having a field called `this_is_a_button` in my structs :smile:

karl-zylinski:
There was a discussion on the discord about making it configurable.
Default behavior is the current "best fit"
and then you can set it into a "variant type name" mode, where the union would be stored like you say:
```
{
    data: {
        // the variant data
    },
    variant: "B",
}
```
the variant key could use the dollar sign you mentioned.

Then there is the question if the variant should store the type name of the variant or the union tag number. I've seen situations where both are more practical:
- If you use type name then you cannot change type names (or you need some json tag that says the old type name)
[...]

PR #4293, opened 2024-09-22
JSON: Optionally write union variant tags and names

karl-zylinski:
Introduces an option to make JSON marshaller write out union variant name and tag. These names and tags are automatically looked for when unmarshalling happens, in a backwards compatible way.

When it is enabled the unions written to JSON look like so:

```
{
	"$data": {
		"b": "Hellope",
	},
	"$name": "B",
	"$tag": 2
}
```

When unmarshalling into a union, it will look for these `$name` and `$tag` fields etc. If found it will use them and try to find a matching union variant.

Test program that shows how it works and how to enable the marshalling option:
```
package main

import "core:fmt"
import "core:encoding/json"
import "core:reflect"

A :: struct {
	a: int,
}

B :: struct {
	b: string,
}

U :: union {
	rawptr,
	A,
	B,
}

main :: proc() {
	um: U = B {
		b = "Hellope",
	}

	bytes, marshal_err := json.marshal(um, opt = {
		write_union_variant_info = true,
		pretty = true,
	})

	if marshal_err != nil {
		fmt.println(marshal_err)
	}

	fmt.println("Marshalled string:")
	fmt.println(string(bytes))

	u: U

	error := json.unmarshal(bytes, &u)

	if error != nil {
		fmt.println(error)
	}

	fmt.println("\nUnmarshalled into:")
	fmt.println(u)
}
```

when run it prints this:
```
Marshalled string:
{
	"$data": {
		"b": "Hellope"
	},
	"$name": "B",
	"$tag": 2
}

Unmarshalled into:
B{b = "Hellope"}
```

The above is similar to the repro in this issue: https://github.com/odin-lang/Odin/issues/3474, but with the desired outcome.

Things to consider:
- Is a best-fit search also wanted? Or perhaps even what we should try to implement before any of this? My motivation for trying this is that I don't trust a JSON marshaller / unmarshaller unless it has this kind of variant info written down.
- Currently it can only write `$name` for variants that have type info `Type_Info_Named`. For basic types like `f32` it would only write the `$tag`. I can use `reflect.write_typeid` in the marshaller to write the "correct" type for something like an f32, but no good way to compare against it in the unmarshaller without doing something similar there, which starts feeling a bit hacky.

### DIFF
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -60,6 +60,11 @@ Marshal_Options :: struct {
 	indentation: int,
 	mjson_skipped_first_braces_start: bool,
 	mjson_skipped_first_braces_end: bool,
+
+	// Write union variant name and tag into data. Always creates a JSON object
+	// for each union value with a $data and $tag field, plus a $name field if
+	// the variant is a named type.
+	write_union_variant_info: bool,
 }
 
 marshal :: proc(v: any, opt: Marshal_Options = {}, allocator := context.allocator, loc := #caller_location) -> (data: []byte, err: Marshal_Error) {
@@ -458,7 +463,42 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err: 
 			tag -= 1
 		}
 		id := info.variants[tag].id
-		return marshal_to_writer(w, any{v.data, id}, opt)
+
+		if opt.write_union_variant_info {
+			opt_write_start(w, opt, '{') or_return
+			opt_write_iteration(w, opt, true) or_return
+
+			opt_write_key(w, opt, "$data") or_return
+			marshal_to_writer(w, any{v.data, id}, opt) or_return
+			opt_write_iteration(w, opt, false) or_return
+
+			/*
+			TODO: Somehow write the name of simple types like f32... But those
+			are hard to match when unmarshalling since they have Type_Info_Named
+
+			opt_write_key(w, opt, "$name") or_return
+
+			io.write_string(w, "\"") or_return
+			reflect.write_typeid(w, info.variants[tag].id) or_return
+			io.write_string(w, "\"") or_return*/
+
+			tin, tin_ok := info.variants[tag].variant.(runtime.Type_Info_Named)
+
+			if tin_ok {
+				opt_write_key(w, opt, "$name") or_return
+				io.write_string(w, "\"") or_return
+				io.write_string(w, tin.name)
+				io.write_string(w, "\"") or_return
+				opt_write_iteration(w, opt, false) or_return
+			}
+
+			opt_write_key(w, opt, "$tag") or_return
+			io.write_i64(w, tag)
+
+			opt_write_end(w, opt, '}') or_return
+		} else {
+			marshal_to_writer(w, any{v.data, id}, opt) or_return
+		}
 
 	case runtime.Type_Info_Enum:
 		if !opt.use_enum_names || len(info.names) == 0 {
--- a/core/encoding/json/parser.odin
+++ b/core/encoding/json/parser.odin
@@ -176,6 +176,97 @@ parse_value :: proc(p: ^Parser, loc := #caller_location) -> (value: Value, err: 
 	return
 }
 
+skip_value :: proc(p: ^Parser, loc := #caller_location) -> (err: Error) {
+	err = .None
+	token := p.curr_token
+	#partial switch token.kind {
+	case .Null:
+		advance_token(p)
+		return
+	case .False:
+		advance_token(p)
+		return
+	case .True:
+		advance_token(p)
+		return
+
+	case .Integer:
+		advance_token(p)
+		return
+	case .Float:
+		advance_token(p)
+		return
+
+	case .Ident:
+		if p.spec == .MJSON {
+			advance_token(p)
+			return
+		}
+
+	case .String:
+		advance_token(p)
+		return
+
+	case .Open_Brace:
+		num_braces := 1
+
+		for num_braces > 0 {
+			advance_token(p)
+
+			ct := p.curr_token
+
+			if ct.kind == .Invalid || ct.kind == .EOF {
+				break
+			}
+
+			if ct.kind == .Open_Brace {
+				num_braces += 1
+			} else if ct.kind == .Close_Brace {
+				num_braces -= 1
+			}
+		}
+
+		advance_token(p)
+		return
+
+	case .Open_Bracket:
+		num_brackets := 1
+
+		for num_brackets > 0 {
+			advance_token(p)
+
+			ct := p.curr_token
+
+			if ct.kind == .Invalid || ct.kind == .EOF {
+				break
+			}
+
+			if ct.kind == .Open_Bracket {
+				num_brackets += 1
+			} else if ct.kind == .Close_Bracket {
+				num_brackets -= 1
+			}
+		}
+
+		advance_token(p)
+		return
+
+	case:
+		if p.spec != .JSON {
+			switch {
+			case allow_token(p, .Infinity):
+				return
+			case allow_token(p, .NaN):
+				return
+			}
+		}
+	}
+
+	err = .Unexpected_Token
+	advance_token(p)
+	return
+}
+
 parse_array :: proc(p: ^Parser, loc := #caller_location) -> (value: Value, err: Error) {
 	err = .None
 	expect_token(p, .Open_Bracket) or_return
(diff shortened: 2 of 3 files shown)


flysand7:
> Is a best-fit search also wanted? Or perhaps even what we should try to implement before any of this? My motivation for trying this is that I don't trust a JSON marshaller / unmarshaller unless it has this kind of variant info written down.

This can get into a grey area of weirdnesses, that if we allow it, that would introduce a whole world of complexity into the marshaller. Constructs like `union {int, str}` are definately useful, and should be unmarshalled correctly, even if variant field isn't specified in the JSON. But on the other hand `union {f32, f64}` shouldn't be allowed to get unmarshalled into.

So it's not a best-fit exactly, but I guess best unambiguous fit, and that's what I would expect the package to be able to do. The best thing here is to only allow unions that only contain specific base types:

- `nil` (most unions contain it anyway, like the `Maybe` one)
- floats
- booleans
- `string`
- `[]T`, where `T` is one of this list
- `map[string]T`, where T is one of this list

Everything else can be disallowed, unless I'm missing something out.

With this, it's not too hard to make something that will "just work" and map well to some of the JSON files with non-uniform structure.

michaelcrichlow:
I updated my local files ('marshal.odin, 'parser.odin' and 'unmarshall.odin') based on your files here. The whole point is to have predictable results when you marshall and unmarshall unions -- so that's what I tried to do. Working directly with unions one at a time works well (as shown in the code example above). So, I tried adding a union to a struct (Entity) to see if it would still work.

```
package test

import "core:encoding/json"
import "core:fmt"
import "core:os"
import "core:reflect"

Entity :: struct {
	id:      int,
	name:    string,
	variant: U,  // <---- See here
}

A :: struct {
	a: int,
}

B :: struct {
	b: string,
}

U :: union {
	rawptr,
	A,
	B,
}

main :: proc() {
	ent: Entity = {
		id = 12,
		name = "Michael",
		variant = B{b = "apple"},
	}

	ent2: Entity = {
		id = 14,
		name = "Elizabeth",
		variant = A{a = 1337},
	}

	bytes, marshal_err := json.marshal(
		ent2,
		opt = {write_union_variant_info = true, pretty = true},
	)

	if marshal_err != nil {
		fmt.println(marshal_err)
	}

	os.write_entire_file("data_00.json", bytes)

	ent_test: Entity
	if data, ok := os.read_entire_file("data_00.json", allocator = context.temp_allocator); ok {
		error := json.unmarshal(data, &ent_test)
		if error != nil {
			fmt.println(error)
			panic("something went wrong")
		}
	}

	fmt.println("Unmarshalled into:")
	fmt.println(ent_test)
	fmt.println("done!")
}
```

That seems to work fine. marshlling 'ent' results in this:

```
Unmarshalled into:
Entity{id = 12, name = "Michael", variant = B{b = "apple"}}
done!
```
and marshalling 'ent2' results in this:

```
Unmarshalled into:
Entity{id = 14, name = "Elizabeth", variant = A{a = 1337}}
done!
```

However, putting them into an array (which I would think would be the end goal of all of this) doesn't seem to work (or I don't know how to do it.) Creating an 'entity_array' and putting both entities into it results in this:

```
package test

import "core:encoding/json"
import "core:fmt"
import "core:os"
import "core:reflect"

Entity :: struct {
	id:      int,
	name:    string,
	variant: U,
}

A :: struct {
	a: int,
}

B :: struct {
	b: string,
}

U :: union {
	rawptr,
	A,
	B,
}

main :: proc() {
	ent: Entity = {
		id = 12,
		name = "Michael",
		variant = B{b = "apple"},
	}

	ent2: Entity = {
		id = 14,
		name = "Elizabeth",
		variant = A{a = 1337},
	}

	// ------------------------------------------------------------------
	// added code
	entity_array: [dynamic]Entity
	append(&entity_array, ent)
	append(&entity_array, ent2)
	// ------------------------------------------------------------------

	bytes, marshal_err := json.marshal(
		entity_array, // <------------------- changed to 'entity_array'
		opt = {write_union_variant_info = true, pretty = true},
	)

	if marshal_err != nil {
		fmt.println(marshal_err)
	}

	os.write_entire_file("data_00.json", bytes)
	fmt.println(string(bytes))
	fmt.println("done!")
}
```

This results in this:
```
[
        {
                "id": 12,
                "name": "Michael",
                "variant": {
                        "$data": {
                                "b": "apple"
                        },
                        "$name": "B",
                        "$tag": 2
                }
        },
        {
                "id": 14,
                "name": "Elizabeth",
                "variant": {
                        "$data": {
                                "a": 1337
                        },
                        "$name": "A",
                        "$tag": 1
                }
        }
]
done!
```

That's also what's in the 'data_00.json' file. So far, so good. I think this looks fine (though you would know better than me). However, trying to unmarshall it doesn't work.

```
package test

import "core:encoding/json"
import "core:fmt"
import "core:os"
import "core:reflect"

Entity :: struct {
	id:      int,
	name:    string,
	variant: U,
}

A :: struct {
	a: int,
}

B :: struct {
	b: string,
}

U :: union {
	rawptr,
	A,
	B,
}

main :: proc() {
	ent: Entity = {
		id = 12,
		name = "Michael",
		variant = B{b = "apple"},
	}

	ent2: Entity = {
		id = 14,
		name = "Elizabeth",
		variant = A{a = 1337},
	}


	entity_array: [dynamic]Entity
	append(&entity_array, ent)
	append(&entity_array, ent2)


	bytes, marshal_err := json.marshal(
		entity_array,
		opt = {write_union_variant_info = true, pretty = true},
	)

	if marshal_err != nil {
		fmt.println(marshal_err)
	}

	os.write_entire_file("data_00.json", bytes)
	// fmt.println(string(bytes))

	// ------------------------------------------------------------------
	// added code
	entity_array_02: [dynamic]Entity
	if data, ok := os.read_entire_file("data_00.json", allocator = context.temp_allocator); ok {
		error := json.unmarshal(data, &entity_array_02)
		if error != nil {
			fmt.println(error)
			panic("something went wrong")
		}
	}
	// ------------------------------------------------------------------

	fmt.println("Unmarshalled into:")
	fmt.println(entity_array_02)
	fmt.println("done!")
}
```

This gives this result:

```
Unsupported_Type_Error{id = Entity, token = Token{pos = Pos{offset = 125, line = 12, column = 2}, kind = "Close_Brace", text = "}"}}
C:/Users/mikec/Documents/Visual Studio Code II/odin_tests/tests_01.odin(66:4) panic: something went wrong
```

So, my current hypothesis is that marshalling from an array to a json file works, but it doesn't know how to unmarshall from the json file back into an array. If you need any help or clarification on the steps I took I'll be happy to assist. Hope you're able to help. Thanks! :)

karl-zylinski:
> So, my current hypothesis is that marshalling from an array to a json file works, but it doesn't know how to unmarshall from the json file back into an array. If you need any help or clarification on the steps I took I'll be happy to assist. Hope you're able to help. Thanks! :)

This is correct. It does not work with an array. Wrap the array in a struct (you can create a temporary struct with the array inside just for marshal/unmarshal) and assign to/from it.

michaelcrichlow:
Ah, okay. Thanks! My current solution is just saving the array as .cbor (which solves all the union issues I was having). I'll have to try your solution out soon. Thanks again! :)

karl-zylinski:
Is there any interest in me trying to finish finish this PR? The json marshal/unmarshal is not very useful with unions right now. It's very easy to end up in state where the current implementation "guesses the wrong type". So I think something like this is needed. I'm happy for any input on my design approach.

LagMeester4000:
I would appreciate if this were implemented. I ran into this today when trying to marshall/unmarshall a union, and it would silently give me differing data.

jfulto98:
Would also like this to be implemented.

d3cr1pt0r:
I have a repo case that doesn't seem to work. Marshal works fine, but unmarshal fails. Here is the repo case.

```
package test

import "core:fmt"
import "core:encoding/json"

A :: struct
{
	a : int,
}

B :: struct
{
	a : int,
	b : int,
}

C :: struct
{
	a : int,
	b : int,
	c : int,
}

Data :: struct
{
	u : union
	{
		A,
		B,
		C,
	}
}

DataArr :: struct
{
	entries : []Data,
}

main :: proc()
{
	data_marshal := DataArr {
		entries = {
			{ u = A { a = 15 } },
			{ u = B { a = 30, b = 45 } },
			{ u = C { a = 60, b = 75, c = 90 } },
		},
	};

	if json_data, err := json.marshal(data_marshal, opt = { write_union_variant_info = true, pretty = true }, allocator = context.temp_allocator); err == nil {
		data_unmarshal : DataArr;
		if json.unmarshal(json_data, &data_unmarshal) == nil {
			fmt.println(data_unmarshal)
		}
		else {
			fmt.println("error: failed to unmarshal")
		}
	}
	else {
		fmt.println("error: failed to marshal")
	}
}
```

PR #4297, opened 2024-09-24
[encoding/ini]: Add missing line terminator at the end of the section

### LINKED ISSUE #4296: INI package not writing section correctly
## Context

The INI package has a bug in `write_section` where it does not 

        Odin:    dev-2024-09:f7d74ff3a
        OS:      Windows 11 Professional (version: 23H2), build 22631.4169
        CPU:     13th Gen Intel(R) Core(TM) i9-13900KS
        RAM:     65243 MiB
        Backend: LLVM 18.1.8

## Expected Behavior

A new line should be written after the section is written to the writer

```ini
[GameConfig]
vsync = true
```

## Current Behavior
Currently section and first value are inline
The result for me personally is that the first value is seemingly ignored.

```ini
[GameConfig] vsync = true
```

### Steps to Reproduce

1. Call `write_section `after setting up a writer.
2. Now call `write_pair` and save to disk, notice how the result has the inline header and value mentioned above.

is it missing `io.write_byte  (w, '\n',  &n) or_return`?
```odin
write_section :: proc(w: io.Writer, name: string, n_written: ^int = nil) -> (n: int, err: io.Error) {
	defer if n_written != nil { n_written^ += n }
	io.write_byte  (w, '[',  &n) or_return
	io.write_string(w, name, &n) or_return
	io.write_byte  (w, ']',  &n) or_return
	return
}
```

flysand7:
Fix #4296

### DIFF
--- a/core/encoding/ini/ini.odin
+++ b/core/encoding/ini/ini.odin
@@ -154,6 +154,7 @@ write_section :: proc(w: io.Writer, name: string, n_written: ^int = nil) -> (n: 
 	io.write_byte  (w, '[',  &n) or_return
 	io.write_string(w, name, &n) or_return
 	io.write_byte  (w, ']',  &n) or_return
+	io.write_byte  (w, '\n',  &n) or_return
 	return
 }
 


DaveTheGameDev:
Thanks @flysand7! Was not 100% sure if anything else was missing. and was a little spooked to do a PR 🤣

PR #4302, opened 2024-09-25
Add static GLFW bindings for Linux

Pariatech:
I static link GLFW for my game but I can't do that with the current bindings for linux. 
I kept the default as true for shared linking for linux to not break anyone else.

### DIFF
--- a/vendor/glfw/bindings/bindings.odin
+++ b/vendor/glfw/bindings/bindings.odin
@@ -41,9 +41,13 @@ when ODIN_OS == .Windows {
 	when GLFW_SHARED {
 		foreign import glfw "system:glfw"
 	} else {
-		foreign import glfw { 
-			"../lib/libglfw3.a",
+		@(private)
+		LIBGLFW3 :: "../lib/libglfw3.a"
+		when !#exists(LIBGLFW3) {
+			#panic("Could not find the static glfw library, add it at \"" + ODIN_ROOT + "vendor/glfw/lib/\"`")
 		}
+
+		foreign import glfw { LIBGLFW3 }
 	}
 }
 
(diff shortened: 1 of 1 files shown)


gingerBill:
And have these been built to work on all distros of Linux? Because if they haven't, I will close this PR.

gingerBill:
I know they are the ones that are in there already, but they might not be linked for a good reason.

flysand7:
> And have these been built to work on all distros of Linux? Because if they haven't, I will close this PR.

He didn't push any binaries, from what I can see

gingerBill:
@flysand7 
> I know they are the ones that are in there already, but they might not be linked for a good reason.

Pariatech:
That's why I didn't push any binary, just a way to static link a binary if so desired. And without breaking any other linux Odin user if they want use the GLFW bindings without building their own binary by making GLFW_SHARED to true by default for them.

Pariatech:
Also I didn't want to be the guy that just put in a binary with a "trust me bro, it works ;-)". Meanwhile the binary that I generate for myself via a debian:latest docker seems to work for the people that tested my game in linux (arch, mint, debian). But even then I can't 100% guarantee that it will work on everyone linux box, especially if the user is running an old kernel / libc system.

vendor/glfw/bindings/bindings.odin
```diff
@@ -38,7 +38,13 @@ when ODIN_OS == .Windows {
 		}
 	}
 } else {
-	foreign import glfw "system:glfw"
+	when GLFW_SHARED {
+		foreign import glfw "system:glfw"
+	} else {
+		foreign import glfw {
```
laytan, line 44: Would be good to add a `when !#exists("path") { #panic("no static library found, add it at ....") }` here.

See https://github.com/odin-lang/Odin/blob/daa3bb9a3beb8a4551ef4c8bdeefcaaaf11ed30d/vendor/stb/image/stb_image_resize.odin#L15 for an example.

Pariatech: Done

PR #4304, opened 2024-09-25
Add linear_search_reverse and linear_search_reverse_proc

kawaii-Code:
Surprised this was missing from the standard library, so here it is. Also added some docs for `linear_search`

### DIFF
--- a/core/slice/slice.odin
+++ b/core/slice/slice.odin
@@ -106,7 +106,7 @@ contains :: proc(array: $T/[]$E, value: E) -> bool where intrinsics.type_is_comp
 */
 @(require_results)
 linear_search :: proc(array: $A/[]$T, key: T) -> (index: int, found: bool)
-	where intrinsics.type_is_comparable(T) #no_bounds_check {
+	where intrinsics.type_is_comparable(T) {
 	for x, i in array {
 		if x == key {
 			return i, true
@@ -122,7 +122,7 @@ linear_search :: proc(array: $A/[]$T, key: T) -> (index: int, found: bool)
 	or -1 if it is not present.
 */
 @(require_results)
-linear_search_proc :: proc(array: $A/[]$T, f: proc(T) -> bool) -> (index: int, found: bool) #no_bounds_check {
+linear_search_proc :: proc(array: $A/[]$T, f: proc(T) -> bool) -> (index: int, found: bool) {
 	for x, i in array {
 		if f(x) {
 			return i, true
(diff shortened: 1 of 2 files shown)


core/slice/slice.odin
```diff
@@ -117,6 +131,61 @@ linear_search_proc :: proc(array: $A/[]$T, f: proc(T) -> bool) -> (index: int, f
...
 }
 
+/*
+	Reverse linear search searches the given element in the given slice in O(n) time,
+	starting from the slice end.
+
+	Returns the last index at which the given element can be found in the slice,
+	or -1 if it is not present
+
+	# Examples
```
laytan, line 141: Examples should be formatted as described here: https://github.com/odin-lang/pkg.odin-lang.org?tab=readme-ov-file#markup-rules

core/slice/slice.odin
```diff
@@ -117,6 +131,61 @@ linear_search_proc :: proc(array: $A/[]$T, f: proc(T) -> bool) -> (index: int, f
 	return -1, false
 }
 
+/*
+	Reverse linear search searches the given element in the given slice in O(n) time,
```
laytan, line 135: searches the -> searches for the (the case in all the added doc comments)

core/slice/slice.odin
```diff
@@ -117,6 +131,61 @@ linear_search_proc :: proc(array: $A/[]$T, f: proc(T) -> bool) -> (index: int, f
...
+
+	index, found = linear_search_reverse(a, 0)
+	assert(index == -1 && found == false)
+	```
+
+	If you need a custom compare procedure, see `linear_search_reverse_proc`
+*/
+@(require_results)
+linear_search_reverse :: proc(array: $A/[]$T, key: T) -> (index: int, found: bool)
+	where intrinsics.type_is_comparable(T) #no_bounds_check {
```
laytan, line 163: `#no_bounds_check` isn't needed, none of the operations add bounds checks (same thing with the reverse proc).

kawaii-Code: What about plain `linear_search`? It does use `#no_bounds_check`

laytan: I see, I am also pretty sure it isn't needed there.

kawaii-Code:
@laytan done. There seem to be more `#no_bounds_check` scattered across `slice.odin`, e.g. in `filter_reverse`, `unique`, etc. Are you sure they aren't needed?

laytan:
> @laytan done. There seem to be more `#no_bounds_check` scattered across `slice.odin`, e.g. in `filter_reverse`, `unique`, etc. Are you sure they aren't needed?

Yes, bounds checks are done when you slice or otherwise index into an array, in these search procs you just loop which afaik doesn't emit a bounds check.

Issue #4306, opened 2024-09-25
Arch amd64: `ptrace_getregs` call causing the porgram stack corruption


eseft:
## Context

* `odin report` output:
```
	Odin:    dev-2024-09:8371ef668
	OS:      Arch Linux, Linux 6.10.10-arch1-1
	CPU:     AMD Ryzen 5 5600G with Radeon Graphics
	RAM:     31460 MiB
	Backend: LLVM 18.1.8
```

## Expected Behavior

Stack is not corrupted and program can continue work after `ptrace(.PTRACE_GETREGS,...)` call. 

## Current Behavior

After successful `ptrace(.PTRACE_GETREGS,...)` program stack getting corrupted.

## Failure Information (for bugs)

Looks like for the amd64 arch the `User_Regs` struct missing couple of fields compared with [https://github.com/bminor/glibc/blob/master/sysdeps/unix/sysv/linux/x86/sys/user.h](https://github.com/bminor/glibc/blob/master/sysdeps/unix/sysv/linux/x86/sys/user.h) which is causing the program stack corruption on `ptrace` syscall

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Compile the following program
```odin
package main

import "core:sys/linux"
import "core:os"


get_regs :: proc(pid: linux.Pid, regs: ^linux.User_Regs) -> bool {
    tmp := linux.User_Regs{}
    err := linux.ptrace(linux.PTRACE_GETREGS, pid, &tmp)
    if err != nil {
        return false
    }
    regs^ = tmp
    return true
}

main :: proc() {
    pid, err := linux.fork()
    assert(err == nil)

    if pid == 0 {
        err_ptrace := linux.ptrace(linux.PTRACE_TRACEME)
        assert(err_ptrace == nil)

        err_exec := os.execvp("/usr/bin/ls", {})
        assert(err_exec == nil)

    } else {
        status: u32
        rusage: linux.RUsage
        linux.wait4(pid, &status, {}, &rusage)
        assert(linux.WIFSTOPPED(status))

        regs: linux.User_Regs
        get_regs(pid, &regs)
    }
}
```
2. Run the program under a debugger and break at `get_regs` proc. Inspect address and value of `regs` local variable:
```
(gdb) p regs
$1 = (struct linux._Arch_User_Regs *) 0x7fffffffdf98
(gdb) p &regs
$2 = (struct linux._Arch_User_Regs **) 0x7fffffffdf68
[...]

PR #4307, opened 2024-09-25
Fix missing registers in User_Regs on AMD64

flysand7:
This PR fixes the OOB stack corruption when calling `ptrace_getregs`, due to `User_Regs` structure not containing segment registers as well as gs and fs register hidden bases.

This PR seems to be fixing #4306, but I'm not entirely sure this fix is correct, because the I copied the definition of the struct from glibc, and not the actual kernel's source code. It's also not clear to me how after the fix segment registers and their bases are all 0.


flysand7:
Ok it was correct

https://elixir.bootlin.com/linux/v6.11/source/arch/x86/include/asm/user_64.h#L91

PR #4327, opened 2024-09-30
fix #4326

### LINKED ISSUE #4326: error reporting for procedures with deferred_<not none> attributes reverses which procedure is deferred and which is initial
odin report:
```
	Odin:    dev-2024-09:38fedf263
	OS:      macOS Sonoma 14.6.1 (build: 23G93, kernel: 23.6.0)
	CPU:     Apple M2
	RAM:     16384 MiB
	Backend: LLVM 18.1.8
```

minimal repro (standalone odin file):

```go
package main

main :: proc() {
    proc0(42)
    proc2(42)
    proc4(42)
}

@(deferred_in=proc1)
proc0 :: proc(param: int) {
    _ = param
}

proc1 :: proc(param: f32) {
    _ = param
}

@(deferred_out=proc3)
proc2 :: proc(param: int) {
    _ = param
}

proc3 :: proc(param: f32) {
    _ = param
}

@(deferred_in_out=proc5)
proc4 :: proc(param: int) -> int {
    return param
}

proc5 :: proc(param: f32) -> f32 {
    return param
}

import "core:fmt"
```

output of `odin check .` (or build or whatever):

```
/private/tmp/foo.odin(10:1) Error: Deferred procedure 'proc0' parameters do not match the inputs of initial procedure 'proc1': 
	(int) =/= (f32) 
	proc0 :: proc(param: int) { 
	^ 
 
/private/tmp/foo.odin(19:1) Error: Deferred procedure 'proc2' parameters do not match the results of initial procedure 'proc3' 
	proc2 :: proc(param: int) { 
	^ 
 
/private/tmp/foo.odin(28:1) Error: Deferred procedure 'proc4' parameters do not match the results of initial procedure 'proc5': 
	(int, int) =/= (f32) 
	proc4 :: proc(param: int) -> int { 
	^ 
 
```

e.g. proc0 is referred to as the "deferred procedure" but its actually proc1 that's deferred

beaumccartney:
just switch the order of things so its printed correctly

fixes #4326

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -6206,7 +6206,7 @@ gb_internal void check_deferred_procedures(Checker *c) {
 				}
 				if ((src_params == nullptr && dst_params != nullptr) ||
 				    (src_params != nullptr && dst_params == nullptr)) {
-					error(src->token, "Deferred procedure '%.*s' parameters do not match the inputs of initial procedure '%.*s'", LIT(src->token.string), LIT(dst->token.string));
+					error(src->token, "Deferred procedure '%.*s' parameters do not match the inputs of initial procedure '%.*s'", LIT(dst->token.string), LIT(src->token.string));
 					continue;
 				}
 
@@ -6219,7 +6219,7 @@ gb_internal void check_deferred_procedures(Checker *c) {
 					gbString s = type_to_string(src_params);
 					gbString d = type_to_string(dst_params);
 					error(src->token, "Deferred procedure '%.*s' parameters do not match the inputs of initial procedure '%.*s':\n\t(%s) =/= (%s)",
-					      LIT(src->token.string), LIT(dst->token.string),
+					      LIT(dst->token.string), LIT(src->token.string),
 					      s, d
 					);
 					gb_string_free(d);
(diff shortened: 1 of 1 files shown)


src/checker.cpp
```diff
@@ -6302,7 +6302,7 @@ gb_internal void check_deferred_procedures(Checker *c) {
 					gbString s = type_to_string(tsrc);
 					gbString d = type_to_string(dst_params);
 					error(src->token, "Deferred procedure '%.*s' parameters do not match the results of initial procedure '%.*s':\n\t(%s) =/= (%s)",
-					      LIT(src->token.string), LIT(dst->token.string),
+					      LIT(dst->token.string), LIT(src->token.string),
 					      s, d
```
Kelimion, line 6306: I would also switch `s, d` in these cases, so it also flips around the initial and deferred input types.
As written it would result in:
```
W:/Odin/bug/bug.odin(10:1) Error: Deferred procedure 'proc1' parameters do not match the inputs of initial procedure 'proc0': 
	(int) =/= (f32) 
	proc0 :: proc(param: int) { 
	^ 
```
When you'd want:
```
W:/Odin/bug/bug.odin(10:1) Error: Deferred procedure 'proc1' parameters do not match the inputs of initial procedure 'proc0': 
	(f32) =/= (int)
	proc0 :: proc(param: int) { 
	^ 
```

beaumccartney: done! good catch

PR #4332, opened 2024-10-01
Static on macosx

Smilex:
I have tested that this works with the iPhone simulator

Basically just call "libtool -static" on the object files

### DIFF
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -663,6 +663,8 @@ gb_internal i32 linker_stage(LinkerData *gen) {
 					static_link_command_line = gb_string_appendc(static_link_command_line, " -static ");
 					static_link_command_line = gb_string_appendc(static_link_command_line, object_files);
 					return system_exec_command_line_app("libtool", static_link_command_line);
+				} else {
+					compiler_error("-build-mode:static on Linux and similar targets");
 				}
 			} else {
 				gbString link_command_line = gb_string_make(heap_allocator(), clang_path);
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -571,10 +571,6 @@ gb_internal i32 linker_stage(LinkerData *gen) {
 				link_settings = gb_string_append_fmt(link_settings, "-nostdlib ");
 			}
 
-			if (build_context.build_mode == BuildMode_StaticLibrary) {
-				compiler_error("TODO(bill): -build-mode:static on non-windows targets");
-			}
-
 			// NOTE(dweiler): We use clang as a frontend for the linker as there are
 			// other runtime and compiler support libraries that need to be linked in
 			// very specific orders such as libgcc_s, ld-linux-so, unwind, etc.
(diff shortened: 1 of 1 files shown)


src/linker.cpp
```diff
@@ -657,39 +653,56 @@ gb_internal i32 linker_stage(LinkerData *gen) {
...
-				// NOTE: macOS links DWARF symbols dynamically. Dsymutil will map the stubs in the exe
-				// to the symbols in the object file
-				result = system_exec_command_line_app("dsymutil", "dsymutil %.*s", LIT(output_filename));
+				gbString link_command_line = gb_string_make(heap_allocator(), clang_path);
+				defer (gb_string_free(link_command_line));
+
+				link_command_line = gb_string_appendc(link_command_line, " -Wno-unused-command-line-argument ");
+				link_command_line = gb_string_append_fmt(link_command_line, " -o \"%.*s\" ", LIT(output_filename));
+
+				link_command_line = gb_string_appendc(link_command_line, object_files);
```
laytan, line 676: why did you move the object files from before the -o to after it?

src/linker.cpp
```diff
@@ -631,7 +627,7 @@ gb_internal i32 linker_stage(LinkerData *gen) {
 					link_settings = gb_string_append_fmt(link_settings, "-mmacosx-version-min=%.*s ", LIT(build_context.minimum_os_version_string));
 				}
 
-				if (build_context.build_mode != BuildMode_DynamicLibrary) {
+				if (build_context.build_mode != BuildMode_DynamicLibrary && build_context.build_mode != BuildMode_StaticLibrary) {
```
laytan, line 630: This does not look needed? `link_settings` isn't used when you do a static library from what I can tell.

src/linker.cpp
```diff
@@ -657,39 +653,56 @@ gb_internal i32 linker_stage(LinkerData *gen) {
...
+			if (build_context.build_mode == BuildMode_StaticLibrary) {
+				if (is_osx) {
+					gbString static_link_command_line = gb_string_make(heap_allocator(), "libtool");
+					defer (gb_string_free(static_link_command_line));
+					static_link_command_line = gb_string_append_fmt(static_link_command_line, " -o \"%.*s\" ", LIT(output_filename));
+					static_link_command_line = gb_string_appendc(static_link_command_line, " -static ");
+					static_link_command_line = gb_string_appendc(static_link_command_line, object_files);
+					return system_exec_command_line_app("libtool", static_link_command_line);
+				} else {
+					compiler_error("-build-mode:static on Linux and similar targets");
```
laytan, line 667:
```suggestion
					compiler_error("TODO: -build-mode:static on this target");
```

laytan (commented):
- I see the .dSYM is not created for a `-debug -build-mode:static`, it should be possible right?
- It must be possible to set an entry/init section in the static library so the `runtime._startup_runtime()` and similar things can be called

Smilex:
> * I see the .dSYM is not created for a `-debug -build-mode:static`, it should be possible right?
> * It must be possible to set an entry/init section in the static library so the `runtime._startup_runtime()` and similar things can be called

Looks like I need to call "dsymutil <your_program>" somehow, but it says that the object files don't have debug data. I'll look into it later

PR #4339, opened 2024-10-02
fix swizzle in for in statement

### LINKED ISSUE #1730: Inconsistent behaviour when using `.xy` after a function call
## Context

I have made this small code that reproduces the bug:
```odin
package main

import "core:fmt"

f :: proc() -> [4][2]string
{
    return {
        {"00", "01"},
        {"10", "11"},
        {"20", "21"},
        {"30", "31"}}
}

main :: proc()
{
    reproBug := true
    if reproBug {        
        for l in f().xy {
            for s in l {
                fmt.println(s)
            }
        }
    }
    else {
        ll := f().xy
        for l in ll {
            for s in l {
                fmt.println(s)
            }
        }
    }
}
```

        Odin: dev-2022-04-nightly:59025b75
        OS:   Windows 10 Education (version: 21H1), build 19043.1645
        CPU:  AMD Ryzen 7 3700X 8-Core Processor
        RAM:  16332 MiB

## Expected Behavior

I should print the following in both cases:
```
00
01
10
11
```

## Current Behavior

When `reproBuf := true` it prints:

```
00
01
10
11
20
21
30
31
```

laytan:
Fixes #1730


laytan:
That broke things :p

laytan:
Hmm think the test failures are actually previously broken things that are now fixed, but the hash checks assume the broken state?

Feoramund:
@laytan I tried this patch for #1730 and it seems to work. The test suite passed locally for me. Is there something else holding this back?

Kelimion:
Checked out this PR branch, but the BMP test fails, @Feoramund. Did you rebase this patch before trying, or are you saying this branch passes for you without any changes?
```
[ERROR] --- [2025-06-01 12:49:47] [test_core_image.odin:2334:run_bmp_suite()] pal8os2 test #1's hash is 4e69aae9, expected 38454155 with Options{}.
[ERROR] --- [2025-06-01 12:49:47] [test_core_image.odin:2334:run_bmp_suite()] pal8os2-sz test #1's hash is 4e69aae9, expected 38454155 with Options{}.
[ERROR] --- [2025-06-01 12:49:47] [test_core_image.odin:2334:run_bmp_suite()] pal8os2-hs test #1's hash is 4e69aae9, expected 38454155 with Options{}.
[ERROR] --- [2025-06-01 12:49:47] [test_core_image.odin:2334:run_bmp_suite()] pal8os2sp test #1's hash is 4e69aae9, expected 38454155 with Options{}.
```
And the compiler segfaults on Linux when I run `odin test speed.odin -file`.

laytan:
My theory was that these tests were actually relying on the bug that this PR fixes, then I forgot about it :p

Feoramund:
I merely performed the diff done in `src/llvm_backend_general.cpp` by hand onto the latest commit on `master`.

Kelimion:
> My theory was that these tests were actually relying on the bug that this PR fixes, then I forgot about it :p

I'll double check whether that's the case, wrote the bmp tests after all. Will apply the fix to `master` (locally) and take a look at the failure case.

Kelimion:
It's actually broken. Saved the files that failed the hash and while they have the correct dimensions, they're entirely black.
Added this debug code:
```odin
if test.hash != img_hash {
	sum: u64
	for p in pixels {
		sum += u64(p)
	}

	log.infof("sum: %v, img: %v", sum, img)
	bug_file := strings.concatenate({TEST_SUITE_PATH_BMP, "/", file.file, ".bug.bmp"}, context.temp_allocator)
	bmp.save(bug_file, img)
	bug_file  = strings.concatenate({TEST_SUITE_PATH_BMP, "/", file.file, ".bug.qoi"}, context.temp_allocator)
	qoi.save(bug_file, img)
}
```
Expected:
image
Got:
image

```
[ERROR] --- [2025-06-01 13:24:23] [test_core_image.odin:2336:run_bmp_suite()] pal8os2 test #1's hash is 4e69aae9, expected 38454155 with Options{}.
[INFO ] --- [2025-06-01 13:24:23] [test_core_image.odin:2343:run_bmp_suite()] sum: 0, img: &Image{width = 127, height = 64, channels = 3, depth = 8, background = nil, metadata = 0x20A27E64450, which = "BMP"}
[ERROR] --- [2025-06-01 13:24:23] [test_core_image.odin:2336:run_bmp_suite()] pal8os2-sz test #1's hash is 4e69aae9, expected 38454155 with Options{}.
[INFO ] --- [2025-06-01 13:24:23] [test_core_image.odin:2343:run_bmp_suite()] sum: 0, img: &Image{width = 127, height = 64, channels = 3, depth = 8, background = nil, metadata = 0x20A27E6C870, which = "BMP"}
[ERROR] --- [2025-06-01 13:24:23] [test_core_image.odin:2336:run_bmp_suite()] pal8os2-hs test #1's hash is 4e69aae9, expected 38454155 with Options{}.
[INFO ] --- [2025-06-01 13:24:23] [test_core_image.odin:2343:run_bmp_suite()] sum: 0, img: &Image{width = 127, height = 64, channels = 3, depth = 8, background = nil, metadata = 0x20A27E74C90, which = "BMP"}
[ERROR] --- [2025-06-01 13:24:23] [test_core_image.odin:2336:run_bmp_suite()] pal8os2sp test #1's hash is 4e69aae9, expected 38454155 with Options{}.
[INFO ] --- [2025-06-01 13:24:23] [test_core_image.odin:2343:run_bmp_suite()] sum: 0, img: &Image{width = 127, height = 64, channels = 3, depth = 8, background = nil, metadata = 0x20A27E7D0A8, which = "BMP"}
```

And the bad hash is the expected hash for 127 * 64 * 3 zero bytes, so it's also not a bug in how it's saved.
```odin
package main

@(require) import "core:fmt"
import "core:hash"

main :: proc() {
	pixels := make([]u8, 127 * 64 * 3)
	fmt.printfln("%08x", hash.crc32(pixels[:])) // 4e69aae9
}
```

Feoramund:
I see now. I only ran test against `normal.odin`, which does not include those tests.

Kelimion:
What's also interesting is that the rest of the BMP's don't fail. They must take a different path through the BMP loader from the other ones, but regardless it's safe to conclude that this fix sadly introduces a bug of its own.

laytan:
Ah, looking at it, it's actually quite obvious why it fails, it now loads the swizzle into a temporary variable, which causes this to fail, need to go back to the drawing board a little here:

```odin
	ll := [4]int{1, 2, 3, 4}
	out: [4]int
	out.yz = ll.yz
	testing.expect_value(t, out, [4]int{0, 2, 3, 0})
```

laytan:
Should be properly fixed now

PR #4342, opened 2024-10-02
fix os2.process_exec on non-windows and add a smoke test

laytan:
draft because ci is failing sporadically on this, need to look into that

### DIFF
--- a/core/os/os2/process.odin
+++ b/core/os/os2/process.odin
@@ -422,6 +422,7 @@ process_exec :: proc(
 				case nil: // nothing
 				case .EOF, .Broken_Pipe:
 					stdout_done = true
+					err = nil
 				case:
 					return
 				}
@@ -437,6 +438,7 @@ process_exec :: proc(
 				case nil: // nothing
 				case .EOF, .Broken_Pipe:
 					stderr_done = true
+					err = nil
 				case:
 					return
 				}
--- a/core/os/os2/process.odin
+++ b/core/os/os2/process.odin
@@ -398,8 +398,11 @@ process_exec :: proc(
 	{
 		defer if err != nil { _ = process_kill(process) }
 
-		stdout_builder := strings.builder_make(allocator) or_return
-		stderr_builder := strings.builder_make(allocator) or_return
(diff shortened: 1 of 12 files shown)


laytan:
Fixed a bunch of bugs here, FreeBSD and RISC-V Linux still misbehaving

laytan:
Ok found and fixed the FreeBSD issue, the RISC-V Linux deadlock is left.

core/os/os2/process.odin
```diff
@@ -392,44 +394,54 @@ process_exec :: proc(
...
-			hangup = true
-		case:
-			return
+
+		if !has_stdout {
+			has_data, err = pipe_has_data(stdout_r)
+			if has_data {
+				n, err = read(stdout_r, buf[:])
+				if strings.write_bytes(&stdout_builder, buf[:n]) != n {
+					err = .Out_Of_Memory
```
flysand7, line 412: This shouldn't return out of memory, because at the end we're pushing data to a string builder. Is there a way we can limit how much bytes we're reading or am I misreading what that does?

laytan: It is returning out of memory when the amount of bytes written to the string builder is not the same as the amount we try to write. I think its safe to assume you are out of memory then (the string builder didn't grow correctly).

flysand7: I realised I probably misread this code. Alright then

laytan: I changed it to just a dynamic array instead of a string builder to be clearer

laytan:
The QEMU we run in CI does not support `execveat` and that call is erroring with `ENOSYS`, then the code tries to reap the process here: https://github.com/odin-lang/Odin/blob/dd7c88989a0d66235ff8cfb0fc482d8503e8ca97/core/os/os2/process_linux.odin#L597

which launches this `waitid`: https://github.com/odin-lang/Odin/blob/dd7c88989a0d66235ff8cfb0fc482d8503e8ca97/core/os/os2/process_linux.odin#L661

and that does not return until the CI timeout of 15 minutes is up.

I do not know why that `waitid` never returns yet.

PR #4350, opened 2024-10-06
[core:encoding/json] Enumerated arrays as objects, bit sets as arrays, fix map sorting, fix bit set endian

jakubtomsu:
This PR has a whole bunch of changes to `core:encoding/json` which I wanted for a long time, as well as a number of bugfixes I found along the way.

Changes:
- option to store enumerated arrays as maps where keys are either enum values or names
- option to store bit sets as arrays of either integers or enum names
- fix map sorting, previously it would return unsupported type when a map isn't sortable. This caused issues in big nested datastructures. Now it sorts the map if possible, otherwise it writes it unsorted. Also currently only maps with string keys are considered sortable, it might be a good idea to add support for all ordered types. But I didn't see a good way to implement that.
- fix bit set endianness, it now swaps if endian is different than the platform

The reason why I need enumerated arrays and bitsets in a more readable form is to better handle changes in the underlying data structures. Currently it's unusable for e.g. game assets when enums can change over time, since enumerated arrays and bitsets will be out of order etc. And since the index in enumerated arrays and values in bitsets are implicit, it's difficult to procedurally upgrade assets from old format to a new one. Now it should be way more flexible since it's possible to always explicitly use the enum field name or value.

### DIFF
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -45,25 +45,23 @@ Marshal_Options :: struct {
 	// If spec is MJSON and this is true, then use '=' as delimiter between
 	// keys and values, otherwise ':' is used.
 	mjson_keys_use_equal_sign: bool,
-	
+
 	// When outputting a map, sort the output by key.
 	//
 	// NOTE: This will temp allocate and sort a list for each map.
-	// TODO: handle non sortable keys!!!
 	sort_maps_by_key: bool,
 
 	// Output enum value's name instead of its underlying value.
 	//
 	// NOTE: If a name isn't found it'll use the underlying value.
 	use_enum_names: bool,
-	
+
 	// Store enumerated arrays as a map instead of a regular array with implicit index.
 	// Set 'enumerated_array_map_use_names' to use the enum name string as a key.
 	enumerated_array_as_map: bool,
 	// when 'enumerated_array_as_map' is true, use the enum value name is used as a key instead of the enum integer value.
 	enumerated_array_map_use_names: bool,
-	
-	
+
 	// Store bit sets as a list of values instead of the raw underlying integer value.
 	// Set 'bit_set_array_use_names' to use the enum names instead of the value.
 	bit_set_as_array: bool,
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -408,16 +408,16 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err: 
 			case runtime.Type_Info_Type_Id:
 				return v.(typeid) == nil
 			case runtime.Type_Info_Pointer,
-			     runtime.Type_Info_Multi_Pointer,
-			     runtime.Type_Info_Procedure:
+				 runtime.Type_Info_Multi_Pointer,
+				 runtime.Type_Info_Procedure:
 				return (^rawptr)(v.data)^ == nil
 			case runtime.Type_Info_Dynamic_Array:
 				return (^runtime.Raw_Dynamic_Array)(v.data).len == 0
 			case runtime.Type_Info_Slice:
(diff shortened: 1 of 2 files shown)


core/encoding/json/marshal.odin
```diff
@@ -215,13 +227,46 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
...
+			for i in 0..<info.count {
+				val := info.min_value + reflect.Type_Info_Enum_Value(i)
+				key: string
+				buf: [40]byte
+				for ev, ei in enum_val.values {
+					if val == ev {
+						if opt.enumerated_array_map_use_names {
+							key = enum_val.names[ei]
+						} else {
+							key = strconv.append_bits(buf[:], cast(u64)enum_val.values[ei], 10, true, 8 * info.elem_size, "0123456789", nil)
```
laytan, line 243:
```suggestion
							key = strconv.append_int(buf[:], cast(i64)enum_val.values[ei], 10)
```

core/encoding/json/marshal.odin
```diff
@@ -215,13 +227,46 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
...
+					}
+				}
+				
+				if key == "" {
+					continue
+				}
+				
+				defer counter += 1
+				opt_write_iteration(w, opt, counter == 0) or_return
+				opt_write_key(w, opt, key)
```
laytan, line 255:
```suggestion
				opt_write_key(w, opt, key) or_return
```

core/encoding/json/marshal.odin
```diff
@@ -488,38 +535,75 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
...
 		}
+		
+		bit_data: u128
+		bit_size := u128(8*ti.size)
 
-		bit_data: u64
-		bit_size := u64(8*ti.size)
-
-		do_byte_swap := is_bit_set_different_endian_to_platform(info.underlying)
+		do_byte_swap := is_bit_set_different_endian_to_platform(ti)
```
laytan, line 542: `info.underlying` -> `ti` seems like a mistake? `ti` is going to be the bit set right? and the `is_bit_set_different...` is checking/expecting an integer.

afaict this is now always going to return false.

core/encoding/json/unmarshal.odin
```diff
@@ -119,7 +119,23 @@ assign_int :: proc(val: any, i: $T) -> bool {
...
+					switch info.endianness {
+					case .Platform: return false
+					case .Little:   return ODIN_ENDIAN != .Little
+					case .Big:      return ODIN_ENDIAN != .Big
+					}
+				}
+				return false
+			}
+			
+			do_byte_swap := is_bit_set_different_endian_to_platform(ti)
```
laytan, line 138: Same comment as in marshal, I think this needs to be given the underlying type, not the bit set itself

core/encoding/json/marshal.odin
jakubtomsu: that would make sense, I'll look into it and fix it. Thank you for the review

jakubtomsu:
@laytan do you know if the `underlying` type info can be something other than `Type_Info_Integer`? I can't think of anything.

Otherwise there could either be a regular union type assertion, or do assert(false) in the default switch case to catch this in a debug build.

laytan:
> @laytan do you know if the `underlying` type info can be something other than `Type_Info_Integer`? I can't think of anything.
> 
> Otherwise there could either be a regular union type assertion, or do assert(false) in the default switch case to catch this in a debug build.

After doing a `type_info_base` to unpack `Type_Info_Named` I think you are right that it can only be `Type_Info_Integer` (or `nil` of course). I agree with adding something that would assert/panic if not the case.

PR #4357, opened 2024-10-07
bindings, types, and constants for some mach vm apis

beaumccartney:
From my own bindings, cleaned up a bit to resemble what I see here.

I added:

- `vm_allocate()`
- `vm_deallocate()`
- `vm_map()`
- `mach_make_memory_entry()`
- `vm_page_size` (constant)
- types for each of the functions

I've reproduced the signatures, types/typedefs, and constants as exactly as
possible from the headers. I also kept doc comments from the original headers
(but they can go of course).

To that end, I wanted to ask about some of the types and procedure already in
`mach_darwin.odin`, as they don't match the original C api exactly. There is
a note about aligning the types to 16 bytes so I wanted to check if there was
a reason for the changes (e.g. in case I need to do something similar for the
type aliases I've given).

type aliases:
| type          | `mach_darwin.odin` | c api       |
|---------------|--------------------|-------------|
|`task_t`       |`u64`               |`mach_port_t`|
|`semaphore_t`  |`u64`               |`mach_port_t`|
|`kern_return_t`|`u64`               |`int`        |
|`thread_act_t` |`u64`               |`mach_port_t`|

procedures:
| procedure                 | `mach_darwin.odin`                                              | c api                                                               |
|---------------------------|-----------------------------------------------------------------|---------------------------------------------------------------------|
|`mach_task_self()`         |`proc() -> task_t`                                               |`proc() -> mach_port_t`                                              |
|`semaphore_signal_thread()`|`proc(semaphore: semaphore_t, thread: thread_t) -> kern_return_t`|`proc(semaphore: semaphore_t, thread: thread_t) -> kern_return_t`|

Note that `mach_port_t` is normally an `unsigned int`

Additionally, `semaphore_signal_thread()` takes a `thread_t`, not
a `thread_act_t` (though they're both typedef-ed to `mach_port_t`)

Also regarding aligning the types, the note referenced another note that said
that aligning a single-field struct didn't work at the time of writing. This
information was committed in 2019, so perhaps the issue is resolved now?

I made a vm-backed ring that maps multiple virtual pages onto the same physical
page to prove that the bindings are working. (I didn't want to clutter here so
I made a gist).
https://gist.github.com/beaumccartney/60cc84799dbdc1e4890ab1ec0a128fcb

I've made a thin wrapper to do make the api usable with more "odin-y" types,
it's included in the gist I linked above. Is a wrapper like mine worth
including?

- enums and bitsets
- `[^]byte` to reference a c pointer that represents the base of a region of
  memory

I didn't wrap `VM_FLAGS_*` in anything cause you can pack more information than
just toggling stuff ("superpage" stuff - its a bit weird)

### DIFF
--- a/core/sys/darwin/mach_darwin.odin
+++ b/core/sys/darwin/mach_darwin.odin
@@ -54,7 +54,7 @@ foreign mach {
 		offset         : vm_offset_t,
 		copy           : boolean_t,
 		cur_protection,
-		max_protection : vm_prot_t,
+		max_protection : VM_Prot_Flags,
 		inheritance    : vm_inherit_t,
 	) -> Kern_Return ---
 
@@ -62,7 +62,7 @@ foreign mach {
 		target_task   : vm_map_t,
 		size          : ^vm_size_t,
 		offset        : vm_offset_t,
-		permission    : vm_prot_t,
+		permission    : VM_Prot_Flags,
 		object_handle : ^mem_entry_name_port_t,
 		parent_entry  : mem_entry_name_port_t,
 	) -> Kern_Return ---
@@ -399,23 +399,17 @@ VM_FLAGS_OVERWRITE          :: 0x00004000  /* delete any existing mappings first
  *	Protection values, defined as bits within the vm_prot_t type
  */
 
-VM_PROT_NONE    : vm_prot_t : 0x00
+VM_Prot :: enum vm_prot_t {
+    Read,
+    Write,
+    Execute,
+}
 
-VM_PROT_READ    : vm_prot_t : 0x01      /* read permission */
-VM_PROT_WRITE   : vm_prot_t : 0x02      /* write permission */
-VM_PROT_EXECUTE : vm_prot_t : 0x04      /* execute permission */
+VM_Prot_Flags :: distinct bit_set[VM_Prot; vm_prot_t]
 
-/*
- *	The default protection for newly-created virtual memory
- */
-
-VM_PROT_DEFAULT :: VM_PROT_READ | VM_PROT_WRITE
-
-/*
- *	The maximum privileges possible, for parameter checking.
- */
-
-VM_PROT_ALL     :: VM_PROT_READ | VM_PROT_WRITE | VM_PROT_EXECUTE
+VM_PROT_NONE    :: VM_Prot_Flags{}
+VM_PROT_DEFAULT :: VM_Prot_Flags{.Read, .Write}
+VM_PROT_ALL     :: VM_Prot_Flags{.Read, .Write, .Execute}
 
 /*
  *	Enumeration of valid values for vm_inherit_t.
--- a/core/sys/darwin/mach_darwin.odin
+++ b/core/sys/darwin/mach_darwin.odin
@@ -4,20 +4,15 @@ foreign import mach "system:System.framework"
 
 import "core:c"
 
-// NOTE(tetra): Unclear whether these should be aligned 16 or not.
-// However all other sync primitives are aligned for robustness.
-// I cannot currently align these though.
-// See core/sys/unix/pthread_linux.odin/pthread_t.
-task_t :: distinct u64
-semaphore_t :: distinct u64
-
-kern_return_t :: distinct u64
-thread_act_t :: distinct u64
+kern_return_t :: distinct c.int
 
 mach_port_t :: distinct c.uint
 vm_map_t :: mach_port_t
 mem_entry_name_port_t :: mach_port_t
 ipc_space_t :: mach_port_t
+thread_t :: mach_port_t
+task_t :: mach_port_t
+semaphore_t :: mach_port_t
 
 vm_size_t :: distinct c.uintptr_t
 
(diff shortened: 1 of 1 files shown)


laytan:
> To that end, I wanted to ask about some of the types and procedure already in
mach_darwin.odin, as they don't match the original C api exactly. There is
a note about aligning the types to 16 bytes so I wanted to check if there was
a reason for the changes (e.g. in case I need to do something similar for the
type aliases I've given).

I don't think any of these types should be different from those in the headers, I don't see anything going on there about aligning, additionally the comment references a linux implementation of pthread? It's all a bit weird.

I would like to see the usage of bit sets and enums here, but you don't need wrappers for that. You can just make the binding return the enums and take the bit sets.

beaumccartney:
I've changed all the types to match the headers, and used enums and bitsets for everything except VM_FLAGS_*

I'm going to figure something out for VM_FLAGS cause there's a little bit of weirdness with stuff e.g. packing data inside the int as well as the flags. For reference the stuff I'm talking about is VM_FLAGS_* in mach/vm_statistics.h.

I'll try and have something for it done today.

beaumccartney:
@laytan I've added enum and bitset wrappers for everything. Let me know if there's anything else I can do.

beaumccartney:
https://github.com/beaumccartney/Odin/blob/b0ff41e673f8b788706488fb061a13a5862f5e96/core/sys/darwin/mach_darwin.odin#L77

is it possible to put that in rodata? when I prefix it with `@rodata` I can still compile it, but I can also just edit the value. Is that a bug?

core/sys/darwin/mach_darwin.odin
```diff
@@ -1,29 +1,514 @@
...
+	semaphore_signal_all :: proc(semaphore: semaphore_t) -> Kern_Return ---
+	semaphore_signal_thread :: proc(semaphore: semaphore_t, thread: thread_t) -> Kern_Return ---
+
+	semaphore_wait :: proc(semaphore: semaphore_t) -> Kern_Return ---
+
+	vm_allocate :: proc (target_task : vm_map_t, address: ^vm_address_t, size: vm_size_t, flags: VM_Flags) -> Kern_Return ---
+
+	vm_deallocate :: proc(target_task: vm_map_t, address: vm_address_t, size: vm_size_t) -> Kern_Return ---
+
+	vm_map :: proc (
```
laytan, line 49: The way of aligning these isn't really how it's normally done in the codebase.
This is usually the way:
```odin
proc(
    target_task: vm_map_t,
    address:     ^vm_address_t,
   ...
```

also it's a nit but could you make sure there is no space between `proc` and `(` anywhere.

core/sys/darwin/mach_darwin.odin
```diff
@@ -1,29 +1,514 @@
...
+		object_handle : ^mem_entry_name_port_t,
+		parent_entry  : mem_entry_name_port_t,
+	) -> Kern_Return ---
+
+	mach_port_deallocate :: proc (
+		task: ipc_space_t,
+		name: mach_port_name_t,
+	) -> Kern_Return ---
+
+	vm_page_size : vm_size_t
```
laytan, line 77:
```suggestion
	vm_page_size: vm_size_t
```

core/sys/darwin/mach_darwin.odin
```diff
@@ -1,29 +1,514 @@
...
+}
+
+Kern_Return :: enum kern_return_t {
```
laytan, line 80: The comments are weird here, can you either put them above their value, or inline with the value, not below like you've done here.

core/sys/darwin/mach_darwin.odin
```diff
@@ -1,29 +1,514 @@
...
+
+mach_port_name_t :: distinct c.uint
 
 @(default_calling_convention="c")
-foreign pthread {
-	mach_task_self :: proc() -> task_t ---
+foreign mach {
+	mach_task_self :: proc() -> mach_port_t ---
+
+	semaphore_create :: proc(task: task_t, semaphore: ^semaphore_t, policy, value: c.int) -> Kern_Return ---
```
laytan, line 36: Could the policy be an enum by chance?

laytan (commented):
Looking good! Few minor things

laytan:
> https://github.com/beaumccartney/Odin/blob/b0ff41e673f8b788706488fb061a13a5862f5e96/core/sys/darwin/mach_darwin.odin#L77
> 
> is it possible to put that in rodata? when I prefix it with `@rodata` I can still compile it, but I can also just edit the value. Is that a bug?

`@(rodata)` is mainly to put the variable in the readonly data section, when it's foreign that is not possible so this may need to be a compile error instead.

laytan:
> Is windows CI broken? From what I can tell, the issue's not on my end. AFAIK `_darwin.odin` is one of those suffixes for stuff that only builds on darwin, and the only thing I've changed has that suffix. Looks like some batch file didn't work.

It does that sometimes, I've ran it again.

> About the rodata thing, should I make an issue?

Yes please

beaumccartney:
https://github.com/odin-lang/Odin/issues/4363

PR #4376, opened 2024-10-14
core:sys/info AVX-512 CPU Features

Barinzaya:
Adds AVX-512 feature detection to `core:sys/info`.

Per a discussion in #package-dev in the Discord server, this adds each individual feature flag of AVX-512 (16 total) as individual entries in `CPU_Feature`, for a current total of 34 `CPU_Feature` values. 4FMAPS and 4VNNIW are excluded, as they don't have corresponding LLVM target features.

XCR0 bits to check were sourced from the [Intel Software Developer Manual](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html), volume 1, section 15.2.

CPUID feature flag bits were sourced from the [Intel Instruction Set Extensions document](https://cdrdv2.intel.com/v1/dl/getContent/671368), section 1.4, table 1-3.

Note that there is an apparently an issue where this detection may report a false negative on Darwin systems; see [a GitHub issue on the Go compiler](https://github.com/golang/go/issues/43089). This issue is *not* addressed by these changes.

### DIFF
--- a/core/sys/info/cpu_intel.odin
+++ b/core/sys/info/cpu_intel.odin
@@ -29,24 +29,22 @@ CPU_Feature :: enum u64 {
 	sse41,     // Streaming SIMD extension 4 and 4.1
 	sse42,     // Streaming SIMD extension 4 and 4.2
 
-	avx512_4fmaps,       // Fused Multiply Accumulation Packed Single precision
-	avx512_4vnniw,       // Vector Neural Network Instructions Word variable precision
-	avx512_bf16,         // Vector Neural Network Instructions supporting bfloat16
-	avx512_bitalg,       // Bit Algorithms
-	avx512_bw,           // Byte and Word instructions
-	avx512_cd,           // Conflict Detection instructions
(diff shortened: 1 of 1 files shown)


laytan:
What do you think of making it for example `avx512bw` instead of `avx512_bw` so they are equivalent to the LLVM naming used for `@(enable_target_features)` and the likes?

Barinzaya:
> What do you think of making it for example `avx512bw` instead of `avx512_bw` so they are equivalent to the LLVM naming used for `@(enable_target_features)` and the likes?

I think the only reason I didn't do that is because of the ones that start with a number, where the numbers run together (e.g. `avx5124fmaps`). In retrospect though, that's not a big deal and I think you're right that making them match the LLVM target features makes more sense. Will make the change.

Barinzaya:
Looking at it with that in mind, it seems like LLVM doesn't even have target features for those two (4FMAPS and 4VNNIW). In that case think I'll actually just remove them. They were only ever on one Intel CPU anyway (Knight's Mill).

PR #4385, opened 2024-10-17
Add debug info for labels to Odin

tf2spi:
This is a PR that adds debug info for labels in the ``Odin`` compiler, so now labels attached to statements in general like ``if``, ``block``, ``for``, etc., have debug info.

```
package labels
import "core:fmt"
main :: proc() {
        myloop: for i in 0..<5 {
                fmt.println("We have debug labels now!")
        }

        myswitch: switch true {
                case:
                        fmt.println("Debug labels on switches but not in cases")
        }
        myif: if true {
                fmt.println("Debug labels on if statements")
        }
        // etc...
}
```

This is useful because debuggers can discern the location of code given a function and label and do actions to it like putting breakpoints on it. It's less finnicky than line breakpoints when it comes to source code changes.
```
# This works on gdb and puts a breakpoint on the 'myif' label in the code above
b labels.main:myif
```

### DIFF
--- a/src/llvm_backend_debug.cpp
+++ b/src/llvm_backend_debug.cpp
@@ -1295,3 +1295,69 @@ gb_internal void add_debug_info_for_global_constant_from_entity(lbGenerator *gen
 		}
 	}
 }
+
+// TODO(tf2spi) ... *sigh* I hate LLVM so much.
+// LLVM-C doesn't have an API to emit DILabel yet, so this is not exported in the dll.
+// For Linux, we can temporarily use the C++ APIs, but we're out of luck for Windows
+// until the DLL gets updated with a definition like LLVMDIBuilderCreateLabel.
+// These might be the 50 most disgusting lines in this entire compiler.
+#ifndef _WIN32
+// Don't you just love a good ol' mangled name? I know I do.
+extern "C" LLVMMetadataRef _ZN4llvm9DIBuilder11createLabelEPNS_7DIScopeENS_9StringRefEPNS_6DIFileEjb(
+	LLVMDIBuilderRef Builder,
+	LLVMMetadataRef Scope,
+	String Name, // Not technically proper to do, but it's ABI compatible for now...
+	LLVMMetadataRef File,
+	unsigned int LineNo,
+	unsigned int AlwaysPreserve);
+#endif
+gb_internal void lb_add_debug_label(lbProcedure *p, Scope *parent, Ast *ast) {
+	if (p == nullptr || p->debug_info == nullptr || p->body == nullptr) {
+		return;
+	}
+	if (parent == nullptr) {
+		return;
+	}
+	if (ast == nullptr || ast->kind != Ast_Label) {
+		return;
+	}
+	Token label_token = ast->Label.token;
+	if (is_blank_ident(label_token.string)) {
+		return;
+	}
+	lbModule *m = p->module;
+	if (m == nullptr) {
+		return;
+	}
+
+	#ifndef _WIN32
+	AstFile *file = ast->file();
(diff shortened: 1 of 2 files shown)


src/llvm_backend_debug.cpp
```diff
@@ -1301,3 +1301,88 @@ gb_internal void add_debug_info_for_global_constant_from_entity(lbGenerator *gen
...
+		llvm_file,
+		label_token.pos.line,
+
+		// NOTE(tf2spi): Defaults to false in LLVM API, but I'd rather not take chances
+		//               Always preserve the label no matter what when debugging
+		true
+	);
+	_ZN4llvm9DIBuilder11insertLabelEPNS_7DILabelEPKNS_10DILocationEPNS_10BasicBlockE(	
+		m->debug_builder,
+		llvm_label,
```
tf2spi, line 1383: Note to self: llvm_label could be nullptr here? Although IIRC, LLVM uses ``new`` in a bunch of places and doesn't specify ``std::nothrow``, even in the C API, so I think exceptions would kill ``odin`` before it returns ``nullptr`` anyways.

src/llvm_backend_debug.cpp
```diff
@@ -1301,3 +1301,88 @@ gb_internal void add_debug_info_for_global_constant_from_entity(lbGenerator *gen
...
+	LLVMDIBuilderRef Builder,
+	LLVMMetadataRef Label,
+	LLVMMetadataRef Location,
+	LLVMBasicBlockRef Block);
+#endif
+gb_internal void lb_add_debug_label(lbProcedure *p, Ast *label, lbBlock *target) {
+	// TODO(tf2spi): Work with gingerBill to see if we can patch LLVM-C
+	//               to export DILabel functions for the C API on Windows
+	//               or to determine whether it is worthwhile to do this.
+	//               Until then, unimplemented, so just return.
```
gingerBill, line 1331: I do not want to change the LLVM-C API. It has to be whatever the official one is.

tf2spi: In retrospect, I now realize this is the more wise approach. Better to ask [llvm-project](https://github.com/llvm/llvm-project) to add a new C API that we can eventually use. [This ORCv2/LLJIT issue](https://github.com/llvm/llvm-project/issues/70666) is a fairly recent example of them adding the API fairly quickly when they detect a valid use case.

tf2spi: [My PR adding DebugInfo labels](https://github.com/llvm/llvm-project/pull/112840) got added recently. Whenever it's convenient and useful to update the ``LLVM-C`` DLL for Windows, I can add a case for Windows to use the new functions.

tf2spi: I saw that, a couple of months ago, LLVM-C in Odin got updated to 20 and this includes the DILabel wrappers I added to LLVM so now I use those in the Windows case. For the POSIX case, I still use the C++ functions to keep compatibility with older libraries.

tf2spi: I looked at the pipeline failures. It looks like they made a slight ABI change to ``llvm::DIBuilder::insertLabel`` to include an ``InsertPosition`` specifically in ``llvm@20`` and not in ``llvm@19`` and before. I suspect I can use ``dlsym`` with ``RTLD_NEXT`` to try and detect if the ``C`` API is present and then fall back to the old ``C++`` API if not.

laytan:
I think it's much better to just put this feature behind a check for llvm version > 20 and not bother with all the c++ and mangling hacks.

laytan:
They do provide that, `LLVM_VERSION_MAJOR` iirc, if you grep the code we use it in a couple of other places.

tf2spi:
Perfect! Great to know for the future and I just put in the version check now

PR #4394, opened 2024-10-19
Suggestion when assigning enum to bit_set

seventh-chord:
This is something I ran into when writing some odin again recently, after not having used the language for >1 year.

Incorrect code:

```
MyEnum :: enum { A, B, C }
bits: bit_set[MyEnum]
bits += .A
```

Old error:
```
C:/dev/tools/Odin/examples/bad_errors.odin(39:14) Error: Invalid type 'bit_set[MyEnum]' for implicit selector expression '.A'
        bits += .A
                 ^
```

New error:
```
C:/dev/tools/Odin/examples/bad_errors.odin(39:14) Error: Cannot convert enum value to 'bit_set[MyEnum]'
        bits += .A
                 ^
        Suggestion: Did you mean '{ .A }'?
```

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -8475,6 +8475,15 @@ gb_internal ExprKind check_implicit_selector_expr(CheckerContext *c, Operand *o,
 			error(node, "Undeclared name '%.*s' for type '%s'", LIT(name), typ);
 
 			check_did_you_mean_type(name, bt->Enum.fields);
+		} else if (is_type_bit_set(th) && is_type_enum(th->BitSet.elem)) {
+			ERROR_BLOCK();
+
+			gbString typ = type_to_string(th);
+			gbString str = expr_to_string(node);
+			error(node, "Cannot convert enum value to '%s'", typ);
+			error_line("\tSuggestion: Did you mean '{ %s }'?\n", str);
+			gb_string_free(typ);
+			gb_string_free(str);
 		} else {
 			gbString typ = type_to_string(th);
 			gbString str = expr_to_string(node);


seventh-chord:
Another error message which has tripped me up is for this code:
```
ambiguous_call :: proc() {
    p0 :: proc(x: int) {}
    p1 :: proc(x: string) {}
    my_group :: proc{p0, p1}
    my_group(does_not_exist)
}
```

I get:
```
C:/dev/tools/Odin/examples/bad_errors.odin(23:5) Error: Ambiguous procedure group call 'my_group' that match with the given arguments
        my_group(does_not_exist)
        ^~~~~~~^
        Given argument types: (invalid type)
        p0 :: proc(x: int) at C:/dev/tools/Odin/examples/bad_errors.odin(20:5)
        p1 :: proc(x: string) at C:/dev/tools/Odin/examples/bad_errors.odin(21:5)
C:/dev/tools/Odin/examples/bad_errors.odin(23:14) Error: Undeclared name: does_not_exist
        my_group(does_not_exist)
                 ^~~~~~~~~~~~~^
```

But I find having the "ambiguous proc group" error first confusing, since the mistake I made in the code was just typing in a wrong variable name.

I was looking at adding an early out to `check_call_arguments_proc_group` if any of the arguments had `type == t_invalid`, but I would have to spend some more time understanding how that function works to be confident that I don't break anything else.

If you think changing this case to remove the "ambiguous proc group" error then I can look into that.

gingerBill:
Please keep these requests separate since they are quite different.

However the first one is possible to add and something I have considered before. The second one is not at all simple. The second one might be "confusing" but it's because people never read the error message.

If you can suggest something actually better for the error message itself (not removing it), then please do. I am rarely sure how to improve an error message, especially when people don't give me good suggestions.

seventh-chord:
That's fine, I'll keep thinking about the "ambiguous proc group" error.

About the first suggestion, anything in particular you'd like me to change about this PR?

PR #4412, opened 2024-10-23
OpenGL: make loading GL less segfault prone

dayllenger:
- it's not obvious that e.g. `load_3_3` doesn't load basic 1.0 functions; if you use it you still get a segfault later
- added an assert to `load_up_to` to make it clear that you can't just pass nil there


gingerBill:
This procedure has existed for nearly 8 years now, so we are not changing the name. 

If you wanted to improve the documentation, that would be nice.

PR #4415, opened 2024-10-24
Added options for preventing Javascript event behavior when the listener callback is invoked.

p2jason:
Hello.

I am writing an app that targets Web and I want to handle Javascript events. Some of them, most notable mouse button events, have default behavior that I want to disable (opening the browser's right-click menu). Odin has a way of doing it currently, by calling `js.event_prevent_default` when the event handler is invoked, but this doesn't work.

My theory is that Javascript expects that you'll call these functions within the scope of the event listener callback. Odin queues up events until the next step, so calling `event.preventDefault` on them has no effect since its done outside the callback's scope.

**Browsers Tested:** Firefox

### DIFF
--- a/core/sys/wasm/js/events.odin
+++ b/core/sys/wasm/js/events.odin
@@ -313,14 +313,14 @@ foreign dom_lib {
 	dispatch_custom_event            :: proc(id: string, name: string, options := Event_Options{}) -> bool ---
 }
 
-add_event_listener :: proc(id: string, kind: Event_Kind, user_data: rawptr, callback: proc(e: Event), use_capture := false) -> bool {
+add_event_listener :: proc(id: string, kind: Event_Kind, user_data: rawptr, callback: proc(e: Event), use_capture := false, prevent_default := false, stop_prop := false, stop_imm_prop := false) -> bool {
 	@(default_calling_convention="contextless")
 	foreign dom_lib {
 		@(link_name="add_event_listener")
-		_add_event_listener :: proc(id: string, name: string, name_code: Event_Kind, user_data: rawptr, callback: proc "odin" (Event), use_capture: bool) -> bool ---
+		_add_event_listener :: proc(id: string, name: string, name_code: Event_Kind, user_data: rawptr, callback: proc "odin" (Event), use_capture: bool, prevent_default: bool, stop_prop: bool, stop_imm_prop: bool) -> bool ---
 	}
 	// TODO: Pointer_Lock_Change etc related stuff for all different browsers
-	return _add_event_listener(id, event_kind_string[kind], kind, user_data, callback, use_capture)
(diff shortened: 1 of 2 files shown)


core/sys/wasm/js/events.odin
```diff
@@ -313,14 +313,14 @@ foreign dom_lib {
 	dispatch_custom_event            :: proc(id: string, name: string, options := Event_Options{}) -> bool ---
 }
 
-add_event_listener :: proc(id: string, kind: Event_Kind, user_data: rawptr, callback: proc(e: Event), use_capture := false) -> bool {
+add_event_listener :: proc(id: string, kind: Event_Kind, user_data: rawptr, callback: proc(e: Event), use_capture := false, prevent_default := false, stop_prop := false, stop_imm_prop := false) -> bool {
```
gingerBill, line 316: I like the idea, but I think with that many flags, it should become a `bit_set`

```odin
Listener_Flag :: enum u32 {
	Use_Capture,
	Prevent_Default,
	Stop_Propagation,
	Stop_Immediate_Propagation,
}
Listener_Flags :: distinct bit_set[Listener_Flag; u32]
```

gingerBill (approved):

laytan:
`js.event_prevent_default` used to work when the callbacks were directly invoked when the event happened, but this has been changed where the callbacks are called at a later point because they are first queued.

p2jason:
Do you think its worth removing the old function (`event_prevent_default` + the other 2 like it)? I don't *think* there are any cases where they would still work

laytan:
Maybe a good idea.

To be honest I don't like the way it's queued now, mainly because there are valid use cases where you decide inside the event handler to prevent the default action or not, which the solution in this pr won't allow.

p2jason:
Adding a callback that gets invoked immediately when the event is received would be the ideal, but that feels like piling on more complexity without actually solving the problem. Do you know why it was changed so that the events are queued up instead of dispatched immediately?

laytan:
I don't remember why Bill added the queue, if I remember correctly it was some issue that he was attempting to fix by adding the queue but it didn't actually end up fixing it.

@gingerBill can you clarify that? Maybe the queue should just be removed so you can dynamically prevent the default action based on logic in the callback, and have it work as it did before.

gingerBill:
@p2jason The queue can be removed now, but it was originally added because I was getting events out of order (e.g. older timestamp AFTER younger timestamp).

p2jason:
Ok. I've removed the event queue. I've also added a callback that can be specified on the JS-side for intercepting events. This is useful when dealing with custom/unsupported event types. Let me know if it needs any more changes.

p2jason:
Cancel that. Just realized that the callback isn't needed now that the event queue is gone.

PR #4421, opened 2024-10-25
core/io: correctly escape runes greater than 0xFFFF

IllusionMan1212:
It seems to me that it's unintentional to escape with only 7 digits

### DIFF
--- a/core/io/util.odin
+++ b/core/io/util.odin
@@ -225,7 +225,7 @@ write_escaped_rune :: proc(w: Writer, r: rune, quote: byte, html_safe := false, 
 			} else {
 				write_byte(w, '\\', &n) or_return
 				write_byte(w, 'U', &n)  or_return
-				for s := 24; s >= 0; s -= 4 {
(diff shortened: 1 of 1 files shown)


Kelimion:
Unicode only goes up to 1,114,111 (0x10FFFF) codepoints, so the choice of 7 hex digits seems overkill already. When writing U+<hex>, encoders and decoders expect 4, 5 or 6 digits.

Unless you can point us to a unicode.org standard making the case for 8 hex digits, I think we'll have to give this a pass.

laytan:
Oh, partly my fault then, it was talked about in Discord and I also assumed 7 was a mistake because you usually see \u with 4 digits and \U with 8, seems like that's not actually a requirement then.

IllusionMan1212:
Well, I can't point you to a standard, but maybe it should still be something to consider since I'm using this while generating Odin code, and anything short of 8 digits is considered an invalid escape sequence.

PR #4423, opened 2024-10-26
math/rand: add `choice_bit_set`

laytan:


core/math/rand/rand.odin
```diff
@@ -687,3 +687,52 @@ choice_enum :: proc($T: typeid, gen := context.random_generator) -> T where intr
...
+	}
+
+Possible Output:
+	A false
+	B true
+	C true
+	5 true
+*/
+@(require_results)
+choice_bit_set :: proc(set: $T/bit_set[$E], gen := context.random_generator) -> (res: E, not_empty: bool) {
```
gingerBill, line 725: The word `not_empty` seems wrong to me but I cannot think of a better word to use.

laytan: Could just make it a more generic `ok`

gingerBill: Sorry for the minor gripe, it just "looked wrong".

laytan: No that's fine, I agree

PR #4424, opened 2024-10-27
path/filepath: fix leak when glob is recursed

gco-bmx:
path/filepath: fix leak when glob is recursed such as when pattern is "/folder/*/*/*.txt"

I was seeing leaks when running tests, of glob, this fixed it
not sure if the typical way to do this or not

### DIFF
--- a/core/path/filepath/match.odin
+++ b/core/path/filepath/match.odin
@@ -243,6 +243,7 @@ glob :: proc(pattern: string, allocator := context.allocator) -> (matches: []str
 
 	m: []string
 	m, err = glob(dir)
+	defer for s in m { delete(s) }; if m != nil { delete(m) }
 	if err != .None {
 		return
 	}


core/path/filepath/match.odin
```diff
@@ -243,6 +243,7 @@ glob :: proc(pattern: string, allocator := context.allocator) -> (matches: []str
 
 	m: []string
 	m, err = glob(dir)
+	defer for s in m { delete(s) }; if m != nil { delete(m) }
```
gingerBill, line 246: This is a use after free bug.

This should be the code:
```odin
defer {
	for s in m { 
		delete(s) 
	}
	delete(m)
}
```

gco-bmx: Sorry are you saying I should add the above to my code? When calling glob? 
As I did, have the code to delete the returned matches,
In this case as I understand, m isn't returned to the caller so it can't be deleted by the caller

Issue #4433, opened 2024-10-29
math/linalg matrix_mul_vector returns a type error on given inputs


blob1807:
## Context

Odin:    dev-2024-10:d8187d1cf
OS:      Windows 11 Home Basic (version: 23H2), build 22631.4317
CPU:     AMD Ryzen 7 6800HS with Radeon Graphics
RAM:     15608 MiB
Backend: LLVM 18.1.8

## Expected Behavior

No type error.

## Current Behavior

Passing a `maritx[I, J]T` & `[J]T` to matrix_mul_vector cause a type error
``Error: Cannot assign value 'a * b' of type 'matrix[2, 1]int' to '[3]int' in return statement``

## Steps to Reproduce
c1 & c2 should both be `matrix[2, 1]int {14, 36}`. 
Changing b's type or commenting out c2's area results in no error & the expected result.

```odin
package main

import "core:fmt"
import "core:math/linalg"

main :: proc() {    
    a := matrix[2, 3]int {
        1, 2, 3,
        4, 5, 6,
    }
    b := [3]int{1, 2, 3}
    //b := matrix[3, 1]int{1, 2, 3}

    c1 := a * b
    fmt.printfln("%T %v", c1, c1)

    c2 := linalg.mul(a, b)
    fmt.printfln("%T %v", c2, c2)
}
```

1. Copy Paste
2. Run
3. Profit...

## Proposed Solution
Add a matrix_mul_vector_differ which returns a matrix & change matrix_mul_vector to only take `maritx[I, I]T`.
Assuming that `maritx[I, J]T * [J]T` should equal  `matrix[I, 1]T` & not `[I]T`

gingerBill:
Why not just do `a * b`?

blob1807:
> Why not just do `a * b`?

True but the same could be said for the other simple mul procs.

gingerBill:
@blob1807 Yes and I saying prefer `a * b` most of the time where it works as expected.

blob1807:
I do use `a * b` normally anyways. I was checking out what linalg has & didn't expect it to have a type error on accepted inputs.

PR #4434, opened 2024-10-29
math/linalg add matrix_mul_vector_differ & update matrix_mul_vector

blob1807:
Currently passing a `maritx[I, J]T` & `[J]T` to matrix_mul_vector (as the inputs want) results in a type error.
`Error: Cannot assign value 'a * b' of type 'matrix[2, 1]int' to '[3]int' in return statement`

Add matrix_mul_vector_differ that returns the expect resulting matrix `maritx[I, 1]T` 
Update matrix_mul_vector to take only take a `maritx[I, I]T`
This change shouldn't effect any current use of matrix_mul_vector

Suggested fix for https://github.com/odin-lang/Odin/issues/4433

### DIFF
--- a/core/math/linalg/general.odin
+++ b/core/math/linalg/general.odin
@@ -212,7 +212,13 @@ matrix_mul_differ :: proc "contextless" (a: $A/matrix[$I, $J]$E, b: $B/matrix[J,
 
 
 @(require_results)
-matrix_mul_vector :: proc "contextless" (a: $A/matrix[$I, $J]$E, b: $B/[J]E) -> (c: B)
+matrix_mul_vector :: proc "contextless" (a: $A/matrix[$I, I]$E, b: $B/[I]E) -> (c: B)
+	where !IS_ARRAY(E), IS_NUMERIC(E) #no_bounds_check {
+	return a * b
+}
+
+@(require_results)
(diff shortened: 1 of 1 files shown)


core/math/linalg/general.odin
```diff
@@ -212,7 +212,13 @@ matrix_mul_differ :: proc "contextless" (a: $A/matrix[$I, $J]$E, b: $B/matrix[J,
...
 
 @(require_results)
-matrix_mul_vector :: proc "contextless" (a: $A/matrix[$I, $J]$E, b: $B/[J]E) -> (c: B)
+matrix_mul_vector :: proc "contextless" (a: $A/matrix[$I, I]$E, b: $B/[I]E) -> (c: B)
+	where !IS_ARRAY(E), IS_NUMERIC(E) #no_bounds_check {
+	return a * b
+}
+
+@(require_results)
+matrix_mul_vector_differ :: proc "contextless" (a: $A/matrix[$I, $J]$E, b: $B/[J]E) -> (c: matrix[I, 1]E)
```
gingerBill, line 221: Have you even tested this code to see if it even works?

blob1807: Yes but see I missed copying something over. Sorry about that.

Issue #4393, opened 2024-10-18
os.args parses \ incorrectly on Windows


clindholm:
## Context

Odin:    dev-2024-10-nightly:af9ae48
OS:      Windows 10 Home Basic (version: 22H2), build 19045.5011
CPU:     AMD Ryzen 7 5800X 8-Core Processor
RAM:     16305 MiB
Backend: LLVM 18.1.8

Given the following program:
```
package main

import "core:os"
import "core:fmt"

main :: proc() {
	for a, i in os.args[1:] {
		fmt.println(i, a)
	}
}
```

Let's say I run it with `odin run . -- test\ test test`
I'd expect:
```
0 test\
1 test
2 test
```

Instead I get:
```
0 test"  test  test
```
`odin run . -- test\ test\ test` gives:
```
0 test"  test"
1 test
```

Quoting makes no difference:
`odin run . -- "test\" 'test\' test` still gives:

```
0 test"  test"
1 test
```
Escaping does work:
`odin run . -- test\\ test`
```
0 test\
1 test
```
But still, that's not necessary with cli programs in general on Windows, and it's not consistent - escaping only escapes the final backslash:
`odin run . -- te\\st\\ test`
```
0 te\\st\
1 test
```

I tried looking at the code for os.args but I haven't worked with the Windows API so I couldn't easily tell what's going on there. I'll still take a closer look, but I thought if I report it someone else would have a fix much faster (or at all 😄).

Just in case, I also tried similar code with nodejs to make sure it's not something weird with my setup.

flysand7:
Which shell you're using? `cmd`, `powershell` or `pwsh`? Does the problem reproduce on all three of them?

Does the problem reproduce if instead `run -- <args>`, you first did `build` and then run the actual executable with the flags?

These are the things I would ask, because I myself have run into the issue with command line arguments being inconsistent across shells, and sometimes depending on whether `--` is used within the command line.

E.g. below I have saved up two screenshots, of a program that prints `os.args` one by one, being ran from cmd and pwsh, with and without `--` as one of the arguments.

![image](https://github.com/user-attachments/assets/4c8f87dd-7ae4-41ae-8216-192bc96d1ff3)

![image](https://github.com/user-attachments/assets/fa0bbaca-d86e-4005-8054-341251d01bd9)

clindholm:
Good points!
I should've said that I did test in different shells and that the problem shows up in all of them.

[...]

PR #4437, opened 2024-10-30
fix windows args parser problem from issue #4393

dpoetzschke:
I stumbled upon the same issue as #4393 
this patch should fix windows based parsing by escaping \ with \\

I tried to copy the existing style etc.

### DIFF
--- a/src/string.cpp
+++ b/src/string.cpp
@@ -411,6 +411,25 @@ gb_internal String concatenate4_strings(gbAllocator a, String const &x, String c
 	return make_string(data, len);
 }
 
+#if defined(GB_SYSTEM_WINDOWS)
+gb_internal String escape_char(gbAllocator a, String s, char cte) {
+	isize buf_len = s.len;
+	u8 *buf = gb_alloc_array(a, u8, buf_len);
+	isize i = 0;
+	for (isize j = 0; j < s.len; j++) {
+		u8 c = s.text[j];
+
+		if (c == cte) {
+			buf[i++] = '\\';
+			buf[i++] = c;
+		} else {
+			buf[i++] = c;
+		}
+	}
+	return make_string(buf, i);
+}
+#endif
+
 gb_internal String string_join_and_quote(gbAllocator a, Array<String> strings) {
 	if (!strings.count) {
 		return make_string(nullptr, 0);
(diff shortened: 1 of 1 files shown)


dpoetzschke:
Not quite sure what I could have broken in the FreeBSD demo check?!

src/string.cpp
```diff
@@ -411,6 +411,25 @@ gb_internal String concatenate4_strings(gbAllocator a, String const &x, String c
...
+#if defined(GB_SYSTEM_WINDOWS)
+gb_internal String escape_char(gbAllocator a, String s, char cte) {
+	isize buf_len = s.len;
+	u8 *buf = gb_alloc_array(a, u8, buf_len);
+	isize i = 0;
+	for (isize j = 0; j < s.len; j++) {
+		u8 c = s.text[j];
+
+		if (c == cte) {
+			buf[i++] = '\\';
```
gingerBill, line 423: This is an OOB problem. You didn't allocate enough memory in the first place.

src/string.cpp
```diff
@@ -411,6 +411,25 @@ gb_internal String concatenate4_strings(gbAllocator a, String const &x, String c
 	return make_string(data, len);
 }
 
+#if defined(GB_SYSTEM_WINDOWS)
+gb_internal String escape_char(gbAllocator a, String s, char cte) {
+	isize buf_len = s.len;
+	u8 *buf = gb_alloc_array(a, u8, buf_len);
```
Kelimion, line 417: The newly allocated String has the same length as input, even though zero or more backslashes will be added. This will result in an out of bounds access if something is actually escaped.

dpoetzschke: Not very familiar with C++.
Would something like work here
```
isize cte_count = 0;
for (isize j = 0; j < s.len; j++)
	if (s.text[j] == cte)
		cte_count++;
u8 *buf = gb_alloc_array(a, u8, buf_len+cte_count);
```

flysand7: Yes

dpoetzschke: changed accordingly

src/string.cpp
dpoetzschke: changed memory allocation in change request below

src/string.cpp
```diff
@@ -411,6 +411,30 @@ gb_internal String concatenate4_strings(gbAllocator a, String const &x, String c
 	return make_string(data, len);
 }
 
+#if defined(GB_SYSTEM_WINDOWS)
+gb_internal String escape_char(gbAllocator a, String s, char cte) {
+	isize buf_len = s.len;
+	isize cte_count = 0;
+	for (isize j = 0; j < s.len; j++)
+		if (s.text[j] == cte)
+			cte_count++;
```
gingerBill, line 420: Please use braces. That's our general style where possible.

dpoetzschke: braces added

PR #4454, opened 2024-11-05
Fix: `build_odin.sh` always runs demo regardless of argument

amekusa:
In the original code, `run_demo()` checks the argument if it's empty or `debug`:
```sh
# original `build_odin.sh` L:154
run_demo() {
	if [ $# -eq 0 ] || [ "$1" = "debug" ]; then
		./odin run examples/demo -vet -strict-style -- Hellope World
	fi
}
```
However, this `if` block always runs since `run_demo()` is always called without argument at L:162 and L:176, which isn't the intended behavior.
Instead, it should run demo only if the argument pass to `build_odin.sh` or `make` was `debug` or empty.

To fix this issue, I simply removed the `if` block from `run_demo()`, and made it gets called only if the argument that passed to `build_odin.sh` was empty or `debug`.

### DIFF
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -152,9 +152,7 @@ build_odin() {
 }
 
 run_demo() {
-	if [ $# -eq 0 ] || [ "$1" = "debug" ]; then
-		./odin run examples/demo -vet -strict-style -- Hellope World
-	fi
+	./odin run examples/demo -vet -strict-style -- Hellope World
 }
 
 if [ $# -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


flysand7:
Isn't it fine if it runs the demo on release builds?

amekusa:
This is the commit that added the `if` condition in `run_demo()`:
https://github.com/odin-lang/Odin/commit/8b98fff98e6447094fa54188c8bd1947ca684307

I quote the reasoning of the said commit:
> add debug build message and align more with build.bat
> 1. if ran without choosing a build type (just `make` or
>   `build_odin.sh`), print out a message about it being a debug build
> 2. Add `make release-native` alongside `make release_native` to align
>   with `build_odin.sh release-native`
> 3. Only run the demo if it is a debug build (just like `build.bat`)

But as I wrote, the `if` condition doesn't work as intended, since `$#` is not a reference of the number of args passed to `make` in cli, but the number of args passed to `run_demo()` function, which is, I think, not the intention of the original committer.

flysand7:
Okay!

laytan:
Thanks! That was my mistake at the time.

PR #4458, opened 2024-11-05
Combine `target` and `subtarget` cli options into one

0dminnimda:
The argument is pretty simple here:

Previously instead of having specific `-os:` and `-arch:` odin had single `-target:` option.  

I am proposing to continue that pattern and combine `-subtarget:` into `-target:`, especially since subtargets are closely tied to the target.  
For example: `-target:windows_amd64 -subtarget:ios`,  
`-target:darwin_arm64 -subtarget:android`, ... are meaningless, why not just bake the subtarget into the target, like `-target:darwin_ios_arm64`? Not like odin will allow to willy-nilly any combination, all of the cases will be checked in one way or another.

As a bonus, the cli is now more concise and the code is simpler! Also this change makes it trivial to implement automatic detection of the subtarget. And it's always a good sign, when the code is removed while preserving functionality.


gingerBill:
No. I told you that it was entirely intentional and I am not accepting this breaking change. `-subtarget:android` is the better option.

There might be loads of meaningless combination even in the target e.g. `-target:js_i386`.

PR #4459, opened 2024-11-06
Fix: Added newline after 'root' command output

davelowlevel:
When executing the `odin root` command, it displayed the root directory's location but without a newline at the end. As a result, the terminal prompt was stuck right next to the output, which caused a cluttered appearance. This change adds a newline after the root directory output to properly separate it from the terminal prompt, improving ux.

Before:
before
After:
after

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3181,7 +3181,7 @@ int main(int arg_count, char const **arg_ptr) {
 			return 0;
 		}
 	} else if (command == "root") {
-		gb_printf("%.*s", LIT(odin_root_dir()));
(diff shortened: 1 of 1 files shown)


karl-zylinski:
Not sure that's a good idea. People rely on it to find the odin executable from scripts, so then those strings may suddenly include a newline. So this is a breaking change I guess.

laytan:
Right, this is intentional and indeed used and relied upon for scripting so you can do `$(odin root)/core/blah` (in shell) to compose a path.

If you really want something like this it should detect if it's running in a tty and add the newline only in that case, but I'm not sure what use the command has outside of scripting in the first place.

PR #4463, opened 2024-11-07
core:log improvements

Lperlind:
There were various issues with core:log, namely that it would silently use context.allocator behind your back. The api has been changed to allow the user to provide an interface to the creation of file/console logger and multiloggers.

Furthermore updated the language used to be more consistent with odin's naming scheme. This renames create -> make and  detroy -> delete.

The old names and behaviour have been deprecated and have had their behaviour retained.

Additionally support for os2 has been added.

Documentation was added too.

### DIFF
--- a/core/log/file_console_logger.odin
+++ b/core/log/file_console_logger.odin
@@ -6,7 +6,10 @@ import "core:encoding/ansi"
 import "core:fmt"
 import "core:strings"
 import "core:os"
+import "core:os/os2"
 import "core:time"
+
+import "base:runtime"
 
 Level_Headers := [?]string{
 	 0..<10 = "[DEBUG] --- ",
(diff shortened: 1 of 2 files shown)


core/log/file_console_logger.odin
```diff
@@ -31,44 +34,158 @@ Default_File_Logger_Opts :: Options{
...
+- ident: An identifier that will be written alongside the logged message
+- close_file_on_delete: Sets the logger to close the file when delete_file_logger is called if `true`
+- allocator: (default: context.allocator)
+- loc: The caller location for debugging purposes (default: `#caller_location`)
+
+Returns:
+- res: The new file logger 
+- res: An allocator error if one occured, `nil` otherwise 
+*/
+make_file_logger_os2 :: proc(f: ^os2.File, lowest := Level.Debug, opt := Default_File_Logger_Opts, ident := "", close_file_on_delete := false, allocator := context.allocator, loc := #caller_location) -> (res: Logger, err: runtime.Allocator_Error) {
```
flysand7, line 107: What's the reason for implementing the of closing the file on the logger's side? Creating a logger isn't what created the file so it's weird that the responsibility for closing it falls onto the logger. Removing this feature will simplify the implementation by a lot, and closing it on the user's side is as simple as simply calling the appropriate function.

core/log/file_console_logger.odin
```diff
@@ -31,44 +34,158 @@ Default_File_Logger_Opts :: Options{
...
 
+@private
+File_Type :: enum {
+	console,
+	os,
+	os2,
+}
 
+/*
+	Implements a logger that writes to a file stream, such as stdout, or a file on the system.
```
flysand7, line 45: Please unindent the doc comments. The indented lines in markdown mark a code block, which is interpreted by Odin's documentation generator as well as other tools (such as editor extensions) as a block of code.

core/log/file_console_logger.odin
```diff
@@ -31,44 +34,158 @@ Default_File_Logger_Opts :: Options{
...
+- ident: An identifier that will be written alongside the logged message
+- close_file_on_delete: Sets the logger to close the handle when delete_file_logger is called if `true`
+- allocator: (default: context.allocator)
+- loc: The caller location for debugging purposes (default: `#caller_location`)
+
+Returns:
+- res: The new file logger 
+- res: An allocator error if one occured, `nil` otherwise 
+*/
+make_file_logger_os :: proc(h: os.Handle, lowest := Level.Debug, opt := Default_File_Logger_Opts, ident := "", close_file_on_delete := false, allocator := context.allocator, loc := #caller_location) -> (res: Logger, err: runtime.Allocator_Error) {
```
flysand7, line 79: I'd advise to break down very long lines.

```suggestion
make_file_logger_os :: proc(
	h: os.Handle,
	lowest := Level.Debug,
	opt := Default_File_Logger_Opts,
	ident := "",
	close_file_on_delete := false,
	allocator := context.allocator,
	loc := #caller_location,
) -> (
	res: Logger,
	err: runtime.Allocator_Error,
) {
```

core/log/multi_logger.odin
```diff
@@ -1,22 +1,53 @@
...
+
+	return Logger{multi_logger_proc, data, Level.Debug, nil}, nil
 }
 
-destroy_multi_logger :: proc(log: Logger) {
+delete_multi_logger :: proc(log: Logger) {
 	data := (^Multi_Logger_Data)(log.data)
-	delete(data.loggers)
-	free(data)
+	free(data, data.allocator)
```
flysand7, line 48: You aren't freeing the `data.loggers` slice here.

Update: Oh I see, you're doing a funny allocation dance. Maybe it's better to keep things simple and do two allocations. This isn't a place that's expected to be a performance bottleneck, so might be worth to simplify the code.

core/log/file_console_logger.odin
```diff
@@ -31,44 +34,158 @@ Default_File_Logger_Opts :: Options{
...
 	ident: string,
+	close_file_on_delete: bool,
+	allocator: runtime.Allocator,
 }
 
-create_file_logger :: proc(h: os.Handle, lowest := Level.Debug, opt := Default_File_Logger_Opts, ident := "") -> Logger {
-	data := new(File_Console_Logger_Data)
-	data.file_handle = h
+/*
+Makes a new logger that will write to the provided `os.Handle`
```
flysand7, line 62: A kind of a nitpick, but the first line of the documentation comment shouldn't describe what the procedure does, as in "Does X", but tells what it is doing in a more imperative way, e.g. "Do X". So for this procedure that would be

```
Make a new logger that will write to the provided Handle.
```

Also I'm not sure, if the usage of the word `make` is correct here per se. Make is used by the builtins that allow you to allocate data structures such as slices, dynamic arrays and maps. For singular objects, the words `new` and `delete` are used.

core/log/file_console_logger.odin
```diff
@@ -31,44 +34,158 @@ Default_File_Logger_Opts :: Options{
...
+- lowest: The lowest level logging to accept
+- opt: The wanted logging options
+- ident: An identifier that will be written alongside the logged message
+- close_file_on_delete: Sets the logger to close the handle when delete_file_logger is called if `true`
+- allocator: (default: context.allocator)
+- loc: The caller location for debugging purposes (default: `#caller_location`)
+
+Returns:
+- res: The new file logger 
+- res: An allocator error if one occured, `nil` otherwise
```
flysand7, line 77: `err:`

core/log/file_console_logger.odin
```diff
@@ -31,44 +34,158 @@ Default_File_Logger_Opts :: Options{
...
+- lowest: The lowest level logging to accept
+- opt: The wanted logging options
+- ident: An identifier that will be written alongside the logged message
+- close_file_on_delete: Sets the logger to close the file when delete_file_logger is called if `true`
+- allocator: (default: context.allocator)
+- loc: The caller location for debugging purposes (default: `#caller_location`)
+
+Returns:
+- res: The new file logger 
+- res: An allocator error if one occured, `nil` otherwise
```
flysand7, line 105: `- err:`

core/log/file_console_logger.odin
```diff
@@ -31,44 +34,158 @@ Default_File_Logger_Opts :: Options{
...
+Inputs:
+- lowest: The lowest level logging to accept
+- opt: The wanted logging options
+- ident: An identifier that will be written alongside the logged message
+- allocator: (default: context.allocator)
+- loc: The caller location for debugging purposes (default: `#caller_location`)
+
+Returns:
+- res: The new file logger 
+- res: An allocator error if one occured, `nil` otherwise
```
flysand7, line 136: - `err:`

core/log/multi_logger.odin
```diff
@@ -1,22 +1,53 @@
...
+*Allocates Using Provided Allocator*
+
+Inputs:
+- logs: The loggers this logger will write to
+- allocator: (default: context.allocator)
+- loc: The caller location for debugging purposes (default: `#caller_location`)
+
+Returns:
+- res: The new multi logger 
+- res: An allocator error if one occured, `nil` otherwise
```
flysand7, line 30: - `err:`

core/log/file_console_logger.odin
flysand7: Also I would avoid overdocumenting these types. You can look at the allocators in `core:mem` to see how I dealt with documenting allocators there, but basically the `_Data` structrs don't need more documentation than to say which objects they hold data for.

The documentation generator will make the correct links, and if you're using a text editor you'd never use this type so you'd never see its documentation. So the correct course of action is to put the documentation that describes that this is data for console and file loggers, and in the procedures that create them do a more detailed documentation.

core/log/file_console_logger.odin
```diff
@@ -31,44 +34,158 @@ Default_File_Logger_Opts :: Options{
...
+delete_console_logger :: proc(log: Logger) {
+	data := cast(^File_Console_Logger_Data)log.data
+	if data.close_file_on_delete {
+		switch data.type {
+		case .console:
+		case .os: os.close(data.file_handle_os)
+		case .os2: os2.close(data.file_handle_os2)
+		}
+	}
+	free(data, data.allocator)
```
flysand7, line 175: Need to pass the `data.loc` parameter.

flysand7 (commented):
Not a final review, I might have missed some obvious things in the docs.

core/log/file_console_logger.odin
Lperlind: It's to retain the old logger behavior until it's removed. I'll probably make it privately set by the create* functions instead.

core/log/multi_logger.odin
Lperlind: It just means there are two places for an error to occur rather than one. It's not for a performance optimisation.

core/log/file_console_logger.odin
flysand7: Leaking a file descriptor on logger destroy shouldn't have terrible side effects. I think it'd make sense to simplify the code right now and not worry about the breakage.

core/log/multi_logger.odin
flysand7: Let's not be weird.

core/log/file_console_logger.odin
```diff
@@ -6,8 +6,11 @@ import "core:encoding/ansi"
 import "core:fmt"
 import "core:strings"
 import "core:os"
+import "core:os/os2"
```
gingerBill, line 9: My issue with importing `os2` here is that it doesn't support everything yet. You might have to split `file_console_logger.odin` into two files. One for the platforms that support `os2`, and one for everything else.

PR #4466, opened 2024-11-07
SIMD Extract MSbs/LSbs Intrinsics

Barinzaya:
This PR adds `simd_extract_msbs`/`simd_extract_lsbs` intrinsics, along with corresponding aliases in `core:simd`. These intrinsics extract the most/least significant bit of each element of a `#simd` vector and packs them into a `bit_set`. This behavior is similar to the SSE/AVX `movemask` intrinsic/`movmsk` instruction. This intrinsic is defined similarly to:

```go
simd_extract_lsbs :: proc(a: #simd[N]T) -> bit_set[0..<N] where type_is_integer(T) || type_is_boolean(T)
simd_extract_msbs :: proc(a: #simd[N]T) -> bit_set[0..<N] where type_is_integer(T) || type_is_boolean(T)
```

For example, this code:
```go
v := #simd[8]i32 { -1, -2, +3, +4, -5, +6, +7, -8 }
fmt.println(simd.extract_msbs(v))
fmt.println(simd.extract_lsbs(v))
```
will print:
```
bit_set[0..=7]{0, 1, 4, 7}
bit_set[0..=7]{0, 2, 4, 6}
```
since elements 0, 1, 4, and 7 have their most significant bits set (due to being negative) and elements 0, 2, 4, and 6 have their least significant bits set (due to being odd)

This intrinsic is particularly useful in conjunction with lane-wise comparison masks, in a few particular use cases that I've found so far.

1. Counting matches. This can be useful on its own or in conjunction with `masked_compress_store`, for instance, where `card(extract_msbs(v))` can be used to determine how many elements will be written. Examples:
```go
count_range :: proc (src: []#simd[16]f32, lower, upper: f32) -> (result: int) {
	// Counts the number of values in `src` that are in the range `[lower, upper].`
	for v in src {
		mask := simd.lanes_ge(v, auto_cast lower) & simd.lanes_le(v, auto_cast upper)
		result += card(simd.extract_msbs(mask))
	}
	return
}

filter_range :: proc (dst: ^[dynamic]f32, src: []#simd[16]f32, lower, upper: f32) {
	// Appends to `dst` the values of `src` that are in the range `[lower, upper].`
	for v in src {
		mask := simd.lanes_ge(v, auto_cast lower) & simd.lanes_le(v, auto_cast upper)
		old_len := len(dst)
		non_zero_resize(dst, old_len + card(simd.extract_msbs(mask)))
		#no_bounds_check { simd.masked_compress_store(&dst[old_len], v, mask) }
	}
}
```
This use case *can* also be done via `-simd_reduce_add_ordered(mask)`, though I find using `extract_msbs` to be clearer on intent.

2. Identifying which elements matched, e.g. for searching or looping over them. Examples:
```go
first_range :: proc (src: []#simd[16]f32, lower, upper: f32) -> (at: int, found: bool) {
	// Returns the (flattened) index of the first value in `src` that is in the range `[lower, upper]`.
	for v, i in src {
		matches := simd.extract_msbs(simd.lanes_ge(v, auto_cast lower) & simd.lanes_le(v, auto_cast upper))
		for j in matches {
			at = 16*i + j
			found = true
			return
		}
	}
	return
}

histogram_range :: proc (bins: []int, src: []#simd[16]f32, lower, upper: f32) {
	// Generates histogram bin counts for the values in `src` in the range `[lower, upper)`.
	// bins[0] will be the number of values that fall in the first 1/Nth of the range,
	// bins[1] will be the number of values that fall in the second 1/Nth of the range, etc.
	span := (upper - lower) / f32(len(bins))
	for v in src {
		matches := simd.extract_msbs(simd.lanes_ge(v, auto_cast lower) & simd.lanes_lt(v, auto_cast upper))

		phases := (v - auto_cast lower) / auto_cast span
		indices := cast(#simd[16]i32)phases

		for j in matches {
			// Can't be done in parallel due to potential intersection
			i := int(simd.extract(indices, j))
			bins[i] += 1
		}
	}
	return
}
```

3. In more specific cases, the `bit_set` itself can also be directly useful for doing boolean logic en masse. Condensing the original checks to bit-masks allows the bit-masks themselves to be used in SIMD vectors to do large amounts of boolean logic at once.

### DIFF
--- a/base/intrinsics/intrinsics.odin
+++ b/base/intrinsics/intrinsics.odin
@@ -285,6 +285,8 @@ simd_reduce_xor         :: proc(a: #simd[N]T) -> T where type_is_integer(T) || t
 simd_reduce_any         :: proc(a: #simd[N]T) -> T where type_is_boolean(T) ---
 simd_reduce_all         :: proc(a: #simd[N]T) -> T where type_is_boolean(T) ---
 
+simd_extract_msbs       :: proc(a: #simd[N]T) -> bit_set[0..<N] where type_is_integer(T) || type_is_boolean(T) ---
+
 
 simd_gather       :: proc(ptr: #simd[N]rawptr, val: #simd[N]T, mask: #simd[N]U) -> #simd[N]T where type_is_integer(U) || type_is_boolean(U) ---
 simd_scatter      :: proc(ptr: #simd[N]rawptr, val: #simd[N]T, mask: #simd[N]U)              where type_is_integer(U) || type_is_boolean(U) ---
--- a/base/intrinsics/intrinsics.odin
+++ b/base/intrinsics/intrinsics.odin
@@ -285,6 +285,7 @@ simd_reduce_xor         :: proc(a: #simd[N]T) -> T where type_is_integer(T) || t
 simd_reduce_any         :: proc(a: #simd[N]T) -> T where type_is_boolean(T) ---
 simd_reduce_all         :: proc(a: #simd[N]T) -> T where type_is_boolean(T) ---
 
+simd_extract_lsbs       :: proc(a: #simd[N]T) -> bit_set[0..<N] where type_is_integer(T) || type_is_boolean(T) ---
 simd_extract_msbs       :: proc(a: #simd[N]T) -> bit_set[0..<N] where type_is_integer(T) || type_is_boolean(T) ---
 
 
--- a/core/simd/simd.odin
+++ b/core/simd/simd.odin
@@ -135,6 +135,7 @@ reduce_xor         :: intrinsics.simd_reduce_xor
 reduce_any         :: intrinsics.simd_reduce_any
 reduce_all         :: intrinsics.simd_reduce_all
 
+extract_lsbs       :: intrinsics.simd_extract_lsbs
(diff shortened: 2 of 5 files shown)


Barinzaya:
Any feedback on this?

flysand7:
I'm struggling to see the reason this couldn't be implemented using other SIMD functions in the package, I mean, why turn to LLVM

Barinzaya:
It technically *can*, but in my attempts, I wasn't able to find a way to implement it in Odin that actually results in remotely decent generated code.

This can be done in hardware in one instruction in x86 using the `movmsk` instructions. The intrinsic implementation uses a vector of `i1` as an intermediate step and bit-casts it to an equivalent integer, which LLVM does translate to `movmsk`. But `i1` can't be represented in Odin code, so the same approach can't be used. The alternatives I tried would avoid `movmsk` and result in terribly inefficient code, defeating the purpose.

Barinzaya:
Since this isn't getting any traction, would it be more acceptable if it were less specific? Maybe `extract_mask` instead, where the bit that it reads is technically implementation-defined?

PR #4471, opened 2024-11-10
List the supported targets using `odin build . -targets:?`

bayo-code:
Added support for listing supported targets using `odin targets`.


Kelimion:
This is asymmetric with `odin build . -microarch:?` and `-target-features:?`

bayo-code:
What do I need to add, please?

I'm new to the project, and I'm still trying to find my way around the code.

FourteenBrush:
I mean `odin build` prints the available targets when you attempt to do something like `odin build . -target:something-invalid`, but having it support `-target:?` doesn't sound bad.

Perhaps this could help?
https://github.com/odin-lang/Odin/blob/da196a2b74f58be1762d496942fea83371c96d03/src/main.cpp#L3280-L3310

bayo-code:
Ah, yes this makes sense now... I think I can do this 🙂

PR #4490, opened 2024-11-16
[runtime] `make(map[K]V)` should not allocate any capacity

Tetralux:
`make(map[K]V)` was resolving to `make_map_cap()` which allocates initial capacity when it wasn't intended to. It now calls `make_map()` which doesn't allocate any capacity.

Both `make(map[K]V)` and `make(map[K]V, allocator)` will NOT allocate initial capacity now.


Tetralux:
Went ahead and added a test for it too.

Tetralux:
Went ahead and added a test for it too.

PR #4494, opened 2024-11-17
[checker] Report error when builtin `min`/`max` has only one numeric parameter

jakubtomsu:
One parameter for `min` or `max` makes only sense when it's a type parameter. But when it's a regular numeric value (constant or variable), it's a no-op. I think in that case it's better to report an error to the user, because there is no practical reason to allow this and it makes it easier to notice a typo.

I ran into this multiple times in cases where I wrote `foo = max(something)` instead of `foo = max(foo, something_else)` to *accumulate* the maximum.


jakubtomsu:
oops forgot enums exist

PR #4497, opened 2024-11-18
-vet-struct-init

jockus:
Tentative pull request - adds a vet flag to ensure all struct values have been initialized when using compound literal value initialization.

Not enabled using any of the regular vet flags, as it will error on code in the core packages. Mainly useful as a file or package scope vet flag. Has been useful when working with Vulkan.

Example:

```
#+vet struct-init
package main

...

	icreateinfo := vk.InstanceCreateInfo {
		sType = .INSTANCE_CREATE_INFO,
		pApplicationInfo = &appinfo,
		enabledLayerCount = u32(len(required_layers)),
		ppEnabledLayerNames = raw_data(required_layers),
		enabledExtensionCount = u32(len(required_extensions)),
		ppEnabledExtensionNames = raw_data(required_extensions),
	}
```

Output:
```
src\vulkan.odin|136 col 17| Error: Incomplete struct initialization, missing members: 
|| 	icreateinfo := vk.InstanceCreateInfo { ... 
|| 	               ^~~~~~~~~~~~~~~~~~~~~~~ ... 
|| 	pNext 
|| 	flags 
````


gingerBill:
This is bad because it enforced EVERY struct to be checked, when in reality you only want certain types to be checked. 

I will not accept this.

jockus:
Figured there'd be some reasonable push back! Thanks.

wrapperup:
I also want to note that Vulkan's API was designed particularly around 0/nullptr being the default, so actually omitting members is perfectly fine in Odin context (assuming you otherwise follow the spec). This would bring a bit of friction.

PR #4507, opened 2024-11-22
Fix bindings for uniform 1iv, 2iv, 3iv, 4iv in webgl

Spad0n:
This PR addresses issues with the bindings for `uniform1iv`, `uniform2iv`, `uniform3iv`, and `uniform4iv` in WebGL. The implementation has been updated to align with the official WebGL API as documented on [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/uniform).

### DIFF
--- a/vendor/wasm/WebGL/webgl.odin
+++ b/vendor/wasm/WebGL/webgl.odin
@@ -158,6 +158,11 @@ foreign webgl {
 	Uniform3i :: proc(location: i32, v0, v1, v2: i32) ---
 	Uniform4i :: proc(location: i32, v0, v1, v2, v3: i32) ---
 	
+	Uniform1iv :: proc(location: i32, count: i32, value: rawptr) ---
+	Uniform2iv :: proc(location: i32, count: i32, value: rawptr) ---
+	Uniform3iv :: proc(location: i32, count: i32, value: rawptr) ---
+	Uniform4iv :: proc(location: i32, count: i32, value: rawptr) ---
+	
 	UseProgram      :: proc(program: Program) ---
 	ValidateProgram :: proc(program: Program) ---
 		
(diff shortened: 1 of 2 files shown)


vendor/wasm/WebGL/webgl.odin
```diff
@@ -174,10 +179,6 @@ Uniform1fv :: proc "contextless" (location: i32, v: f32)       { Uniform1f(locat
 Uniform2fv :: proc "contextless" (location: i32, v: glm.vec2)  { Uniform2f(location, v.x, v.y) }
 Uniform3fv :: proc "contextless" (location: i32, v: glm.vec3)  { Uniform3f(location, v.x, v.y, v.z) }
 Uniform4fv :: proc "contextless" (location: i32, v: glm.vec4)  { Uniform4f(location, v.x, v.y, v.z, v.w) }
-Uniform1iv :: proc "contextless" (location: i32, v: i32)       { Uniform1i(location, v) }
-Uniform2iv :: proc "contextless" (location: i32, v: glm.ivec2) { Uniform2i(location, v.x, v.y) }
-Uniform3iv :: proc "contextless" (location: i32, v: glm.ivec3) { Uniform3i(location, v.x, v.y, v.z) }
-Uniform4iv :: proc "contextless" (location: i32, v: glm.ivec4) { Uniform4i(location, v.x, v.y, v.z, v.w) }
```
gingerBill, line 180: Note you've just changed it to make it work for your use case and haven't touched the `UniformNfv` code above it to be "correct" either.

Spad0n: Indeed, I could have also updated the `UniformNfv` bindings. Would you prefer that I include these fixes in this PR, or create a separate one ?

PR #4541, opened 2024-11-30
Use a proper Queue in thread.Pool

Waqar144:
With lots of tasks the dynamic array takes a big performance hit as its allocating all the time on pop_front


gingerBill:
When you say it takes a big performance hit, did you hit this problem in reality or is this more a theoretical concern?

Waqar144:
I hit it in reality for a small tool I am working on. I had a list of filepaths and I wanted to run a search on those files. The quickest solution that I came to was using the threadpool by just adding a task for each file:

```odin
	for file in files {
		thread.pool_add_task(&pool, context.allocator, search_fn, file)
	}
	thread.pool_start(&pool)
	thread.pool_finish(&pool)
	thread.pool_destroy(&pool)
```

there were around ~70K files -> 70k tasks and the search was taking around 2 seconds. Quick profiling showed that the tasks buffer in thread pool was being reallocated too much. Changing it to a queue brought the time down to < .25 seconds, the actual time that the search was taking. This might be misusing the threadpool a bit though, but it was really easy to setup and get working.

PR #4554, opened 2024-12-03
Add misc\get-date.c

### LINKED ISSUE #4540: Error while building on Windows 11 ("misc\odin.rc(41) : error RC2104 : undefined keyword or key name: All")
My System: Windows 11x64

I've downloaded the master version (and also the dev-2024-11 version) and I can't build Odin because the following error occurs:

```bat
misc\odin.rc(41) : error RC2104 : undefined keyword or key name: All
main.cpp
libtommath.cpp
LINK : fatal error LNK1181: cannot open input file 'misc\odin.res'

mt : general error c101008c: Failed to read the manifest from the resource of file "\\?\R:\Sources\Odin\Odin\odin.exe".
```

I've made some tests, to find the cause of the issue and I found that the `build.bat file` of the `dev-2024-10` works well and if I replace the master version by this one, I can build Odin.

The changes in the file are very small. The line that causes the error in the build is :

In the master version (build is HS):
```bat
for /f %%i in ('powershell get-date -format "{yyyyMMdd}"') do (
    set CURR_DATE_TIME=%%i
)
```

In the dev-2024-10 version (build is OK):
```bat
for /f "usebackq tokens=1,2 delims=,=- " %%i in (`wmic os get LocalDateTime /value`) do @if %%i==LocalDateTime (
    set CURR_DATE_TIME=%%j
)
```
(note that this line of script is buggy, because the loop uses %%i and the CURR_DATE_TIME uses %%j,I've tested to put `%%i` instead of `%%j` and nothing changes

To debug the code deeper, I've added the line:
```bat
echo %odin_version_full%
```
after the line:
```bat
set odin_version_full="%V1%.%V2%.%V3%.%V4%"
```

in both files (master and dev-2024-10) and these are the results:

with the dev-2024-10 version -> "2024.11.0.0" is printed
with the master version -> "Loca.lD.0.0" is printed

I ran the command from cmd.exe.
```bat
powershell get-date -format "{yyyyMMdd}" 
```

I've got the error message:

```bat
The argument 'get-date' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
same error in a powershell terminal or in a command prompt console (cmd.exe)
```

Kelimion:
Prints the current date as YYYYMMDD without relying on PowerShell.

Hopefully fixes #4540

### DIFF
--- a/misc/get-date.c
+++ b/misc/get-date.c
@@ -1,0 +1,13 @@
+/*
+	Prints the current date as YYYYMMDD
+
+	e.g. 2024-12-25
+*/
+#include <stdio.h>
+#include <time.h>
+
+int main(int arg_count, char const **arg_ptr) {
+	time_t t = time(NULL);
+	struct tm* now = localtime(&t);
+	printf("%04d%02d%02d", now->tm_year + 1900, now->tm_mon + 1, now->tm_mday);
+}
--- a/build.bat
+++ b/build.bat
@@ -19,7 +19,11 @@ if "%VSCMD_ARG_TGT_ARCH%" neq "x64" (
 	)
 )
 
-for /f %%i in ('powershell get-date -format "{yyyyMMdd}"') do (
+pushd misc
+cl get-date.c
+popd
+
+for /f %%i in ('misc\get-date') do (
 	set CURR_DATE_TIME=%%i
 )
 set curr_year=%CURR_DATE_TIME:~0,4%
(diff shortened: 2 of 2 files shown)


Kelimion:
Huzzah!

misc/get-date.c
```diff
@@ -0,0 +1,13 @@
...
+	Prints the current date as YYYYMMDD
+
+	e.g. 2024-12-25
+*/
+#include <stdio.h>
+#include <time.h>
+
+int main(int arg_count, char const **arg_ptr) {
+	time_t t = time(NULL);
+	struct tm* now = localtime(&t);
```
laytan, line 11: Sure you want that in local time?

PR #4568, opened 2024-12-08
docs: improved `slice.reinterpret` docstring

giuliano-macedo:
I've noticed that the documentation for the procedure `slice.reinterpret` is malformed at pkg.odin-lang.org because it was using the ``` notation instead of `Example` and `Output`.
While I was at it, for the second example docstring, I ended up changing it a bit to avoid having a formatting incompatible with `ols` by  initializing an array and setting two values.

## Docs before:
reinterpret_before

## Docs now:
image

### DIFF
--- a/core/slice/slice.odin
+++ b/core/slice/slice.odin
@@ -48,22 +48,41 @@ to_type :: proc(buf: []u8, $T: typeid) -> (T, bool) #optional_ok {
 }
 
 /*
-	Turn a slice of one type, into a slice of another type.
+Turn a slice of one type, into a slice of another type.
 
-	Only converts the type and length of the slice itself.
-	The length is rounded down to the nearest whole number of items.
+Only converts the type and length of the slice itself.
+The length is rounded down to the nearest whole number of items.
 
-	```
-	large_items := []i64{1, 2, 3, 4}
-	small_items := slice.reinterpret([]i32, large_items)
-	assert(len(small_items) == 8)
-	```
-	```
-	small_items := []byte{1, 0, 0, 0, 0, 0, 0, 0,
-						  2, 0, 0, 0}
-	large_items := slice.reinterpret([]i64, small_items)
-	assert(len(large_items) == 1) // only enough bytes to make 1 x i64; two would need at least 8 bytes.
-	```
+Example:
+
+	import "core:fmt"
+	import "core:slice"
+
+	i64s_as_i32s :: proc() {
+		large_items := []i64{1, 2, 3, 4}
+		small_items := slice.reinterpret([]i32, large_items)
+		assert(len(small_items) == 8)
+		fmt.println(large_items, "->", small_items)
+	}
+
+	bytes_as_i64s :: proc() {
+		small_items := [12]byte{}
+		small_items[0] = 1
+		small_items[8] = 2
+		large_items := slice.reinterpret([]i64, small_items[:])
+		assert(len(large_items) == 1) // only enough bytes to make 1 x i64; two would need at least 8 bytes.
+		fmt.println(small_items, "->", large_items)
+	}
+
+	reinterpret_example :: proc() {
+		i64s_as_i32s()
+		bytes_as_i64s()
+	}
+
+Output:
+	[1, 2, 3, 4] -> [1, 0, 2, 0, 3, 0, 4, 0]
+	[1, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0] -> [1]
+
 */
 @(require_results)
 reinterpret :: proc "contextless" ($T: typeid/[]$U, s: []$V) -> []U {


laytan:
doc tests don't like the blocks I think:
```
--- Generating documentation test for "slice.reinterpret"
(5:1): you cannot use a block statement in the file scope
(11:1): you cannot use a block statement in the file scope
Errors on the following code generated for "reinterpret":
package documentation_verification

import "core:slice"
import "core:fmt"
{
	large_items := []i64{1, 2, 3, 4}
	small_items := slice.reinterpret([]i32, large_items)
	assert(len(small_items) == 8)
	fmt.println(large_items, "->", small_items)
}
{
	small_items := [12]byte{}
	small_items[0] = 1
	small_items[8] = 2
	large_items := slice.reinterpret([]i64, small_items[:])
	assert(len(large_items) == 1) // only enough bytes to make 1 x i64; two would need at least 8 bytes.
	fmt.println(small_items, "->", large_items)
}
```

I think you can just remove those (and separate by a comment or a few newlines if necessary), I also don't think we need to add the import statements in examples.

cstrachan88:
> I also don't think we need to add the import statements in examples.

The packages with documentation I have seen include the imports. See [`core:strconv`](https://github.com/odin-lang/Odin/blob/master/core/strconv/strconv.odin#L57) for instance. They also define a procedure with procedure name + "_example" instead of the blocks:

```odin
import "..."
reinterpret_example :: proc() { /* ... */ }
```

laytan:
That's not really a rule, if you want to add them you can, but I probably won't do it for procedures I document.

Things I've documented I like to keep short and sweet and only show what is relevant to the example. IMO it adds 5 lines of code to an example that is really 2 lines of code.

giuliano-macedo:
@laytan I didn't do it without adding blocks because I couldnt think of better name for the variables without duplicating them, I could only add the truncating example, but I do like the happy case documented as well...

I could this way as well, what do you think?

```odin
large_items1 := []i64{1, 2, 3, 4}
small_items1 := slice.reinterpret([]i32, large_items1)
assert(len(small_items1) == 8)
fmt.println(large_items1, "->", small_items1)

small_items2 := [12]byte{}
small_items2[0] = 1
small_items2[8] = 2
large_items2 := slice.reinterpret([]i64, small_items2[:])
assert(len(large_items2) == 1) // only enough bytes to make 1 x i64; two would need at least 8 bytes.
fmt.println(small_items2, "->", large_items2)
```

giuliano-macedo:
> I believe you can also do Example: multiple times, so you could split it up that way too.
I swear I tried this. Done! thanks for the comments @laytan @cstrachan88

laytan:
Sorry, one more thing and then it's perfect, you need to either indent the examples etc one more tab, or dedent the description by one tab.

You can see in your screenshot now that the description's font changed to a monospace font, because it is interpreted as preformatted text.

So to be clear the `Turn a slice` text needs to start at the same indentation level as the `Example:` "headers".

giuliano-macedo:
Done, thx!

laytan:
Ok I don't think this failure is your fault, I feel like it should be fine with that.

giuliano-macedo:
@laytan apparently it is mandatory to have imports and a "PROCEDURE_NAME_example" procedure to make `documentation_tester.odin` run properly.

I've changed accordingly should work in the CI now

core/slice/slice.odin
```diff
@@ -48,22 +48,41 @@ to_type :: proc(buf: []u8, $T: typeid) -> (T, bool) #optional_ok {
...
+	import "core:slice"
+
+	i64s_as_i32s :: proc() {
+		large_items := []i64{1, 2, 3, 4}
+		small_items := slice.reinterpret([]i32, large_items)
+		assert(len(small_items) == 8)
+		fmt.println(large_items, "->", small_items)
+	}
+
+	bytes_to_i64s :: proc(){
```
laytan, line 68:
```suggestion
	bytes_as_i64s :: proc() {
```

core/slice/slice.odin
```diff
@@ -48,22 +48,41 @@ to_type :: proc(buf: []u8, $T: typeid) -> (T, bool) #optional_ok {
...
+		small_items := [12]byte{}
+		small_items[0] = 1
+		small_items[8] = 2
+		large_items := slice.reinterpret([]i64, small_items[:])
+		assert(len(large_items) == 1) // only enough bytes to make 1 x i64; two would need at least 8 bytes.
+		fmt.println(small_items, "->", large_items)
+	}
+
+	reinterpret_example :: proc(){
```
laytan, line 77:
```suggestion
	reinterpret_example :: proc() {
```

core/slice/slice.odin
```diff
@@ -48,22 +48,41 @@ to_type :: proc(buf: []u8, $T: typeid) -> (T, bool) #optional_ok {
...
+		i64s_as_i32s();
```
laytan, line 78:
```suggestion
		i64s_as_i32s()
```

core/slice/slice.odin
```diff
@@ -48,22 +48,41 @@ to_type :: proc(buf: []u8, $T: typeid) -> (T, bool) #optional_ok {
...
+		bytes_to_i64s();
```
laytan, line 79:
```suggestion
		bytes_as_i64s()
```

laytan:
> @laytan apparently it is mandatory to have imports and a "PROCEDURE_NAME_example" procedure to make `documentation_tester.odin` run properly.
> 
> I've changed accordingly should work in the CI now

Hmm I am not a fan of that requirement, but that's besides the point, this looks good besides some code style issues.

Also weird that I did not fail CI on any of the procs/packages I've contributed because I haven't ever had to do this, this isn't new either.

giuliano-macedo:
Done!

giuliano-macedo:
Oops, the output is incorrect, I will fix it now

giuliano-macedo:
Done, there was a trailling tab sry

laytan:
Thanks and sorry for all the back and forth!

PR #4569, opened 2024-12-09
add `math.pow2` overloads

MaliusArth:
add overloads for `math.pow2_f{16, 32, 64}` in the vein of the `math.pow`/`math.pow10`s overloads

`odin test tests\internal`, which includes `test_pow.odin`, ran successfully on my `x86_64-win11` device. Given the nature of the changes I don't expect differing behavior on other targets.

### DIFF
--- a/core/math/math.odin
+++ b/core/math/math.odin
@@ -205,8 +205,21 @@ pow10_f64 :: proc "contextless" (n: f64) -> f64 {
 	return 0
 }
 
+@(require_results) pow2_f16le :: proc "contextless" (x: f16le) -> f16le { return #force_inline f16le(pow2_f16(f16(x))) }
+@(require_results) pow2_f16be :: proc "contextless" (x: f16be) -> f16be { return #force_inline f16be(pow2_f16(f16(x))) }
+@(require_results) pow2_f32le :: proc "contextless" (x: f32le) -> f32le { return #force_inline f32le(pow2_f32(f32(x))) }
+@(require_results) pow2_f32be :: proc "contextless" (x: f32be) -> f32be { return #force_inline f32be(pow2_f32(f32(x))) }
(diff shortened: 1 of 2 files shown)


core/math/math.odin
```diff
@@ -205,8 +205,21 @@ pow10_f64 :: proc "contextless" (n: f64) -> f64 {
...
+@(require_results) pow2_f64be :: proc "contextless" (x: f64be) -> f64be { return #force_inline f64be(pow2_f64(f64(x))) }
+pow2 :: proc{
+	pow2_f16, pow2_f16le, pow2_f16be,
+	pow2_f32, pow2_f32le, pow2_f32be,
+	pow2_f64, pow2_f64le, pow2_f64be,
+}
+
 @(require_results)
-pow2_f64 :: proc "contextless" (#any_int exp: int) -> (res: f64) {
+pow2_f64 :: proc "contextless" (n: f64) -> (res: f64) {
```
gingerBill, line 221: So the reason this took an integer and not a float in the first place is that `2^(non_integer)` is a valid expression which is not the same as doing `2^(integer)`.

MaliusArth: I think you mixed `non_integer` and `integer` but I have since realized my mistake, which is that those `pow2` procs are specifically meant for fast integer exponentiation and not general purpose `pow2`

MaliusArth: Comparing with the `pow10` procs I realized that `pow10`s shouldn't accept floats either given that the exponent is truncated.
(`testing.expect_value(t, math.pow10_f64(0.12345), math.pow_f64(10, 0.12345))` fails)

MaliusArth: I'd be happy to create a PR to fix `pow10`s signatures

PR #4580, opened 2024-12-14
Fix parsing of integer w/ 'e' notation

### LINKED ISSUE #4485: Incorrect parsing of integer constants like `2e1`
## Context

Originally mentioned in https://github.com/odin-lang/Odin/issues/4481#issuecomment-2473762184 , moving to a separate issue.

### Steps to Reproduce

This program

```
main :: proc() {
	i: int = 2e1
	j: int = 2.2e1
	fmt.printfln("%v, %v", i, j)
}
```

gives me this result:

```
10, 22
```

I'd  expect `2e1` to be `20`.

tf2spi:
Closes #4485 by making floating point values with 'e' and without a '.' treated as floats rather than ints.

### DIFF
--- a/src/big_int.cpp
+++ b/src/big_int.cpp
@@ -251,7 +251,10 @@ gb_internal void big_int_from_string(BigInt *dst, String const &s, bool *success
 			exp *= 10;
 			exp += v;
 		}
-		big_int_exp_u64(dst, &b, exp, success);
+		BigInt tmp = {};
+		mp_init(&tmp);
+		big_int_exp_u64(&tmp, &b, exp, success);
+		big_int_mul_eq(dst, &tmp);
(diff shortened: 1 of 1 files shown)


tf2spi:
Marked as draft b/c there may be other ways that floating point numbers are incorrectly parsed as big ints (like w/ '0x' and 'p', though IIRC, it's taken care of in other code above) and there may be a better way, like parsing a big int and observing the status code rather than doing syntactical checks.

tf2spi:
The original problem actually wasn't that ints were being parsed with 'e' notation. They were supposed to be but the destination got overwritten with ``10**exponent`` rather than ``number * 10**exponent`` like it was supposed to.

PR #4583, opened 2024-12-15
Draft: Convert expressions to typed bool where appropriate

### LINKED ISSUE #4244: Switch statement not type-checking
## Context
```
Odin:    dev-2024-08-nightly:8359995
OS:      CachyOS, Linux 6.10.6-2-cachyos
CPU:     Quad-Core Processor (up to 1.4GHz)
RAM:     3354 MiB
Backend: LLVM 18.1.8
```
## Expected Behavior

```go
r := 'a'
switch r {
    case r > '0':
}
```
Should throw a mismatched-types compile error, since we are comparing `rune` to `bool`
## Current Behavior

`r > '0'` is converted to `true`/`1`/`'0x01'` and then compared against `'a'`.
```go
switch "12" { case [2] int { 50, 51 }: } // causes an inconvertable type error.
switch 1 { case [2] int { 50, 51 }: } // compiles and runs (but may be because of array-programming)
switch true { case 1: } // similar error to previous
```

tf2spi:
Closes #4244 by making sure expressions that should be typed bools are typed as such

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -2849,6 +2849,7 @@ gb_internal void check_comparison(CheckerContext *c, Ast *node, Operand *x, Oper
 		switch (op) {
 		case Token_CmpEq: /* comp = comp; */ break;
 		case Token_NotEq: comp = !comp; break;
+			// TODO(tf2spi): Error out on the rest of the operators?
 		}
 		x->mode  = Addressing_Constant;
 		x->type  = t_untyped_bool;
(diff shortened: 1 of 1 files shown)


tf2spi:
Want to investigate TODOs and review whether conversions to ``t_bool`` are appropriate

gingerBill:
This is a bodge and not solving the original issue. You haven't considered why those were untyped in the first place.

Issue #4581, opened 2024-12-14
Unused variable error despite not using -vet-unused etc.


mtarik34b:
If an unused variable is declared inside a for-loop or an if-cond, a *'declared but not used'* error occurs despite compiling without `-vet-unused` and alike. This one triggers: https://github.com/odin-lang/Odin/blob/8b1c9b0ff5b7ad392fa48e050eae460da8edb982/src/check_stmt.cpp#L1594

**Odin Version:**
Odin:    dev-2024-12:8b1c9b0ff

### Steps to Reproduce
Build below program with `odin build .`
```odin
package pkg

main :: proc() {
	for i in 0 ..< 2 {
		var: int
	}

	if true { x := 123 }
}
```
### Failure Logs
```
Error: 'var' declared but not used 
        var: int 
        ^~^ 
Error: 'x' declared but not used 
        if true { x := 123; } 
                  ^ 
```

gingerBill:
So this isn't a bug and actually by design. It might seem like a bug, but it's fundamentally because code like this is pretty much always a bug or unintended.

PR #4593, opened 2024-12-18
Fix unused variables error without -vet flags enabled.

Creativty:
Fix for #4581.
It only adds a couple of lines.
A store for checking vet flags using the global `build_context.vet_flags` and a compare before starting the loop to check for errors.


gingerBill:
As described, this is not a bug.

PR #4600, opened 2024-12-19
Get arm64 up to speed in os2 linux

jasonKercher:
I saw @laytan removed the arm specific `Open_Flags_Bits` while working on `sys/posix`.  That looked like an accident, so I re-added them.  Please correct me if I'm wrong.

Re-added open flags for arm64 to sys/linux/bits.odin. Make process_start name based instead of descriptor based to allow running of scripts. Fix bug in heap_linux. Fix and simplify os2.remove.


laytan:
I have no clue why I did that anymore :p

PR #4603, opened 2024-12-20
Haiku: fix build and add initial `core:sys/posix` support

avanspector:


core/sys/haiku/errno.odin
```diff
@@ -1,11 +1,13 @@
 #+build haiku
```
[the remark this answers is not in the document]
avanspector: During a bit of refactor renamed it to be more consistent with other errno files since Haiku errors are superset of POSIX errno

core/sys/posix/signal.odin
```diff
@@ -185,6 +185,16 @@ foreign lib {
 	*/
 	sigwait :: proc(set: ^sigset_t, sig: ^Signal) -> Errno ---
 
+	when ODIN_OS != .Darwin {
+		/*
+		Wait for queued signals.
+
+		[[ More; https://pubs.opengroup.org/onlinepubs/9699919799/functions/sigtimedwait.html ]]
+		*/
+		sigtimedwait :: proc(set: ^sigset_t, info: ^siginfo_t, timeout: ^timespec) -> result ---
```
laytan, line 194: I don't love this, I tried making the posix package a common base where all defined procs are implemented on all targets it supports, now one could use sigtimedwait on a target expecting it to work when compiling to darwin.

I think I'd rather see it defined in the `sys/haiku` package instead.

laytan (commented):
Could you also try adding `haiku` to the test suite in `tests/core/sys/posix`?

I guess it might not work well because it's a start on the posix package so there's still things undefined that are used there though.

Would also be great if we could get everything for os2 implemented. Currently missing a lot of symbols though.

If you don't have time for these things we can just merge this and worry about those later though, at least it will build and run on haiku again.

core/sys/posix/signal.odin
avanspector: done

avanspector:
I added the test suite.
I think os2 needs its own PR, this PR is just to make Odin and most of core work on Haiku.
About the CI: Haiku package manager has its own CI infrastructure, i'll try make use of it at some point

laytan:
Thanks, looking good, one thing I'm wondering is why you chose `uintptr` instead of `uint` for what I am guessing is `size_t` in C?

tests/core/sys/posix/structs/structs.odin
```diff
@@ -70,8 +73,10 @@ main :: proc() {
 
 	fmt.println("utimbuf", size_of(posix.utimbuf), align_of(posix.utimbuf))
 
-	fmt.println("wordexp_t", size_of(posix.wordexp_t), align_of(posix.wordexp_t))
-
+	when ODIN_OS != .Haiku {
```
laytan, line 76: Can this be done with `when #defined(posix.wordexp_t)`? That would make it fail the test when the struct is eventually added so we don't forget.

avanspector: wordexp as well as sys/shm are not implemented on haiku afaict

avanspector:
> Thanks, looking good, one thing I'm wondering is why you chose `uintptr` instead of `uint` for what I am guessing is `size_t` in C?

Good idea, I still use them interchangeably, but `uint` is better here

tests/core/sys/posix/posix.odin
```diff
@@ -140,15 +158,15 @@ test_langinfo :: proc(t: ^testing.T) {
...
-		{ "//usr//lib//",     "//usr",      "lib" },
-		{ "/home//dwc//test", "/home//dwc", "test" },
+		{ "usr",              ".",                                                 "usr" },
+		{ "usr/",             ".",                                                 "usr" },
+		{ "",                 ".",                                                 "." },
+		{ "/",                "/",                                                 "/" },
+		{ "///",              "/",                                                 "/" },
+		{ "/usr/",            "/",                                                 "usr" },
+		{ "/usr/lib",         "/usr",                                              "lib" },
+		{ "//usr//lib//",     "//usr"      + ("/" when ODIN_OS == .Haiku else ""), "lib" },
```
laytan, line 168: Unfortunate that Haiku has non-standard behaviour here

tests/core/sys/posix/posix.odin
```diff
@@ -65,8 +65,17 @@ test_dirent :: proc(t: ^testing.T) {
 		for entry in entries {
 			defer posix.free(entry)
 
-			if entry.d_type != .REG {
-				continue
+			when ODIN_OS == .Haiku {
```
laytan, line 68: Unfortunate Haiku doesn't do the d_type convention like everyone else

PR #4619, opened 2024-12-24
vendor:raylib: add options to link with raylib/raygui from system

spitulax:
Odin ships the libraries for raylib and raygui and links to them when using their bindings from vendor. But this does not play well with Nix, presumably because the libraries are not compiled for Nix.
To use `vendor:raylib` in Nix, you have to also compile the libraries for Nix and link with it. 
This added an option to link the libraries from system instead of the libraries shipped into the compiler.
May fix <https://github.com/NixOS/nixpkgs/issues/338972>.

### DIFF
--- a/vendor/raylib/raygui.odin
+++ b/vendor/raylib/raygui.odin
@@ -3,8 +3,11 @@ package raylib
 import "core:c"
 
 RAYGUI_SHARED :: #config(RAYGUI_SHARED, false)
+RAYGUI_SYSTEM :: #config(RAYGUI_SYSTEM, false)
 
-when ODIN_OS == .Windows {
+when RAYGUI_SYSTEM {
+	foreign import lib "system:raygui"
+} else when ODIN_OS == .Windows {
 	foreign import lib {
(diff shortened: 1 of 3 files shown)


vendor/raylib/raygui.odin
```diff
@@ -3,8 +3,11 @@ package raylib
 import "core:c"
 
 RAYGUI_SHARED :: #config(RAYGUI_SHARED, false)
+RAYGUI_SYSTEM :: #config(RAYGUI_SYSTEM, false)
 
-when ODIN_OS == .Windows {
+when RAYGUI_SYSTEM {
+	foreign import lib "system:raygui"
+} else when ODIN_OS == .Windows {
```
gingerBill, line 10: This will break on Windows.

spitulax: `RAYLIB_SYSTEM` and `RAYGUI_SYSTEM` allows building in some systems that wouldn't accept bundled libraries like Nix. If you don't enable them on Windows it should just use the bundled libraries.

PR #4626, opened 2024-12-25
linux: os2._process_start: skip non-ELF files when searching in PATH

jkenda:
os2.process_start was trying to execute bash scripts that were flagged executable so I added a check for the ELF magic.

### DIFF
--- a/core/os/os2/process_linux.odin
+++ b/core/os/os2/process_linux.odin
@@ -392,6 +392,21 @@ _process_start :: proc(desc: Process_Desc) -> (process: Process, err: Error) {
 		return linux.access(strings.to_cstring(&b), linux.X_OK) == .NONE
 	}
 
+    is_elf_file :: proc(fd: linux.Fd) -> bool {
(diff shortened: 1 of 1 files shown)


jasonKercher:
I believe you *want* it to execute bash scripts in `$PATH`?  It currently fails to because of the `CLOEXEC` flag on the descriptor causes `ENOENT`.  So, for bash scripts we want to leak the file descriptor, and it currently does not.

I snuck a solution for this into #4600 and made it name based (maybe I should have made it separate).  I had trouble getting `execveat` with that `AT_EMPTY_PATH` flag to work within a container.  So, maybe the issue is more complex than just elf vs shebang?  The name based approach just works in all cases and makes the file descriptor the kernel's problem.

laytan:
I merged @jasonKercher 's solution to this because it allows executing shebangs while this skips them, the posix implementation also works with shebangs so this brings that up to the same standard, people also expect that behavior.

PR #4628, opened 2024-12-25
Add json.unparse: RTTI-less conversion from json.Value to []u8

bullyingteen:
This is a follow up PR to the smalltalk we had in Discord/General/Beginners.

### Introduction
While trying out odin's encoding/json module I've found out that the only way to convert json.Value to []u8/string was to use json.marshal (docs lack that information).
Which feels very unnatural and misleading to me as a C/C++ developer. So I did implement json.unparse. It probably has a tiny bit of performance boost due to it being RTTI-less.

Added test case for json.unparse, tested it on my windows 11.

Named it json.unparse for consistency with json.parse, but naming suggestions are wellcome.

### Bonus
Fixed an Unsupported_Type error in json.marshal on (rawptr)nil values.

### P.S. 
json.marshal uses Raw_Map routines when converting json.Object, but it requires runtime.Type_Info for accessing Map_Info hence I used simple k/v for-loop over map. Let me know if traversing Raw_Map buckets is preferred for some reason. I might implement it that way instead.

### DIFF
--- a/core/encoding/json/unparse.odin
+++ b/core/encoding/json/unparse.odin
@@ -29,8 +29,12 @@ unparse_to_builder :: proc(b: ^strings.Builder, v: Value, opt: ^Marshal_Options)
 }
 
 unparse_to_writer :: proc(w: io.Writer, v: Value, opt: ^Marshal_Options) -> io.Error {
+	if v == nil {
+		return unparse_null_to_writer(w, opt)
+	}
+
 	switch uv in v {
-		case Null: return unparse_null_to_writer(w, uv, opt)
+		case Null: return unparse_null_to_writer(w, opt)
 		case Integer: return unparse_integer_to_writer(w, uv, opt)
 		case Float: return unparse_float_to_writer(w, uv, opt)
 		case Boolean: return unparse_boolean_to_writer(w, uv, opt)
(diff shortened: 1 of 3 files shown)


gingerBill (commented):
There are loads of styling issues.

core/encoding/json/unparse.odin
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	switch uv in v {
+		case Null: return unparse_null_to_writer(w, opt)
+		case Integer: return unparse_integer_to_writer(w, uv, opt)
+		case Float: return unparse_float_to_writer(w, uv, opt)
+		case Boolean: return unparse_boolean_to_writer(w, uv, opt)
+		case String: return unparse_string_to_writer(w, uv, opt)
+		case Array: return unparse_array_to_writer(w, uv, opt)
+		case Object: return unparse_object_to_writer(w, uv, opt)
```
gingerBill, line 43: Not consistent with the Odin core style.

core/encoding/json/unparse.odin
```diff
@@ -0,0 +1,120 @@
...
+		for k,v in m {
+			opt_write_iteration(w, opt, first_iteration) or_return
+			opt_write_key(w, opt, k) or_return
+			unparse_to_writer(w, v, opt) or_return
+			first_iteration = false
+		}
+
+		opt_write_end(w, opt, '}') or_return
+	}
+	else {
```
gingerBill, line 98: } else {

karl-zylinski:
BTW, some of the styling issues Bill pointed out can be caught by compiler flag `-strict-style`

PR #4636, opened 2024-12-29
update cgltf from 1.13 to 1.14

2nafish117:
PR updates cgltf bindings from 1.13 to 1.14


2nafish117:
The wasm binary hasnt been compiled with the latest changes, im not sure how i can do that on windows.
the CI seems to be failing also, looks like a random fail?

laytan:
I've rebuild the libs and yeah those were random ci failures.

PR #4639, opened 2024-12-30
`vendor:vulkan` Add support for xlib and xcb

flga:
We need forward declarations for the WSI types, but xlib doesn't use any prefixes.
To make it clear and consistent with the other WSI types, I've added the `Xlib` prefix:

`Display` => `XlibDisplay`
`Window` => `XlibWindow`
`VisualID` => `XlibVisualID`

However, one could argue keeping the names consistent with xlib would be better, so let me know your thoughts on that.

### DIFF
--- a/vendor/vulkan/structs.odin
+++ b/vendor/vulkan/structs.odin
@@ -36,8 +36,7 @@ when ODIN_OS == .Windows {
 	}
 }
 
-// We want to use `vendor:x11/xlib` types so we need to match their build constraints.
-when ODIN_OS == .Linux || ODIN_OS == .FreeBSD || ODIN_OS == .OpenBSD {
+when xlib.IS_SUPPORTED {
 	XlibDisplay  :: xlib.Display
 	XlibWindow   :: xlib.Window
 	XlibVisualID :: xlib.VisualID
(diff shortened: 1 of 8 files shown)


laytan:
Could you not use `vendor:x11/xlib` for the types?

flga:
Done.

I thought that importing `x11/xlib` would imply xlib gets linked in, but anecdotally that seems to not be the case, the dependency gets elided if no procedures are called.

Can this be relied on will things like `reflect` get in the way of this "optimization"?

`xlib` has build constraints that have to be either matched or lifted. I chose the former for no other reason than trying to keep it consistent with win32.
This does mean that they need to be kept in sync but I suspect that's not a big deal.

laytan:
Yeah foreign dependencies don't get linked if none of their procs are called.

vendor/vulkan/structs.odin
```diff
@@ -32,7 +36,20 @@ when ODIN_OS == .Windows {
 	}
 }
 
-CAMetalLayer :: struct {}
+// We want to use `vendor:x11/xlib` types so we need to match their build constraints.
+when ODIN_OS == .Linux || ODIN_OS == .FreeBSD || ODIN_OS == .OpenBSD {
```
laytan, line 40: Instead of copying over targets and keeping them in sync. I would love to see a `xlib.IS_SUPPORTED` constant in xlib itself, similar to `time.IS_SUPPORTED` and `thread.IS_SUPPORTED`.

flga: Done

PR #4640, opened 2024-12-31
Made raylib bindings easier to use with emscripten

karl-zylinski:
Made raylib work with emscripten by importing `env.o` lib when ARCH is WASM. Also added raylib WASM lib, but that is for use with emcc after compilation of object file. This makes the raylib bindings work "out-of-the-box" with emscripten (i.e. you can compile, but it needs to be freestanding WASM target, which makes `core:os` and some other packages not work).

Also added the wasm raylib lib file.

TODO: Test raygui and perhaps compile a lib for it.


karl-zylinski:
I will probably rework this so we can override the lib path with a config flag. That way we don't need explicit emscripten fixes (having explicit emscripten fixes is a slippery slope of adding hacks).

karl-zylinski:
We merged this instead, which provides the same functionality without explicit emscripten hacks: https://github.com/odin-lang/Odin/pull/4647

PR #4641, opened 2024-12-31
encoding/base32: Add RFC-compliant error handling and improve reliability

zolk3ri:
The current implementation of base32.decode() uses assertions for (incomplete) validation, which causes programs to crash on invalid input. This PR replaces assertions with proper error handling according to RFC 4648, allowing users of the package to handle invalid input gracefully.

Motivation: While implementing a pure Odin implementation of TOTP (Time-based One-Time Password), I discovered that base32.decode() would segfault on invalid input due to assertions. As a core library package, it should provide proper error handling to allow users to gracefully handle decode errors, which these changes enable.

### Problem Description
- **Programs crash with segfaults** on invalid base32 input due to assertions
- Explicit disabling of bounds checking (`#no_bounds_check`) prevents proper error handling
- Buffer allocation issues can cause out-of-bounds access
- Non-compliance with RFC 4648 padding rules

The original implementation prioritized performance by disabling bounds checking, but this tradeoff is not appropriate for a core library package. The small or potentially negligible performance overhead from bounds checking is worth the improved safety and usability.

### Implementation Details

The error enum:
```odin
Error :: enum {
  None,
  Invalid_Character, // Input contains characters outside the specified alphabet
  Invalid_Length,    // Input length is not valid for base32 (must be a multiple of 8 with proper padding)
  Malformed_Input,   // Input has improper structure (wrong padding position or incomplete groups)
}
```

Before (crashes on invalid input):
```odin
data := decode(input)
if data == nil { // Only checks for empty input
  return nil
}
defer delete(data)
```

After (error handling available):
```odin
data, err := decode(input)
if err != .None {
  #partial switch err {
    case .Invalid_Character:
    // Handle error where input contains characters not in base32 alphabet
    case .Malformed_Input:
    // Handle error where invalid padding structure or position
    case .Invalid_Length:
    // Handle error where input length is not valid for base32
  }
}
defer delete(data)
```

Key implementation changes:
- Function now returns `([]byte, Error)` instead of just `[]byte`
- Removed `#no_bounds_check` directive for safety
- Proper buffer size calculation
- Added bounds validation throughout
- Proper padding validation
- **RFC 4648 compliant** error handling
- Support for custom alphabets with custom validation

### Design Decisions

1. Bounds Checking:
   - Original code used `#no_bounds_check` for performance
   - Removed this directive despite small (if any) performance impact
   - Proper error handling requires bounds validation
   - Safety and usability outweigh minor performance cost
   - Follows Odin's philosophy of being explicit and safe by default

2. Error Handling:
   - Simple enum with distinct error cases
   - Clear separation between character, length, and structure errors
   - RFC 4648 compliant validation
   - Empty input returns `(nil, .None)` by convention

3. Testing:
   - RFC 4648 test vectors
   - Tests all error conditions
   - Verifies correct handling of edge cases

### Breaking Changes

1. Function Signature:
   - **decode() now returns `([]byte, Error)` instead of `[]byte`**
   - Error handling available but not required

2. Performance:
   - Minor overhead from bounds checking
   - No additional allocations for error handling

3. Error Handling:
   - Programs can handle errors if desired
   - *No more unexpected crashes on invalid input*

### Testing

Test coverage includes:
- RFC 4648 test vectors and validation rules
- Character validation (invalid characters and case sensitivity)
- Padding validation (placement and length requirements)
- Length validation
- Roundtrip encoding/decoding
- Custom alphabet support

All tests pass with 100% coverage of the new functionality and error cases.

### Future Work
- Add similar error handling improvements to encoding/base64 package
- Consider adding error handling to encode() for input validation

Thank you for considering this contribution.

Happy New Year! :)


bullyingteen:
encoding/base64 needs the same treatment I guess?
See base64.decode_into

bullyingteen:
> > encoding/base64 needs the same treatment I guess? See base64.decode_into
> 
> 
> 
> Yup! That is why I mentioned it under "Future Work". If this happens to get merged, I will work on encoding/base64 as well.

Missed it ngl 🙂

core/encoding/base32/base32.odin
```diff
@@ -1,148 +1,228 @@
...
+}
+
+Validate_Proc :: #type proc(c: byte) -> bool
+
+@private
+_validate_default :: proc(c: byte) -> bool {
+	return (c >= 'A' && c <= 'Z') || (c >= '2' && c <= '7')
+}
+
+ENC_TABLE := [32]byte {
```
zen3ger, line 31: It would be nice to have `@(rodata)` on both default tables.

core/encoding/base32/base32.odin
```diff
@@ -1,148 +1,230 @@
...
+	out_len := input_chars * 5 / 8
+	out = make([]byte, out_len, allocator)
+	defer if err != .None {
+		delete(out)
+	}
+
+	// Process input in 8-byte blocks
+	outi := 0
+	for i := 0; i < input_chars; i += 8 {
+		buf: [8]byte
```
Waqar144, line 200: can skip 0 init here by using `---`

core/encoding/base32/base32.odin
```diff
@@ -1,148 +1,230 @@
...
+		block_size := min(8, input_chars - i)
+
+		// Decode block
+		for j := 0; j < block_size; j += 1 {
```
Waqar144, line 204: style: you can use range style loop: `for j in 0 ..< block_size` if you want

core/encoding/base32/base32.odin
```diff
@@ -1,148 +1,230 @@
...
+		return nil, .None
+	}
+
+	// Check minimum length requirement first
+	if len(data) < 2 {
+		return nil, .Invalid_Length
+	}
+
+	// Validate characters using provided validation function
+	for i := 0; i < len(data); i += 1 {
```
Waqar144, line 135: style: you can use range style loop: `for c in data` if you want.

alternatively perhaps `bytes.index_byte` from `core:bytes` can be used to find the index of first `PADDING`. and then use that for this loop and the one below to eliminate the `PADDING` branch.

core/encoding/base32/base32.odin
```diff
@@ -1,148 +1,230 @@
...
+		}
+		if !validate(c) {
+			return nil, .Invalid_Character
+		}
+	}
+
+	// Validate padding and length
+	data_len := len(data)
+	padding_count := 0
+	for i := data_len - 1; i >= 0; i -= 1 {
```
Waqar144, line 148: `#no_bounds_check` can be used here

core/encoding/base32/base32.odin
```diff
@@ -1,148 +1,230 @@
...
+		if data[i] != byte(PADDING) {
+			break
+		}
+		padding_count += 1
+	}
+
+	// Check for proper padding and length combinations
+	if padding_count > 0 {
+		// Verify no padding in the middle
+		for i := 0; i < data_len - padding_count; i += 1 {
```
Waqar144, line 158: `#no_bounds_check` can be used here

core/encoding/base32/base32.odin
```diff
@@ -1,148 +1,230 @@
...
+
+encode :: proc(data: []byte, ENC_TBL := ENC_TABLE, allocator := context.allocator) -> string {
+	out_length := (len(data) + 4) / 5 * 8
+	out := make([]byte, out_length, allocator)
+	_encode(out, data, ENC_TBL)
+	return string(out[:])
+}
+
+@private
+_encode :: proc(out, data: []byte, ENC_TBL := ENC_TABLE, allocator := context.allocator) {
```
Waqar144, line 69: can `#no_bounds_check` can be used for this proc?

core/encoding/base32/base32.odin
```diff
@@ -1,148 +1,230 @@
...
+		for i := 0; i < data_len - padding_count; i += 1 {
+			if data[i] == byte(PADDING) {
+				return nil, .Malformed_Input
+			}
+		}
+
+		content_len := data_len - padding_count
+		mod8 := content_len % 8
+		required_padding: int
+		switch mod8 {
```
Kelimion, line 167: I suppose this works fine, but I might be a bit more explicit about the values here:
```odin
		// Remainders of 2, 4, 5, and 7 need to be padded to 8 bytes.
		switch mod8 {
		case 2, 4, 5, 7: required_padding = 8 - mod8
		case: required_padding = 0
		}
```
Alternatively, you could use a LUT:
```odin
// Remainders of 2, 4, 5, and 7 need to be padded to 8 bytes.
PAD_LENGTH := [8]int{0, 0, 6, 0, 4, 3, 0, 1}
required_padding := PAD_LENGTH[content_len % 8]
```

Kelimion (commented):
Two minor nitpicks. Looks good to me otherwise. Also thanks for adding a test suite for the package.

core/encoding/base32/base32.odin
```diff
@@ -1,148 +1,230 @@
...
+		case 2: required_padding = 6 // 2 chars need 6 padding chars
+		case 4: required_padding = 4 // 4 chars need 4 padding chars
+		case 5: required_padding = 3 // 5 chars need 3 padding chars
+		case 7: required_padding = 1 // 7 chars need 1 padding char
+		case: required_padding = 0
+		}
+
+		if required_padding > 0 {
```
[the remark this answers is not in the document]
Kelimion: And the `mod8 != 0` branch could maybe even go away? Because that's only ever true if `required_padding` is 2, 4, 5 or 7. 🤔

core/encoding/base32/base32.odin
Waqar144: Clearer is subjective, but you are the coder  here so its your call.

core/encoding/base32/base32.odin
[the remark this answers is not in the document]
Kelimion: Never mind. I wrote a nested loop iterating over combinations of `content_len` and `padding_count`, and the "simplified" test for malformed isn't really all that simplified. So let's keep your nested conditions. :-)
```odin
			is_malformed := false
			if required_padding > 0 {
				if padding_count != required_padding {
					is_malformed = true
				} else if mod8 != 0 {
					is_malformed = true
				}
			}

			// Simplified check
			alt_malformed := required_padding > 0 &&
				((padding_count != required_padding) ||
				 (mod8 > 0 && padding_count == required_padding))
```

Kelimion:
> Motivation: While implementing a pure Odin implementation of TOTP (Time-based One-Time Password), I discovered that base32.decode() would segfault on invalid input due to assertions.

I've been considering writing a TOTP implementation for `core`, as it fits in nicely with `core:net` and `core:crypto`, and the work underway by Laytan to support http(s). So if you wanted to contribute TOTP, I'd certainly be willing to give it a fair shake.

Kelimion:
> I would be interested in contributing my TOTP implementation to `core`. My current implementation is available at [totp.odin](https://zolk3ri.name/cgit/odin-totp/plain/totp.odin). As noted in the [README](https://zolk3ri.name/cgit/odin-totp/about/), this was my first Odin code and first experience with the language, so there might be areas that could use improvement.

Great. I'll give it a read soon.
 
> I haven't yet pushed the changes that utilize the improved error handling for the `encoding/base32` package from my PR. Initially, I planned to add base32 validation in my TOTP library, but I thought it would be more beneficial to improve the `encoding/base32` package directly instead.

Indeed.
 
> As a side note, I noticed there are many uses of `assert()` throughout `core` (some likely valid). If you're open to exploring more robust error handling approaches (allowing users to handle errors gracefully), I would be happy to help improve packages as needed, in collaboration with others.

Many of those are intentional, as they deal with programmer/user error. If a certain state is only possible because of misuse of an API, not untrusted input, then an assert is sensible. It's the kind of feedback a programmer can't ignore and will helpfully tell them they've neglected to pass an appropriately sized buffer, for example. So the rule of thumb we've settled on for newer implementations is to have asserts for "contract violations", and errors for well, errors.

> I've also been working on a minimal libcurl wrapper, available at [odin-curl](https://zolk3ri.name/cgit/odin-curl/). The goal is to provide libcurl functionality with idiomatic Odin features where appropriate, rather than just a direct FFI binding.
 
 We'll have to have a think about whether that fits. As a binding it would have to live in `vendor`, not `core`, but it may be better situated in your own repository and added to Mjolnir's package index and other places for people to find it.
 
> I'm happy to collaborate on Discord (I've joined but haven't been very active), though I should mention that I may be away for a couple of months due to health reasons. If the TOTP work can't wait, feel free to use what you need from my code - it's completely fine. :)

At your leisure. Health comes first.

core/encoding/base32/base32.odin
Kelimion: I haven't read the RFC, so I'll defer to you on that. But to me at least, 6, 4, 3 and 1 felt like magic numbers, even though they were clearly derived from `8 - mod8`, so the `8 - mod8` felt more explicit than "these are the magic numbers". If you follow my train of thought.

But that could be added to the comments instead?
```odin
// Requirement is for 2, 4, 5 and 7 remaining characters to be padded to 8. So the values here are `8 - mod8`.
switch mod8 {
case 2: required_padding = 6 // 2 chars need 6 padding chars
case 4: required_padding = 4 // 4 chars need 4 padding chars
case 5: required_padding = 3 // 5 chars need 3 padding chars
case 7: required_padding = 1 // 7 chars need 1 padding char
case: required_padding = 0
}
```

Kelimion:
> Thank you a lot! :)

Thank you for the package's cleanup. :-)

Don't mind the CI error. That spurious one occasionally happens on the RISCV target, and I don't really feel like special casing the test to ignore it on that target.

PR #4647, opened 2025-01-01
Make it easier to do Raylib WASM builds

karl-zylinski:
Added WASM libs for raylib and raygui.

The new libs are added to the foreign import blocks, with optional ways to override them. These override makes it possible to use the bindings with emscripten (so you can use `-define:RAYLIB_WASM_LIB=env.o`)

This can be compared to my other PR: https://github.com/odin-lang/Odin/pull/4640 -- that one imports `env.o` directly, but having such emscripten specifics hacks in there wasn't popular (although, the raylib WASM libs won't work without emscripten anyways. The default imports of `wasm/libraylib.a` in this PR can be seen as documentation of which libs to use with emcc).

Initially I wanted to add a global raylib library file override, but the foreign import blocks have multiple items, such as this one:
```
foreign import lib {
	"windows/raylibdll.lib" when RAYLIB_SHARED else "windows/raylib.lib" ,
	"system:Winmm.lib",
	"system:Gdi32.lib",
	"system:User32.lib",
	"system:Shell32.lib",
}
```
So adding in an override for the just the lib file would get messy.


vendor/raylib/raygui.odin
```diff
@@ -3,6 +3,7 @@ package raylib
 import "core:c"
 
 RAYGUI_SHARED :: #config(RAYGUI_SHARED, false)
+RAYGUI_WASM_LIB :: #config(RAYGUI_WASM_LIB, "")
```
gingerBill, line 6: Why not just `RAYGUI_WASM_LIB :: #config(RAYGUI_WASM_LIB, "wasm/libraygui.a")`?

karl-zylinski: Because I'm stupid! Thanks :D

PR #4649, opened 2025-01-02
add default -I and -L paths for OpenBSD

dspx-plcr:
These can be added manually via environment variables when building, but it's probably sensible to have them as default on OpenBSD.

### DIFF
--- a/build_odin.sh
+++ b/build_odin.sh
@@ -110,8 +110,8 @@ Linux)
 	LDFLAGS="$LDFLAGS -Wl,-rpath=\$ORIGIN"
 	;;
 OpenBSD)
-	CXXFLAGS="$CXXFLAGS $($LLVM_CONFIG --cxxflags --ldflags)"
-	LDFLAGS="$LDFLAGS -liconv"
+	CXXFLAGS="$CXXFLAGS -I/usr/local/include $($LLVM_CONFIG --cxxflags --ldflags)"
+	LDFLAGS="$LDFLAGS -L/usr/local/lib -liconv"
 	LDFLAGS="$LDFLAGS $($LLVM_CONFIG --libs core native --system-libs)"
 	;;
 Haiku)


laytan:
Ditto, never heard anyone say it doesn't build with the current settings, and it shouldn't need it.

Are you perhaps confusing this build script with the actual Odin linking code?

dspx-plcr:
Hiya,

Sorry, I should have explained why I was suggesting this. The symptom I'm trying to solve is that odin does not build on OpenBSD 7.6.

```sh
[despereaux@polacre ~/build/Odin]$ git status
HEAD detached at origin/master
nothing to commit, working tree clean
[despereaux@polacre ~/build/Odin]$ git remote get-url origin
https://github.com/odin-lang/Odin.git
[despereaux@polacre ~/build/Odin]$ LLVM_CONFIG=/usr/local/llvm17/bin/llvm-config make release-native
./build_odin.sh release-native
+ clang++ src/main.cpp src/libtommath.cpp -Wno-switch -Wno-macro-redefined -Wno-unused-value -DODIN_VERSION_RAW="dev-2025-01" -DGIT_SHA="fdd3b46" -std=c++14 -I/usr/local/llvm17/include -std=c++17 -fno-exceptions -funwind-tables -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -L/usr/local/llvm17/lib -O3 -march=native -pthread -lm -lstdc++ -liconv -lLLVM -o odin
In file included from src/main.cpp:2:
In file included from src/common.cpp:68:
src/string.cpp:504:11: fatal error: 'iconv.h' file not found
        #include <iconv.h>
                 ^~~~~~~~~
1 error generated.
*** Error 1 in /home/despereaux/build/Odin (Makefile:19 'release-native')
[despereaux@polacre ~/build/Odin]$ uname -a
OpenBSD polacre 7.6 GENERIC.MP#338 amd64
```
(I have to use the OpenBSD llvm-17 package as 7.6 ships with llvm 16, which isn't supported by odin's build script)

The problem is that `iconv` is a third party package and hence is installed in the `/usr/local` prefix, whereas llvm is maintained as part of the base system, so gets special treatment. As a side note, even if llvm16 did work, this special treatment means that,
```sh
[despereaux@polacre ~/build/Odin]$ llvm-config --cxxflags --ldflags
-I/usr/include -std=c++17   -fno-exceptions -fno-rtti -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS
-L/usr/lib
```
so compiling odin would still fail.

Fundamentally, `llvm-config` tells you about how to compile and link against LLVM, not any other library, and `iconv` is some other library.

If you just add the `-I` path, then linking fails, so the `-L` path is also necessary.

dspx-plcr:
I agree that the `pkg-config` (or `pkgconf`) option is better semantically, but practically, the gnu `libiconv` package doesn't ship with a pkg config file, so it doesn't help. In terms of maintainability, I can't speak first hand for the history, but I took a look at [the man pages](https://man.openbsd.org/OpenBSD-2.8/bsd.port.mk.5) from OpenBSD 2.8 in 2000, and `/usr/local` is the prefix for third party ports :slightly_smiling_face: 

I understand the aversion though, and ultimately it's the the Odin project's call as to if it's acceptable or not. Either way I'll be keeping the diff on my local branch because I find it useful.

Indeed the problem does persist with llvm-17, as the prefix paths for `libiconv` and `llvm-17` are still different.

PR #4654, opened 2025-01-03
Minor adjustments in documentation-only files for better LSP support 

obiwan87:
Make some adjustments to "documentation only" files. This helps LSPs provide better code completion.

### DIFF
--- a/base/intrinsics/intrinsics.odin
+++ b/base/intrinsics/intrinsics.odin
@@ -2,6 +2,7 @@
 #+build ignore
 package intrinsics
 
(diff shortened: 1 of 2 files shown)


base/intrinsics/intrinsics.odin
```diff
@@ -2,6 +2,7 @@
 #+build ignore
 package intrinsics
 
+import "base:runtime"
 // Package-Related
```
laytan, line 6:
```suggestion
import "base:runtime"

// Package-Related
```

PR #4665, opened 2025-01-08
Add SetWindowSubclass and RegisterHotKey functions

GrumpyLion:
Added some missing functions from User32.lib and Comctl32.lib that I needed to use for peeking window messages and registering hotkeys.

### DIFF
--- a/core/sys/windows/user32.odin
+++ b/core/sys/windows/user32.odin
@@ -2,6 +2,7 @@
 package sys_windows
 
 import "base:intrinsics"
+import "core:c"
 foreign import user32 "system:User32.lib"
 
 @(default_calling_convention="system")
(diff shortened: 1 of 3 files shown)


core/sys/windows/user32.odin
```diff
@@ -32,6 +32,8 @@ foreign user32 {
 	RegisterClassExW :: proc(^WNDCLASSEXW) -> ATOM ---
 	UnregisterClassW :: proc(lpClassName: LPCWSTR, hInstance: HINSTANCE) -> BOOL ---
 
+	RegisterHotKey :: proc(hnwd: HWND, id: int, fsModifiers: UINT, vk: UINT) -> BOOL ---
```
laytan, line 35: This should be `c.int` and add `import "core:c"` in the file. An Odin `int` is not the same as a c `int`.

PR #4674, opened 2025-01-10
make max array count I64_MAX

### LINKED ISSUE #4658: 'Array count too large' for large inline arrays
## Context

      Odin:    dev-2024-12-nightly:cf53404
      OS:      Windows 10 Enterprise N (version: 20H2), build 19042.1466   
      CPU:     Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
      RAM:     65470 MiB
      Backend: LLVM 18.1.8

## Expected Behavior

This 
```odin
package main

A :: struct {
	a : [8]u8, // good 
	b : [1 * 1024 * 1024 * 1024]u8, // not good 
}

main :: proc() {  }
```
should compile without issue. While instantiating such a struct would most likely blow the stack, just declaring it should not be a problem.

## Current Behavior

The build errors with `Array count too large, 1073741824`

### Steps to Reproduce

1. copy the provided snippet into a new file
2. run odin build on the file 
3. error

there is some ongoing discussion on the discord about the exact cause of this;  
https://discord.com/channels/568138951836172421/568871298428698645/1325284384969986099  
but im going to open a bug report for now

laytan:
Closes #4658

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -4841,9 +4841,19 @@ gb_internal bool check_index_value(CheckerContext *c, Type *main_type, bool open
 
 			} else { // NOTE(bill): Do array bound checking
 				i64 v = -1;
-				if (i.used <= 1) {
-					v = big_int_to_i64(&i);
+
+				switch (i.used) {
+				case 0: v = 0; break;
+				case 1: v = big_int_to_i64(&i); break;
+				default:
+					static BigInt max = big_int_make_i64(I64_MAX);
+					if (big_int_cmp(&max, &i) > 0) {
+						v = big_int_to_i64(&i);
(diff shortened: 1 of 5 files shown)


laytan:
So only MacOS arm seems to be fine with this massive array. Linux and BSDs crash fail linking, Windows crashes, MacOS Intel gets stuck compiling or linking the program.

laytan:
Should figure out a max they all generally like and go with that.

gingerBill:
The original restriction was probably arbitrary but it was to prevent a few bugs in LLVM due to initialization issues. Large arrays in LLVM tend to make the entire thing go extremely slow. It's hard to replicate because it's been quite a long time now and I cannot remember the precise circumstances to do it.

PR #4677, opened 2025-01-11
Add no-core/no-vendor build settings

bangbangsheshotmedown:
Odin reserves common module names like `io` and `fmt`

This is annoying for people who don't like to rely on other people standard library and instead want a custom/tiny one

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -488,6 +488,9 @@ struct BuildContext {
 
 	bool   no_rtti;
 
+	bool   no_core;
+	bool   no_vendor;
+
 	bool   dynamic_map_calls;
 
 	bool   obfuscate_source_code_locations;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -370,6 +370,8 @@ enum BuildFlagKind {
 	BuildFlag_StrictStyle,
 	BuildFlag_ForeignErrorProcedures,
 	BuildFlag_NoRTTI,
+	BuildFlag_NoCore,
+	BuildFlag_NoVendor,
 	BuildFlag_DynamicMapCalls,
 	BuildFlag_ObfuscateSourceCodeLocations,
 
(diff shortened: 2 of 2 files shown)


gingerBill:
This PR is useless. You can literally redefine those imports yourself if you want to do it with `-collection:core="path/to/my/core"`

As for "common module names", that's on purpose. 

1) Note the name of the directory has nothing to do with the name of the `package` declaration. The latter is for ABI consistency: https://odin-lang.org/docs/faq/#why-are-package-declarations-required-in-every-file
2) They are packages that are commonly used by loads of people, and for good reason.

bangbangsheshotmedown:
> This PR is useless.

It' not useless

I don't have a "core" collection, i don't want the compiler to give me a default one

I want control

Ultimately i want

-no-rtti
-no-entry-point
-no-crt

See, this is why Odin is not compatible with embedded, you need to remove friction for that kind of things and separate the language from the rest that's builtin

bangbangsheshotmedown:
> If you don't want anything from core or vendor, don't import from it.

i don't import them, my module is called `io` 

> What you ultimately want exists, you just stated it.

where?

> You wanting to arbitrarily just delete those directories is just that—arbitrary.

I don't want to delete anything, i want to tell the compiler to not include them by default for this specific project

I think there is a misunderstanding, i'll try it differently

How can i have my module named `io` without getting this error message:

``` 
Error: Duplicate declaration of 'package io' 
	package io 
```

structure:
```
project/
    io/
    entry.odin
    main_flash.odin
```

karl-zylinski:
@bangbangsheshotmedown I think there is a misunderstanding on your side about how packages work. Your folder can have the name `io`. The `io` on the `package io` line can be anything. So you pick anything else there and it will work. That name is only used for ABI linking and doesn't affect your code.

bangbangsheshotmedown:
@karl-zylinski 

that's the thing, i don't want to rename it, hence the PR

Kelimion:
And you don't have to, as long as you don't import `core:io`. No complaints whatsoever from the compiler about a conflicting `io` here.

test.odin
```odin
package test

import "io"

main :: proc() {
	io.hellope()
}
```
io/io.odin
```odin
package io

import "base:runtime"

hellope :: proc() {
	runtime.print_string("Hellope!\n")
}
```

gingerBill:
@bangbangsheshotmedown I'm just going to be blunt.

You have not tried anything you are complaining about. As I said in my initial reply:

* The name of the directory can be anything
* The package name at the top of each file in the directory can be anything
* The package name does not have to match the name of the directory either
* The package name (NOT the directory name) needs to be unique project-wide, so if you never import e.g. `core:io`, you can call your package `io`.


Please stop wasting our time when you are clearing not even bothering to try out even the absolute basics.

karl-zylinski:
@Kelimion Ah, I didn't know that. It must have been `fmt` that pulled in `io`. So if you are truly "stand-alone" and not having core in there, then you can use the name `io`. 

@bangbangsheshotmedown  so I guess the confusion came from thinking that `io` wasn't imported, but it it is: By importing `core:fmt` you pulled in `core:io` and caused a package name conflict. As long as you skip everything from `core` you're good to go.

gingerBill:
Or just not name your package `io`. It's just for consistent ABI for the link names, which is unironically brilliant for embedded, the opposite of what you arrogantly and ignorantly asserted initially. You can STILL call the directory `io`.

gingerBill:
> ...there is close to no documentation, so it's hard to figure things out

This is documented, and even if it wasn't, you could have asked people before trying to make a PR. 

> Watch your mouth when you deal with strangers...

This is the arrogant ignorance I am talking about. I have to be blunt when dealing with strangers because I do not have the resources at hand to deal with such things. I am sorry if you are talking this the wrong way, but you are acting like I am describing.

PR #4681, opened 2025-01-12
Add net.dial_tcp_from_host{_or_endpoint} and unify them

haesbaert:
The main motivation for this is to have sinergy with flags parsing, currently flags for a sockaddr returns a net.Host_Or_Endpoint, but we can't just dial from it since there isn't a variant.

Consider the following:

```
Options :: struct {
	target: net.Host_Or_Endpoint `args:"pos=0,required" usage:"host:port"`,
}

before :: proc() -> (sock: net.TCP_Socket, err: net.Network_Error) {
	opt: Options

	flags.parse_or_exit(&opt, os.args)
	switch t in opt.target {
	case net.Host:
		sock, err = net.dial_tcp(t.hostname, t.port)
	case net.Endpoint:
		sock, err = net.dial_tcp(t)
	}
	return
}

after :: proc() -> (sock: net.TCP_Socket, err: net.Network_Error) {
	opt: Options

	flags.parse_or_exit(&opt, os.args)
	sock, err = net.dial_tcp(opt.target)
	return
}

```

For completion, add dial_tcp_from_host() and define the upper functions in terms of the newly added ones, cuts one repeated block, now:

from_hostname_and_port_string is parse + from_host_or_endpoint from_hostname_with_port_override is parse + override + from_host_or_endpoint from_host is to_endpoint + from_endpoint
from_host_or_endpoint is from_endpoint or from_host

### DIFF
--- a/core/net/socket.odin
+++ b/core/net/socket.odin
@@ -34,23 +34,12 @@ any_socket_to_socket :: proc "contextless" (socket: Any_Socket) -> Socket {
     Expects both hostname and port to be present in the `hostname_and_port` parameter, either as:
     `a.host.name:9999`, or as `1.2.3.4:9999`, or IP6 equivalent.
 
-    Calls `parse_hostname_or_endpoint` and `resolve`, then `dial_tcp_from_endpoint`.
+    Calls `parse_hostname_or_endpoint` and `dial_tcp_from_host_or_endpoint`.
 */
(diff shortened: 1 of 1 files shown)


haesbaert:
Any feedback on this? I have more PRs for net in the pipeline but don't want to flood.

Kelimion:
Sorry for the delay.

haesbaert:
> Sorry for the delay.

Not at all, many thanks :heart: 

Keeping the pipeline full: https://github.com/odin-lang/Odin/pull/4813

PR #4687, opened 2025-01-13
Micro UI text wrapping bug

Binasaurus-Hex:
image

currently microui makes an assumption in the text procedure that the width of a word + space individually is the same as both together. I have found that for certain fonts this is not the case, and causes text to be cut off before the line wraps. This change removes ambiguity by just measuring the whole line each time we increment a character.

### DIFF
--- a/vendor/microui/microui.odin
+++ b/vendor/microui/microui.odin
@@ -921,25 +921,26 @@ text :: proc(ctx: ^Context, text: string) {
 	layout_row(ctx, {-1}, ctx.text_height(font))
 
 	outer: for len(text) > 0 {
-		last_space: int
 		r := layout_next(ctx)
+		last_space: int
 
 		for ch, i in text {
 			if ch == ' ' {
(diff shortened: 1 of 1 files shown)


laytan:
Is this issue also present on the original at https://github.com/rxi/microui ?

Binasaurus-Hex:
> Is this issue also present on the original at https://github.com/rxi/microui ?

yep, just made a small test project in C and has the same issue

Binasaurus-Hex:
I've had a play around with it some more and it seems like its just to do with the fact that microui measures the space width in isolation, when its surrounded by other characters you get an extra 2x the glyph padding. and this discrepancy is what stacks up and causing the lines to not wrap correctly. 

So you can also fix it by modifying your own text width function to treat spaces differently and add the extra padding, instead of making this change, but it seems a bit of a hack. Especially for a new person just trying to draw some text on the screen it could be confusing.

PR #4693, opened 2025-01-15
Output debug log messages, when debug level is "debug"

obiwan87:
Allow debug messages to be displayed when debug level is set to debug. This allows users to create a release build of the test runner while also having the possibility to get debug messages. 

This is particularly useful when integrating the test runner in an IDE. In my test integration I use the command line output as a signal for determining when a test has started, failed, succeeded, etc.

image

This wouldn't be possible without the command line output. However, now I'm forced to build a debug build in order for this to be possible. With this fix, I can also just set the log level to debug

### DIFF
--- a/core/testing/runner.odin
+++ b/core/testing/runner.odin
@@ -611,7 +611,7 @@ runner :: proc(internal_tests: []Internal_Test) -> bool {
 						total_done_count += 1
 					}
 
-					when ODIN_DEBUG {
+					when ODIN_DEBUG || LOG_LEVEL == "debug" {
 						log.debugf("Test #%i %s.%s changed state to %v.", task_channel.test_index, it.pkg, it.name, event.new_state)
 					}
 


laytan:
What do you think @Feoramund ?

Feoramund:
That message was a holdover from when I was first implementing the test runner and I had forgotten to take it out. It does serve a niche purpose for me when I'm debugging the runner itself with respect to individual tests possibly stalling it, but it's since served its purpose.

>Allow debug messages to be displayed when debug level is set to debug. This allows users to create a release build of the test runner while also having the possibility to get debug messages.

This patch doesn't do anything about outputting debug log messages in general, because we already have `ODIN_TEST_LOG_LEVEL` as a config which can be set when in any non`-debug` mode.

`odin test . -o:speed -define:ODIN_TEST_LOG_LEVEL="debug"` works just fine for the original intent written in this post.

EDIT:
> This is particularly useful when integrating the test runner in an IDE. In my test integration I use the command line output as a signal for determining when a test has started, failed, succeeded, etc.

With regards to this concern, the patch accomplishes its goal.

Feoramund:
After some thought, I think if people are going to rely on this message, it should be given its own `#config` and sent to the info channel.

laytan:
> After some thought, I think if people are going to rely on this message, it should be given its own `#config` and sent to the info channel.

Agreed, something like `ODIN_TEST_LOG_STATE_TRANSITIONS`.

Kelimion:
Superceded by #5413.

PR #4694, opened 2025-01-15
Add objc_send declaration

obiwan87:
For transparency and LSP support, add documentation only declaration objc_send to intrinsics.odin


gingerBill:
So that's not the signature of `objc_send`, and in fact quite a bit weirder in practice too. I'd argue the LSP might have to add an edge case for this directly.

obiwan87:
Yeah, I guess there's no way express the `objc_send` signature in Odin. I just added a very generic/abstract declaration so that there's at least a reference point for gathering more information on how to use it.

PR #4697, opened 2025-01-15
Allow captures in gfind and gmatch to be used in-loop

JamesDSource:
As it is currently. Trying to slice the iterated string in gfind and gmatch will not work. The string is sliced at the end of the function, after the captures are found.

```
import "core:fmt"
import "core:text/match"

main :: proc() {
	src := `12356789123456789
		a cat with a hat
		a dog with a bone
	123456789123456789`
	caps: [match.MAX_CAPTURES]match.Match
	for _ in match.gmatch(&src, "a (%w+) with a (%w+)", &caps) {
		fmt.println(
			src[caps[1].byte_start:caps[1].byte_end],
			src[caps[2].byte_start:caps[2].byte_end],
		)
	}
	// Output:
	// th 123
	// 345 5678
}
```

I just moved that slice to the beginning of the next iteration. This does add an implicit requirement to have the captures array zeroed. I also made it automatically zero the captures array on the last iteration in case of reuse.

### DIFF
--- a/core/text/match/strlib.odin
+++ b/core/text/match/strlib.odin
@@ -698,11 +698,11 @@ gmatch :: proc(
 			first := length > 1 ? 1 : 0
 			cap := captures[first]
 			res = haystack[cap.byte_start:cap.byte_end]
-		} else {
-			captures^ = {}
 		}
 	} 
-
+	if !ok {
+		captures^ = {}
+	}
 	return
 }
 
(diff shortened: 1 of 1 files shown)


Skytrias:
Looks good to me, the captures array reset on `!ok` is smart 👍

Skytrias (approved):

PR #4703, opened 2025-01-16
Do not warn about stack overflow in range loops 'by reference'

### LINKED ISSUE #4685: Addressable loop variable causes stack overflow when it shouldn't
## Context

        Odin:    dev-2025-01:fdd3b46c1
        OS:      Windows 10 Professional (version: 22H2), build 19045.5247
        CPU:     Intel(R) Core(TM) i7-6950X CPU @ 3.00GHz
        RAM:     32674 MiB
        Backend: LLVM 18.1.8

## Expected Behavior

`for &s in scenes  {}` should never cause a stack overflow warning due to type of `s` being too big.

## Current Behavior

If the struct type used by the loop is big, then you get a stack overflow warning, even though `&` is there.

### Steps to Reproduce

Run this (note the `&`  in `for &s in big_scenes {`, this shouldn't copy)
```
package buggy_bug

Scene :: struct {
	big_things: [1500000]int,
}

big_scenes: [10]Scene

main :: proc() {
	for &s in big_scenes {

	}
}
```

Get this:
```
C:/code/playground/playground.odin(10:7) Warning: Declaration of 's' may cause a stack overflow due to its type 'Scene' having a size of 12000000 bytes 
	for &s in big_scenes { 
	     ^ 
[Finished in 218ms]
```

flysand7:
This PR fixes the stack overflow warning that pops up in ranged loops, when looping over large arrays by reference. Since we're looping by reference, there are no stack copies of the entire loop, which makes the warning useless.

This PR resolves #4685.

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -749,9 +749,15 @@ gb_internal void check_scope_usage_internal(Checker *c, Scope *scope, u64 vet_fl
 			// TODO(bill): When is a good size warn?
 			// Is >256 KiB good enough?
 			if (sz > 1ll<<18) {
-				gbString type_str = type_to_string(e->type);
-				warning(e->token, "Declaration of '%.*s' may cause a stack overflow due to its type '%s' having a size of %lld bytes", LIT(e->token.string), type_str, cast(long long)sz);
-				gb_string_free(type_str);
+				bool is_ref = false;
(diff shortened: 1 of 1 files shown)


laytan:
You left a binary in there

flysand7:
Binary removed

PR #4716, opened 2025-01-18
os/os2: wasi target support

laytan:
Implements everything that can be implemented with the `core:sys/wasm/wasi` bindings.

Also adds partial support for path/filepath and implements crypto.rand_bytes.


gingerBill:
Looks good! I'll have to try and test it next.

PR #4719, opened 2025-01-19
remove `-L/` and `-l:` from linker flags

spitulax:
This is the actual fix to <https://github.com/NixOS/nixpkgs/issues/338972> and many other issues with linking on Nix build system. The issue may have been caused by using `-l:` to specify the path to libraries and `-L/` to set the linker search path. The linker could not find the libraries only when building using Nix `stdenv`. Building inside a nix shell does not trigger this issue. Probably it has something to do with build sandboxing, but I'm not too knowledgeable on this matter.

I found the solution when I was trying to isolate the issue using a simple C program and a static library. Building this with nix `stdenv`, resulting in the linker not being able to found the library.
```nix
buildPhase = ''
    gcc -o main main.c -L/ -l:$(pwd)/lib/libh.a
'';
```
```
/nix/store/j7p46r8v9gcpbxx89pbqlh61zhd33gzv-binutils-2.43.1/bin/ld: cannot find -l:/build/source/lib/libh.a: No such file or directory
```
But removing `-L/` and `-l:` does the trick.
```nix
buildPhase = ''
    gcc -o main main.c $(pwd)/lib/libh.a
'';
```

I also tested this patch on Debian and it seemed to work fine.

### DIFF
--- a/src/gb/gb.h
+++ b/src/gb/gb.h
@@ -278,6 +278,11 @@ extern "C" {
 	#include <semaphore.h>
 #endif
 
+#if defined(GB_SYSTEM_LINUX)
+	#include <grp.h>
+	#include <pwd.h>
+#endif
+
 
 ////////////////////////////////////////////////////////////////
 //
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -615,7 +615,7 @@ try_cross_linking:;
 							//                local to the executable (unless the system collection is used, in which case we search
 							//                the system library paths for the library file).
 							if (string_ends_with(lib, str_lit(".a")) || string_ends_with(lib, str_lit(".o")) || string_ends_with(lib, str_lit(".so")) || string_contains_string(lib, str_lit(".so."))) {
-								lib_str = gb_string_append_fmt(lib_str, " -l:\"%.*s\" ", LIT(lib));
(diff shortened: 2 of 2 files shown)


flysand7:
@gingerBill There are some differences between specifying `-l:` and `-L`, but yes, these should be disabled *just* for Nix, if we want to disable them.

I also believe that we need to update the way these flags are handled in the presence of

* `foreign import "lib"`; and
* `foreign import "system:lib"`

Since `system:` explicitly asks for it to be a system dependency. Currently `system:` directive is only used to special-case some common libraries -- `libc`, `m` and `thread`, which I think needs to be fixed in the future at some point.

gfaster:
> Wouldn't it better to disable them just for Nix?
Perhaps it would be better to have it disabled via configuration flag so other unconventionally structured environment (e.g. [Guix](https://guix.gnu.org/)) can benefit as well

asumbek:
Any update on this? I seems to also hit a roadblock while writing a derivation for a Raylib project of mine.

asumbek:
I think I found a workaround for this.

By not using some certain linkers such as the default `bfd` linker and lld, it seems to link pretty fine.

What I mean is that passing `-extra-linker-flags="-fuse-lld=gold"` (or whatever linker that's not listed above), the linker will recognize `-l:` under Nix build environment.

PR #4732, opened 2025-01-20
os/os2: Properly update CWD on Linux when using _process_start()

zen3ger:
The `dir_fd` argument to `execveat()` is not for setting the current working directory. It is used to resolve relative executable paths, hence explicit `chdir/fchdir` call is required to set CWD.

### DIFF
--- a/core/os/os2/process_linux.odin
+++ b/core/os/os2/process_linux.odin
@@ -547,6 +547,10 @@ _process_start :: proc(desc: Process_Desc) -> (process: Process, err: Error) {
 		if _, errno = linux.dup2(stderr_fd, STDERR); errno != .NONE {
 			write_errno_to_parent_and_abort(child_pipe_fds[WRITE], errno)
 		}
+		if dir_fd != linux.AT_FDCWD {
+			errno = linux.fchdir(dir_fd)
+			assert(errno == nil)
+		}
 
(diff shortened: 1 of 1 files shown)


core/os/os2/process_linux.odin
```diff
@@ -547,6 +547,10 @@ _process_start :: proc(desc: Process_Desc) -> (process: Process, err: Error) {
 		if _, errno = linux.dup2(stderr_fd, STDERR); errno != .NONE {
 			write_errno_to_parent_and_abort(child_pipe_fds[WRITE], errno)
 		}
+		if dir_fd != linux.AT_FDCWD {
+			errno = linux.fchdir(dir_fd)
+			assert(errno == nil)
```
laytan, line 552: Please handle the error by writing to parent as done with the dup2 calls above

zen3ger: afaik, it can't happen. if `open` succeeds then you have a directory fd, and correct access rights so `fchdir` can't fail either.

laytan: It could always happen because of some weird case, where in general I would be fine with an assert, but in this case you are in the forked process, and an assert will write to stderr, which can be set by the user or be inherited, or just be dev/null, and then it traps, making the user that calls wait on the process later think their process trapped, but their process was never actually started.

PR #4733, opened 2025-01-20
os/os2: add get_executable_path and get_executable_directory

laytan:


core/os/os2/path_darwin.odin
```diff
@@ -0,0 +1,31 @@
...
+
+import "base:runtime"
+
+import "core:sys/darwin"
+import "core:sys/posix"
+
+_get_executable_path :: proc(allocator: runtime.Allocator) -> (path: string, err: Error) {
+	size: u32
+
+	ret := darwin._NSGetExecutablePath(nil, &size)
```
gingerBill, line 11: This means it is internal and allowed to change behaviour. We need to use something else.

gingerBill: I might be wrong, I am not sure. Maybe `NSBundle` is a better choice?

```
NSBundle.mainBundle()->bundlePath()
```
maybe?

laytan: It is a weird situation, because this is a documented (https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dyld.3.html) function intended to be used, and also the recommended way by everybody I see talking about/doing this.

The NS.Bundle would mean we depend on objective C so using core:os would link all of those objective C libraries, not sure we want that?

Another option is just making this do `current_process_info({.Executable_Path})` and return that. Which does this:
https://github.com/odin-lang/Odin/blob/867af80bff2956178ac72d9afbc9327b67cd4ae8/core/os/os2/process_posix_darwin.odin#L105:110 , this should also be fine and doesn't use objective c or underscored symbols.

laytan: I've changed it to use that method instead, seems like the best option

gingerBill (approved):

PR #4743, opened 2025-01-23
Fix 2x2 matrix inverses in specific.odin

prescientmoon:
I think some of the indices were flipped (they were right in `general.odin`). I'm not sure why the functions are still there while they also exist in `general.odin`, but they did pop up when searching through the website, which had me waste hours debugging the issue.

### DIFF
--- a/core/math/linalg/specific.odin
+++ b/core/math/linalg/specific.odin
@@ -1207,8 +1207,8 @@ matrix2_inverse_f16 :: proc "contextless" (m: Matrix2f16) -> (c: Matrix2f16) #no
 	d := m[0, 0]*m[1, 1] - m[0, 1]*m[1, 0]
 	id := 1.0/d
 	c[0, 0] = +m[1, 1] * id
-	c[0, 1] = -m[1, 0] * id
-	c[1, 0] = -m[0, 1] * id
+	c[0, 1] = -m[0, 1] * id
+	c[1, 0] = -m[1, 0] * id
 	c[1, 1] = +m[0, 0] * id
 	return c
 }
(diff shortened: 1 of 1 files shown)


gingerBill:
No. That's correct because that's how an inverse is defined. You have to flip those around.

prescientmoon:
> No. That's correct because that's how an inverse is defined. You have to flip those around.

A very quick example shows that this is wrong. Consider the following code:
```odin
package repro

import "core:fmt"
import "core:math/linalg"

main :: proc() {
	a: matrix[2, 2]f32 = {0, 1, 2, 0}
	b := linalg.matrix2_inverse_f32(a)
	fmt.println(a * b)
	fmt.println(b * a)
}
```

This is the output:
```
matrix[0.5, 0; 0, 2]
matrix[2, 0; 0, 0.5]
```

Now consider this code:
```odin
package repro

import "core:fmt"

matrix2_inverse_f32 :: proc "contextless" (
	m: matrix[2, 2]f32,
) -> (
	c: matrix[2, 2]f32,
) #no_bounds_check {
	d := m[0, 0] * m[1, 1] - m[0, 1] * m[1, 0]
	id := 1.0 / d
	c[0, 0] = +m[1, 1] * id
	c[0, 1] = -m[1, 0] * id
	c[1, 0] = -m[0, 1] * id
	c[0, 1] = -m[0, 1] * id
	c[1, 0] = -m[1, 0] * id
	c[1, 1] = +m[0, 0] * id
	return c
}

main :: proc() {
	a: matrix[2, 2]f32 = {0, 1, 2, 0}
	b := matrix2_inverse_f32(a)
	fmt.println(a * b)
	fmt.println(b * a)
}
```

This is the output:
```
matrix[1, 0; 0, 1]
matrix[1, 0; 0, 1]
```

prescientmoon:
Thanks for merging this!

skorianez:
Maybe add some unit tests?

PR #4751, opened 2025-01-24
Grow-in-place for some arenas

Barinzaya:
Added support for growing the last allocation in place to `runtime.Arena` and `virtual.Arena`.

These changes allow the last allocation that has been made to be resized to a larger size by just extending their allocation in-place, when there's sufficient room in the memory block to do so.

Shrinking in place and re-using the rest of the allocation can be supported using almost the same logic, but would require the memory to be zeroed. Since this would add a additional cost that isn't currently present, shrinking has not been changed.

### DIFF
--- a/base/runtime/default_temp_allocator_arena.odin
+++ b/base/runtime/default_temp_allocator_arena.odin
@@ -210,10 +210,24 @@ arena_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
 		case size == 0:
 			err = .Mode_Not_Implemented
 			return
-		case (uintptr(old_data) & uintptr(alignment-1) == 0) && size < old_size:
-			// shrink data in-place
-			data = old_data[:size]
-			return
+		case uintptr(old_data) & uintptr(alignment-1) == 0:
+			if size < old_size {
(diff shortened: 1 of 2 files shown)


gingerBill (approved):

base/runtime/default_temp_allocator_arena.odin
```diff
@@ -210,10 +210,24 @@ arena_allocator_proc :: proc(allocator_data: rawptr, mode: Allocator_Mode,
...
-			data = old_data[:size]
-			return
+		case uintptr(old_data) & uintptr(alignment-1) == 0:
+			if size < old_size {
+				// shrink data in-place
+				data = old_data[:size]
+				return
+			}
+
+			if block := arena.curr_block; block != nil {
```
JesseRMeyer, line 220: Is `block` here and down below intended to be a pointer to `curr_block`?  `block.used` is mutated, but that change would only effect the copy here, not the original.

Barinzaya: `curr_block` is a `^Memory_Block`. It's a pointer, so it affects the same place

PR #4766, opened 2025-01-26
correct Raylib CameraMoveRight signature

Goldenlion5648:
Without this change, when attempting to export to web, a warning is given (and if the function is actually called in the web, it crashes).

This PR corrects the signature to the one used in the raylib C file https://github.com/raysan5/raylib/blob/f6f31a9f216c54fe8bf661e042a4cfc9a6892e33/src/rcamera.h#L148


nadako:
#4748 fixes the same thing, so merge either that or this, but close both :)

gingerBill:
Already fixed in #4748

Issue #2348, opened 2023-02-26
strings.split_*_iterators() do not return "" after last trailing separator


MarenFayre:
## Version
```
Odin: dev-2023-02:de6c0f68
OS:   Windows 10 Professional (version: 22H2), build 19045.2604
```

## Current Behavior
When splitting `"a\nb\nc\n"`, the `split_lines_iterator()` returns `["a", "b", "c"]` ignoring the trailing '\n'.  
This behaviour is inconsistent with `split_lines()` which returns `["a", "b", "c", ""]` as expected.  

The same behaviour is displayed by `split_iterator()`/`split_by_byte_iterator()`/... with arbitrary trailing separators.

```odin
package main
import "core:fmt"
import "core:mem"
import "core:strings"
main :: proc() {
    s := "a\nb\nc\n"

    lines := strings.split_lines(s)
    fmt.println(lines) 
    // ["a", "b", "c", ""] -- Expected

    lines_iter := s 
    for l in strings.split_lines_iterator(&lines_iter) {
        fmt.println(">>", l)
    }
    // >> a
    // >> b
    // >> c  -- Missing the last empty "" after '\n'
}
```

## Additional information

The string `"a\nb\nc\n\n\n"` similarly becomes `["a", "b", "c", "", ""]` instead of  `["a", "b", "c", "", "", ""]`, showing that only the last trailing separator is skipped.

gingerBill:
I don't think this is solvable without making the iterator be something more complex. The reason it isn't return what you expected is that the only information the `split_lines_iterator` uses is a string and it just subslices it as it goes along. When it comes to the last `\n`, there is nothing after it so it assumes that is the last of the iteration and stops; it cannot know it should continue.

MarenFayre:
Oh I see. Then I would suggest that the behaviour of `split_lines` should be changed instead.
While my intuition was that `"a\nb\nc\n"` should become `["a", "b", "c", " "]`, a quick look at Python `splitlines()` and Rust `lines()` showed that they both ignore trailing `\n` to do this: `["a", "b", "c"]`
So for example:
```odin
lines := strings.split_lines("\n")   // returns ["", ""] right now, but should return [""]
```

Sanian-Creations:
Looking into this made me realize that there is one other huge difference with the split and split iterator procs (separate from split line) and that is that `strings.split` with `""` as separator will return an array containing every individual character, but `strings.split_iterator` will return the whole string in one go without splitting it at all.

[...]

PR #4789, opened 2025-02-04
[core/strings] Fix _split_iterator when separator is empty

jfcode4:
Currently `split_iterator()` returns the entire string at once if the separator is empty, but it should rather split the string by runes. This was also mentioned in #2348.  The function also keeps returning ok=true which causes an infinite loop. Here is an example where this happens:
```odin
package test

import "core:fmt"
import "core:strings"

main :: proc() {
	mystr := "Hello world"
	for s in strings.split_iterator(&mystr, "") {
		fmt.printfln("(%s)", s)
	}
}
```
This PR fixes these issues

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -1031,14 +1031,10 @@ Returns:
 */
 @private
 _split_iterator :: proc(s: ^string, sep: string, sep_save: int) -> (res: string, ok: bool) {
-	if sep == "" {
-		res = s[:]
-		ok = true
-		s^ = s[len(s):]
-		return
-	}
-
 	m := index(s^, sep)
+	if sep == "" {
+		m = 1 if len(s) > 0 else -1
+	}
 	if m < 0 {
 		// not found
 		res = s[:]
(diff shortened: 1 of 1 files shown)


gingerBill:
So here's the thing, why should it do that?

jfcode4:
Because this is what the `split()` function does. If the separator is empty the string gets split into individual runes. The same behavior should happen in `split_iterator()`.
Also an iterator shouldn't keep returning true, that was a bug.

core/strings/strings.odin
```diff
@@ -1031,14 +1031,15 @@ Returns:
 */
 @private
 _split_iterator :: proc(s: ^string, sep: string, sep_save: int) -> (res: string, ok: bool) {
+	m := index(s^, sep)
```
laytan, line 1034: I suggest a `m: int` and then an else branch attached to the `if sep == ""` that does `m = index(s^, sep)` instead. That's a small optimisation because you aren't using the result of `index` if the separator is empty here.

jfcode4: Thank you for the suggestion :)

PR #4797, opened 2025-02-05
improve abs() on floats for more correct and faster results

### LINKED ISSUE #4787: Absolute value of negative zero returns negative zero should be positive zero
### minimal example
```odin
package test
import "core:fmt"
import "core:math"
main :: proc() {
	k := -0.0
	fmt.printfln("abs(-0.0) = %v", math.abs(k)) // abs(-0.0) = -0
}
```
### provides incorrect answer to absolute value of negative zero, should be +0
<https://en.wikipedia.org/wiki/Absolute_value>
<https://en.m.wikipedia.org/wiki/Signed_zero>
<https://www.calculatorsoup.com/calculators/algebra/absolute-value-calculator.php>

### compiler command
odin run . or odin build .

### system info
Odin 2025-01
Windows
LLVM 18.1.8

### suggested labels
incorrect results
bug (possibly?)

laytan:
Fixes #4787 

Seems like this is more correct (abs of -0 is +0), and also optimises better (into fabs and similar instructions, instead of a cmp&select)

~~draft because not handling endian types, want to add some tests, and need to apply it to gb_abs so constants also show the same behaviour.~~

### DIFF
--- a/src/llvm_backend_proc.cpp
+++ b/src/llvm_backend_proc.cpp
@@ -2175,9 +2175,9 @@ gb_internal lbValue lb_build_builtin_proc(lbProcedure *p, Ast *expr, TypeAndValu
 			}
 			GB_PANIC("Unknown complex type");
 		} else if (is_type_float(t)) {
-			Type *t_float;
-			Type *t_unsigned;
-			lbValue mask;
+			Type *t_float = nullptr;
+			Type *t_unsigned = nullptr;
+			lbValue mask = {0};
 			switch (type_size_of(t)) {
 			case 2:
 				t_float = t_f16;
--- a/src/llvm_backend_proc.cpp
+++ b/src/llvm_backend_proc.cpp
@@ -2174,7 +2174,35 @@ gb_internal lbValue lb_build_builtin_proc(lbProcedure *p, Ast *expr, TypeAndValu
 			case 128: return lb_emit_runtime_call(p, "abs_complex128", args);
 			}
 			GB_PANIC("Unknown complex type");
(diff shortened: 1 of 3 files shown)


src/llvm_backend_proc.cpp
```diff
@@ -2174,7 +2174,32 @@ gb_internal lbValue lb_build_builtin_proc(lbProcedure *p, Ast *expr, TypeAndValu
 			case 128: return lb_emit_runtime_call(p, "abs_complex128", args);
 			}
 			GB_PANIC("Unknown complex type");
+		} else if (is_type_float(t)) {
```
gingerBill, line 2177: Is there an LLVM intrinsic we could use rather than doing this manually?

laytan: There is (https://llvm.org/docs/LangRef.html#llvm-fabs-intrinsic), I thought you didn't use it initially for a good reason, I can use that instead.

gingerBill: I cannot remember why I didn't use it. Maybe it required linking with libm?

laytan: I've tried it and it does not require libm, it does generate the exact same code as what I do here. But for the endian types we would need to byteswap, call intrinsic, and byteswap back, which may be worse.

What do you want me to do, call the intrinsic, or keep it like this?

gingerBill: Keep it like this, I guess.

gingerBill (approved):

PR #4812, opened 2025-02-08
debuginfo: ensure named types are package prefixed

flga:
This allows debug scripts/debugg*ers* to do proper type lookups.
Lots of whitespace :(

### DIFF
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -161,10 +161,10 @@ struct TypeStruct {
 
 struct TypeUnion {
 	Slice<Type *> variants;
-	
+
 	Ast *         node;
 	Scope *       scope;
-	
+
 	i64           variant_block_size;
 	i64           custom_align;
 	Type *        polymorphic_params; // Type_Tuple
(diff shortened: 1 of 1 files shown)


gingerBill:
I think the reason this was not done originally is because you cannot refer to that type due to the `.`. It might have to be `pkg::name` instead to match C++.

flga:
@gingerBill types are [already](https://github.com/odin-lang/Odin/blob/584fdc0d4ab9dcc01e218e499fdaef9969223d65/src/llvm_backend_debug.cpp#L992) declared in `pkg.type` form, just not consistently. Hence the change, I bumped into this while inspecting enum array debug info.

I'm not sure under which contexts it wouldn't work, but at least both lldb and gdb don't complain (DWARF).

Just wanted to clarify that if we want to move away from it we need to _also_ change current logic even if we drop this change.

Lperlind:
You can't have duplicate package names in Odin so that sidenote won't be a problem

flga:
Ah, you are right, I assumed that if they had different roots it would be fine. Thanks!

laytan:
Note that this function is also used for some error messages and probably other things.
(just a note, I don't know if we want that or not)

gingerBill:
This PR is going to be redundant with the new name canonicalization rules.

PR #4813, opened 2025-02-08
Fix some compression bugs in dns.

haesbaert:
- A compression pointer is when the two higher bits are set, the code was considering only 0xC0 as a pointer, where in reality anything from 0xC0-0xFF is a pointer, probably went unnoticed since you need big packets to have long pointers.
 - Make sure we can access the lower byte of the pointer by checking len, the code was careful to not access past the first byte, but ignored the second.
 - As per RFC9267 make sure a pointer only points backwards, this one is not so bad, as the code had a iteration_max that ended up guarding against infinite jumps.

Lightly tested, some eyes are welcome, but these are likely remote DOSable.

### DIFF
--- a/core/net/dns.odin
+++ b/core/net/dns.odin
@@ -533,18 +533,21 @@ decode_hostname :: proc(packet: []u8, start_idx: int, allocator := context.alloc
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


haesbaert:
I'm writing a recursive resolver in odin and eyeballed the compression code and noticed these.
Did some very light testing.

PR #4826, opened 2025-02-11
core:image add support for option .flip_vertical

DerTee:
This is basically the same as the option in stb_image to flip an image. One common use case is loading images for OpenGL which expects pixel data to be upside down, unlike most other graphics APIs and image formats.

I'm not super sure if this is useful to most people, plus it's trivial to write yourself if you do need it, so no problem if you decide not to accept this PR just for irrelevance :)

### DIFF
--- a/core/image/common.odin
+++ b/core/image/common.odin
@@ -5,6 +5,7 @@
 	List of contributors:
 		Jeroen van Rijn: Initial implementation, optimization.
 		Ginger Bill:     Cosmetic changes.
+		DerTee:          Vertical flip.
 */
 
 // package image implements a general 2D image library to be used with other image related packages
@@ -12,6 +13,7 @@ package image
 
 import "core:bytes"
 import "core:mem"
+import "core:slice"
 import "core:io"
 import "core:compress"
 import "base:runtime"
@@ -146,6 +148,7 @@ Option :: enum {
 	alpha_drop_if_present,         // Unimplemented for QOI. Returns error.
 	alpha_premultiply,             // Unimplemented for QOI. Returns error.
 	blend_background,              // Ignored for non-PNG formats
+	flip_vertical,
 
 	// Unimplemented
 	do_not_expand_grayscale,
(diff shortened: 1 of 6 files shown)


laytan:
My opinion on this, (but would love a review from @Kelimion) is that the helper function is a nice addition but having it as an option on top isn't needed.

core/image/common.odin
```diff
@@ -146,6 +148,7 @@ Option :: enum {
 	alpha_drop_if_present,         // Unimplemented for QOI. Returns error.
 	alpha_premultiply,             // Unimplemented for QOI. Returns error.
 	blend_background,              // Ignored for non-PNG formats
+	vertical_flip,                 // flip image vertically on load
```
Kelimion, line 151: For consistency with the other comments, I'd capitalize `// Flip`.

core/image/common.odin
```diff
@@ -1438,6 +1441,17 @@ expand_grayscale :: proc(img: ^Image, allocator := context.allocator) -> (ok: bo
 	return true
 }
 
+vertical_flip :: proc(img: ^Image) {
+    pixels := img.pixels.buf[:]
+    bpp := img.depth/8 * img.channels
```
Kelimion, line 1446: This proc should probably have a comment that it only supports 8 and 16 bpp, and an assert at the top that enforces this.

Kelimion:
For completeness, it would be great if this new option had some testing as well. Maybe a handful of images of 8 and 16 bpp that we load twice, once with `.flip_vertical` and once plain, which are then compared.

DerTee:
@laytan 
> My opinion on this, (but would love a review from Kelimion) is that the helper function is a nice addition but having it as an option on top isn't needed.

I agree in principle. Still there are reasons, both very minor, admittedly, to include the option:
 - BMPs can be encoded in a flipped format (negative height, iirc). There is already code in the BMP loader that flips the image. By making the flip an option, we can detect that during load and avoid unnecessarily flipping the image twice. Not a big deal of course, but it's slightly nicer.
 - People use stb_image a lot and it has a vertical_flip option, plus it might add to discoverability.

DerTee:
@Kelimion 
I still need to write the tests, no need to review yet.

Kelimion:
> * BMPs can be encoded in a flipped format (negative height, iirc). There is already code in the BMP loader that flips the image. By making the flip an option, we can detect that during load and avoid unnecessarily flipping the image twice. Not a big deal of course, but it's slightly nicer.

TGA as well. See the variable `origin_is_top`, if the image height is negative.

laytan:
Alright, I am fine with the option then

Issue #4834, opened 2025-02-12
Virtual growing arena Out_Of_Memory bug


karl-zylinski:
## Context

        Odin:    dev-2025-01:1613728a6
        OS:      Windows 10 Professional (version: 22H2), build 19045.5371
        CPU:     Intel(R) Core(TM) i7-6950X CPU @ 3.00GHz
        RAM:     32674 MiB
        Backend: LLVM 18.1.8

## Current Behavior

Run this program. It crashes with:
`C:/Projects/Odin Playground/playground.odin(15:2) runtime assertion: Out_Of_Memory`

```
package virtual_growing_arena_bug

import vmem "core:mem/virtual"
import "core:mem"
import "core:fmt"

main :: proc() {
	arena: vmem.Arena
	arena_allocator := vmem.arena_allocator(&arena)

	err: mem.Allocator_Error
	_, err = mem.alloc(4193371, 1, arena_allocator)
	fmt.assertf(err == nil, "%v", err)
	_, err = mem.alloc(896, 64, arena_allocator)
	fmt.assertf(err == nil, "%v", err)
}
```

karl-zylinski:
I think it's something with how 896 + 64 + 4193371 just barely goes over 4 MB, but only because of the 64 bytes alignment. For some reason it then blows up.

karl-zylinski:
Actually, now I got this again, with the exact same repro as above. Either I messed something up or it's slightly system-dependent.

Barinzaya:
> Actually, this no longer happens in latest version of Odin. It was fixed by this: [98b3a9e](https://github.com/odin-lang/Odin/commit/98b3a9eacd9b95a5db75fb001da0bfb0c7a18645)
That change only affects resizes. There are no resizes in your repro, so I wouldn't expect that change to have any impact.

karl-zylinski:
> > Actually, this no longer happens in latest version of Odin. It was fixed by this: [98b3a9e](https://github.com/odin-lang/Odin/commit/98b3a9eacd9b95a5db75fb001da0bfb0c7a18645)
> 
> That change only affects resizes. There are no resizes in your repro, so I wouldn't expect that change to have any impact.

Yeah, I think it only made it slightly more rare or hid it in a certain circumstance. Not sure. I got the bug again now in any case.

Barinzaya:
It does seem to reproduce consistently for me using the current repro code.

jasonKercher:
This fixes it:
```diff
[...]

PR #4835, opened 2025-02-12
Fix issue #4834

jasonKercher:
The logic that determines whether we need to allocate more space does not account having to align the end forward for requested alignment.


laytan:
(quick thought) we could also flip it around and try to allocate_from_memory_block, and if that errors, allocate a new block. Trying to avoid duplicating the calculations.

jasonKercher:
Yeah, I like that better.  This is bad.  I missed that `needed` is used below.  We are botching that value with extra padding it may not need.  I won't have time to do this right until later today if you want to take this one?

PR #4836, opened 2025-02-12
fix wrong out of memory in edge cases, just try allocate from block for one source of truth

### LINKED ISSUE #4834: Virtual growing arena Out_Of_Memory bug

laytan:
Follow up on #4835

Fixes #4834 

Also fixes an already existing bug where on NetBSD and FreeBSD we are doing `mmap` with no protection flags set, on these targets we need to specify a special "maximum" protection value in order to properly `mprotect` with more permissions.

### DIFF
--- a/base/runtime/default_temp_allocator_arena.odin
+++ b/base/runtime/default_temp_allocator_arena.odin
@@ -104,13 +104,15 @@ arena_alloc :: proc(arena: ^Arena, size, alignment: uint, loc := #caller_locatio
 	if size == 0 {
 		return
 	}
-	
-	needed := align_forward_uint(size, alignment)
-	if arena.curr_block == nil || (safe_add(arena.curr_block.used, needed) or_else 0) > arena.curr_block.capacity {
+
+	prev_used := 0 if arena.curr_block == nil else arena.curr_block.used
+	data, err = alloc_from_memory_block(arena.curr_block, size, alignment)
+	if err == .Out_Of_Memory {
 		if arena.minimum_block_size == 0 {
 			arena.minimum_block_size = DEFAULT_ARENA_GROWING_MINIMUM_BLOCK_SIZE
 		}
 
+		needed := align_forward_uint(size, alignment)
 		block_size := max(needed, arena.minimum_block_size)
 
 		if arena.backing_allocator.procedure == nil {
@@ -121,10 +123,9 @@ arena_alloc :: proc(arena: ^Arena, size, alignment: uint, loc := #caller_locatio
 		new_block.prev = arena.curr_block
 		arena.curr_block = new_block
 		arena.total_capacity += new_block.capacity
+		prev_used = 0
+		data, err = alloc_from_memory_block(arena.curr_block, size, alignment)
 	}
-
-	prev_used := arena.curr_block.used
-	data, err = alloc_from_memory_block(arena.curr_block, size, alignment)
 	arena.total_used += arena.curr_block.used - prev_used
 	return
 }
--- a/core/mem/virtual/arena.odin
+++ b/core/mem/virtual/arena.odin
@@ -107,8 +107,9 @@ arena_alloc :: proc(arena: ^Arena, size: uint, alignment: uint, loc := #caller_l
 
 	switch arena.kind {
 	case .Growing:
-		needed := mem.align_forward_uint(size, alignment)
-		if arena.curr_block == nil || (safe_add(arena.curr_block.used, needed) or_else 0) > arena.curr_block.reserved {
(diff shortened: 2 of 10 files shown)


Feoramund:
I can reproduce the NetBSD segfault. However, when I hook a debugger up to it (lldb or gdb), the segfault ceases to happen, making this harder to pin down.

Feoramund:
Looks like it's segfaulting on this line in `platform_memory_alloc`: https://github.com/odin-lang/Odin/blob/f0b1357132de38e9dad0964ebe4c21db1ed4f430/core/mem/virtual/virtual_platform.odin#L21

The value of `block` looks like a valid pointer at the time, as in not zero and evenly divisible by 4096, but accessing it causes a segfault as if the pointer was bogus.

Digging further, it looks like `posix.mprotect` is returning a value != `.OK` and `_commit` returns `.Out_of_Memory`, but we don't check for that (this should probably be changed and panic accordingly or propagate the error up). This is starting to make more sense. The pointer's valid, but its protection hasn't been set correctly yet.

Checking `errno`, I see that the actual error is 13 which should be `EACCES`, and referencing `mprotect` on NetBSD 10.0's manpages:
```
ERRORS
     [EACCES]           A memory protection violation occurred.

                        The PROT_EXEC flag was attempted on pages which belong
                        to a file system mounted with the NOEXEC flag.

                        The new protection is less restrictive than the
                        protection originally set with mmap(2).

                        PaX mprotect restrictions prohibit the requested
                        protection.
```

After seeing that we use `mmap` with no protection flags in `_reserve`, I suspect it's due to the protection being less restrictive, and in fact, when I change it to be `{ .READ, .WRITE }` by default, it works.

laytan:
@Feoramund thanks so much for looking into this!

> Digging further, it looks like posix.mprotect is returning a value != .OK and _commit returns .Out_of_Memory, but we don't check for that (this should probably be changed and panic accordingly or propagate the error up). This is starting to make more sense. The pointer's valid, but its protection hasn't been set correctly yet.

I think the fact it is actually returning `.Out_Of_Memory` is "fine", especially if we now fix the misuse here.

> After seeing that we use mmap with no protection flags in _reserve, I suspect it's due to the protection being less restrictive, and in fact, when I change it to be { .READ, .WRITE } by default, it works.

I've looked at the manpages too, and I am thinking we may need to do this for NetBSD, to allow any protection later being enabled with mprotect? (I don't think we can just give the mmap call `.Read, .Write` because it can also be used by others and they may want executable too, even though `commit` only needs read and write.)

> As a NetBSD extension, the PROT_MPROTECT macro can be used to request
     additional permissions for later use with mprotect(2).  For example
     PROT_MPROTECT(PROT_READ) requests that future PROT_READ mappings are
     allowed and can be enabled using [mprotect(2)](https://man.netbsd.org/mprotect.2), but does not currently
     grant read mappings to the returned memory segment.

FreeBSD seems to have a similar thing:

> In addition to these protection flags, FreeBSD provides the ability  to
       set  the	maximum	protection of a	region allocated by mmap and later al-
       tered by	[mprotect(2)](https://man.freebsd.org/cgi/man.cgi?query=mprotect&sektion=2&apropos=0&manpath=FreeBSD+14.2-RELEASE+and+Ports).  This is accomplished by or'ing one or more PROT_
       values wrapped in the PROT_MAX()	macro into the prot argument.

Feoramund:
Interesting extensions. That looks like the real fix here then, to use that for both FreeBSD (just in case) and NetBSD.

I mentioned raising a panic on `commit` in `platform_memory_alloc` because at the very least having an assert to make sure that it's returning nil would help track down similar bugs in the future.

laytan:
@Feoramund Thanks again, the changes seem to work. This is ready to merge.

PR #4840, opened 2025-02-13
Reordering procedure declaration in the example demo so that it matches the order in which main calls them

Gabr1313:
Moving declaration of the procedure `threading_example`


gingerBill:
Why not change the order in `main`? Do you just want to have a higher contribution amount?

Gabr1313:
No, sincerly it does not matter anything to me. It is my first PR and I don't even know what the contribution amount is, nor what I could do with it. As a begginer I was simply reading `demo.odin`  from top to bottom, and found it difficult to understand the syntax used in the procedure I moved. Reading it later was easier. If that was a mistake, I apologize, and I will be more careful in the future.

PR #4846, opened 2025-02-15
Fix opengl's `check_error` helper procedure not printing errors in debug builds

Memresable:
`check_error` procedures were flipped when defined in the `when GL_DEBUG { ... } else { ... }` blocks, so it didn't print errors by default in debug builds


RoboMage:
Bumping this. Not sure why this one got passed over.

PR #4853, opened 2025-02-17
wgpu: update to v24.0.0.2

laytan:
Lots of changes to webgpu here.

I did not implement futures in JS, neither did wgpu-native in native, so that should be fine.

https://github.com/gfx-rs/wgpu-native/releases/tag/v24.0.0.2


jimhub:
One quick note from testing this out. It looks like `UncapturedErrorCallbackInfo` should not have the callback mode field: https://github.com/webgpu-native/webgpu-headers/blob/bac520839ff5ed2e2b648ed540bd9ec45edbccbc/webgpu.h#L1366-L1371

pietrovismara:
It seems that for this to work on windows you need to import `system:propsys.lib` and `system:runtimeobject.lib` [here ](https://github.com/odin-lang/Odin/blob/77c7087de369e9b82e8a93fea7bccf19e2e1f923/vendor/wgpu/wgpu.odin#L32), otherwise odin build will fail with a bunch of LNK2019 errors.

laytan:
> One quick note from testing this out. It looks like `UncapturedErrorCallbackInfo` should not have the callback mode field: https://github.com/webgpu-native/webgpu-headers/blob/bac520839ff5ed2e2b648ed540bd9ec45edbccbc/webgpu.h#L1366-L1371

@jimhub  Thanks! I've fixed it.

vendor/wgpu/wgpu.odin
```diff
@@ -73,7 +79,11 @@ QUERY_SET_INDEX_UNDEFINED :: max(u32)
 WHOLE_MAP_SIZE :: max(uint)
 WHOLE_SIZE :: max(u64)
 
-Flags :: u32
+Flags :: u64
+
+StringView :: string
```
Valakor, line 84: It's unclear to me whether this is a safe translation of `WGPUStringView`. While it seems likely the binary layout is probably the same (though I don't actually know this since I'm not sure where Odin's `string` type is defined) the wgpu-native API defines a couple special values / interpretations of values that I don't think are possible to represent in Odin.

This is taken from the wgpu-native doc comment for the `WGPUStringView` type:
```
/*
 * Values are encoded as follows:
 * - `{NULL, WGPU_STRLEN}`: the null value.
 * - `{non_null_pointer, WGPU_STRLEN}`: a null-terminated string view.
 * - `{any, 0}`: the empty string.
 * - `{NULL, non_zero_length}`: not allowed (null dereference).
 * - `{non_null_pointer, non_zero_length}`: an explictly-sized string view with
 *   size `non_zero_length` (in bytes).
 */
```

In particular, it seems to me that the following two encodings will cause problems:
- The `{NULL, WGPU_STRLEN}` "null" value encoding
- The `{non_null_pointer, WGPU_STRLEN}` value encoding

I don't believe either of these encodings can be safely represented by the Odin `string` type, and trying to use them (e.g. from an output provided by wgpu-native) may crash your program unless you explicitly know to always destructure and check for these cases beforehand, and mean you can't trivially print any string retrieved from the webgpu-native API.

cstrachan88: The [string type](https://pkg.odin-lang.org/base/runtime/#Raw_String) in Odin is just a pointer and a length, which is the same as the wgpu string view. However, the only two encodings that are for sure safe are the `{any, 0}` and `{non_null_pointer, non_zero_length}`. 

- `{NULL, non_zero_length}` and `{NULL, WGPU_STRLEN}` will crash when you try to use the string
- `{non_null_pointer, WGPU_STRLEN}` will print garbage and/or crash

Valakor: All the callback functions (`BufferMapCallback`, `CompilationInfoCallback`, `CreateComputePipelineCallback`, etc. and the native `LogCallback`) return `StringView` to the user, and functions like `wgpuAdapterGetInfo` do as well. Based on the doc comment from wgpu-native it seems mostly safe to use these output values except it's unclear whether the "null" encoding might be used (which would break Odin programs):

```
 * Values passed into the API may use the special length value @ref STRLEN
 * to indicate a null-terminated string.
 * Non-null values passed out of the API (for example as callback arguments)
 * always provide an explicit length and **may or may not be null-terminated**.
 * 
 * Some inputs to the API accept null values. Those which do not accept null
 * values "default" to the empty string when null values are passed.
```
This also means it's impossible (or at least difficult) for an Odin program to explicitly specify the "null" value which might have meaning as an input parameter separate from empty.

laytan: I don't think any of these concerns happen in practice.

> Non-null values passed out of the API (for example as callback arguments)
 always provide an explicit length and **may or may not be null-terminated**.
 
Means anything passed from the api to us in a callback has a length set.

> Some inputs to the API accept null values. Those which do not accept null
 values "default" to the empty string when null values are passed.

I haven't seen anything about this, nor can I imagine a use case.

PR #4855, opened 2025-02-18
Deterministic Canonical Naming for Link Names and Types

gingerBill:
The ability to generate deterministic naming conventions to be used for link names and types.

This is the first step towards fully deterministic builds.


gingerBill:
Nesting suffix example:
```odin
package pkg 

bar :: proc() {
    Foo :: distinct int // pkg::bar::Foo
    {
        Foo :: distinct int // pkg::bar::Foo[1]
        {
            Foo :: distinct int // pkg::bar::Foo[2]
        }
    }
    {
        Foo :: distinct int // pkg::bar::Foo[3]
    }
}
```

PR #4856, opened 2025-02-18
Update core_builtin.odin

Eugenenoble2005:
fix type in ensure proc documentation


laytan:
That's not a typo, `iff` is a common abbreviation of "if and only if"

PR #4857, opened 2025-02-18
added always_omitempty and always_omitnil flags to encoding_json.Marshal_Options

jkenda:
Not sure if this belongs in the core library but I got sick of writing "omitempty" next to every Maybe so I added these flags.

### DIFF
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -56,6 +56,15 @@ Marshal_Options :: struct {
 	// NOTE: If a name isn't found it'll use the underlying value.
 	use_enum_names: bool,
 
+	// Always omit empty values including empty (c)strings, (dynamic) arrays,
+	// slices, unions, bit sets, maps, and nil values.
+	always_omitempty: bool,
+
+	// Always omit nil values.
+	//
+	// Doesn't omit empty values, unlike always_omitempty.
+	always_omitnil: bool,
+
 	// Internal state
 	indentation: int,
 	mjson_skipped_first_braces_start: bool,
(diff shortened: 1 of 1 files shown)


core/encoding/json/marshal.odin
```diff
@@ -345,6 +354,27 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
 
 	case runtime.Type_Info_Struct:
 		is_omitempty :: proc(v: any) -> bool {
+			v := v
```
gingerBill, line 357: Use `reflect.length` instead.

core/encoding/json/marshal.odin
```diff
@@ -345,6 +354,27 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
...
+			case runtime.Type_Info_Slice:
+				return (^runtime.Raw_Slice)(v.data).len == 0
+			case runtime.Type_Info_Map:
+				return (^runtime.Raw_Map)(v.data).len == 0
+			}
+			return false
+		}
+
+		is_omitnil :: proc(v: any) -> bool {
 			v := v
```
gingerBill, line 378: Use `reflect.is_nil`

jkenda:
reflect.length(x) == 0 and is_omitempty(x) are not always equal which will cause backwards incompatibility. reflect.length gives a more accurate result so I'm guessing that's okay?

jkenda:
This pull request was premature, sorry for wasting your time. Going to keep it on my fork for a while, I might do another pull request when it's polished.

PR #4862, opened 2025-02-20
Added functionality for mem.Arena to support growing in-place. 

Byron-Hodel:
Added `arena_resize*` procedures for `mem.Arena`, supporting:  
- In-place growth when no memory is allocated after the block.  
- Freeing allocations if no subsequent allocations exist (`size == 0`).  
- Allocating a new block and copying data when in-place growth isn't possible.  

In addition to these changes, I've modified the `arena_allocator_proc` to utilize these new procedures.

### DIFF
--- a/core/mem/allocators.odin
+++ b/core/mem/allocators.odin
@@ -234,6 +234,154 @@ arena_free_all :: proc(a: ^Arena) {
 	a.offset = 0
 }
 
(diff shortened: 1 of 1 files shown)


Byron-Hodel:
I was messing around on my branch and it messed up my pr.

PR #4876, opened 2025-02-24
#4873 issue fixed

### LINKED ISSUE #4873: LLVM backend panic when comparing array of strings with empty string literal
## Context

```
Odin:    dev-2025-02
OS:      NixOS 25.05 (Warbler), Linux 6.6.71
CPU:     AMD Ryzen 7 5800H with Radeon Graphics         
RAM:     15847 MiB
Backend: LLVM 18.1.8
```

## Current Behavior

```go
package test

main :: proc() {
    x: [4]string
    if x == "" {}
}
```

`src/llvm_backend_expr.cpp(3195): Panic: Unknown handled type: [4]string -> [4]string`

This does not happen when using anything else than an empty string literal

Relevant gdb output:
```
src/llvm_backend_expr.cpp(3195): Panic: Unknown handled type: [4]string -> [4]string

Thread 13 ".odin-wrapped" received signal SIGILL, Illegal instruction.
[Switching to Thread 0x7fffe86a56c0 (LWP 759814)]
0x000055555570489e in lb_emit_comp_against_nil(lbProcedure*, TokenKind, lbValue) ()
(gdb) 
```

### Steps to Reproduce

Use the above code, and attempt to build it with `odin build main.odin -file`

rezwanahmedsami:
Fixed: #4873

### DIFF
--- a/src/llvm_backend_expr.cpp
+++ b/src/llvm_backend_expr.cpp
@@ -3110,6 +3110,17 @@ gb_internal lbValue lb_emit_comp_against_nil(lbProcedure *p, TokenKind op_kind, 
 			}
 		}
 		break;
+	case Type_Array: {
+			// For arrays where the element type is nil-comparable (like string),
+			// perform a memory comparison of the entire array.
+			auto args = array_make<lbValue>(permanent_allocator(), 2);
+			lbValue lhs = lb_address_from_load_or_generate_local(p, x);
+			args[0] = lb_emit_conv(p, lhs, t_rawptr);
+			args[1] = lb_const_int(p->module, t_int, type_size_of(x.type));
+			lbValue mem_cmp = lb_emit_runtime_call(p, "memory_compare_zero", args);
+			return lb_emit_comp(p, op_kind, mem_cmp, lb_const_int(p->module, t_int, 0));
+	} break;
+		
 
 	case Type_DynamicArray:
 		{


laytan:
This is not correct, `==` on strings does not just check the pointer, but also the content.

rezwanahmedsami:
> This is not correct, `==` on strings does not just check the pointer, but also the content.

what you think ?

What will be the correct way?

laytan:
Also check the content?

laytan:
Or error in the frontend that this is not a valid thing

rezwanahmedsami:
> Or error in the frontend that this is not a valid thing

If you can give me an explanation, what can be wrong?

Then it will be easy for me to catch it.

Because I stay busy with office work most of the time.

So I can't get much time to investigate it.

So, can you help me with that?


Just point out the right thing for me.

laytan:
I'm going to close this because I am not interested in holding your hand here

FourteenBrush:
> This is not correct, `==` on strings does not just check the pointer, but also the content.

@laytan in this case a comparison is being made between an array and the nil value of an array (all zeroes) (`lb_emit_comp_against_nil`). Even though the array might store strings, we should still emit a memcmp 0 over the whole array right?

Could this be reopened if so?

laytan:
Ah you are right, my bad, I didn't clock that this was only for comparing against nil.

laytan:
One thing that would still be inconsistent here is that `str == ""` doesn't care about the pointer, just the length.

For example:
```odin
val := transmute(string)runtime.Raw_String{
	data = auto_cast uintptr(0xDEADBEEF),
	len  = 0,
}
assert(val == "")
```

holds true, no assertion, while with this patch, this would assert:

```odin
vals := [1]string{val}
assert(vals == "")
```

So this may still not be correct.

FourteenBrush:
Hmm I see that counts for other types too, consider an array of maps, dynamic arrays etc, those also only check their data pointer So yes sadly still incorrect

EDIT: never mind, these types are not simply comparable

FourteenBrush:
Works fine on dev-2026-07, corresponding issue was already closed, this may also be closed.

"Works fine" meaning none of this panics:
```go
package test

import "base:runtime"

main :: proc() {
    val := transmute(string)runtime.Raw_String{
	data = auto_cast uintptr(0xDEADBEEF),
	len  = 0,
}
    assert(val == "")

    vals := [1]string{val}
    assert(vals == "")
}

```

PR #4884, opened 2025-02-26
Add DWM_WINDOW_CORNER_PREFERENCE enum for window corner preferences

Dzentsetsu:
This enum directly corresponds to the DWM_WINDOW_CORNER_PREFERENCE values in the Windows API, which allow specifying the rounding preference for window corners.


core/sys/windows/dwmapi.odin
```diff
@@ -38,6 +38,13 @@ DWMNCRENDERINGPOLICY :: enum {
 	DWMNCRP_LAST,
 }
 
+DWM_WINDOW_CORNER_PREFERENCE :: enum c_int{
```
laytan, line 41:
```suggestion
DWM_WINDOW_CORNER_PREFERENCE :: enum c_int {
```

PR #4888, opened 2025-02-27
[WIP] wgpu: Update bindings from v22.1.0.1 to v24.0.0.1

Valakor:
I'm marking this Draft because of a few design questions, and because I haven't yet figured out how to update `wgpu.js` properly.

Notable differences and things I'd like feedback on:
- There are a handful of API-level bugs in the current release that have been fixed since the release of v24.0.0.1 that will need to be included in an official release before merging this PR. Currently the issues are with the return values of `wgpuSurfacePresent`, `wgpuAdapterGetInfo`, and `wgpuSurfaceGetCapabilities`.
- Some API's in WGPU now effectively return a struct+WGPUStatus tuple. I've retained that behavior exactly, but `Status` only has the values `Success` and `Error` at the moment. This fits nicely with the (value, ok) return idiom in Odin if I were to translate the API values to bools instead, but would break if Status ever gains additional values.
- API's no longer accept c-style strings, but instead a new `StringView` type. I've added some helpers (extra constants, creation functions, and a user formatter) that don't exist in the actual WGPU API but I wasn't sure what general policy was on API extras like these.
- Still working on updating the existing examples.
- I'm completely unfamiliar with JavaScript and WASM so I have yet to even attempt to update `wgpu.js`. Maybe @laytan has thoughts here? Some notable changes:
  - The new `StringView` type
  - Flags enums are now `u64` instead of `u32`


Valakor:
Hah I had been working on this on and off for a bit and only noticed @laytan 's similar pull request in #4853... Maybe worth comparing to see if either of us have missed anything?

Valakor:
No worries, it happens. I'll add notes to your original PR.

PR #4893, opened 2025-03-01
Use CTZ in `for in bit_set` codegen

Barinzaya:
This changes the implementation of `for in` over a `bit_set` to use an alternative approach. Instead of looping over every possible value that the `bit_set` could have and testing each one, this instead:

1. Sets up an initial mask using the bit set ANDed with the mask of all bits that are allowed values (the compiler already has a function to construct this mask).
2. As long as there are bits remaining, use CTZ to find the index of the least significant 1 and clear it via `m & (m - 1)`.
3. Adjust the bit index to the corresponding element that it represents (i.e. `Elem_Type(index + lower)`).

For obvious reasons, this approach performs better for `bit_set`s that are sparsely populated, but in my testing it also performs better for densely populated and even full `bit_set`s. The code is also simpler than the previous code (which had to have special cases for enums and ranges--this implementation doesn't).

**But**, is there any platform on which this approach will produce less optimal results? I've been able to test on ARM (a Raspberry Pi) and WASM (via `wasmtime`), and both appear to be improvements in performance without requiring a specific target feature. On the other hand, RISC-V does seem to require an extension (`zbb`) for this. I don't have any RISC-V hardware to compare actual performance on with/without `zbb`, though. Should this logic include a special case to fall back to the original approach on RISC-V without `zbb`?

To-do:

- [x] Implement `#reverse` (note: existing implementation doesn't handle it either!)
- [ ] Fallback to avoid CTZ?

### DIFF
--- a/src/llvm_backend_stmt.cpp
+++ b/src/llvm_backend_stmt.cpp
@@ -1119,6 +1119,10 @@ gb_internal void lb_build_range_stmt(lbProcedure *p, AstRangeStmt *rs, Scope *sc
 			lbValue all_mask = lb_const_value(p->module, mask, exact_bit_set_all_set_mask(et));
 			lbValue initial_mask = lb_emit_arith(p, Token_And, the_set, all_mask, mask);
 
+			if (rs->reverse) {
+				initial_mask = lb_emit_reverse_bits(p, initial_mask, mask);
+			}
+
 			lbAddr remaining = lb_add_local_generated(p, mask, false);
 			lb_addr_store(p, remaining, initial_mask);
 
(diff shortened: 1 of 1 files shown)


gingerBill:
Very cool to see! I just wrote the dumb thing first to get it working.

Barinzaya:
Considering switching the `#reverse` implementation to use CLZ instead of reversing the bits, but this approach was very simple to implement and still performs reasonably well.

Anyway, this all seems to be functional as far as I can tell. I've tested as many corner cases as I can think of (enums with a positive minimum, a negative minimum, sparse enums, ranges with a positive lower limit, ranges with a negative lower limit, ranges with a non-power-of-two size, etc. Aside from one unrelated issue that I found (which is fixed separately in #4900), I think these changes should be solid.

I guess the remaining question is: how does this fare on RISC-V? I don't have a RISC-V system to test on. Is it worth having a separate codepath that does this the old way and avoids CTZ to accommodate RISC-V without `zbb`?

Kelimion:
I'd treat RISC-V without `zbb` as e-waste and not bother with fallback code for it. If the backend has to synthesize the equivalent of `ctz` for it and that has some overhead, so be it. I see no reason to complicate the compiler to special case hardware that doesn't care about such utter basics.

But at the end of the day it's @gingerBill's call to make.

Barinzaya:
Going to pull this out of draft, then. I think it should be ready to go, unless a non-CTZ fallback is desired.

PR #4894, opened 2025-03-01
Add support for SDL3_image.

0xafbf:
I started from the existing port of SDL2_image, updated the APIs and removed things that were also removed in SDL3.

I tried to keep the style I see in other files. also included the DLLs that SDL releases for win_x64.


vendor/sdl3/image/sdl_image.odin
```diff
@@ -0,0 +1,107 @@
+package sdl3_image
+
+import "core:c"
+import SDL ".."
```
gingerBill, line 4: Prefer `import SDL "vendor:sdl3"`

PR #4896, opened 2025-03-02
Fixed json null value parsing

### LINKED ISSUE #4895: JSON package doesn't roundtrip `null` values
## Context

```
Odin:    dev-2025-02-nightly:584fdc0
OS:      macOS Sequoia 15.3.1 (build 24D70, kernel 24.3.0)
CPU:     Apple M4
RAM:     24576 MiB
Backend: LLVM 18.1.8
```

I needed a small tool to combine several JSON files into one that I wanted to try Odin for, but got met with a marshaling error "Unsupported type". This is because the original files contain objects with fields of value `null` that need to be roundtripped (see code example below). I suspect this is because the parsed value is a `json.Null` union variant, especially because using `nil` works, but directly using `json.Null {}` does not. The unsupported type error gets raised here https://github.com/odin-lang/Odin/blob/f390598b403eb336276ef9161753bf26d24d0d01/core/encoding/json/marshal.odin#L178

Let me know should there be a setting I haven't found.

## Expected Behavior

`null` values can be roundtripped when first parsed and then marshaled.

## Current Behavior

Unsupported type error.

## Example

The original roundtripping code I wanted to write is in the last example block. The first 2 showcase the error when manually building a JSON object where `nil` works but `json.Null {}` doesn't.

```odin
package main

import "core:fmt"
import "core:encoding/json"

main :: proc() {
	// works
	{
		parsed: json.Object
		parsed["id"] = 234
		parsed["title"] = "asdf"
		parsed["runtime"] = 56.9
		parsed["creator"] = nil
		bytes, err := json.marshal(parsed)
		assert(err == nil)
		str := cast(string)bytes
		fmt.println(str)
	}

	// does not work
	{
		parsed: json.Object
		parsed["id"] = 234
		parsed["title"] = "asdf"
		parsed["runtime"] = 56.9
		parsed["creator"] = json.Null {} // this is what you get when parsing null
		bytes, err := json.marshal(parsed)
		assert(err == nil)
		str := cast(string)bytes
		fmt.println(str)
	}

	// does not work - original roundtripping code I wanted to write
	{
		content := `
{
	"id": 234,
	"title": "asdf",
	"runtime": 56.9,
	"creator": null
}
`
		parsed, _ := json.parse(transmute([]u8)content)
		bytes, err := json.marshal(parsed)
		assert(err == nil)
		str := cast(string)bytes
		fmt.println(str)
	}
}
```

cstrachan88:
Fixes #4895


laytan:
I think we should check and only allow explicitly the `json.Null` type here, not any other pointer type

AmedeoAlf:
This is still broken in master, is there any hope of it being merged?

Kelimion:
OP hasn't responded to @laytan's feedback. I don't think it will be merged without that change.

cstrachan88:
> OP hasn't responded to @laytan's feedback. I don't think it will be merged without that change.

Sorry I dropped the ball on this. Fix is in.

PR #4899, opened 2025-03-02
wgpu: Some small fixes after update to v24.0.0.2

Valakor:
- The native type of the "device" argument in `DeviceLostCallback` and `UncapturedErrorCallback` is `WGPUDevice const * device`. At first I had `Maybe(Device)` but I believe this pointer is always non-null, though it might itself point to a null device.
- `BindGroupLayoutReference` should be `BindGroupLayoutAddRef`
- The wrapper `SurfaceGetCapabilities` should return `Status` as part of its return value
- Missing `/* const */` annotations
- A handful of `ChainedStruct` vs. `ChainedStructOut` fixes
- Change `SurfaceConfigurationExtras.desiredMaximumFrameLatency` to `u32` to match wgpu-native
- Fix enumeration `SpirvShaderPassthrough` (was `SprivShaderPassThrough`)

Side note: There are a handful of API functions in `wgpu-native` that aren't implemented and just panic if you attempt to call them. They're all the functions stubbed here: https://github.com/gfx-rs/wgpu-native/blob/trunk/src/unimplemented.rs
I wonder if all these functions should just be commented out so someone can't try to call them?


laytan:
Thanks!

> The native type of the "device" argument in DeviceLostCallback and UncapturedErrorCallback is WGPUDevice const * device. At first I had Maybe(Device) but I believe this pointer is always non-null, though it might itself point to a null device.

Aha, this needs to be changed in the js implementation too then.

> I wonder if all these functions should just be commented out so someone can't try to call them?

I don't like just not having them because people will think the bindings are incomplete. Additionally, most of them are actually implemented in the js implementation, so somebody just doing js could still want them, or conditionally call them only in js. They panic anyway so I think it's fine.

Valakor:
I haven't quite figured out how exactly the JS bindings work, so I'd appreciate you taking a look at those when you get the chance.

vendor/wgpu/wgpu.odin
```diff
@@ -762,100 +762,100 @@ BufferMapCallback :: #type proc "c" (status: MapAsyncStatus, message: StringView
...
 
 ChainedStructOut :: struct {
-	next:  ^ChainedStructOut,
+	next: ^ChainedStructOut,
 	sType: SType,
 }
 
 BufferMapCallbackInfo :: struct {
-	nextInChain: ^ChainedStructOut,
+	nextInChain: /* const */ ^ChainedStruct,
```
gingerBill, line 783: Why are you adding these comments? This doesn't help anyone.

vendor/wgpu/wgpu.odin
```diff
@@ -762,100 +762,100 @@ BufferMapCallback :: #type proc "c" (status: MapAsyncStatus, message: StringView
...
 	mode: CallbackMode,
 	callback: BufferMapCallback,
-	userdata1: rawptr,
-	userdata2: rawptr,
+	userdata1: /* NULLABLE */ rawptr,
```
gingerBill, line 786: Does this actually matter? And if so, `Maybe(rawptr)` might be a better way to indicate the nil-ability

vendor/wgpu/wgpu.odin
laytan: I added them initially to the bindings when I started them. I agree they aren't helpful but it makes sense to add them here because they are added everywhere else already.

If you want me to remove them everywhere, say so, and I'll get a PR going to remove them everywhere.

vendor/wgpu/wgpu.odin
laytan: It doesn't really matter here but is useful in a few places, especially in proc definitions to see what needs to be passed.

laytan: For the `Maybe` thing, maybe? :p 

But just like with the const stuff it would make sense to do everywhere so the whole package is consistent, we can do that in a different PR if wanted, I don't have a strong opinion on either way.

PR #4910, opened 2025-03-05
Remove unneeded slicing in `strings.clone`

Feoramund:
Unless I'm mistaken, there should be no need to slice a newly allocated byte slice by its exact length for a regular `string`. (This differs from `clone_from_bytes` which appends a null byte and does need to re-slice `c` according to its API description.)

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -25,7 +25,7 @@ Returns:
 clone :: proc(s: string, allocator := context.allocator, loc := #caller_location) -> (res: string, err: mem.Allocator_Error) #optional_allocator_error {
 	c := make([]byte, len(s), allocator, loc) or_return
 	copy(c, s)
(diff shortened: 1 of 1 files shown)


laytan:
Depends if we can assume `make` with say 8 bytes does not return a slice of length > 8 bytes.

I feel like we should be able to assume/enforce that.

laytan:
Ah I see `make_slice` already always returns a slice of the given length and doesn't care about whatever the allocator implementation might return.

Feoramund:
I was working under the assumption that if any allocator returns a slice not equal to the number of bytes requested, then that's a bug (assuming an otherwise successful/non-erroneous return value). Possibly something worth adding to #4216. You're right though in that `make_slice` disregards the length of what the allocator returns; I wasn't aware of that.

PR #4911, opened 2025-03-06
New core:container type - Pool

RoboMage:
Adding in Pool as a new core container type.
Everybody loves data pools, so having a generic version in the core library will likely be very useful. It's dynamically resizable, and has single bit flags to track active slots in the pool.

### DIFF
--- a/core/container/pool/pool.odin
+++ b/core/container/pool/pool.odin
@@ -1,0 +1,96 @@
+package container_pool
+
+import "core:fmt"
+
+// Single bit to mark a slot as active
+PoolFlag :: bit_set[0..<64;u64]
+
+// Dynamically resizable generic data pool
+Pool :: struct($type: typeid) {
+    data : [dynamic]type,
+    flag : [dynamic]PoolFlag,
+    free : [dynamic]int
+}
+
+// Reserve memory for the pool
+init :: proc($type: typeid, data_cnt : uint, free_cnt : uint = 8) -> Pool(type) {
+    pool : Pool(type)
+    reserve(&pool.data, data_cnt)
+    reserve(&pool.flag, (data_cnt / 64) + 1)
+    reserve(&pool.free, free_cnt)
+    return pool
+}
+
+is_index_valid :: proc(pool : ^Pool($T), index : int) -> (is_valid : bool) {
+    if index >= 0 && index < len(pool.data) {
+        flag := pool.flag[index / 64]
+        
+        is_valid = index % 64 in flag
+    }
+    return
+}
+
+active_count :: proc(pool : ^Pool($T)) -> int {
+    total : int
+    for f in pool.flag do total += card(f)
+    return total
+}
+
+get_ref :: proc(pool : ^Pool($T), #any_int index : int, loc := #caller_location ) -> (value : ^T = nil, ok : bool) #optional_ok {
+    if is_index_valid(pool, index) {
+        value = &pool.data[index]
+        ok = true
+    } else {
+        fmt.eprintfln("%v: Index is out of bounds", loc)
+    }
+    return
+}
+
+get_value :: proc(pool : ^Pool($T), #any_int index : int, loc := #caller_location ) -> (value : T, ok : bool) #optional_ok {
+    if is_index_valid(pool, index) {
+        value = pool.data[index]
+        ok = true
+    } else {
+        fmt.eprintfln("%v: Index is out of bounds", loc)
+    }
+    return
+}
+
+insert :: proc(pool : ^Pool($T), value: T) -> int {
+    idx : int = -1
+    if len(pool.free) > 0 {
+        idx = pop(&pool.free)
+        pool.data[idx] = value
+    } else {
+        idx = len(pool.data)
+        append(&pool.data, value)
+        
+        // If we're out of flags, add a new set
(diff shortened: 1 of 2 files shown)


Feoramund:
At first I thought this might be a Pool allocator. After reviewing the code, I'm not sure I see what the algorithmic benefit to using this over a plain `[dynamic]T` array would be. Assuming most people use `unordered_remove` on their dynamic arrays, is there some advantage that I'm missing that this structure affords? I see that it keeps a list of freed entries and remove operations work by flipping bits on a bitmap to indicate an empty or filled state.

gingerBill:
The implementation of this pool isn't that great in the first place (too many dynamic arrays, and you could have used an in-place free-list). 

Nor have you adhered to any of the core library's style either. 

I will not accept this whatsoever.

core/container/pool/pool.odin
```diff
@@ -0,0 +1,96 @@
+package container_pool
+
+import "core:fmt"
+
+// Single bit to mark a slot as active
+PoolFlag :: bit_set[0..<64;u64]
```
flysand7, line 6: `bit_set[0..<64; u64]`

core/container/pool/pool.odin
```diff
@@ -0,0 +1,96 @@
...
+
+// Dynamically resizable generic data pool
+Pool :: struct($type: typeid) {
+    data : [dynamic]type,
```
flysand7, line 10: The core library style does not use a space before `:` in a declaration, unless it is directly followed by `=`.

core/container/pool/pool.odin
```diff
@@ -0,0 +1,96 @@
...
+    flag : [dynamic]PoolFlag,
+    free : [dynamic]int
+}
+
+// Reserve memory for the pool
+init :: proc($type: typeid, data_cnt : uint, free_cnt : uint = 8) -> Pool(type) {
```
flysand7, line 16: Please avoid declaring identifiers for types that start with a lowercase letter. It should be either `Type` or `T`, but `T` should have been preferred.

core/container/pool/pool.odin
```diff
@@ -0,0 +1,96 @@
...
+        flag := pool.flag[index / 64]
+        
+        is_valid = index % 64 in flag
+    }
+    return
+}
+
+active_count :: proc(pool : ^Pool($T)) -> int {
+    total : int
+    for f in pool.flag do total += card(f)
```
flysand7, line 35: Avoid `do` in the core library.

core/container/pool/pool.odin
```diff
@@ -0,0 +1,96 @@
...
+    return total
+}
+
+get_ref :: proc(pool : ^Pool($T), #any_int index : int, loc := #caller_location ) -> (value : ^T = nil, ok : bool) #optional_ok {
+    if is_index_valid(pool, index) {
+        value = &pool.data[index]
+        ok = true
+    } else {
+        fmt.eprintfln("%v: Index is out of bounds", loc)
```
flysand7, line 44: This function should either:

- Panic if an illegal state has been reached
- Return `ok = false`, if the operation could not be performed

You are printing to the error stream *AND* returning `false`. This does not really make any sense. If you provide an index that is out of bounds, it is clearly a bug in your usage code. You should not provide an index that is out of bounds when accessing the pool. The best behaviour for a program if it detects a bug is to panic (Which is to crash, but here the error is simply ignored).

It does not make sense to return `ok = false` here either by the same reasoning. A bug should not be returned as a boolean false.

core/container/pool/pool.odin
```diff
@@ -0,0 +1,96 @@
...
+        value = pool.data[index]
+        ok = true
+    } else {
+        fmt.eprintfln("%v: Index is out of bounds", loc)
+    }
+    return
+}
+
+insert :: proc(pool : ^Pool($T), value: T) -> int {
+    idx : int = -1
```
flysand7, line 60: `idx := -1`

core/container/pool/pool.odin
```diff
@@ -0,0 +1,96 @@
...
+}
+
+remove :: proc(pool : ^Pool($T), #any_int index : int, loc := #caller_location ) {
+    if is_index_valid(pool, index) {
+        pool.data[index] = 0
+        append(&pool.free, index)
+
+        // Get the flag set, then invalidate the active flag
+        flag_set  := &pool.flag[index / 64]
+        flag_set^ -= PoolFlag{index % 64}
```
flysand7, line 86: `flag_set &~= index%64 `

Or `PoolFlag { index%64 }`

core/container/pool/pool.odin
flysand7, line 6: The types should be named using Ada case, that is words are separated by an underscore and each word starts with a capital letter: `Pool_Flag`.

core/container/pool/pool.odin
flysand7, line 16: Core library uses `int` type for all indices and sizes. This makes computations involve less casts.

flysand7 (commented):
In case you are curious, I have pointed out most of your stylistic mistakes. Don't take them as an insult, but you do need to adhere to the style to some capacity. But like the others said, this isn't the main reason why this isn't getting merged -- this version of a pool is, to put it lightly, not much more useful than a plain dynamic array.

PR #4912, opened 2025-03-06
[datetime]: Remove the definition of EPOCH

flysand7:
This PR removes the unnecessary declaration and semantically-incorrect usages of EPOCH.

This affects the following places:
- Converting ordinals to year: `day - 1` simply offsets the day count such that the zeroeth day starts at `CE 0001-01-01`, instead of `BCE 0001-12-31`, because the leap year cycle starts one day later than "Ordinal Day 0".
- Weekday calculations: `CE 0001-01-01` is Ordinal Day 1, and is defined as Monday. It follows that the week day number is simply Ordinal Day mod 7.

I have also made a little bit more sense with the constants in `ordinal_to_year`, making it appear more like the number of days 400, 100 and 4 year intervals would have. I made sure I didn't make a mistake by calculating these expressions and comparing the answer to the negative part of the diff.

### DIFF
--- a/core/time/datetime/constants.odin
+++ b/core/time/datetime/constants.odin
@@ -5,12 +5,10 @@ Type representing a mononotic day number corresponding to a date.
 
 	Ordinal 1 = Midnight Monday, January 1, 1 A.D. (Gregorian)
 	        |   Midnight Monday, January 3, 1 A.D. (Julian)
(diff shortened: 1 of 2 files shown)


flysand7:
@Kelimion Here it is :)

PR #4919, opened 2025-03-07
Add missing x86 syscalls from map_shadow_stack to removexattrat

latedeployment:


latedeployment:
@gingerBill it seems the MacOS arm CI is broken it also fails here on the same reason, https://github.com/odin-lang/Odin/pull/4915, when I run the CI commands locally on my mac it works fine.

PR #4928, opened 2025-03-11
Added simple test coverage overview

andreas-jonsson:
Added [kcov](https://github.com/SimonKagstrom/kcov) to the build pipeline and upload the result to [codecov.io](https://codecov.io/) for easy visualization of the test coverage in Odin.

You can find the results here: https://app.codecov.io/gh/andreas-jonsson/Odin/

_Consider this more of a demo then something that should be merged as is._

### DIFF
--- a/.github/workflows/cover.yml
+++ b/.github/workflows/cover.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Normal Core library tests
         run: |
           ./odin build tests/core/normal.odin -build-mode:test -debug -file -all-packages -vet -strict-style -disallow-do -define:ODIN_TEST_FANCY=false -define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true -target:linux_amd64
-          kcov kcov-out ./tests .
+          kcov kcov-out ./core .
 
       - name: Optimized Core library tests
         run: |
--- a/.github/workflows/cover.yml
+++ b/.github/workflows/cover.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Normal Core library tests
         run: |
           ./odin build tests/core/normal.odin -build-mode:test -debug -file -all-packages -vet -strict-style -disallow-do -define:ODIN_TEST_FANCY=false -define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true -target:linux_amd64
-          kcov kcov-out ./core .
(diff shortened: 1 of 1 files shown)


Feoramund:
This is awesome. I didn't know you could check code coverage without leveraging the compiler in some way or making an ad-hoc inline collector that ticked some flags when certain paths were crossed. I remember seeing [some LLVM documentation](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html) for their coverage generation with the hopes that we could one day use it, but if something like this exists, then maybe we don't need to.

Really cool demo, even if it's not merged.

Kelimion:
I'll be looking into this, @andreas-jonsson. Just made a codecov.io account to have a play with it.

I think it could help guide us in prioritizing tests for packages missing any coverage, and densifying coverage on packages that have spotty test coverage.

Is there a technical reason you're using such old Ubuntu and LLVM versions in your CI action?

andreas-jonsson:
Cool!

I think the reason was that kcov was not in the apt repo of any later versions of Ubuntu at the time I did the PR and I was too lazy to install it from source. :)

Kelimion:
@laytan mentioned something about not all lines being picked up, which skews the overall %, so we're open to running automatic coverage tests`*`, but we won't publish one of those badges on the README.

`*` for the benefit of those with an interest to furthering the tests.

I think we'll just install the build-essentials and build `kcov` from source, and use this PR as a reference.

Feoramund:
I can confirm seeing some strange output, even on the line-by-line.
strange_output

I wouldn't rule out Odin's debug info generation as being partly responsible, because I will run into this sort of thing even with `lldb` on rare occasion.

However, this is still an incredibly useful tool as a potential guide for where more tests could be added.

I have to advise anyone using this at home that you should be aware of procedures taking polymorphic parameters; they will be ignored, because they are technically not compiled in. You have to use them at least once (and it's a good idea to do so, to make sure they compile in the first place).

Kelimion:
We have working [code coverage](https://app.codecov.io/gh/odin-lang/Odin) now. Took some doing because it needed `ubuntu-latest` instead of '22, and so forth, and it intially didn't recognize I had set a CODECOV_TOKEN secret (because fork), but we have it working.

Closing this PR with thanks for the inspiration, @andreas-jonsson.

PR #4929, opened 2025-03-12
Add bit_set to `core:sys/posix` with all OS types that should be POSIX compliant

RaphGL:
I've been writing a library and I needed to conditionally compile a function so if the OS was POSIX I could do a syscall otherwise provide a fallback. 

With this bit_set I could do this:
```odin
when ODIN_OS in posix.ALL_VALID_OS_TYPES {
    my_func :: proc() { /* do something */ }
} else {
    my_func :: proc() { /* do something */ }
}
```

### DIFF
--- a/core/sys/posix/posix.odin
+++ b/core/sys/posix/posix.odin
@@ -50,11 +50,12 @@ Unimplemented headers:
 package posix
 
 import "base:intrinsics"
+import "base:runtime"
 
 import "core:c"
 
 result :: enum c.int {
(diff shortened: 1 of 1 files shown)


laytan:
This should probably just be solved like we do in `core:thread` and `core:time` among others, where we have an `IS_SUPPORTED` constant that is true or false based on the OS. Your code would then be `when posix.IS_SUPPORTED {`

laytan:
I am going to close this, feel free to make another PR with my comment's suggestion.

PR #4935, opened 2025-03-12
core/fmt: add nprint family of procs to get required buffer sizes

mailgerigk:
This pr adds nprint* style procs to fmt to be easily query how many bytes are needed to store the formatted output. This is useful if you want to acquire a buffer with the correct size to use with the bprint* procs.


gingerBill:
I see no need for this to exist in `fmt` but rather you want to create a stream that just counts the number of bytes written and then pass it to the `fmt.wprint*` procedures.

PR #4946, opened 2025-03-19
Added note for vkguide-compatible perspective projection fix

nevertrimnuts:
Im not sure whether something like this can or should be included in the core math library (or perhaps there's another better place to put it), but for me, and I expect for others as well following popular vulkan guides such as vkguide, it would be nice to have at least one note like this that fixes a bug like this, if only at least to advertise that the odin glsl math library is not a drop-in replacement for glm.

Also note: this is the first instance of a math bug happening because of a discrepancy between vkguide's math library of choice (glm) and the odin glsl library.

### DIFF
--- a/core/math/linalg/glsl/linalg_glsl.odin
+++ b/core/math/linalg/glsl/linalg_glsl.odin
@@ -1373,6 +1373,34 @@ not :: proc{
 
 @(require_results) identity :: proc "c" ($M: typeid/matrix[$N, N]$T) -> M { return 1 }
 
+
+// NOTE: vkguide, in chapter 3, subchapter "Mesh Loading" uses a view and 
+// projection matrix in order to render "Suzanne", the monkey head.
+// this module advertises itself as being a replacement, at least to some
+// extent of the popular "glm" c++ math library. As such, I tried to use
+// the `mat4Perspective` function in order to do perspective projection,
+// exactly how vkguide does it, and was disappointed to see nothing rendered.
+// If, from that point, you pass in the near & far parameters to 
+// `mat4Perspective` in the reverse order, you will see the monkey head, 
+// but perspective will be reversed: things farther away will be LARGER
+// and things closer will be smaller.
+// vkguide does the reverse-z depth buffer trick to increase rendering accuracy
+// of far-away objects, which is why this occurs.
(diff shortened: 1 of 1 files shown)


JesseRMeyer:
GLM can be configured to support Vulkan's coordinate system, which is incompatible with OpenGL's.  I'm not sure where in vkguide this is being done, as it's not mentioned anywhere explicitly.  I think this is the source of the confusion.

https://community.khronos.org/t/confused-when-using-glm-for-projection/108548/2

Our GLSL compatibility assumes OpenGL, not Vulkan.  I'm in favor of supplying a vulkan compatible perspective transform, but the note referring to vkguide isn't appropriate in core.

nevertrimnuts:
@JesseRMeyer where would be the best place to put something like a vulkan compatible perspective transform function? If I made a directory + file "core:math/linalg/vulkan/linalg_vulkan.odin" and started work on implementing vulkan-compatible math functions, would that be welcome?

> GLM can be configured to support Vulkan's coordinate system, which is incompatible with OpenGL's. I'm not sure where in vkguide this is being done, as it's not mentioned anywhere explicitly.  

The only way vkguide influences the compilation of glm is by defining the macro #GLM_FORCE_DEPTH_ZERO_TO_ONE, which ends up compiling this code in glm:
```
#ifdef GLM_FORCE_DEPTH_ZERO_TO_ONE
#	ifdef GLM_FORCE_LEFT_HANDED
#		define GLM_CONFIG_CLIP_CONTROL GLM_CLIP_CONTROL_LH_ZO
#	else
#		define GLM_CONFIG_CLIP_CONTROL GLM_CLIP_CONTROL_RH_ZO
#	endif
#else
#	ifdef GLM_FORCE_LEFT_HANDED
#		define GLM_CONFIG_CLIP_CONTROL GLM_CLIP_CONTROL_LH_NO
#	else
#		define GLM_CONFIG_CLIP_CONTROL GLM_CLIP_CONTROL_RH_NO
#	endif
#endif
```
which ends up compiling [this function](https://github.com/g-truc/glm/blob/2d4c4b4dd31fde06cfffad7915c2b3006402322f/glm/ext/matrix_clip_space.inl#L233) for perspective projection.

PR #4963, opened 2025-03-23
Added proc last_index_multi

SrMordred:
Added on `core:strings` the procedure `last_index_multi`
following the same conventions that `index_multi` have + tests.

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -1891,6 +1891,48 @@ index_multi :: proc(s: string, substrs: []string) -> (idx: int, width: int) {
 	return
 }
 /*
+Finds the last occurrence of any substring in `substrs` within `s`
+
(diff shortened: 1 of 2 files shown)


core/strings/strings.odin
```diff
@@ -1891,6 +1891,48 @@ index_multi :: proc(s: string, substrs: []string) -> (idx: int, width: int) {
...
+	for substr in substrs {
+		if len(substr) == 0 {
+			return
+		}
+	}
+
+	highest_index := -1
+	found := false
+	for substr in substrs {
+        haystack_slice_start := max( 0 , highest_index )
```
laytan, line 1919: This line is indented differently from the others, also has some weird formatting where you added a space after and before the parentheses and comma which we don't do.

PR #4969, opened 2025-03-25
Added golden ratio constant to math package

gorevojd:
Added golden ratio constant to math package


gingerBill:
No.

flysand7:
Just let me clarify on why we're not accepting this constant: Simply put, its not useful. PI is useful in trigonometric function, and Euler's constant has a connection to logarithms. The golden ratio, on the other hand, does have a connection to pentagons and the fibonacci sequence, but those aren't anywhere in the math package.

If you want to use this constant for graphics programming, you'd better stick with plain old `1.6` or `1.618`, as you don't need crazy precision to deal with them, but note that your choice of that constant would be coming from aesthetics rather than from math.

PR #4981, opened 2025-03-28
core/encoding/endian: enumerate platform byte order

avanspector:
Enables for
`i, ok := endian.get_u32(a[:], .Platform)`
instead of
`i, ok := endian.get_u32(a[:], endian.PLATFORM_BYTE_ORDER)`,
and due to how `fmt`/`reflect` work,
`fmt.println(endian.Byte_Order.Platform)` will still output/produce `.Little` or `.Big`, which is quite elegant.

### DIFF
--- a/core/encoding/endian/endian.odin
+++ b/core/encoding/endian/endian.odin
@@ -6,6 +6,7 @@ import "core:math/bits"
 Byte_Order :: enum u8 {
 	Little,
 	Big,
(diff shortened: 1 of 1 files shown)


gingerBill:
No. You could just do `Platform :: endian.PLATFORM_BYTE_ORDER` in your code. And if you want platform specific endianness, it sounds like you are misusing the library.

Issue #10, opened 2017-01-19
LLVM Error when returning a ? type


JonBoyleCoding:
I was just trying to implement some basic 'file' reading (actually stdin) functionality just for my own testing of the language, and was thinking about how one could use your ? 'maybe' functionality.

My goal was to write a function that returns a ?byte instead of having to return an EOF like in the C fgetc implementation.

```
#import "fmt.odin";
#import "os.odin";

fgetc :: proc(fd : os.Handle) -> ?byte
{
	c : [1]byte;

	if bytes_read, _ := os.read(fd, c[:]); bytes_read == 0
	{
		return nil;
	}

	return c[0];
}

main :: proc ()
{
	b := fgetc(os.stdin);

	if bv, ok := b?; ok
	{
		fmt.println(b);
	}
}
```

However, when compiling, I end up with the following (llvm?) error

> test.ll:1128:9: error: expected type
>         store  zeroinitializer, * %19

I believe the relevant segment in the LLVM bytecode is as follows:
```
if.then-1:
	; ReturnStmt
	store {i8, i1} zeroinitializer, {i8, i1}* %4
	%19 = getelementptr inbounds {i8, i1}, {i8, i1}* %4, i64 0, i32 0
	%20 = getelementptr inbounds {i8, i1}, {i8, i1}* %4, i64 0, i32 1
	store  zeroinitializer, * %19
	store i1 true, i1* %20
	%21 = load {i8, i1}, {i8, i1}* %4, align 1
	ret {i8, i1} %21
```

gingerBill:
This is now fixed. It was caused by a printing error in ir_print.c

PR #4988, opened 2025-04-01
Integer Concatenation Operator `##`

gingerBill:
The ability to concatenate two integers.

* `10 ## 20 == 1020`
* `123 ## 456 == 123456`

Implemented as the equivalent of the following code:

```odin
integer_concatenate :: proc(x, y: uint) -> uint {
    pow: uint = 10
    for y >= pow {
        pow *= 10
    }
    return x*pow + y
}
```

Also supports `##=` operator:

```
a ##= b
// equivalent to
a = a ## b
```

Useful for those cases when you need to join two numbers together for those recreational mathematics problems.

### DIFF
--- a/core/odin/parser/parser.odin
+++ b/core/odin/parser/parser.odin
@@ -1627,7 +1627,8 @@ token_precedence :: proc(p: ^Parser, kind: tokenizer.Token_Kind) -> int {
 	case .Mul, .Quo,
 	     .Mod, .Mod_Mod,
 	     .And, .And_Not,
-	     .Shl, .Shr:
+	     .Shl, .Shr,
+	     .Concat:
 		return 7
 	}
 	return 0
--- a/core/odin/tokenizer/token.odin
+++ b/core/odin/tokenizer/token.odin
@@ -64,6 +64,8 @@ Token_Kind :: enum u32 {
 		Shl,      // <<
 		Shr,      // >>
 
+		Concat,   // ##
+
 		Cmp_And,  // &&
 		Cmp_Or,   // ||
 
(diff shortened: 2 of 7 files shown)


pankkor:
Thanks! Long awaited feature!
What would happen when `y == -1`?

flysand7:
I think for consistency's sake we should use the same operator as for concatenating strings. The `+` operator should work like a concatenation operator, if both inputs are constants, and like the addition operator if both inputs are variables. If one input is a constant and the other is a variable, that should lead to a compiler error, suggesting to cast one of the inputs to a variable.

```odin
2 + 2 == 22

x := 2
y := 2
x + y == 4

x + 2 == (compiler error)
x + int(2) == 4
```

peachey2k2:
One caveat I feel like you skipped is when you need it for bases other than 10, so maybe we should modify it to also work for those purposes.

`31 #10# 69 == 3169`

`0xcafe #16# 0xbabe == 3405691582`

Lperlind:
I think it would be much easier to use if we require an extra levels of indirection similar to the C preprocessor.

Lots of devs coming to Odin are from C/C++ so this can make it easier for them to understand as it will feel more familiar.

burak-efe:
LGTM

korvahkh:
Clearly the operator should be `..` (Lua does this and it just makes sense)

As an aside, you should be able to iterate over the digits of an integer with `for in`

For example:

```odin
x := 5
y := 28
for d in x..y {
    fmt.println(d)
}
// Output:
// 5
// 2
// 8 
```

gingerBill:
This was an April Fool's Joke.

pankkor:
I hope we can still have it! Won't it be fun to have some infinite looping going on, when second operand is big enough (`y == -1`)? :)

PR #4992, opened 2025-04-02
ci: update to LLVM 20 on MacOS and Linux CI and releases

laytan:


laytan:
huh, why does brew on the ci not have llvm 20 but it is released and available locally for me, weird

PR #4993, opened 2025-04-02
[DOC] Add documentation for package core:container/small_array

Sojamann:
I wrote a (hopefully) nearly complete documentation of the `container/small_array` package with a bunch of examples. Let me know what needs changing ✌️.

### DIFF
--- a/core/container/small_array/small_array.odin
+++ b/core/container/small_array/small_array.odin
@@ -4,36 +4,171 @@ import "base:builtin"
 import "base:runtime"
 _ :: runtime
 
+/*
+A fixed-size stack-allocated array operated on in a dynamic fashion.
+
+Fields:
+- `data`: The underlying array
+- `len`: Amount of items that the `Small_Array` currently holds
+
+Example:
+
+	import "core:container/small_array"
+
+	example :: proc() {
+		a: small_array.Small_Array(100, int)
+		small_array.push_back(&a, 10)
+	}
+*/
 Small_Array :: struct($N: int, $T: typeid) where N >= 0 {
 	data: [N]T,
 	len:  int,
 }
 
+/*
+Returns the amount of items in the small-array.
 
+**Inputs**
+- `a`: The small-array
+
+**Returns**
+- the amount of items in the array
+*/
 len :: proc "contextless" (a: $A/Small_Array) -> int {
 	return a.len
 }
 
+/*
+Returns the capacity of the small-array.
+
+**Inputs**
+- `a`: The small-array
+
+**Returns** the capacity
+*/
 cap :: proc "contextless" (a: $A/Small_Array) -> int {
 	return builtin.len(a.data)
 }
 
+/*
+Returns how many more items the small-array could fit.
+
+**Inputs**
+- `a`: The small-array
+
+**Returns**
+- the number of unused slots
+*/
 space :: proc "contextless" (a: $A/Small_Array) -> int {
 	return builtin.len(a.data) - a.len
 }
 
+/*
+Returns a slice of the data.
+
+**Inputs**
+- `a`: The pointer to the small-array
+
+**Returns**
+- the slice
+
+Example:
+
+	import "core:container/small_array"
+	import "core:fmt"
+
+	slice_example :: proc() {
+		print :: proc(a: ^small_array.Small_Array($N, int)) {
+			for item in small_array.slice(a) {
+				fmt.println(item)
+			}
+		}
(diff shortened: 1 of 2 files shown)


core/container/small_array/small_array.odin
```diff
@@ -4,59 +4,313 @@ import "base:builtin"
...
+		// the length of the small-array remains unchanged
+		small_array.set(&a, 4, 'X')
+		fmt.println(small_array.slice(&a))
+
+		// resizing makes the change visible
+		small_array.resize(&a, 100)
+		fmt.println(small_array.slice(&a))
+	}
+
+Output:
```
Kelimion, line 234: This test appears to fail:
```
Test "set_example" got unexpected output:
"[A, B]\n[Z, B]\n[Z, B]\n[Z, B, "
Expected:
"[A, B]\n[Z, B]\n[Z, B]\n[Z, B, , , X]\n"
```

Kelimion: Looks quite good apart from this test failure. 👍

Sojamann: Alright I will address this.

The log output also mentions that not all defined functions can be executed e.g.

> We could not find the procedure "slice_example :: proc()" needed to test the example created for "container_small_array.slice"
> The following procedures were found:
> 	slice_example :: proc(a: ^small_array.Small_Array($N, int))
> 	main :: proc()

Is the goal for all examples to be executable or just not fail in the pipeline? I would then also have to adapt the examples. Just let me know.

Kelimion: Main priority is not to have any of them that are executable fail, like the one I mkentioned. The more examples that are executable, the better. If not all of them are executable, that itself isn't a showstopper.

Sojamann: I fixed all examples, hoping for the pipeline to pick up on all examples and passing at the end. If it fails again I will spend more time getting the _documentation tester_ to run locally.

Sojamann: :c

Kelimion: Almost seems as if the output is getting interrupted: `"[A, B]\n[Z, B]\n[Z, B]\n[Z, B, "`. Where's the closing bracket and newline? Running it locally may indeed be a good idea.

Sojamann: No small_array example/test is failing locally. "it works on my machine"

[test.log](https://github.com/user-attachments/files/19594949/test.log)

Sojamann:
I tested it on a Windows machine indeed getting the same error as the pipeline showed ...

Changing the example like this:

```diff
diff --git a/core/container/small_array/small_array.odin b/core/container/small_array/small_array.odin
index d86058620..aee3bc1a3 100644
--- a/core/container/small_array/small_array.odin
+++ b/core/container/small_array/small_array.odin
@@ -224,7 +224,9 @@ Example:
 
 		// updates index 4, which is not visible since
 		// the length of the small-array remains unchanged
-		small_array.set(&a, 4, 'X')
+		small_array.set(&a, 2, 'X')
+		small_array.set(&a, 3, 'Y')
+		small_array.set(&a, 4, 'Z')
 		fmt.println(small_array.slice(&a))
 
 		// resizing makes the change visible
@@ -237,7 +239,7 @@ Output:
 	[A, B]
 	[Z, B]
 	[Z, B]
-	[Z, B, , , X]
+	[Z, B, X, Y, Z]
 
 */
 set :: proc "contextless" (a: ^$A/Small_Array($N, $T), index: int, item: T) {

```

Resulted in passing the tests: [log.txt](https://github.com/user-attachments/files/19615509/log.txt)

The default value of the rune might make the documentation runner from continue reading the test output on Windows only?
I'm going to adapt the example accordingly since this is a different Issue if this is ok

Kelimion:
> The default value of the rune might make the documentation runner from continue reading the test output on Windows only? I'm going to adapt the example accordingly since this is a different Issue if this is ok

Right, best to adapt the example so it passes. There may be a bug in `read_*` from `stdin` on Windows that's nothing to do with this.

Kelimion:
If this passes I'll merge it. Thanks for the documentation. :-)

Sojamann:
Should I squash the changes before the merge?

Kelimion:
That'd be lovely. A bit friendlier on `git bisect`.

Kelimion:
Thanks, @Sojamann.

PR #5016, opened 2025-04-07
Fix `math.nextafter` skipping from 0 to 1

Barinzaya:
When `x == 0`, the `math.nextafter` procs were skipping straight from 0 to 1/-1. I think these were intended to be a bitcasted *integer* 1, which will be the floating-point value with the smallest magnitude.

The three `nextafter` implementations (`f16`, `f32`, and `f64`) have all been updated to use their respective smallest float values rather than `1` in the `copy_sign` call.

### DIFF
--- a/core/math/math.odin
+++ b/core/math/math.odin
@@ -2296,7 +2296,7 @@ nextafter_f16 :: proc "contextless" (x, y: f16) -> (r: f16) {
 	case x == y:
 		r = x
 	case x == 0:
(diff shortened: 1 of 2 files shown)


Kelimion:
Would it be too much to ask you to add a test for this to `tests/core/math`?
Something that has a `[][2]f64` with hex floats `f64(0h........)`, one being `val` and the other the expected `val` + 1 ulp returned by `nextafter`.

PR #5032, opened 2025-04-13
core:sync/chan: add package documentation

Sojamann:
This PR provides the documentation for the `core:sync/chan` package.

### Code Ordering
I would like to move the code for `Raw_Queue` to the bottom so that all *chan* related procedures are grouped together.

### Make 'Raw_Queue' private
The `Raw_Queue`  should not be exported since it does not really match purpose of the package. Therefore my proposal is to deprecate all functions related to the `Raw_Queue` to make it `@(private)` in the future.

PR #5035, opened 2025-04-13
core:sync/chan: maintainance and package clarity

Sojamann:
This implements the proposals from https://github.com/odin-lang/Odin/pull/5032

### DIFF
--- a/core/sync/chan/chan.odin
+++ b/core/sync/chan/chan.odin
@@ -1037,139 +1037,6 @@ is_closed :: proc "contextless" (c: ^Raw_Chan) -> bool {
 	return bool(c.closed)
 }
 
-
-
-/*
-`Raw_Queue` is a non-thread-safe queue implementation designed to store messages
-of fixed size and alignment.
-
-Note: For most use cases, it is recommended to use `core:container/queue` instead,
-as `Raw_Queue` is used internally by `Raw_Chan` and may not provide the desired
-level of convenience for typical applications.
-*/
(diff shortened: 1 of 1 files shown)


Kelimion:
I don't think it's a good idea to use the `@(deprecated)` tag.

`raw_queue_init` is used by `create_raw_buffered` for example. You don't want regular use of the channels themselves to spam you with deprecation warnings.

And with `-warnings-as-errors`, you can't even use the package at all. Please go with `@(private)`, as per your initial suggestion.

Kelimion:
Try compiling this, with and without `-warnings-as-errors`. As a user of `create_raw_unbuffered`, we really don't need to know that something *it* calls is going private, let alone have it fail to compile.
```odin
package scratch

import "core:sync/chan"

main :: proc() {
	c, err := chan.create_raw_buffered(size_of(int), align_of(int), 10, context.allocator)
	assert(err == .None)
	defer chan.destroy(c)
}
```

Sojamann:
I see .... I was not aware of the consequences thanks for letting me know 🙏.

My train of thought was that potential users would, by deprecating it, have the time to use something else but yeah I was on the wrong track there.

Kelimion:
The core packages need to be able to compiled with `-vet -strict-style -vet-style -warnings-as-errors -disallow-do`, and the CI enforces this. That the deprecation didn't cause the PR to fail surprises me. We may need to add `-warnings-as-errors` there. I'll have a look.

PR #5045, opened 2025-04-16
fix not resolving to alias in a recursive declaration

### LINKED ISSUE #5043: switch statement over union does not see first case if the first case is a map.
## Context

```
Odin:    dev-2025-04-nightly:d9f990d
OS:      Windows 10 Professional (version: 22H2), build 19045.4529   
CPU:     AMD Ryzen 5 3600 6-Core Processor
RAM:     16283 MiB
Backend: LLVM 20.1.0
```

Originally found by arp242 [here](https://github.com/Up05/toml_parser/issues/5#issuecomment-2807800831) but it seems to be easily reproducible, so this is just more of a credit.

## Expected Behavior

The compiler should be able to detect that a switch statement has a case.

## Current Behavior

It doesn't detect it, if the first case is a map/pointer to a map

BTW: It works if we move the case with map down by 1 case or use `#partial`

Just:
```odin
Table :: map [string] Type
List  :: [dynamic] Type

Type :: union {
    ^Table,
    ^List,
    i64,
}


main :: proc() {
    v: Type = 5

    switch t in v {
    case ^Table: // or case ^map [string] Type:
    case ^List:
    case i64:

    }
}
```

### Failure Logs

```
D:/src/odin/toml_parser/test.odin(16:5) Error: Unhandled switch case: ^Table 
        Suggestion: Was '#partial switch' wanted? 
```

laytan:
Fixes #5043 

The union variant type was assigned `^Table` instead of the aliassed `^map[string]Type` because we set `is_type_alias = true` after checking the rhs (containing the union to check) of the declaration.

### DIFF
--- a/src/check_decl.cpp
+++ b/src/check_decl.cpp
@@ -468,6 +468,10 @@ gb_internal void check_type_decl(CheckerContext *ctx, Entity *e, Ast *init_expr,
 	}
 	e->type = named;
 
+	if (!is_distinct) {
+		e->TypeName.is_type_alias = true;
+	}
+
 	check_type_path_push(ctx, e);
 	Type *bt = check_type_expr(ctx, te, named);
 	check_type_path_pop(ctx);
@@ -502,9 +506,9 @@ gb_internal void check_type_decl(CheckerContext *ctx, Entity *e, Ast *init_expr,
 	if (!is_distinct) {
 		e->type = bt;
 		named->Named.base = bt;
-		e->TypeName.is_type_alias = true;
 	}
 
+	e->TypeName.is_type_alias = !is_distinct;
 
 	if (decl->type_expr != nullptr) {
 		Type *t = check_type(ctx, decl->type_expr);


graphitemaster:
This PR breaks the following code.
```
package main
Node_Ptr :: ^Node
Node :: struct {
	node: Node_Ptr,
}
main :: proc() {
	node: Node_Ptr
}
```

When resolving the type for `node` in the `Node` struct, via `lb_type_internal`
```
(gdb) print base->Struct.fields.data[0]->type->Basic->kind
$8 = Basic_Invalid
```

You can see that the `base->Struct.fields.data[0]` `Entity` has an unresolved state.
```
(gdb) print base->Struct.fields.data[0]->state
$9 = std::atomic<EntityState> = { EntityState_Unresolved }
```

For some reason this change prevents the entity from being resolved correctly. It seems to be because of the initial setting of `is_type_alias = true` in `check_type_decl`
```cpp
if (!is_distinct) {
    e->TypeName.is_type_alias = true;
}
```

Removing (or commenting this out) fixes it. Following the call-stack it seems that `is_type_alias` changes the behavior of `add_type_info_dependency` here
```cpp
if (type->kind == Type_Named) {
    Entity *e = type->Named.type_name;
    if (e->TypeName.is_type_alias) {
        type = type->Named.base;
    }
}
```
Which doesn't have the `Node_Ptr` type but rather `^Node` and so the `Node_Ptr` type is never added as a dependency to the `d->type_info_deps` map and thus the `Entity` stays unresolved.

Since this fix actually breaks a bunch of code I'm reverting it, reopening the issue it was meant to fix and going to increase the priority of finding a proper fix.

PR #5064, opened 2025-04-21
Add initial support for Objective-C class implementation

harold-b:
# Add support for Objective-C class implementation

The purpose of this work is to add full Odin-native support for creating Objective-C classes, not just consuming them.
Doing so without modifying syntax, only leveraging attributes and new compiler intrinsics. I also attempt to leverage existing attributes and augment them when necessarry to add the desired new functionality.

This facilitates creating new classes where needed, for things like delegates, event handlers, etc. without ugly runtime workarounds, or leaving Odin to provide it in a different language.

This PR has the following requirements:

- Automatically register user-made classes with the Objective-C runtime at startup
- Allow specifying a superclass
- Allow association with a single type that would serve as a singular Ivar that is added to the new Objective-C class
- Allow specifying a selector name manually to be used with the implemented class
- Provide an easy way to get the ivar value for an Objective-C class.

What it does not aim to do (at least not this PR):
- Add new syntax
- Register associated protocols with a class.
- Force protocol conformance or register protocols during class generation. Supports for things like [conformsToProtocol](https://developer.apple.com/documentation/objectivec/nsobject-swift.class/conforms(to:)?language=objc) won't be available.
- Attempt to automatically propagate the Odin state (context) into the Objective-C runtime flow.

### Requirements
- [x] Implement `@objc_implement` attribute for structs
- [x] Implement `@objc_superclass` attribute for procs
    - [x] Disallow superclass cycles
    - [x] Ensure sane ordering during generation so that superclasses get generated first
- [x] Implement `@objc_ivar` attribute
- [ ] Limit generating the global ivar offset to intrinsic usage?
- [x] Implement `@objc_selector` attribute

- [ ] Apply a legitimate strategy for export/linkage
    - [ ] If the `@export` attribute is applied to an Objective-C class, then all methods are exported. Currently all are exported unconditionally (`@export` is implied)

- [ ] Checker:
    - [x] Basic preliminary checks
    - [ ] Proper checks specification
    - [ ] Proper checks applied across all changes
- [ ] Proper, understandable error messages
- [ ] Resolve all pending TODOs

### State and discussion
At the time of this writing the full example below works as all the main requirements have been implemented. It will only work with a single module as there's still issues with the changes. Due to lack of experience with the codebase there needs to be some guidance from the team to get this to a proper shippable state.

I'd like to get feedback on attribute and intrinsic naming in addition to the code changes itself.

### Practical Explanation & Usage

Creating a new Objective-C class looks like it does when consuming an existing one, except we add the `@objc_implement` attribute:
```odin
@(objc_class="HBAppDelegate", objc_implement)
HBAppDelegate :: struct {
    using _: NSF.Object,
    using _: NS.ApplicationDelegate,
}
```

Typically you also want to specify a super/base class as well:
```odin
@(objc_class="HBAppDelegate", objc_implement, objc_superclass=NS.Object)
HBAppDelegate :: struct { ...
```

Finally you may want to associate a type to serve as the class' Ivar:
```odin
@(objc_class="HBAppDelegate", objc_implement, objc_superclass=NS.Object, objc_ivar=HBAppDelegateT)
HBAppDelegate :: struct {
    using _: NSF.Object,
    using _: NS.ApplicationDelegate,
}

// This serves as the only Ivar for HBAppDelegate, it's just a regular struct. 
HBAppDelegateT :: struct {
    ctx: runtime.Context,
    // ...
}
```

You can add methods to it the same way you do when consuming existing Objective-C methods, simply add the same `@objc_implement` attribute:
```odin
@(objc_type=HBAppDelegate, objc_implement, objc_name="alloc", objc_is_class_method=true)
HBAppDelegate_alloc :: proc "c" () -> ^HBAppDelegate {
    return msgSend(^HBAppDelegate, HBAppDelegate, "alloc")
}
```

The compiler will generate a hidden function that conforms to the signature that the Objective-C runtime expects, which simply forwards the call to our method `proc`:
```odin
__$objc_method::MyFoo::alloc :: proc "c" ( self: objc_id, _cmd: objc_SEL ) -> ^HBAppDelegate {
    return HBAppDelegate_alloc()
}
```

Any arguments will be forwarded:
```odin
@(objc_type=HBAppDelegate, objc_implement, objc_name="initWithContext")
HBAppDelegate_initWithContext :: proc "c" ( self: ^HBAppDelegate, ctx: runtime.Context ) -> ^HBAppDelegate {
    s := intrinsics.ivar_get(self, HBAppDelegateT)
    s.ctx = ctx

    return self
}

// Generates:
__$objc_method::MyFoo::alloc :: proc "c" ( self: objc_id, _cmd: objc_SEL, ctx: runtime.Context ) -> ^HBAppDelegate {
    return HBAppDelegate_initWithContext(self, ctx)
}
```

Notice the use of the new intrinsic `intrinsics.ivar_get`. This intrinsic will obtain a pointer to the Type's data of the Ivar from an instance, if one is associated with the class. This is implemented by recording the offset to the ivar at startup, then it simply offsets from the instance pointer during the call to the intrinsic.

You can also specify the selector name to be used for a method. By default the value of `@objc_name` will be used, if one is not specified. But there will be many cases when you may want to specify one manually, as it is when implementing a protocol. Specify it via the `@objc_selector` attribute:

```odin
@(objc_type=HBAppDelegate, objc_implement, objc_name="applicationWillFinishLaunching", objc_selector="applicationWillFinishLaunching:")
HBAppDelegate_applicationWillFinishLaunching :: #force_inline proc "c" (self: ^HBAppDelegate, notification: ^NS.Notification) {
...
}

@(objc_type=HBAppDelegate, objc_implement, objc_name="applicationDidFinishLaunching", objc_selector="applicationDidFinishLaunching:")
HBAppDelegate_applicationDidFinishLaunching :: #force_inline proc "c" (self: ^HBAppDelegate, notification: ^NS.Notification) {
...
}
```

Full working example of an `NSAppDelegate` (partially) implemented in Odin:

```odin
package objcsandbox

import "base:intrinsics"
import "base:runtime"
import "core:log"
import NSF "libs:darwodin/Foundation"
import NS  "libs:darwodin/AppKit"

msgSend  :: intrinsics.objc_send
ivar_get :: intrinsics.ivar_get


main :: proc() {
    context.logger = log.create_console_logger()

    app := NS.Application.sharedApplication()
    app->setActivationPolicy(.Regular)
    app->activate()
    app->setDelegate(HBAppDelegate.alloc()->initWithContext(context))
    app->run()
}

State :: enum {
    Initializing,
    Launching,
    FinishedLaunching,
}
HBAppDelegateT :: struct {
    ctx:   runtime.Context,
    state: State,
}

@(objc_class="HBAppDelegate", objc_superclass=NSF.Object, objc_implement, objc_ivar=HBAppDelegateT)
HBAppDelegate :: struct {
    using _: NSF.Object,
    using _: NS.ApplicationDelegate,
}

@private
get_self :: #force_inline proc "c" ( self: ^HBAppDelegate ) -> ^HBAppDelegateT {
    return ivar_get(self, HBAppDelegateT)
}

@(objc_type=HBAppDelegate, objc_implement, objc_name="alloc", objc_is_class_method=true)
HBAppDelegate_alloc :: proc "c" () -> ^HBAppDelegate {
    return msgSend(^HBAppDelegate, HBAppDelegate, "alloc")
}

@(objc_type=HBAppDelegate, objc_implement, objc_name="initWithContext")
HBAppDelegate_initWithContext :: proc "c" ( self: ^HBAppDelegate, ctx: runtime.Context ) -> ^HBAppDelegate {
    s := get_self(self)
    s.ctx = ctx

    return self
}

@(objc_type=HBAppDelegate, objc_implement, objc_name="applicationWillFinishLaunching", objc_selector="applicationWillFinishLaunching:")
HBAppDelegate_applicationWillFinishLaunching :: #force_inline proc "c" (self: ^HBAppDelegate, notification: ^NSF.Notification) {
    s := get_self(self)
    context = s.ctx
    log.infof("applicationWillFinishLaunching: %v -> %v", s.state, State.Launching)
    s.state = .Launching
}

@(objc_type=HBAppDelegate, objc_implement, objc_name="applicationDidFinishLaunching", objc_selector="applicationDidFinishLaunching:")
HBAppDelegate_applicationDidFinishLaunching :: #force_inline proc "c" (self: ^HBAppDelegate, notification: ^NSF.Notification) {
    s := get_self(self)
    context = s.ctx
    log.infof("applicationDidFinishLaunching: %v -> %v", s.state, State.FinishedLaunching)
    s.state = .FinishedLaunching
}
```

### DIFF
--- a/base/intrinsics/intrinsics.odin
+++ b/base/intrinsics/intrinsics.odin
@@ -353,15 +353,18 @@ x86_xgetbv :: proc(cx: u32) -> (eax, edx: u32) ---
 objc_object   :: struct{}
 objc_selector :: struct{}
 objc_class    :: struct{}
+objc_ivar     :: struct{}
+
 objc_id    :: ^objc_object
 objc_SEL   :: ^objc_selector
 objc_Class :: ^objc_class
+objc_Ivar  :: ^objc_ivar
 
 objc_find_selector     :: proc($name: string) -> objc_SEL   ---
 objc_register_selector :: proc($name: string) -> objc_SEL   ---
 objc_find_class        :: proc($name: string) -> objc_Class ---
 objc_register_class    :: proc($name: string) -> objc_Class ---
-
+ivar_get			   :: proc(self: ^$T, $U: typeid) -> ^U ---
 
 valgrind_client_request :: proc(default: uintptr, request: uintptr, a0, a1, a2, a3, a4: uintptr) -> uintptr ---
 
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -387,6 +387,80 @@ gb_internal bool check_builtin_objc_procedure(CheckerContext *c, Operand *operan
 		try_to_add_package_dependency(c, "runtime", "objc_allocateClassPair");
 		return true;
 	} break;
+
+    case BuiltinProc_objc_ivar_get:
+    {
+        Type *self_type = nullptr;
+        Type *ivar_type = nullptr;
+
+        Operand self {};
+        check_expr_or_type(c, &self, ce->args[0]);
+
+        if (!is_operand_value(self) || !check_is_assignable_to(c, &self, t_objc_id)) {
+            gbString e = expr_to_string(self.expr);
+            gbString t = type_to_string(self.type);
+            error(self.expr, "'%.*s' expected a type or value derived from intrinsics.objc_object, got '%s' of type %s", LIT(builtin_name), e, t);
+            gb_string_free(t);
+            gb_string_free(e);
+            return false;
+        }
+        else if (!is_type_pointer(self.type)) {
+            gbString e = expr_to_string(self.expr);
+            gbString t = type_to_string(self.type);
+            error(self.expr, "'%.*s' expected a pointer of a value derived from intrinsics.objc_object, got '%s' of type %s", LIT(builtin_name), e, t);
+            gb_string_free(t);
+            gb_string_free(e);
+            return false;
+        }
+
+        self_type = type_deref(self.type);
+
+        if (!(self_type->kind == Type_Named &&
+            self_type->Named.type_name != nullptr &&
+                self_type->Named.type_name->TypeName.objc_class_name != "")) {
+            gbString t = type_to_string(self_type);
+            error(self.expr, "'%.*s' expected a named type with the attribute @(obj_class=<string>) , got type %s", LIT(builtin_name), t);
+            gb_string_free(t);
+            return false;
+        }
+
+        if (self_type->Named.type_name->TypeName.objc_ivar == nullptr) {
+            gbString t = type_to_string(self_type);
+            error(self.expr, "'%.*s' requires that type %s have the attribute @(obj_ivar=<ivar_type_name>).", LIT(builtin_name), t);
+            gb_string_free(t);
+            return false;
+        }
+
+        Operand ivar {};
+        check_expr_or_type(c, &ivar, ce->args[1]);
+        if (ivar.mode == Addressing_Type) {
+            ivar_type = ivar.type;
+        } else {
+            return false;
+        }
+
+        if (self_type->Named.type_name->TypeName.objc_ivar != ivar_type) {
+            gbString name_self     = type_to_string(self_type);
+            gbString name_expected = type_to_string(self_type->Named.type_name->TypeName.objc_ivar);
+            gbString name_given    = type_to_string(ivar_type);
+            error(self.expr, "'%.*s' ivar type %s does not match @obj_ivar type %s on Objective-C class %s.",
+                  LIT(builtin_name), name_given, name_expected, name_self);
+            gb_string_free(name_self);
+            gb_string_free(name_expected);
+            gb_string_free(name_given);
+            return false;
+        }
+
+        if (type_hint != nullptr && type_hint->kind == Type_Pointer && type_hint->Pointer.elem == ivar_type) {
+            operand->type = type_hint;
+        } else {
+            operand->type = alloc_type_pointer(ivar_type);
+        }
+
+        operand->mode = Addressing_Value;
+
+        return true;
+    } break;
 	}
 }
 
@@ -2132,7 +2206,8 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 	case BuiltinProc_objc_find_selector: 
 	case BuiltinProc_objc_find_class: 
 	case BuiltinProc_objc_register_selector: 
-	case BuiltinProc_objc_register_class: 
+	case BuiltinProc_objc_register_class:
+    case BuiltinProc_objc_ivar_get:
 		return check_builtin_objc_procedure(c, operand, call, id, type_hint);
 
 	case BuiltinProc___entry_point:
--- a/src/check_decl.cpp
+++ b/src/check_decl.cpp
@@ -526,6 +526,54 @@ gb_internal void check_type_decl(CheckerContext *ctx, Entity *e, Ast *init_expr,
 		check_decl_attributes(ctx, decl->attributes, type_decl_attribute, &ac);
 		if (e->kind == Entity_TypeName && ac.objc_class != "") {
 			e->TypeName.objc_class_name = ac.objc_class;
+            e->TypeName.objc_superclass = ac.objc_superclass;
+            e->TypeName.objc_ivar       = ac.objc_ivar;
+
+            if (ac.objc_is_implementation) {
+                e->TypeName.objc_is_implementation = true;
+                mpsc_enqueue(&ctx->info->objc_class_implementations, e);    // TODO(harold): Don't need this for anything. Remove.
+
+                GB_ASSERT(e->TypeName.objc_ivar == nullptr || e->TypeName.objc_ivar->kind == Type_Named);
+
+                // Ensure superclass hierarchy are all Objective-C classes and does not cycle
+                Type *super = ac.objc_superclass;
+                if (super != nullptr) {
+                    TypeSet super_set{};
+                    type_set_init(&super_set, 8);
+                    defer (type_set_destroy(&super_set));
+
+                    type_set_update(&super_set, e->type);
+
+                    for (;;) {
+                        if (type_set_update(&super_set, super)) {
+                            error(e->token, "@(objc_superclass) Superclass hierarchy cycle encountered");
+                            break;
+                        }
+
+                        if (super->kind != Type_Named) {
+                            error(e->token, "@(objc_superclass) References type must be a named struct.");
+                            break;
+                        }
+
+                        Type* named_type = base_type(super->Named.type_name->type);
+                        if (!is_type_objc_object(named_type)) {
+                            error(e->token, "@(objc_superclass) Superclass must be an Objective-C class.");
+                            break;
+                        }
+
+                        super = super->Named.type_name->TypeName.objc_superclass;
+                        if (super == nullptr) {
+                            break;
+                        }
+
+                        // TODO(harold): Is this the right way to do this??? The referenced entity must be already resolved
+                        //               so that we can access its objc_superclass attribute
+                        check_single_global_entity(ctx->checker, super->Named.type_name, super->Named.type_name->decl_info);
+                    }
+                }
+            } else if (e->TypeName.objc_superclass != nullptr) {
+                error(e->token, "@(objc_superclass) can only be applied when the obj_implement attribute is also applied");
+            }
 
 			if (type_size_of(e->type) > 0) {
 				error(e->token, "@(objc_class) marked type must be of zero size");
@@ -942,6 +990,31 @@ gb_internal void check_objc_methods(CheckerContext *ctx, Entity *e, AttributeCon
 			if (tn->scope != e->scope) {
 				error(e->token, "@(objc_name) attribute may only be applied to procedures and types within the same scope");
 			} else {
+
+                if (ac.objc_is_implementation) {
+                    GB_ASSERT(e->kind == Entity_Procedure);
+
+                    CheckerInfo *info = ctx->info;
+                    mutex_lock(&info->objc_method_mutex);
+                    defer (mutex_unlock(&info->objc_method_mutex));
+
+                    auto method = ObjcMethodData{ ac, e };
+
+                    if (ac.objc_selector == "") {
+                        method.ac.objc_selector = ac.objc_name;
+                    }
+
+                    Array<ObjcMethodData>* method_list = map_get(&info->objc_method_implementations, t);
+                    if (method_list) {
+                        array_add(method_list, method);
+                    } else {
+                        auto list = array_make<ObjcMethodData>(permanent_allocator(), 1, 8);
+                        list[0] = method;
+
+                        map_set(&info->objc_method_implementations, t, list);
+                    }
+                }
+
 				mutex_lock(&global_type_name_objc_metadata_mutex);
 				defer (mutex_unlock(&global_type_name_objc_metadata_mutex));
 
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -1351,10 +1351,12 @@ gb_internal void init_universal(void) {
 		t_objc_object   = add_global_type_name(intrinsics_pkg->scope, str_lit("objc_object"),   alloc_type_struct_complete());
 		t_objc_selector = add_global_type_name(intrinsics_pkg->scope, str_lit("objc_selector"), alloc_type_struct_complete());
 		t_objc_class    = add_global_type_name(intrinsics_pkg->scope, str_lit("objc_class"),    alloc_type_struct_complete());
+        t_objc_ivar     = add_global_type_name(intrinsics_pkg->scope, str_lit("objc_ivar"),     alloc_type_struct_complete());
 
 		t_objc_id       = alloc_type_pointer(t_objc_object);
 		t_objc_SEL      = alloc_type_pointer(t_objc_selector);
 		t_objc_Class    = alloc_type_pointer(t_objc_class);
+        t_objc_Ivar     = alloc_type_pointer(t_objc_ivar);
 	}
 }
 
@@ -1387,6 +1389,9 @@ gb_internal void init_checker_info(CheckerInfo *i) {
 	array_init(&i->defineables, a);
 
 	map_init(&i->objc_msgSend_types);
+    mpsc_init(&i->objc_class_implementations, a);
+    map_init(&i->objc_method_implementations);
+
 	string_map_init(&i->load_file_cache);
 	array_init(&i->all_procedures, heap_allocator());
 
@@ -3345,6 +3350,11 @@ gb_internal DECL_ATTRIBUTE_PROC(proc_decl_attribute) {
 		ac->test = true;
 		return true;
 	} else if (name == "export") {
+        if (ac->objc_is_implementation) {
+            error(value, "Setting @(export) explicitly is not allowed when @(objc_implement) is set. It is exported implicitly.");
+            return false;
+        }
+
 		ExactValue ev = check_decl_attribute_value(c, value);
 		if (ev.kind == ExactValue_Invalid) {
 			ac->is_export = true;
@@ -3356,6 +3366,12 @@ gb_internal DECL_ATTRIBUTE_PROC(proc_decl_attribute) {
 		}
 		return true;
 	} else if (name == "linkage") {
+
+        if (ac->objc_is_implementation) {
+            error(value, "Explicit linkage not allowed when @(objc_implement) is set. It is set implicitly");
+            return false;
+        }
+
 		ExactValue ev = check_decl_attribute_value(c, value);
 		if (ev.kind != ExactValue_String) {
 			error(value, "Expected either a string 'linkage'");
@@ -3662,6 +3678,35 @@ gb_internal DECL_ATTRIBUTE_PROC(proc_decl_attribute) {
 			}
 		}
 		return true;
+	} else if (name == "objc_implement") {
+        ExactValue ev = check_decl_attribute_value(c, value);
+        if (ev.kind == ExactValue_Bool) {
+            ac->objc_is_implementation = ev.value_bool;
+        } else if (ev.kind == ExactValue_Invalid) {
+            ac->objc_is_implementation = true;
+        } else {
+            error(elem, "Expected a boolean value, or no value, for '%.*s'", LIT(name));
+        }
+
+        // This implies exported, strongly linked
+        if (ac->objc_is_implementation) {
+            ac->is_export = true;
+            ac->linkage   = str_lit("strong");
+        }
+
+        return true;
+    } else if (name == "objc_selector") {
+		ExactValue ev = check_decl_attribute_value(c, value);
+		if (ev.kind == ExactValue_String) {
+			if (string_is_valid_identifier(ev.value_string)) {
+				ac->objc_selector = ev.value_string;
+			} else {
+				error(elem, "Invalid identifier for '%.*s', got '%.*s'", LIT(name), LIT(ev.value_string));
+			}
+		} else {
+			error(elem, "Expected a string value for '%.*s'", LIT(name));
+		}
+		return true;
 	} else if (name == "require_target_feature") {
 		ExactValue ev = check_decl_attribute_value(c, value);
 		if (ev.kind == ExactValue_String) {
@@ -3901,8 +3946,36 @@ gb_internal DECL_ATTRIBUTE_PROC(type_decl_attribute) {
 			ac->objc_class = ev.value_string;
 		}
 		return true;
-	}
-	return false;
+	} else if (name == "objc_implement") {
+        ExactValue ev = check_decl_attribute_value(c, value);
+        if (ev.kind == ExactValue_Bool) {
+            ac->objc_is_implementation = ev.value_bool;
+        } else if (ev.kind == ExactValue_Invalid) {
+            ac->objc_is_implementation = true;
+        } else {
+            error(elem, "Expected a boolean value, or no value, for '%.*s'", LIT(name));
+        }
+        return true;
+    } else if (name == "objc_superclass") {
+        Type *objc_superclass = check_type(c, value);
+
+        if (objc_superclass != nullptr) {
+            ac->objc_superclass = objc_superclass;
+        } else {
+            error(value, "'%.*s' expected a named type", LIT(name));
+        }
+        return true;
+    } else if (name == "objc_ivar") {
+        Type *objc_ivar = check_type(c, value);
+
+        if (objc_ivar != nullptr) {
+            ac->objc_ivar = objc_ivar;
+        } else {
+            error(value, "'%.*s' expected a named type", LIT(name));
+        }
+        return true;
+    }
+    return false;
 }
 
 
--- a/src/checker.hpp
+++ b/src/checker.hpp
@@ -148,8 +148,12 @@ struct AttributeContext {
 
 	String  objc_class;
 	String  objc_name;
-	bool    objc_is_class_method;
+    String  objc_selector;
 	Type *  objc_type;
+    Type *  objc_superclass;
+    Type *  objc_ivar;
+	bool    objc_is_class_method   : 1;
+    bool    objc_is_implementation : 1;     // This struct or proc provides a class/method implementation, not a binding to an existing type.
 
 	String require_target_feature; // required by the target micro-architecture
 	String enable_target_feature;  // will be enabled for the procedure only
@@ -365,6 +369,11 @@ struct ObjcMsgData {
 	Type *proc_type;
 };
 
+struct ObjcMethodData {
+    AttributeContext ac;
+    Entity *proc_entity;
+};
+
 enum LoadFileTier {
 	LoadFileTier_Invalid,
 	LoadFileTier_Exists,
@@ -478,6 +487,12 @@ struct CheckerInfo {
 
 	BlockingMutex objc_types_mutex;
 	PtrMap<Ast *, ObjcMsgData> objc_msgSend_types;
+
+    MPSCQueue<Entity *> objc_class_implementations;
+
+    BlockingMutex objc_method_mutex;
+    PtrMap<Type *, Array<ObjcMethodData>> objc_method_implementations;
+
 
 	BlockingMutex load_file_mutex;
 	StringMap<LoadFileCache *> load_file_cache;
--- a/src/checker_builtin_procs.hpp
+++ b/src/checker_builtin_procs.hpp
@@ -331,6 +331,7 @@ BuiltinProc__type_end,
 	BuiltinProc_objc_find_class,
 	BuiltinProc_objc_register_selector,
 	BuiltinProc_objc_register_class,
+    BuiltinProc_objc_ivar_get,
 
 	BuiltinProc_constant_utf16_cstring,
 
@@ -673,6 +674,7 @@ gb_global BuiltinProc builtin_procs[BuiltinProc_COUNT] = {
 	{STR_LIT("objc_find_class"),        1, false, Expr_Expr, BuiltinProcPkg_intrinsics},
 	{STR_LIT("objc_register_selector"), 1, false, Expr_Expr, BuiltinProcPkg_intrinsics, false, true},
 	{STR_LIT("objc_register_class"),    1, false, Expr_Expr, BuiltinProcPkg_intrinsics, false, true},
+    {STR_LIT("ivar_get"),               2, false, Expr_Expr, BuiltinProcPkg_intrinsics, false, true},
 
 	{STR_LIT("constant_utf16_cstring"), 1, false, Expr_Expr, BuiltinProcPkg_intrinsics},
 
--- a/src/entity.cpp
+++ b/src/entity.cpp
@@ -235,6 +235,9 @@ struct Entity {
 			Type * type_parameter_specialization;
 			String ir_mangled_name;
 			bool   is_type_alias;
+            bool   objc_is_implementation;
+            Type*  objc_superclass;
+            Type*  objc_ivar;
 			String objc_class_name;
 			TypeNameObjCMetadata *objc_metadata;
 		} TypeName;
--- a/src/llvm_backend.cpp
+++ b/src/llvm_backend.cpp
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
 	return p;
 }
 
+// TODO(harold): Move this out of here and into a more suitable place.
+// TODO(harold): Should not take an allocator, but always use temp, as we return string literals as well.
+String lb_get_objc_type_encoding(Type *t, gbAllocator allocator, isize pointer_depth = 0) {
+    // NOTE(harold): See https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html#//apple_ref/doc/uid/TP40008048-CH100
+
+    // NOTE(harold): Darwin targets are always 64-bit. Should we drop this and assume "q" always?
+    #define INT_SIZE_ENCODING (build_context.metrics.ptr_size == 4 ? "i" : "q")
+    switch (t->kind) {
+    case Type_Basic: {
+        switch (t->Basic.kind) {
+        case Basic_Invalid:
+            return str_lit("?");
+
+        case Basic_llvm_bool:
+        case Basic_bool:
+        case Basic_b8:
+            return str_lit("B");
+
+        case Basic_b16:
+            return str_lit("C");
+        case Basic_b32:
+            return str_lit("I");
+        case Basic_b64:
+            return str_lit("q");
+        case Basic_i8:
+            return str_lit("c");
+        case Basic_u8:
+            return str_lit("C");
+        case Basic_i16:
+        case Basic_i16le:
+        case Basic_i16be:
+            return str_lit("s");
+        case Basic_u16:
+        case Basic_u16le:
+        case Basic_u16be:
+            return str_lit("S");
+        case Basic_i32:
+        case Basic_i32le:
+        case Basic_i32be:
+            return str_lit("i");
+        case Basic_u32le:
+        case Basic_u32:
+        case Basic_u32be:
+            return str_lit("I");
+        case Basic_i64:
+        case Basic_i64le:
+        case Basic_i64be:
+            return str_lit("q");
+        case Basic_u64:
+        case Basic_u64le:
+        case Basic_u64be:
+            return str_lit("Q");
+        case Basic_i128:
+        case Basic_i128le:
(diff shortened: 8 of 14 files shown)


harold-b:
After playing around a bit with the last state of things, using an intrinsic manually to access fields with the object's ivar was quite cumbersome.
Having to set the context manually each time also became cumbersome quickly. 

I solved both issues by adding 2 new features to complete the desired functionality:

- Pseudo field support for accessing the Ivar from Objective-C object instances, much like one does in native Objective-C.
- Allow non-class methods to use the Odin calling convention by specifying a context provider procedure via an attribute on the declaration of the class struct.

The former works by simply emitting the `objc_ivar_get` intrinsic during Selector expression resolution for Objective-C structs that have an ivar set. Then replacing the original address with the adjusted one.

The second uses a new attribute `@objc_context_provider` which takes a procedure with a specific signature:
 - It must be "c" calling convention
 - It must have a single parameter, which is either a pointer to the Objective-C object, or its Ivar type.
 - It must return a single value, which is a context

During generation, when a method with Odin calling convention is encountered, the wrapper procedure is adjusted to first set a context by calling the context provider procedure with the 'self' argument. If the self argument is the Ivar type, it first emits an `objc_ivar_get` and passes that instead. The reason for allowing the Ivar type is so that a base struct with a context and any other related data may be used as an embedded field in differing, but related, ObjC Class' Ivar.

Example:

```odin

main :: proc() {

    // Set initial context
    context.user_index = 777
    foo := MyFoo.alloc()->initWithContext(context)

    // Ivar pseudo-field access
    name := foo.name
    fmt.printfln("I am '%v'", name)

    foo.name = "Not_Fooey"
    fmt.printfln("Now I am '%v'", foo.name)

    // Odin calling convention call
    foo->sayHello()
}

@private
get_ctx :: proc "c" ( self: ^MyFooT ) -> runtime.Context {
    return self.ctx
}

@(objc_class="MyFoo", objc_implement, objc_superclass=NS.Object, objc_ivar=MyFooT, objc_context_provider=get_ctx)
MyFoo  :: struct { using _: NS.Object }
MyFooT :: struct {
    ctx:  runtime.Context,
    name: string,
    age:  int,
}

@(objc_type=MyFoo, objc_name="alloc", objc_implement, objc_is_class_method=true)
MyFoo_alloc :: proc "c" () -> ^MyFoo {
    return msgSend(^MyFoo, MyFoo, "alloc")
}

@(objc_type=MyFoo, objc_implement, objc_name="initWithContext")
MyFoo_initWithContext :: proc "c" (self: ^MyFoo, ctx: runtime.Context) -> ^MyFoo {
    // Set initial context via ivar explicitly
    context = ctx
    vself: ^MyFooT = ivar_get(self, MyFooT)
    self.ctx  = ctx
    self.name = "Fooey"
    self.age  = 32
    fmt.printfln("Set name to %v", self.name)

    return self
}

@(objc_type=MyFoo, objc_implement, objc_name="sayHello")
MyFoo_sayHello :: proc (self: ^MyFoo) {
    fmt.printfln("Hello, I am %v :: %v", self.name, context.user_index)
}
```

Lperlind:
I am guessing this PR could completely replace the code written such as Application delegate that created a class during runtime

https://github.com/odin-lang/Odin/blob/master/core%2Fsys%2Fdarwin%2FFoundation%2FNSApplication.odin#L253

harold-b:
Indeed. And it facilitates being able to have more than 1 protocol implemented in a single class, as well as replacing base class functionality in the hierarchy. You could do the latter with [darwodin](https://github.com/harold-b/darwodin) but it's painful and bloated.

There are also cases where you need to provide a separate object with a specific selector, such as when using GestureRecognizers, and this would have meant generating it all and registering it at runtime manually.

Lperlind:
That will be super cool. I would mark the those procedures as deprecated and point people to this stuff instead

src/check_builtin.cpp
```diff
@@ -387,6 +387,80 @@ gb_internal bool check_builtin_objc_procedure(CheckerContext *c, Operand *operan
...
+
+        if (!is_operand_value(self) || !check_is_assignable_to(c, &self, t_objc_id)) {
+            gbString e = expr_to_string(self.expr);
+            gbString t = type_to_string(self.type);
+            error(self.expr, "'%.*s' expected a type or value derived from intrinsics.objc_object, got '%s' of type %s", LIT(builtin_name), e, t);
+            gb_string_free(t);
+            gb_string_free(e);
+            return false;
+        }
+        else if (!is_type_pointer(self.type)) {
```
gingerBill, line 407: Prefer `} else if` style rather than that Bjarne style

src/check_builtin.cpp
```diff
@@ -387,6 +387,80 @@ gb_internal bool check_builtin_objc_procedure(CheckerContext *c, Operand *operan
 		try_to_add_package_dependency(c, "runtime", "objc_allocateClassPair");
 		return true;
 	} break;
+
+    case BuiltinProc_objc_ivar_get:
```
gingerBill, line 391: Use tabs for indentation, not spaces.

src/check_decl.cpp
```diff
@@ -526,6 +526,68 @@ gb_internal void check_type_decl(CheckerContext *ctx, Entity *e, Ast *init_expr,
...
+                // Ensure superclass hierarchy are all Objective-C classes and does not cycle
+                Type *super = ac.objc_superclass;
+                if (super != nullptr) {
+                    TypeSet super_set{};
+                    type_set_init(&super_set, 8);
+                    defer (type_set_destroy(&super_set));
+
+                    type_set_update(&super_set, e->type);
+
+                    for (;;) {
```
gingerBill, line 555: `while (super != nullptr) {`

src/check_decl.cpp
```diff
@@ -526,6 +526,68 @@ gb_internal void check_type_decl(CheckerContext *ctx, Entity *e, Ast *init_expr,
...
+                        }
+
+                        super = super->Named.type_name->TypeName.objc_superclass;
+                        if (super == nullptr) {
+                            break;
+                        }
+
+                        // TODO(harold): Is this the right way to do this??? The referenced entity must be already resolved
+                        //               so that we can access its objc_superclass attribute
+                        check_single_global_entity(ctx->checker, super->Named.type_name, super->Named.type_name->decl_info);
```
gingerBill, line 579: Checking it again should not be a problem, but is this what you wanted since you reassigned `super` above. This seems a bit unsafe.

src/checker.cpp
```diff
@@ -1351,10 +1351,12 @@ gb_internal void init_universal(void) {
 		t_objc_object   = add_global_type_name(intrinsics_pkg->scope, str_lit("objc_object"),   alloc_type_struct_complete());
 		t_objc_selector = add_global_type_name(intrinsics_pkg->scope, str_lit("objc_selector"), alloc_type_struct_complete());
 		t_objc_class    = add_global_type_name(intrinsics_pkg->scope, str_lit("objc_class"),    alloc_type_struct_complete());
+        t_objc_ivar     = add_global_type_name(intrinsics_pkg->scope, str_lit("objc_ivar"),     alloc_type_struct_complete());
```
gingerBill, line 1354: Use tabs for indentation.

src/checker.cpp
```diff
@@ -6389,6 +6481,47 @@ gb_internal void check_deferred_procedures(Checker *c) {
...
+	for (Entity *e = nullptr; mpsc_dequeue(&c->procs_with_objc_context_provider_to_check, &e); /**/) {
+		GB_ASSERT(e->kind == Entity_TypeName);
+
+		Entity *proc_entity = e->TypeName.objc_context_provider;
+		GB_ASSERT(proc_entity->kind == Entity_Procedure);
+
+		Type *proc_type = proc_entity->type;
+
+		// TODO(harold): Give better errors here (specify exactly what's wrong)
+		const char* signature_error = "The procedure for @(objc_context_provider) has an incorrect signature.";
```
gingerBill, line 6494: `char const *signature_error = ...` to be consistent

src/checker.hpp
```diff
@@ -148,8 +148,13 @@ struct AttributeContext {
 
 	String  objc_class;
 	String  objc_name;
-	bool    objc_is_class_method;
+    String  objc_selector;
```
gingerBill, line 151: Use tabs for indentation.

src/checker.hpp
```diff
@@ -148,8 +148,13 @@ struct AttributeContext {
...
 	Type *  objc_type;
+    Type *  objc_superclass;
+    Type *  objc_ivar;
+	Entity *objc_context_provider;
+	bool    objc_is_class_method   : 1;
```
gingerBill, line 156: Only use bit fields when you have enough to minimize padding. This isn't worth it.

src/checker_builtin_procs.hpp
```diff
@@ -331,6 +331,7 @@ BuiltinProc__type_end,
 	BuiltinProc_objc_find_class,
 	BuiltinProc_objc_register_selector,
 	BuiltinProc_objc_register_class,
+    BuiltinProc_objc_ivar_get,
```
gingerBill, line 334: Use tabs for indentation.

src/entity.cpp
```diff
@@ -235,6 +235,10 @@ struct Entity {
 			Type * type_parameter_specialization;
 			String ir_mangled_name;
 			bool   is_type_alias;
+            bool   objc_is_implementation;
```
gingerBill, line 238: Use tabs for indentation.

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
 	return p;
 }
 
+// TODO(harold): Move this out of here and into a more suitable place.
+// TODO(harold): Should not take an allocator, but always use temp, as we return string literals as well.
+String lb_get_objc_type_encoding(Type *t, gbAllocator allocator, isize pointer_depth = 0) {
```
gingerBill, line 1178: Use tabs for indentation.

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
...
+            return build_context.metrics.ptr_size == 4 ? str_lit("I") : str_lit("Q");
+        case Basic_uintptr:
+        case Basic_rawptr:
+            return str_lit("^v");
+
+        case Basic_string:
+            return build_context.metrics.ptr_size == 4 ? str_lit("{string=*i}") : str_lit("{string=*q}");
+
+        case Basic_cstring: return str_lit("*");
+        case Basic_any:     return str_lit("{any=^v^v");  // rawptr + ^Type_Info
```
gingerBill, line 1272: Missing `}`?

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
...
+        case Basic_f32:
+        case Basic_f32le:
+        case Basic_f32be:
+            return str_lit("f");
+        case Basic_f64:
+        case Basic_f64le:
+        case Basic_f64be:
+            return str_lit("d");
+
+        // TODO(harold) These:
```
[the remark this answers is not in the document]
gingerBill: So would it be? `{complex64=ff}`, etc?

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
...
+
+        args.count = 3;
+        args[0] = superclass == nullptr ? lb_const_nil(m, t_objc_Class) : superclass_value;
+        args[1] = class_name;
+        args[2] = lb_const_int(m, t_uint, 0);
+        class_ptr = lb_emit_runtime_call(p, "objc_allocateClassPair", args);
+
+        array_add(&class_impls, lbObjCGlobalClass{g, class_ptr});
+    }
+    else {
```
gingerBill, line 1488: Don't use Bjarne style clauses please.

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
...
+    lbValue class_ptr{};
+    lbValue class_name = lb_const_value(m, t_cstring, exact_value_string(g.name));
+
+    // If this class requires an implementation, save it for registration below.
+    if (g.class_impl_type != nullptr) {
+
+        // Make sure the superclass has been initialized before us
+        lbValue superclass_value{};
+
+        auto& tn = g.class_impl_type->Named.type_name->TypeName;
```
gingerBill, line 1470: `auto &tn = ...`

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
...
+    } else {
+        lbValue v = {};
+        LLVMTypeRef t = lb_type(m, g.type);
+        v.value = LLVMAddGlobal(m->mod, t, g.global_name);
+        v.type = alloc_type_pointer(g.type);
+        addr = lb_addr(v);
+        LLVMSetInitializer(v.value, LLVMConstNull(t));
+    }
+
+    lbValue class_ptr{};
```
gingerBill, line 1461: `lbValue class_ptr = {};` is the more common style convention.

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
...
+    lbValue class_name = lb_const_value(m, t_cstring, exact_value_string(g.name));
+
+    // If this class requires an implementation, save it for registration below.
+    if (g.class_impl_type != nullptr) {
+
+        // Make sure the superclass has been initialized before us
+        lbValue superclass_value{};
```
gingerBill, line 1468:
```
lbValue superclass_value = lb_const_nil(m, t_objc_Class);
```

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
...
+        if (superclass != nullptr) {
+            auto& superclass_global = string_map_must_get(&class_map, superclass->Named.type_name->TypeName.objc_class_name);
+            lb_register_objc_thing(handled, m, args, class_impls, class_map, p, superclass_global.g, call);
+            GB_ASSERT(superclass_global.class_value.value);
+
+            superclass_value = superclass_global.class_value;
+        }
+
+        args.count = 3;
+        args[0] = superclass == nullptr ? lb_const_nil(m, t_objc_Class) : superclass_value;
```
gingerBill, line 1481: See the default initialization above.
```
args[0] = superclass_value;
```

src/llvm_backend.cpp
```diff
@@ -1186,38 +1512,308 @@ gb_internal void lb_finalize_objc_names(lbGenerator *gen, lbProcedure *p) {
...
+    		is_context_provider_ivar = ivar_type != nullptr && internal_check_is_assignable_to(contex_provider_self_named_type, ivar_type);
+    	}
+
+
+        Array<ObjcMethodData>* methods = map_get(&m->info->objc_method_implementations, class_type);
+        if (!methods) {
+            continue;
+        }
+
+        for (const ObjcMethodData& md : *methods) {
```
gingerBill, line 1630:
```cpp
for (ObjcMethodData const &md : *methods) {
```

src/llvm_backend.cpp
```diff
@@ -1186,38 +1512,308 @@ gb_internal void lb_finalize_objc_names(lbGenerator *gen, lbProcedure *p) {
...
+            String proc_name = make_string_c("__$objc_method::");
+            proc_name = concatenate_strings(temporary_allocator(), proc_name, g.name);
+            proc_name = concatenate_strings(temporary_allocator(), proc_name, str_lit("::"));
+            proc_name = concatenate_strings( permanent_allocator(), proc_name, md.ac.objc_name);
+
+            wrapper_args.count = 2;
+            wrapper_args[0] = md.ac.objc_is_class_method ? t_objc_Class : class_ptr_type;
+            wrapper_args[1] = t_objc_SEL;
+
+            auto method_param_count  = (isize)method_type->Proc.param_count;
```
gingerBill, line 1643:
```
isize method_param_count = method_type->Proc.param_count;
```

src/llvm_backend.cpp
```diff
@@ -1186,38 +1512,308 @@ gb_internal void lb_finalize_objc_names(lbGenerator *gen, lbProcedure *p) {
...
+            i32  method_param_offset = 0;
```
gingerBill, line 1644: Why isn't this just `isize` to begin with?

src/llvm_backend.cpp
```diff
@@ -1186,38 +1512,308 @@ gb_internal void lb_finalize_objc_names(lbGenerator *gen, lbProcedure *p) {
...
+        } // End methods
+
+        // Add ivar if we have one
+        if (ivar_type != nullptr) {
+            // Register a single ivar for this class
+            Type *ivar_base = ivar_type->Named.base;
+
+            const i64 size      = type_size_of(ivar_base);
+            const i64 alignment = type_align_of(ivar_base);
+            // TODO(harold): Checker: Alignment must be compatible with ivar rules. Or we should increase the alignment if needed.
```
gingerBill, line 1761: If an `ivar` needs to have a minimum alignment, I'd enforce it on the declaration, if possible.

src/llvm_backend_proc.cpp
```diff
@@ -3290,6 +3290,7 @@ gb_internal lbValue lb_build_builtin_proc(lbProcedure *p, Ast *expr, TypeAndValu
 	case BuiltinProc_objc_find_class:        return lb_handle_objc_find_class(p, expr);
 	case BuiltinProc_objc_register_selector: return lb_handle_objc_register_selector(p, expr);
 	case BuiltinProc_objc_register_class:    return lb_handle_objc_register_class(p, expr);
+    case BuiltinProc_objc_ivar_get:          return lb_handle_objc_ivar_get(p, expr);
```
gingerBill, line 3293: Use tabs for indentation.

src/llvm_backend_utility.cpp
```diff
@@ -2148,13 +2148,75 @@ gb_internal lbAddr lb_handle_objc_find_or_register_class(lbProcedure *p, String
...
+	mpsc_enqueue(&m->gen->objc_classes, lbObjCGlobal{m, global_name, name, t_objc_Class, class_impl_type});
 
 	lbAddr addr = lb_addr(g);
 	string_map_set(&m->objc_classes, name, addr);
 	return addr;
 }
 
+gb_internal lbAddr lb_handle_objc_find_or_register_ivar(lbModule *m, Type *self_type) {
+
+    String name = self_type->Named.type_name->TypeName.objc_class_name;
```
gingerBill, line 2160: Use tabs for indentation.

src/llvm_backend.cpp
harold-b: It probably should be. I did it to save space a tiny bit of space in the executable, but perhaps its not worth it, or there may be, in the unlikely case, classes with sizes larger than 4GiB. I'll fix it in the next round.

harold-b:
# Objective-C Classes Checker Specification
This is what I've come up with for the checker specification. Most of these are probably already handled, but I'll go through them carefully. I also need to ensure they are handled in the correct location.

Any comments and feedback is welcomed.

@(objc_implement):
- On structs, if set to true:
	- [ ] Must either have no value (defaults to `true`, or if it does, it must be a bool value.
	- [ ] Struct must have a @(objc_class) attribute.
	- [ ] Class name must not already exist (Already handled for us?).

- On procedures, if set to true:
	- [ ] Must either have no value (defaults to `true`, or if it does, it must be a bool value.
	- [ ] Must have @(objc_type) attribute.
	- [ ] Struct pointed to by @(objc_type) must have the @(objc_implement) attribute.
	- [ ] If the @(objc_is_class_method) attribute is not set to true, then the procedure MUST the first parameter be a pointer the the struct specified in @(objc_type)
	- [ ] The procedure MUST return at most 1 value (or no value at all).
	- [ ] If the procedure uses the Odin calling convention, the struct specified in @(objc_type) MUST have satisfy  @(objc_context_provider) attribute.
	- [ ] If the procedure has the @(objc_is_class_method) attribute, then the calling convention must always be "c", unless the @(objc_context_provider) points to a procedure with no arguments.

@(objc_superclass):
- [ ] The struct must satisfy the @(objc_class) attribute
- [ ] The value must be set to a type identifier that also satisfies the @(objc_class) attribute.
- [ ] No hierarchy cycles must be present.

@(objc_context_provider):
- [ ] The struct must satisfy the @(objc_class) attribute
- [ ] The value must point to a procedure with the correct signature:
	- [ ] The procedure must return a `runtime.Context`.
	- [ ] The procedure must have the "c" or "contextless" calling convention.
	- [ ] The procedure must take either 1 parameter or no parameter. If it takes 1 parameter, it must be one of the following:
		- A pointer to the declaring struct.
		- A pointer to the ivar type of the declaring struct, ONLY IF the @(objc_ivar) attribute is satisfied.
	- [ ] If the procedure takes a parameter, only instance methods, that is methods without the @(objc_is_class_method) attribute, can be declared with the Odin calling convention.

@(objc_ivar):
- [ ] The struct must satisfy the @(objc_class) attribute.
- [ ] The value must be set to a named type (should it be restricted only to a struct?).

@(objc_selector):
- [ ] The proc must satisfy the @(objc_type) attribute.
- [ ] The value must be a string.


intrinsics.obj_get_ivar():
- [ ] Must be called with 1 argument only:
- [ ] The argument must be a pointer to an objective-c struct which satisfies the  @(objc_ivar) attribute.
- [ ] NOTE: We can remove the second argument (which it currently takes). We can satisfy it by ourselves in the compiler. (For a better experience I'll have to do more work on ols to be able to infer the return type, which should be fine.)

base/intrinsics/intrinsics.odin
```diff
@@ -353,15 +353,18 @@ x86_xgetbv :: proc(cx: u32) -> (eax, edx: u32) ---
...
 objc_SEL   :: ^objc_selector
 objc_Class :: ^objc_class
+objc_Ivar  :: ^objc_ivar
 
 objc_find_selector     :: proc($name: string) -> objc_SEL   ---
 objc_register_selector :: proc($name: string) -> objc_SEL   ---
 objc_find_class        :: proc($name: string) -> objc_Class ---
 objc_register_class    :: proc($name: string) -> objc_Class ---
-
+ivar_get			   :: proc(self: ^$T, $U: typeid) -> ^U ---
```
laytan, line 367: This should be prefixed with `objc_` imo

laytan (commented):
I am not sure about the context provider functionality, it seems a bit too much, I also wonder if having the psuedo field support already is a good solution to the problem?

```odin
@(objc_class="MyFoo", objc_implement, objc_superclass=NS.Object, objc_ivar=MyFooT)
MyFoo  :: struct { using _: NS.Object }
MyFooT :: struct {
    ctx:  runtime.Context,
}

@(objc_type=MyFoo, objc_implement, objc_name="sayHello")
MyFoo_sayHello :: proc "c" (self: ^MyFoo) {
    context = self.ctx // this should use the psuedo field and work right?
    fmt.printfln("Hello, I am %v :: %v", self.name, context.user_index)
}
```

that does not seem bad to me, and "feels" like it fits better.

And then an observation I have, all the procs in your examples are annotated with `@obc_implement`, is that necessary, can't you check if the proc's struct is `@objc_implement`?  Or is this just your examples and are there cases you want to attach a proc that isn't an implement?

base/intrinsics/intrinsics.odin
harold-b: I agree. I reverted to that already locally, but haven't pushed yet.

harold-b:
> I am not sure about the context provider functionality, it seems a bit too much, I also wonder if having the psuedo field support already is a good solution to the problem?
>
> that does not seem bad to me, and "feels" like it fits better.

I really like the ease of use the auto-context provides. After changing all the classes in my personal project, it became tiresome to have to specify the context each time, so I came up with this solution. I am hoping to to make writing Objective-C in odin mostly nicer than native Objective-C (for those that don't use ARC or property lifetime semantics). I think not having to preface your body with an explicit context each time, if one does not want to, adds to that goal. It's optional, of course, you *don't have to* provide one. In which case the checker errors if you didn't use C calling conv.

> And then an observation I have, all the procs in your examples are annotated with `@obc_implement`, is that necessary, can't you check if the proc's struct is `@objc_implement`? Or is this just your examples and are there cases you want to attach a proc that isn't an implement?

Having to write this each time is something I found bothersome too, so I like your suggestion. `@objc_implement` can default to `true` if the `@objc_type` already has `@objc_implement` enabled. If the user wants a `proc` to not be  `@objc_implement`, then he can explicitly set it to false: `@objc_implement=false`. That would simplify things plenty.

We still need to allow the `@objc_implement=false` option for procs like `alloc` which need to fallback to `msgSend` (if `@objc_implement=false` it would recurse and overflow).

I'll update the checker spec with that change.

laytan:
> I really like the ease of use the auto-context provides. After changing all the classes in my personal project, it became tiresome to have to specify the context each time, so I came up with this solution. I am hoping to to make writing Objective-C in odin mostly nicer than native Objective-C (for those that don't use ARC or property lifetime semantics). I think not having to preface your body with an explicit context each time, if one does not want to, adds to that goal. It's optional, of course, you don't have to provide one. In which case the checker errors if you didn't use C calling conv.

I get your point, but I don't think one very simple assignment is much of a hassle.

It is also much clearer where the context is coming from, if you don't see it being assigned it is always coming from the caller atm, which this breaks.

Also from the pov of the caller, you are calling an "odin" calling convention procedure, but it doesn't get the context from your scope, it gets a magic context from some other place.

From your own example, changed a bit:

```odin

// Set initial context
context.user_index = 777

foo := MyFoo.alloc()->initWithContext(context)

context.user_index = 666
// Odin calling convention call
foo->sayHello()
```

In any other code, and what you expect, that will print 666, but it prints 777. Having that made clear by it being a "c" calling convention proc, and seeing that the body of the proc first sets a context, is imo quite valuable.

---

I noticed @gingerBill didn't say anything about it and he ultimately decides, but this is my pov.

harold-b:
> I get your point, but I don't think one very simple assignment is much of a hassle.
> 
> It is also much clearer where the context is coming from, if you don't see it being assigned it is always coming from the caller atm, which this breaks.
> 
> Also from the pov of the caller, you are calling an "odin" calling convention procedure, but it doesn't get the context from your scope, it gets a magic context from some other place.
> ...

That's a good point...
As a counterpoint, the user has to explicitly opt-in to having the context provided via the attribute and procedure. Plus 99% of the time one would expect these procedures to be called from the Objective-C runtime (where no context exists), not via direct dispatch from odin itself.

But you've raised a very good point, I have to think about this a bit more... And yes, Bill's input and decision here is important.

harold-b:
I have more thoughts on this, and still have to think about it more. But my preliminary conclusion is that all selector call expressions on an `@objc_implement` method should resolve to Objective-C dispatch always (`msgSend`), never direct dispatch. This way there is always only one predictable path. If the user happens to need direct dispatch for performance reasons, then he should have an unwrapped version of the `proc` that is callable via direct dispatch, or even a `@objc_implement=false` version.

This also would work in favor of being able to provide the `objc_super()` intrinsic later.

src/llvm_backend_expr.cpp
```diff
@@ -5174,18 +5172,43 @@ gb_internal lbAddr lb_build_addr_internal(lbProcedure *p, Ast *expr) {
...
+
+				// If it's a deep pointer, dereference it first
+				// TODO(harold): Ensure this is save to do here. lb_emit_deep_field_gep() has several derefs, once per index.
+				//				 Not sure what multiple indices represent...
+				Type* type = type_deref(addr.addr.type);
+
+				// TODO(harold): Checker: Must NOT allow ivar dereferencing on non-pointer types.
+				//					      this would access memory outside the size of the value.
+				//						  In fact, locals/globals of Objective-C types ought not be allowed at all.
+				GB_ASSERT(is_type_pointer(type));
```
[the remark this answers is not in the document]
harold-b: I can't remember exactly. I think I had to redo this at some point when I realized it was always a deep pointer (`^NSObject` -> `^objc_object`) and had that left over. I think its enough to just assert its a deep pointer and then deref twice unconditionally. Any other state should be invalid and handled during checker for the intrinsic.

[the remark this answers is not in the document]
harold-b: Actually I lost context here after a couple of weeks of not touching it. Let me get back into it and ensure I understand what's going on here again.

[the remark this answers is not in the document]
harold-b: Scratch the previous nonsense comment. I recalled how this worked and what was going on. Had leftover garbage there as I was working through it at the time.

It's fixed now, only allows indirect access, for ivar pseudofields.

src/check_decl.cpp
harold-b: Fixed—But is this the correct location for this check?

base/intrinsics/intrinsics.odin
harold-b: It seems I was wrong. I had renamed it only in the docs. Fixed now.

src/llvm_backend.cpp
harold-b: I revisited Apple's docs: https://developer.apple.com/documentation/objectivec/class_addivar(_:_:_:_:_:)?language=objc

> The instance variable’s minimum alignment in bytes is 1<<align. The minimum alignment of an instance variable depends on the ivar’s type and the machine architecture. For variables of any pointer type, pass log2(sizeof(pointer_type)).

I think in that word salad they meant to say to simply pass the log2 of the alignment, not the actual alignment, which after testing seems to be correct. I validated it by recreating my test case directly in Objective-C and inspecting the ivar offset at runtime.

Fixed now

src/llvm_backend.cpp
harold-b: Sorry, I misread here. I thought this referred to the size of the ivar offset global.
Which I changed to native size here: https://github.com/odin-lang/Odin/pull/5064/commits/1505edef011949e10b31335577ea344e5da2b623

Should I revert that or allow sizes above 32-bit address space?

harold-b: Probably ok to leave the ivar offset as native word size. But let me know if you differ.

I fixed the original issue now as well.

harold-b:
All checker spec implemented except for ` Class name must not already exist (Already handled for us?).`

I don't know if this is already handled somewhere since that attribute predates this PR.

I will try to finish that off tomorrow.

What's left is unconditional exporting and TODOs cleanup/clarifications.

harold-b:
>  that all selector call expressions on an @objc_implement method should always resolve to Objective-C dispatch (msgSend), never direct dispatch.

I am leaving what I mentioned there as outside of the scope of this PR as I'd like to not keep increasing the scope to an already massive PR. Hopefully we can discuss more at length later and agree on what is to be done.

harold-b:
> All checker spec implemented except for ` Class name must not already exist (Already handled for us?).`
> 
> I don't know if this is already handled somewhere since that attribute predates this PR.
> 
> I will try to finish that off tomorrow.
> 
> What's left is unconditional exporting and TODOs cleanup/clarifications.

I've implemented both. There's now a class name re-use check  for all classes using `@objc_implement`.

I've moved setting the implicit export and linkage to a better location outside of the attribute resolution stage, where it won't get overwritten.

All that's left is clearing up a couple of minor TODOs

harold-b:
This is now ready for review.

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
 	return p;
 }
 
+String lb_get_objc_type_encoding(Type *t, isize pointer_depth = 0) {
+	// NOTE(harold): See https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html#//apple_ref/doc/uid/TP40008048-CH100
+
+	// NOTE(harold): Darwin targets are always 64-bit. Should we drop this and assume "q" always?
+	#define INT_SIZE_ENCODING (build_context.metrics.ptr_size == 4 ? "i" : "q")
```
gingerBill, line 1180: `#define INT_SIZE_ENCODING (build_context.metrics.int_size == 4 ? "i" : "q")`

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
...
+		case Basic_int:
+			return str_lit(INT_SIZE_ENCODING);
+		case Basic_uint:
+			return build_context.metrics.ptr_size == 4 ? str_lit("I") : str_lit("Q");
+		case Basic_uintptr:
+		case Basic_rawptr:
+			return str_lit("^v");
+
+		case Basic_string:
+			return build_context.metrics.ptr_size == 4 ? str_lit("{string=*i}") : str_lit("{string=*q}");
```
gingerBill, line 1265: `int_size` not `ptr_size`

src/llvm_backend.cpp
```diff
@@ -1173,6 +1173,332 @@ gb_internal lbProcedure *lb_create_objc_names(lbModule *main_module) {
...
+	case Type_Proc:
+		return str_lit("?");
+	case Type_BitSet:
+		return lb_get_objc_type_encoding(t->BitSet.underlying, pointer_depth);
+	case Type_SimdVector:
+		break;
+	case Type_Matrix:
+		break;
+	case Type_BitField:
+		return lb_get_objc_type_encoding(t->BitField.backing_type, pointer_depth);
```
gingerBill, line 1415: You could do the `b`num thing but it's probably not necessary.

harold-b: I left it up to the backing type simply because I did not know if there are any semantic differences/and or limits between the expected bitfields from the ObjC-side But this pointed me to the missing Matrix and SimdVector types, which I fixed.

harold-b:
> I'm not a huge fan of the `@(objc_context_provider)` idea since it seems kind of unnecessary if the concept of an `ivar` already exists. It isn't clear when or where this context callback is being invoked, compared doing a manual `context = self.ctx`.

That's unfortunate. It is certainly useful when filling out large protocols or implementing full classes completely in Odin.

I modeled it after the `deferred_in/out` mechanism——which one could argue is also unclear as to where they are invoked/how it works barring documentation. I understand there's distinction here as we're also talking about context propagation, and completely see where you guys' concern is coming from.

What do you think about having calls to ObjC methods be ObjC-runtime dispatch only as I specified [here](https://github.com/odin-lang/Odin/pull/5064#issuecomment-2843142456)? Meaning, whenever these methods are invoked directly from the Odin side (which should be the rare case), they would unconditionally be dispatched via `msgSend(...)`, that is, through ObjC runtime dispatch. This way the context is **always** coming from the `@objc_context_provider` proc.

gingerBill:
I'm merging this PR even if I still not sure about the `@(objc_context_provider)` attribute.

PR #5066, opened 2025-04-22
Fill in missing build flags for `core/sys/windows` package

ekliot:
These files are triggering compilation errors on my Linux machine, which are fixed with these build flags.

Thanks for reviewing.

### DIFF
--- a/core/sys/windows/ole32.odin
+++ b/core/sys/windows/ole32.odin
@@ -1,3 +1,4 @@
+#+build windows
 package sys_windows
 
 foreign import "system:Ole32.lib"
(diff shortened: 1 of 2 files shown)


ekliot:
Okay, so I've sourced my _real_ issue: `vendor/vulkan/structs.odin`

I copied it into `vendor/vulkan/structs_windows.odin`, then removed `win32` and such from `structs.odin`, and added a `#+build !windows` tag to the file.

My wasm build went normally, without this PR's patch (so, the only changes to the source tree was `structs.odin`.

However, that file is autogenerated, so hacking this seems unwise. Please advise.

PR #5081, opened 2025-04-27
Add -vet-explicit-allocators

Lperlind:
This vet flag will make it so that allocators must be explicitly used in places where context.allocator and context.temp_allocator are a procedure parameter.

The goal of this flag is to prevent using the context.allocator in cases where a different allocator was meant to be used. Some code bases default context.allocator to nil/panic allocator to catch this at runtime. This effectively makes it a compile time error instead.

This change intentionally just targets `context.temp_allocator` and `context.allocator` rather than all `runtime.Allocator` default values as I feel that if an explicit value is set that is not `context` then some much more intentional default value was intended.

Example code:
```odin
#+vet explicit-allocators
package alloc

import "core:fmt"

foo :: proc(allocator := context.allocator) {
}

foo_temp :: proc(allocator := context.temp_allocator) {
}

main :: proc() {
    foo(context.allocator)
    foo_temp(context.allocator)
    result := fmt.aprintf("hello %v", "world", allocator = context.allocator)

    foo() // Error: Parameter 'allocator' of type 'Allocator' must be explicitly provided in procedure call
    foo_temp() // Error: Parameter 'allocator' of type 'Allocator' must be explicitly provided in procedure call
    result := fmt.aprintf("hello %v", "world") // Error: Parameter 'allocator' of type 'Allocator' must be explicitly provided in procedure call
}
```

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -305,6 +305,7 @@ enum VetFlags : u64 {
 	VetFlag_Cast            = 1u<<8,
 	VetFlag_Tabs            = 1u<<9,
 	VetFlag_UnusedProcedures = 1u<<10,
+	VetFlag_ExplicitAllocators = 1u<<11,
 
 	VetFlag_Unused = VetFlag_UnusedVariables|VetFlag_UnusedImports,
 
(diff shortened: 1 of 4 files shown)


gingerBill:
The idea and implementation is good.

My issue is that `-vet-explicit-allocators` is pretty much never going to be a good thing in practice since loads of core library will break. So `#+vet explicit-allocators` is probably one of the main use cases of it.

herohiralal:
are there any plans to merge this?

tbh i'd have a bunch of use for this, and especially the commandline flag as well

one way to use it is to just add it to OLS's configuration but not have it as a part of the actual compilation process

call it a skill issue, but i have definitely messed up the whole new<->free and make<->delete thing by calling them with different allocators just because the whole thing compiles with implicit context allocators

PR #5088, opened 2025-04-29
vendor:wasm/WebGL: Fix incorrect parameter types

leecommamichael:
I was playing with WebGL today, noticed there were some errant Enums.

### DIFF
--- a/vendor/wasm/WebGL/webgl.odin
+++ b/vendor/wasm/WebGL/webgl.odin
@@ -61,10 +61,10 @@ foreign webgl {
 	BufferData    :: proc(target: Enum, size: int, data: rawptr, usage: Enum) ---
 	BufferSubData :: proc(target: Enum, offset: uintptr, size: int, data: rawptr) ---
 
-	Clear         :: proc(bits: Enum) ---
+	Clear         :: proc(bits: u32) ---
 	ClearColor    :: proc(r, g, b, a: f32) ---
-	ClearDepth    :: proc(x: Enum) ---
-	ClearStencil  :: proc(x: Enum) ---
+	ClearDepth    :: proc(x: f32) ---
+	ClearStencil  :: proc(x: int) ---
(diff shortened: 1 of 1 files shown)


vendor/wasm/WebGL/webgl.odin
```diff
@@ -61,10 +61,10 @@ foreign webgl {
 	BufferData    :: proc(target: Enum, size: int, data: rawptr, usage: Enum) ---
 	BufferSubData :: proc(target: Enum, offset: uintptr, size: int, data: rawptr) ---
 
-	Clear         :: proc(bits: Enum) ---
+	Clear         :: proc(bits: u32) ---
 	ClearColor    :: proc(r, g, b, a: f32) ---
-	ClearDepth    :: proc(x: Enum) ---
-	ClearStencil  :: proc(x: Enum) ---
+	ClearDepth    :: proc(x: f32) ---
+	ClearStencil  :: proc(x: int) ---
```
gingerBill, line 67: Is it `int` or `i32`?

leecommamichael: https://registry.khronos.org/webgl/specs/latest/1.0/ has it as GLbitfield which is unsigned long, so I'd say int.

leecommamichael: https://registry.khronos.org/webgl/specs/latest/1.0/ has it as GLint, which I see marked up as `long` on that page. Assuming ILP32/LP64 I'd say "register size" (int) is good.

leecommamichael: I've updated it. Seems to be a bug in WebGL's spec. I've looked up recent standards for GLES and GL and they all agree with you.

> Table 2.2: GL data types. GL types are not C types. Thus, for example, GL
type int is referred to as GLint outside this document, and is not necessarily
equivalent to the C type int. An implementation must use exactly the number of
bits indicated in the table to represent a GL type

PR #5092, opened 2025-04-30
Revert "fix not resolving to alias in a recursive declaration"

### LINKED ISSUE #5061: Typedef of a struct pointer placed before the struct that contains a field of this type crashes the compiler
## Context

```
        Odin:    dev-2025-04:ab5ca087a
        OS:      Windows 10 Professional (version: 22H2), build 19045.5737
        CPU:     AMD Ryzen 9 5950X 16-Core Processor
        RAM:     65447 MiB
        Backend: LLVM 20.1.0
```

## Expected Behavior

The code compiles without problems, or, if this is disallowed intentionally, a nice error message is printed.

## Current Behavior

The compiler asserts:
```
C:\dev\Odin\src\llvm_backend_general.cpp(1625): Assertion Failure: `type != t_invalid`
```

The code below does not compile and crashes the compiler in the provided form.

When the typedef is placed **after** the struct definition it compiles just fine.
When the field's type is replaced with `^Node`, the code also compiles.

I believe it worked fine before, because I found this in a freetype bindings package, which I used for a while without problems and **did not update it** at all (so the code must have stayed the same). The only thing that I updated was my Odin installation, however, unfortunately I don't know which version/commit broke it.

### Steps to Reproduce

1. Run this code normally ("odin run .")
```
package main

Node_Ptr :: ^Node // the typedef...

Node :: struct {
	prev, next: Node_Ptr, // replacing the type with ^Node also fixes it
}

// ...if placed here, it works just fine

main :: proc() {
	node: Node_Ptr
}
```

graphitemaster:
Reverts odin-lang/Odin#5045 and fixes #5061

### DIFF
--- a/src/check_decl.cpp
+++ b/src/check_decl.cpp
@@ -468,10 +468,6 @@ gb_internal void check_type_decl(CheckerContext *ctx, Entity *e, Ast *init_expr,
 	}
 	e->type = named;
 
-	if (!is_distinct) {
-		e->TypeName.is_type_alias = true;
-	}
-
 	check_type_path_push(ctx, e);
 	Type *bt = check_type_expr(ctx, te, named);
 	check_type_path_pop(ctx);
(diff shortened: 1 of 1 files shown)


graphitemaster:
See referenced PR, issue for rationale for this.

laytan:
Don't think we should be reverting, I can look at a proper fix later today

PR #5096, opened 2025-04-30
os2: Fix reading directory on Windows sometimes failing to decode a file name following one with exotic unicode characters

jdennis9:
Windows doesn't zero the cFileName buffer in FindFirstFile, so there are sometimes extra characters after the null-terminator. Passing the whole buffer to WideCharToMultiByte causes it to try decode these extra characters which might be messed up by the new null-terminator and it fails to decode.

### DIFF
--- a/core/os/os2/dir_windows.odin
+++ b/core/os/os2/dir_windows.odin
@@ -14,7 +14,7 @@ find_data_to_file_info :: proc(base_path: string, d: ^win32.WIN32_FIND_DATAW, al
 	if d.cFileName[0] == '.' && d.cFileName[1] == '.' && d.cFileName[2] == 0 {
 		return
 	}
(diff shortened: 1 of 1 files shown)


Kelimion:
Can you give an example of what kind of exotic runes should be part of a previous filename for the original to fail?

jdennis9:
> Can you give an example of what kind of exotic runes should be part of a previous filename for the original to fail?

In my case I had a file with the characters: 𝙐𝙣𝙙𝙚𝙧𝙯𝙤𝙣𝙚_𝙐𝙕 𝙋𝙧𝙚𝙢𝙞𝙚𝙧𝙚
I was to reproduce the issue using a file name just containing those characters, followed by shorter file names

Kelimion:
Thanks. I pondered whether it might be better to zero out the find_data at the call site after conversion to utf8, but that's about the same amount of work as finding the zero in this approach.

PR #5108, opened 2025-05-03
Alternate `reduce_add`/`reduce_mul` intrinsics

Barinzaya:
The new `reduce_add`/`reduce_mul` intrinsics perform the corresponding arithmetic reduction, in different orders than the sequential order of the existing `reduce_add_ordered`. These alternative orders can often be faster to calculate for floats, as they can offer better SIMD hardware utilization when the compiler is strict about the order of addition.

Two different orders are added for these: pairwise (operating on pairs of adjacent elements) or bisecting (operating element-wise on the first and last N/2 elements of the vector). In both cases, the reduction is performed until only a single element remains.


Kelimion:
Beautiful ASCII art.

Barinzaya:
> Beautiful ASCII art.

Big thanks to @flysand7 for setting a fantastic example to follow. I referenced the existing ones heavily!

core/simd/simd.odin
```diff
@@ -2510,3 +2510,457 @@ Example:
 recip :: #force_inline proc "contextless" (v: $T/#simd[$LANES]$E) -> T where intrinsics.type_is_float(E) {
 	return T(1) / v
 }
+
+/*
+Creates a vector where each lane contains the index of that lane.
```
flysand7, line 2515: The heading line of the doc should start with an imperative verb, just like they do with git commit messages.

```suggestion
Create a vector where each lane contains the index of that lane.
```

core/simd/simd.odin
```diff
@@ -2510,3 +2510,457 @@ Example:
...
+	} else {
+		return extract(v, 0)
+	}
+}
+
+/*
+Reduce a vector to a scalar by adding up all the lanes in a binary fashion.
+
+This procedure returns a scalar that is the sum of all lanes, calculated by
+splitting the vector in two parts and adding the two halves together
```
flysand7, line 2663: This might need a description of *how* the vector is split into two halves.

core/simd/simd.odin
```diff
@@ -2510,3 +2510,457 @@ Example:
...
+	} else {
+		return extract(v, 0)
+	}
+}
+
+/*
+Reduce a vector to a scalar by multiplying up all the lanes in a binary fashion.
+
+This procedure returns a scalar that is the product of all lanes, calculated by
+splitting the vector in two parts and multiplying the two halves together
```
flysand7, line 2868: This also needs a description of how the vector is split into two halves.

core/simd/simd.odin
```diff
@@ -2510,3 +2510,457 @@ Example:
...
+		n = n / 2
+		for i in 0 ..< n {
+			a[i] = a[2*i+0] * a[2*i+1]
+		}
+	}
+	res := a[0]
+
+Graphical representation of the operation for N=4:
+
+	   +-----------------------+
```
flysand7, line 2793: Can you check that you got the graphical representation right? It seems to me that you have confused the graphical representation for `reduce_mul_pairs` and `reduce_mul_split`, because the implementation seems to disagree with this graphic somewhat.

flysand7: Solid.

core/simd/simd.odin
```diff
@@ -2510,3 +2510,461 @@ Example:
...
+Result:
+- A vector of the given type, where each lane contains the index of that lane.
+
+**Operation**:
+
+	for i in 0 ..< N {
+		res[i] = i
+	}
+*/
+indices :: #force_inline proc "contextless" ($V: typeid/#simd[$N]$E) -> V where intrinsics.type_is_numeric(E) {
```
gingerBill, line 2529: This might be better implemented as an intrinsic so that it works for any sized `#simd` vector.

gingerBill: I have now made this an intrinsic, which will void this specific part of the PR, sorry.

gingerBill:
In general, I think most of this is better solved with implementing the intrinsics directly rather than this approach.

core/simd/simd.odin
Barinzaya: Is the goal eventually to increase the limit of remove the power-of-two restriction?

gingerBill:
@Barinzaya, this is effectively the case Odin does suffer due to the lack of metaprogramming. However, this is such a rare case that just adding a new intrinsic might help to begin with. We already have the `simd.reduced_*_ordered` set, why not add the `*_pairs` and `_bitsect` forms?

Barinzaya:
> @Barinzaya, this is effectively the case Odin does suffer due to the lack of metaprogramming. However, this is such a rare case that just adding a new intrinsic might help to begin with. We already have the `simd.reduced_*_ordered` set, why not add the `*_pairs` and `_bitsect` forms?

I'll work on implementing them as intrinsics instead, then.

gingerBill (approved):

PR #5121, opened 2025-05-05
XAudio2 Implementation

Badaxis:
Here are the bindings for XAudio2 v2.9 for Windows.
Took a few days to test. Kept all comments, implemented helper procedures.
Kept all files separate, don't know if I should fuse them.
Omitted xapobase.h as it's basically a class constructor of xapo.h on a different lib.
Omitted hrtfapoapi.h for now, it's a bit niche and separate dll. I'll look it up later.
https://learn.microsoft.com/en-us/windows/win32/xaudio2/xaudio2-introduction


Kelimion:
Looks quite nice at first glance. Will have to take some more time later to perform an proper review.

PR #5122, opened 2025-05-06
Add asan support for various allocators and stack unpoisoning

Lperlind:
Adds address sanitizer poisoning to various allocators. The general strategy for poisoning/unpoisoning is as follows:

* When an allocator is given a block of memory to be used for allocation it should poison that entire memory region.
* When an allocator allocates it should unpoison the handed out memory region.
* When an allocator frees it should poison the reclaimed memory.
* As allocators can be composed with other allocators, they need to be able to handle and deal with already poisoned memory blocks, i.e. use `@(no_sanitize_address)` and temporarily unpoison memory if required.

Additionally the stack is unpoisoned manually as this feature is only available in LLVM 21 (`use-after-scope` asan option cleans up poisoned stack variables). This is needed to prevent poisoned memory from buffer arenas persisting across scopes.

core test suite passes with sanitizers enabled with PR https://github.com/odin-lang/Odin/pull/5095

Thanks to @laytan for helping out with the stack poisoning issues


base/runtime/internal.odin
```diff
@@ -1106,3 +1106,12 @@ __read_bits :: proc "contextless" (dst, src: [^]byte, offset: uintptr, size: uin
...
 	}
 }
+
+@(no_sanitize_address)
+__asan_unpoison_memory_region :: #force_inline proc "contextless" (address: rawptr, size: uint) {
+	foreign {
+		__asan_unpoison_memory_region :: proc "system" (address: rawptr, size: uint) ---
+	}
+	__asan_unpoison_memory_region(address, size)
+}
```
gingerBill, line 1116: Why is this wrapped?

src/llvm_backend_stmt.cpp
```diff
@@ -2917,6 +2917,22 @@ gb_internal void lb_emit_defer_stmts(lbProcedure *p, lbDeferExitKind kind, lbBlo
...
+				lbValue local = p->asan_stack_locals[i];
+
+				auto args = array_make<lbValue>(temporary_allocator(), 2);
+				args[0] = lb_emit_conv(p, local, t_rawptr);
+				args[1] = lb_const_int(p->module, t_int, type_size_of(local.type->Pointer.elem));
+				lb_emit_runtime_call(p, "__asan_unpoison_memory_region", args);
+			}
+		}
+	#else
+		#error "Need to implement LLVM 21 'use-after-scope' asan option"
```
gingerBill, line 2933: I would do `compiler_error(...)` if an `p->asan_stack_locals` exists rather than prevent people from compiling the compiler.

base/runtime/internal.odin
laytan: I wasn't sure and didn't try to find out if `lb_emit_runtime_call` could directly call foreign procedures.

src/llvm_backend_stmt.cpp
Lperlind: Hmm I think if you want that it is better if I just have no error and just stay vigilant for the llvm 21 release

base/runtime/internal.odin
Lperlind: Turns out it is not needed.

Issue #2358, opened 2023-03-01
Odin Core Package Documentation


Lperlind:
## Current Status
Currently we're using `core:strings` as the playing ground for documentation. It's quite a large package so most of the groundworks on how things should be styled, minimum requirements, etc. are being tested there. Once completed it will be the  de facto reference on writing documentation for the core libraries.

# Odin Core Package Documentation

This issue is mainly to help track overall progress on documentation of Odin's core packages.
Bill stated that he really needs others to contribute if we want extensive coverage of all the packages.

## Goals

Overall the simplest goal we can is to have all public facing procedure and type have a description of what they do (no matter how trivial the function is). This is any easy first pass that seems pretty achievable.

Ideally we extend this further and have at least 1 example of usage code with expected outputs for every procedure, etc.

## Documentation Language

It would be really good if we can stick to some consistent set of rules as a goto for how to word things, we should come to an agreement at some point. I've noticed different packages explain in very different ways from one another it would be good to iteratively make things more the same.

Because everything is so different, there is no obvious example to refer to.

For now the simplest advice is "make it clear and terse, don't be clever".

## Packages Checklist

Below are all the packages that need to be documented. Many packages contain a sizeable amount of documentation but are missing many functions, these are marked as `incomplete`.
Some packages have very little to no documentation, these are marked as `not documented`.
I largely glanced through https://pkg.odin-lang.org/core/ to determine the state of current things.
You can use this as some sort of a guide to what is best to work on.

We can update the checklist below as work is slowly merged into master.

- [ ] bufio
- [ ] bytes
- [ ] compress
- [ ] container
- [ ] crypto
- [x] dynlib
- [x] fmt
- [ ] hash
- [ ] image
- [ ] io
- [ ] log
- [ ] math
- [x] mem
- [ ] odin
- [ ] path
- [ ] reflect
- [ ] runtime
- [x] simd
- [ ] slice
- [ ] sort
- [ ] strconv
- [x] strings
- [ ] sync
- [ ] text
- [x] time
- [x] thread
- [ ] unicode
- [x] flags

Packages that should not be documented
- c: users should refer to lib c documentation
- sys: users should refer to their platform specific documentation
- os: planned to be deprecated for os2
- thread: planned to have changes

Lperlind:
Steps to standardise example code has been added in this PR: https://github.com/odin-lang/pkg.odin-lang.org.
If merged in I'll outline example requirements in a future comment.

flysand7:
[...]

PR #5126, opened 2025-05-06
 Document core:slice package 

sz-piotr:
Tries to follow https://github.com/odin-lang/Odin/discussions/5039 as close as possible.
If this is merged https://github.com/odin-lang/Odin/issues/2358 should be updated.

In addition to adding and updating doc comments the following changes are made:
- Add #optional_allocator_error to map utilities
- Improve procedure argument naming

### DIFF
--- a/core/slice/map.odin
+++ b/core/slice/map.odin
@@ -6,7 +6,20 @@ import "base:runtime"
 _ :: intrinsics
 _ :: runtime
 
-map_keys :: proc(m: $M/map[$K]$V, allocator := context.allocator) -> (keys: []K, err: runtime.Allocator_Error) {
+/*
+Return all keys of a map.
+
+*Allocates Using Provided Allocator*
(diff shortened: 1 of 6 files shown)


core/slice/map.odin
```diff
@@ -25,19 +52,37 @@ map_values :: proc(m: $M/map[$K]$V, allocator := context.allocator) -> (values:
...
+/*
+A map entry including a key and a value.
+*/
 Map_Entry :: struct($Key, $Value: typeid) {
 	key:   Key,
 	value: Value,
 }
 
+/*
+Similar to `Map_Entry` but also includes the hash of the key.
```
deadwanderer, line 64: Should probably just document this struct fully here; don't make the user go looking somewhere else to find a full definition?

So, something like `A map entry, including a key, a value, and the hash of the key`

PR #5130, opened 2025-05-07
RAD Debugger support through the custom `.raddbg` section

gingerBill:


src/llvm_backend.cpp
```diff
@@ -3343,6 +3365,72 @@ gb_internal bool lb_generate_code(lbGenerator *gen) {
 	if (build_context.ODIN_DEBUG) {
 		TIME_SECTION("LLVM Debug Info Complete Types and Finalize");
 		lb_debug_info_complete_types_and_finalize(gen);
+
+		// Custom `.raddbg` section for its debugger
+		if (build_context.metrics.os == TargetOs_windows) {
```
gingerBill, line 3370: This is because raddbg only works on Windows currently.

PR #5132, opened 2025-05-07
Add `@(public)`

karl-zylinski:
This PR adds `@(public)`, which can be used to revert `#+private file` and `#+private package` for a single entity.

## Rationale
This makes it possible to export things to the outside of a package, directly from files tagged with for example `#+private file`. But that is not my main concern.

The main thing I'm trying to fix is that I sometimes use `#+private file` at the top of a file within a bigger program. This program is not used as a package by anyone (it's a video game, all of it is a single package). I then use `@(private="package")` to expose a few API functions, from that file, to the rest of the program. In this specific case, when I see it as just a program, and not a package, it is very strange to put `@(private="package")` to make something "less private": What I want to signal is "it's public to the program". There is no "package thinking" going on at all at this point.

With this organizational method I can prefix the "API procs", and make utility procs without any prefixes, with little concern for name collisions etc. It makes my programming more joyful.

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -3866,7 +3866,7 @@ gb_internal DECL_ATTRIBUTE_PROC(const_decl_attribute) {
 			error(elem, "Expected a string value for '%.*s'", LIT(name));
 		}
 		return true;
-	} else if (name == "private") {
+	} else if (name == "private" || name == "public") {
 		// NOTE(bill): Handled elsewhere `check_collect_value_decl`
 		return true;
 	} else if (name == "static" ||
@@ -3890,7 +3890,7 @@ gb_internal DECL_ATTRIBUTE_PROC(type_decl_attribute) {
 			error(elem, "Expected a string value for '%.*s'", LIT(name));
 		}
 		return true;
-	} else if (name == "private") {
+	} else if (name == "private" || name == "public") {
 		// NOTE(bill): Handled elsewhere `check_collect_value_decl`
 		return true;
 	} else if (name == "objc_class") {
@@ -4178,6 +4178,7 @@ gb_internal void check_collect_value_decl(CheckerContext *c, Ast *decl) {
 	ast_node(vd, ValueDecl, decl);
 
 	EntityVisiblityKind entity_visibility_kind = c->foreign_context.visibility_kind;
+	bool force_public = false;
 	bool is_test = false;
 	bool is_init = false;
 	bool is_fini = false;
@@ -4239,6 +4240,14 @@ gb_internal void check_collect_value_decl(CheckerContext *c, Ast *decl) {
 				}
 				slice_unordered_remove(elems, j);
 				j -= 1;
+			} else if (name == "public") {
+				if (value != nullptr) {
+					error(value, "'%.*s' expects no parameter", LIT(name));
+				}
+				entity_visibility_kind = EntityVisiblity_Public;
+				force_public = true;
+				slice_unordered_remove(elems, j);
+				j -= 1;
 			} else if (name == "test") {
 				is_test = true;
 			} else if (name == "init") {
@@ -4254,7 +4263,7 @@ gb_internal void check_collect_value_decl(CheckerContext *c, Ast *decl) {
 		return;
 	}
 
-	if (entity_visibility_kind == EntityVisiblity_Public &&
+	if (!force_public && entity_visibility_kind == EntityVisiblity_Public &&
 	    (c->scope->flags&ScopeFlag_File) &&
 	    c->scope->file) {
 	    	if (c->scope->file->flags & AstFile_IsPrivateFile) {
@@ -4267,7 +4276,6 @@ gb_internal void check_collect_value_decl(CheckerContext *c, Ast *decl) {
 	if (entity_visibility_kind != EntityVisiblity_Public && !(c->scope->flags&ScopeFlag_File)) {
 		error(decl, "Attribute 'private' is not allowed on a non file scope entity");
 	}
-
 
 	if (vd->is_mutable) {
 		if (!(c->scope->flags&ScopeFlag_File)) {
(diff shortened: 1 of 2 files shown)


gingerBill:
I am very hesitant on this as a concept, and it's why I've never implemented it myself. 

If it was to exist, I'd also want it to be an _error_ in files which do not have the `#+private` directive, as I know many people will use it willy-nilly.

There is also my other concern that people will default to `#+private` and then use `@(public)` which completely defeats why I wanted public-by-default in the first place. People who do private-by-default in others languages are honestly just being defensive against themselves for no real reason, and it makes it an absolute pain to use for others who actually need the private stuff—the programmer cannot know what the users of their code actually need and hiding it behind something private is usually a very bad idea. Sometimes there is a good reason to keep some things "private" but it should be treated as the edge case.

And with the thing you are implementing, I am still not sure if it is warranted. It might be a leaky abstraction of `#+private file` where you just needed a @(private="package")` interface first and then be public in a public file. I understand it is cumbersome, but if you are using `#+private file`, is that need actually necessary?

karl-zylinski:
I agree that using `#+private` to hide things from users of a package is bad. Not sure how much the language should enforce stuff like that though, since being opinionated at that detail level is bound to cause friction in edge cases.

I don't really need this for packages. I just want a way to make something "program-public" from a private file to the rest of my program, without using `@(private="package")`, because that naming felt very confusing. So in a sense I'm just using this as a less ugly way to to write `@(private="package")`.

Not sure I completely understand the last paragraph. I just wanted a single `level_editor.odin` file that implements it all, where I can use non-prefixed proc names that don't collide with my other editors. Think of it as its own little "subprogram" within the bigger package. There is no interface as it does not add any benefit.

Think like this: You want to make a system for a game that does _a lot of stuff_ but it has a very small API surface. Procedural generation systems come to mind. Perhaps it has 1-2 procs that are "public" and 50 procs that are "private". So you put `#+private file` at the top so you can code a bit more freely within that file. The only issue I had was that using `@(private="package")` felt like a strange way to make the API public to the rest of the program.

gingerBill:
I do think that the use of `@(public)` outside of `#+private` files should be banned though since it is redundant.

ghost-x47:
Somehow, I ended up here, and I am not sure if i ever allowed to comment on this)
I do think that using public to reverse private file is not very useful. What is the point of private file in the first place if you still need to think - "oh, right, it still might contain public code. Need to search for that." It is complication. I would endure the struggle to rewrite or change the code to remove private file or separate into other file, rather than always struggle to read the file. Is there will be a feature to mark a file to be really private? +private_enforced or something along the lines?

karl-zylinski:
@ghost-x47 You can already use `@(private="package")` to reverse `#+private file`. In fact, a file that is `#+private file` with no `@(private="package")` or `@(public)` is mostly useless since nothing in it can ever be run by anything else.

karl-zylinski:
@Kelimion Yeah, that was why I wrote "mostly useless" :D 

@gingerBill I added the error you mentioned. Do as you please with the PR!

Feoramund:
Do I understand this correctly that it is a shorter way of making every `proc`, global variable, and constant as `@(private)`, except for the elements someone deems as `@(public)`?

karl-zylinski:
@Feoramund Yeah. You can use it for packages in that way. But you can also use `@(public)` as an easy way to override `#+private file`, picking just a few procs from a file to be available to the rest of your program (I mostly make programs, not reusable packages). That way you can go a bit more bananas within that file. You can already do this today using `@(private="package")`, but it feels a bit weird and has bad discoverability due to it just being another variant of "private".

Feoramund:
I will say that in the past, I have wanted to make many procs and constants in a package private, while leaving only a couple public. However, the verbosity and friction of adding `@(private="...")` to all those definitions gave me pause to think that maybe I should restructure what I'm doing, and it's usually ended up with a better API where I don't make everything so private. I think that's what Odin implicitly encourages: letting programmers do as they want with state.

I can't say I've had the situation that you described in your original post. I've never "upgraded" or overridden a privacy level before (I did not even know of `@(private="package")` before now; looks like we only use it in `os2` and `core:debug`), so this may not be for me, but I do like the friction aspect of how we currently do things since it made me stop and think.

karl-zylinski:
@Feoramund `@(private)` is the same as `@(private="package")`. That's the default. The only other option is `(@private="file")`

So you could put `#+private file` at the top and write `@private` in front of procs to make them not file private. That's perhaps the most confusion version of all of this. `@(private="package")` makes a bit more sense, but it's still not that pretty.

In any case, what I wanted this for was not to create an API for a package, but to not clutter the global namespace in a big project. So I put `#+private file` at top of some files that have a lot of procs and "unprivate" a few using `@(package="package")`, which I found a bit obtuse.

Apis035:
I just got in a situation where I actually needed this PR.

```odin
#+private file
package editor

// Public ---------------------------------------------- //

@private SkinSetStyle :: proc(style: SkinStyle) { skinStyle = style }
@private SkinUpdate   :: proc() { SkinUpdateProc[skinStyle]() }
@private SkinDraw     :: proc() { SkinDrawProc[skinStyle]() }

// Private --------------------------------------------- //

SkinStyle :: enum { /* ... */ }

skinStyle: SkinStyle
skinUpdateProc := [SkinStyle]proc() { /* ... */ }
skinDrawProc   := [SkinStyle]proc() { /* ... */ }

// ...
```

Same reason as Karl's, to avoid cluttering global namespace, as in this file only these 3 procs should be used. But it looks so awkward that the exported procs is marked as private when it is actually public. The `@(public)` sugar would be nice to have so I really hope this PR would be accepted!

gingerBill:
I am not going to accept this idea. The more I have mulled this over, the more I thought it's going to be a bad idea. To quote what @Feoramund said:

> I think that's what Odin implicitly encourages: letting programmers do as they want with state.

This is exactly the design principle here and why I wanted public-by-default, at the behest of what many people's "common wisdom" is on the internet.

PR #5159, opened 2025-05-13
New core library: core:container/segtree

0xrsp:
Implement segment tree data structure (without lazy propagation) in core:container.


gingerBill:
This is honestly way too niche to be in the core library.

PR #5160, opened 2025-05-14
Faster math.next_power_of_two and refined API

0xrsp:
Replace math.next_power_of_two with a faster implementation similar to that used in clang/gcc and refined API.

OLD API:
`next_power_of_two :: proc "contextless" (x: int) -> int `

NEW API:
`next_power_of_two :: proc "contextless" (#any_int x: int) -> uint `


gingerBill:
Have you actually tested this to see how much faster it is?

0xrsp:
Benchmarked with LLVM backend x86_64 it is significantly faster on most optimization levels being at least ~2x faster and on average ~4-5x faster for most inputs, LLVM seems to produce. It also behaves better than the original implementation due to the return type being unsigned i.e. 
orig(1<<62+1)=-1<<63
updated(1<<62+1)=1<<63

The reason I left this as a draft is the polymorphic int input could be wider than the uint so need a way to get an unsigned int that is the same width as a signed type.

Despite the initial branch seems to be much faster due to just less operations required and smaller code size.

Disassembly of new version:
image

Disassembly of old version:
image

JesseRMeyer:
That's non-optimized codegen.

https://godbolt.org/z/KM3PfPGfe

PR #5172, opened 2025-05-17
sys/linux: Improve documentation for Dirent and related procedures

krnowak:
I know that `sys` package is generally excluded from [the documentation effort](https://github.com/odin-lang/Odin/issues/2358), but I suppose it still makes sense to document procedures that are not a 1:1 translation of OS API. For the procedures, I tried to follow the [docs conventions](https://github.com/odin-lang/Odin/discussions/5039).

Here I wanted to address couple issues:

- There is no such procedure like `dirent_iterator`.
- Example code ignores return value from `getdents()`, does not declare the `offs` variable.
- [Docs on odin website](https://pkg.odin-lang.org/core/sys/linux/#dirent_iterate_buf) are barely legible.

### DIFF
--- a/core/sys/linux/wrappers.odin
+++ b/core/sys/linux/wrappers.odin
@@ -54,22 +54,45 @@ WCOREDUMP :: #force_inline proc "contextless" (s: u32) -> bool {
 // TODO: sigaddset etc
 
 
-/// Iterate the results of getdents
-/// Only iterates as much data as loaded in the buffer
-/// In case you need to iterate *all* files in a directory
-/// consider using dirent_get_iterate
-///
-/// Example of using dirent_iterate_buf
-///   // Get dirents into a buffer
-///   buf: [128]u8
-///   sys.getdents(dirfd, buf[:])
-///   // Print the names of the files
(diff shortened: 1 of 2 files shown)


core/sys/linux/wrappers.odin
```diff
@@ -54,22 +54,45 @@ WCOREDUMP :: #force_inline proc "contextless" (s: u32) -> bool {
...
+
+**Inputs**:
+- `buf` - byte buffer with data from `getdents()`
+- `offs` - offset to the next possible directory entry in `buf`
+
+**Returns**:
+- A pointer to a directory entry in `buf`, or `nil`.
+- A bool value denoting if a valid directory entry is returned.
+
+**Example**:
```
laytan, line 73: This should be `Example:` or the documentation generator needs to be changed to pick up on `**Example:**` too.

Did we actually decide on the `**`? I still think it's a bad idea.

krnowak: Welp, the toplevel post in https://github.com/odin-lang/Odin/discussions/5039 uses `**Example**:`, but grepping the sources shows me that `Example:` is used most of the time. Will update.

PR #5173, opened 2025-05-17
Keep shared libraries from calling main program's startup/cleanup procs on Linux

### LINKED ISSUE #5169: Odin shared libraries call the main program's `_startup_runtime` on Linux
Given the following main program:

`main.odin`
```odin
package main

import "base:runtime"

foreign import alpha "alpha.so"
foreign import beta "beta.so"

foreign alpha {
	get_alpha :: proc() -> int ---
}

foreign beta {
	get_beta :: proc() -> int ---
}

@(init)
program_init :: proc() {
	runtime.println_any("Init from the main program.")
}

@(init)
program_init_2 :: proc() {
	runtime.println_any("Another init from the main program.")
}

main :: proc() {
	alpha := get_alpha()
	beta := get_beta()

	runtime.println_any(alpha)
	runtime.println_any(beta)
}
```

And the two following libraries:

`alpha/lib.odin`
```odin
package alpha

import "base:runtime"

alpha: int

@(init)
init_alpha :: proc() {
	runtime.println_any("Init from inside alpha.")
	alpha = 32
}

@(export)
get_alpha :: proc() -> int {
	return alpha
}
```

`beta/lib.odin`
```odin
package beta

import "base:runtime"

beta: int

@(init)
init_beta :: proc() {
	runtime.println_any("Init from inside beta.")
	beta = 16
}

@(export)
get_beta :: proc() -> int {
	return beta
}
```

Compiled with:
```
odin build alpha -build-mode:dll -debug
odin build beta -build-mode:dll -debug
odin build . -debug
```

Running the main program produces this output:
```
Init from the main program.
Another init from the main program.
Init from the main program.
Another init from the main program.
Init from the main program.
Another init from the main program.
0
0
```

I'm curious if this also happens on Windows, or if this is another SysV problem.

Feoramund:
Fixes #5169 

I've tested this and it works on my machine. I'm not sure if it needs to be done for any other procs, or if this is the ideal way to do it, but if not, hopefully it helps expand the understanding on the issue.

### DIFF
--- a/src/llvm_backend.cpp
+++ b/src/llvm_backend.cpp
@@ -1907,6 +1907,10 @@ gb_internal lbProcedure *lb_create_startup_runtime(lbModule *main_module, lbProc
 	lb_add_attribute_to_proc(p->module, p->value, "optnone");
 	lb_add_attribute_to_proc(p->module, p->value, "noinline");
 
+	// Make sure shared libraries call their own runtime startup on Linux but
+	// still allow it to be visible to other objects.
+	LLVMSetVisibility(p->value, LLVMProtectedVisibility);
+
 	lb_begin_procedure_body(p);
 
 	lb_setup_type_info_data(main_module);
@@ -2015,6 +2019,10 @@ gb_internal lbProcedure *lb_create_cleanup_runtime(lbModule *main_module) { // C
 	p->is_startup = true;
 	lb_add_attribute_to_proc(p->module, p->value, "optnone");
 	lb_add_attribute_to_proc(p->module, p->value, "noinline");
+
+	// Make sure shared libraries call their own runtime cleanup on Linux but
+	// still allow it to be visible to other objects.
+	LLVMSetVisibility(p->value, LLVMProtectedVisibility);
 
 	lb_begin_procedure_body(p);
 


Kelimion:
That it's practically a one-liner is encouraging.

Feoramund:
Please if you could, make sure to test that this doesn't break anything on Windows as a result. I chose `Protected` visibility instead of `Hidden`, since it seemed more permissive and still worked. If it breaks Windows, it should be okay to just stuff it behind an architecture-based condition.

Kelimion:
With this patch applied, I get the following.

Windows:
```
Init from inside alpha.
Init from inside beta.
Init from the main program.
Another init from the main program.
32
16
```
Linux:
```
Init from inside beta.
Init from inside alpha.
Init from the main program.
Another init from the main program.
32
16
```

I don't know why, but on Linux it consistently prints `beta` before `alpha`, and on Windows it's consistently `alpha`, then `beta`. But it's progress.

Feoramund:
Interesting. I just noticed the order is actually different too as you pointed out, so it's technically not mirroring Windows.

I would consider this a bug in its own right, as I can imagine people may need to rely on certain linking order behaviors in the future, especially if `init` procs are involved.

Kelimion:
That may be down to the difference between msvc and clang's linkers, who knows. Would have to be looked into. Other than that the fix looks fine to me, but let's let Bill take a gander first.

laytan:
Looks similar to a patch I did recently that was having the same symptoms with other procs and globals being used from other modules: https://github.com/odin-lang/Odin/commit/1053ec30518486e3feb523d1a37e814ce992e79c

Weird that doesn't apply it to these procs too.

Feoramund:
It looks like the startup/cleanup procs are never seen by either `lb_find_ident` or `lb_find_value_from_entity`, which are the two ways you get to `lb_set_entity_from_other_modules_linkage_correctly`.

It looks like they getting skipped on `lb_build_expr_internal` due to them being constant values, so we don't get to `case_ast_node(i, Ident, expr)` which could lead to a call to `lb_find_ident`.

Is this possibly where another check for it should be?

Feoramund:
I updated the patch to use the same visibility/linkage as `lb_correct_entity_linkage`.

gingerBill (approved):

PR #5175, opened 2025-05-18
Use `--sysroot` instead of `-Wl,-syslibroot` on Darwin

Feoramund:
This keeps the linker from using the wrong SDK and mirrors how we build the Odin compiler itself in `build_odin.sh`.

Specifically, I needed this fix in order to keep my old macOSX 11 from using the version 10 SDK, which kept it from being able to find `__ulock_wait2` for `core:sync` and the demo.


laytan:
I tried this out and `l /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk` points to 15.4 while `xcrun --sdk macosx --show-sdk-path` points to 15.2, weird, but not the end of the world, and we do the same in the compiler build so it should be fine.

PR #5183, opened 2025-05-20
Do not call disabled deferred procedures

### LINKED ISSUE #5083: `@(disabled=!ODIN_DEBUG)` doesn't work if procedure called using `@(deferred_*=foo_debug)`
## Context

Odin:    dev-2025-04:d463aba7d
OS:      Ubuntu 24.04.2 LTS, Linux 6.8.0-58-generic
CPU:     Intel(R) Core(TM) i7-4790S CPU @ 3.20GHz
RAM:     15901 MiB
Backend: LLVM 20.1.4

## Expected Behavior

Expect `foo_debug` to be disabled on debug builds even if called via a `deferred_*` attribute instead of directly.
## Current Behavior

Direct calls are disabled correctly, deferred call runs as normal

### Steps to Reproduce

Compile with and without `-debug`.
`debug_check` prints twice for `-debug`, once otherwise (because called directly). Expected 

```odin
package main

import "core:fmt"

@(disabled=!ODIN_DEBUG)
debug_check :: proc () {
	fmt.printfln("The code is safe. DEBUG: %v.", ODIN_DEBUG)
}

@(deferred_in=debug_check)
foreign_routine :: proc () {
	fmt.printfln("Running evil code. DEBUG: %v.", ODIN_DEBUG)
}

main :: proc () {
	foreign_routine()
}
```
```
Running evil code. DEBUG: false. // Expected
The code is safe. DEBUG: false.  // This shouldn't have printed

Running evil code. DEBUG: true. // Expected
The code is safe. DEBUG: true.  // Expected
```

### Failure Logs

Feoramund:
Fixes #5083 

I put the warning token on the `proc` with the deferral as it's the first step to the cause of the issue, but I can see a case for either proc as being a valid warning source.

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -6325,6 +6325,12 @@ gb_internal void check_deferred_procedures(Checker *c) {
 			continue;
 		}
 
+		if (dst->flags & EntityFlag_Disabled) {
+			// Prevent procedures that have been disabled from acting as deferrals.
+			src->Procedure.deferred_procedure = {};
+			continue;
+		}
+
 		GB_ASSERT(is_type_proc(src->type));
 		GB_ASSERT(is_type_proc(dst->type));
 		Type *src_params = base_type(src->type)->Proc.params;


Feoramund:
I can toss that out. I made it a warning because that's what I did when I fixed disabled `@(init)`. Should that also not be a warning? My intent was consistency.

Kelimion:
> I can toss that out. I made it a warning because that's what I did when I fixed disabled `@(init)`. Should that also not be a warning? My intent was consistency.

That's more likely to be an unintentional error. You run things under `@(init)` for a reason. If you want to disable it, you could also add a `when DISABLED_REASON { return }` as the first thing in those rare few exceptions where you don't want the warning, but also for that particular init to not do anything.

Kelimion:
(These things are hard, aren't they, deciding sane defaults? But we can always revisit these decisions if we get feedback on these particular obscure use cases.)

Feoramund:
> (These things are hard, aren't they, deciding sane defaults? But we can always revisit these decisions if we get feedback on these particular obscure use cases.)

Yeah, I have a tendency to go for caution and informing the user as much as possible, but I can also see this being a less annoying result if it's commonplace for someone to have disabled deferrals for some reason.

Feoramund:
I've had issues with downloading the test assets locally today too.

Kelimion:
> > (These things are hard, aren't they, deciding sane defaults? But we can always revisit these decisions if we get feedback on these particular obscure use cases.)
> 
> Yeah, I have a tendency to go for caution and informing the user as much as possible, but I can also see this being a less annoying result if it's commonplace for someone to have disabled deferrals for some reason.

I can see wanting to be cautious, but I also figure that if you're going out of your way to use `deferred` and `disabled` together, you're telling the compiler you at least think you know what you're doing.

Kelimion:
And FreeBSD caps it off with a "VM is booting" loop. I think I'll merge the patch. :-)

Kelimion:
Confirmed the fix works locally. Thanks as always, Feoramund.

PR #5184, opened 2025-05-20
Add `ODIN_BUILD_FLAGS` environment variable

Feoramund:
This change appends each space-separated flag in the `ODIN_BUILD_FLAGS` environment variable to build-mode commands, allowing users to have default flags for the build/test/run actions.

Note that it will complain if you explicitly use a flag that is already in your environment variable, just as if you typed it. This doesn't do anything too magical.

If there's a need for an `ODIN_CHECK_FLAGS`, I can add that too.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -647,6 +647,19 @@ gb_internal bool parse_build_flags(Array<String> args) {
 		flag_args = array_slice(args, 3, args.count);
 	}
 
+	if (build_context.command_kind & Command__does_build) {
+		String env_flags = make_string_c(gb_get_env("ODIN_BUILD_FLAGS", heap_allocator()));
+		if (env_flags != "") {
+			flag_args.allocator = heap_allocator();
+			String_Iterator env_flag_it = {env_flags, 0};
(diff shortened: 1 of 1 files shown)


Kelimion:
I've been thinking on and off to add a `-flags=<filename>` option for the same purpose, so you can express your common flags in one go. Obviously the `-flags` option would be disallowed if encountered in a file, or this could go ad infinitum.

Let's see what @gingerBill makes of it.

gingerBill:
It is a good question whether an environment variable or a file is better. I know many people want this but I do worry it'll make things a little more complicated.

Feoramund:
> I know many people want this but I do worry it'll make things a little more complicated.

I can see that happening now that I think about it, whether it's a file or an environment variable.

I'm okay with dropping this PR. Most shells I've used have some support for aliases, and failing that, people can put an executable script in their `PATH` as a shortcut that adds certain flags they use often.

Feoramund:
Closing as unneeded; shells with variable expansion and aliases are suitable enough for now.

Issue #4923, opened 2025-03-10
`odin test` doesn't respect terminfo/`TERM=dumb`


diniamo:
## Context

* Operating System & Odin Version: NixOS, odin version dev-2025-01
* Please paste `odin report` output:

```
	Odin:    dev-2025-01
	OS:      NixOS 25.05 (Warbler), Linux 6.13.0
	CPU:     Intel(R) Core(TM) i5-10400F CPU @ 2.90GHz
	RAM:     48113 MiB
	Backend: LLVM 18.1.8
```

## Expected Behavior

When printing, Odin ideally respects `terminfo(5)`, but at least TERM=dumb, and doesn't print any unsupported (in the case of a dumb terminal, that would be all) control sequences.

## Current Behavior

Odin doesn't respect either of these, and the output ends up unreadable.

### Steps to Reproduce

1. Run `TERM=dumb odin test package`
2. Output is still colored

laytan:
Duplicate of https://github.com/odin-lang/Odin/issues/4100 (it's fine to leave this open)

diniamo:
Ah, I didn't find that.

The fact that the internal logging package doesn't handle this is a pretty big deal.

Kelimion:
I'd also like it so I can run `odin test foo` from Sublime and not get the ANSI spew in my build output.

Issue #4100, opened 2024-08-18
[Odin Test] Ignore ANSI escape codes for non tty outputs


MrStevns:
## Context

When using `odin test` where the output is not printed to a terminal but say the output window in Sublime Text or in my case, Focus, the output is unnecessarily messy. It would be nice if we could ignore the ANSI escape codes or have a flag to disable them.

### Odin report
* Odin:    dev-2024-08:ac483f72e
* OS:      macOS Monterey 12.7.2 (build: 21G1974, kernel: 21.6.0)
* CPU:     Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz
* RAM:     16384 MiB
* Backend: LLVM 18.1.8

## Expected Behavior
When building a project from either Sublime text or Focus, or any other non tty output, i'd expect something like this:
```
[INFO ] --- [2024-08-17 07:52:29] Starting test runner with 1 thread. Set with -define:ODIN_TEST_THREADS=n.
[INFO ] --- [2024-08-17 07:52:29] The random seed sent to every test is: 39114654177527. Set with -define:ODIN_TEST_RANDOM_SEED=n.
[INFO ] --- [2024-08-17 07:52:29] Memory tracking is enabled. Tests will log their memory usage if there's an issue.
[INFO ] --- [2024-08-17 07:52:29] < Final Mem/ Total Mem> <  Peak Mem> (#Free/Alloc) :: [package.test_name]
[INFO ] --- [2024-08-17 07:52:29] [rendering.odin:9:example()] Debug test
rendering  [|                       ]         1 :: [package done]

Finished 1 test in 270µs. The test was successful.
```

## Current Behavior
Where as right now the output is this.
```
�[?25l�[?7lrendering  [�[90m|                       �[0m]    0/   1 :: �[90mexample�[0m
�[?7h1 thread                                 0/   1 :: total
�[2F�[J�]2;Odin test runner (0/1)�\�[?7lrendering  [�[33m|                       �[0m]    0/   1 :: �[33mexample�[0m
�[?7h1 thread                                 0/   1 :: total
�[2F�[J�]2;Odin test runner (1/1)�\�[?7lrendering  [�[32m|                       �[0m]         1 :: [package done]
�[?7h
�[?25h�[0m[INFO ] --- �[0m[2024-08-17 07:53:51] Starting test runner with 1 thread. Set with -define:ODIN_TEST_THREADS=n.
[...]

PR #5185, opened 2025-05-20
Add ODIN_TEST_PLAIN_TEXT

Kelimion:
Allow running tests with `-define:ODIN_TEST_PLAIN_TEXT=true`.

This removes all ANSI format codes (and thus `ODIN_TEST_FANCY_OUTPUT`), so the test runner output is more friendly for build output windows in an editor, whose ANSI support may be broken or lacking entirely.

On Windows, we now also restore the console mode so the cursor doesn't disappear after the tests are done.

### DIFF
--- a/core/testing/logging.odin
+++ b/core/testing/logging.odin
@@ -17,21 +17,39 @@ import "core:strings"
 import "core:sync/chan"
 import "core:time"
 
-when USING_SHORT_LOGS {
-	Default_Test_Logger_Opts :: runtime.Logger_Options {
-		.Level,
-		.Terminal_Color,
-		.Short_File_Path,
-		.Line,
+when PLAIN_TEXT {
+	when USING_SHORT_LOGS {
+		Default_Test_Logger_Opts :: runtime.Logger_Options {
+			.Level,
+			.Short_File_Path,
+			.Line,
+		}
+	} else {
+		Default_Test_Logger_Opts :: runtime.Logger_Options {
+			.Level,
+			.Short_File_Path,
+			.Line,
+			.Procedure,
+			.Date, .Time,
+		}
 	}
 } else {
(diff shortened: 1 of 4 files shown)


Kelimion:
Would appreciate your input, @Feoramund. I've found the inability to switch off ANSI processing altogether previously to be limiting. Now I can add `odin test` to my regular build script and have it look correctly in an editor.

If you can think of a more structured way to add this functionality, that'd even better.

laytan:
I think the solution we do in the compiler as also outlined in https://github.com/odin-lang/Odin/issues/4923 is a better way. Could also have both but I think there are not many situations you want this PRs solution when you have the automatic one.

Kelimion:
> I think the solution we do in the compiler as also outlined in #4923 / #4100 is a better way. Could also have both but I think there are not many situations you want this PRs solution when you have the automatic one.

Right, and someone might want to tee it to a file.

Kelimion:
`tty_is_dumb(s)` knocking out `.Color` in `core:log` and taking out the window title, progress prints and color in `core:testing` so it works automatically would be preferable.

If @Feoramund wants first crack at it, I'm okay with that. Otherwise I'll replace this PR with one that does this for `core:log` and `core:testing`.

So for now I'll close this and just push a fix for the test runner eating the blinking cursor on Windows (by failing to restore the console mode).

Feoramund:
> > I think the solution we do in the compiler as also outlined in #4923 / #4100 is a better way. Could also have both but I think there are not many situations you want this PRs solution when you have the automatic one.
> 
> Right, and someone might want to tee it to a file.

Yes, from what I recall, this is the better way to do it, but the last couple times I looked at resolving this, it wasn't a straightforward fix. I could take another look at it.

PR #5189, opened 2025-05-20
Add `core:terminal`, fix test runner/`core:log` ANSI code issues

### LINKED ISSUE #4100: [Odin Test] Ignore ANSI escape codes for non tty outputs

### LINKED ISSUE #4923: `odin test` doesn't respect terminfo/`TERM=dumb`

Feoramund:
Fixes #4100
Fixes #4923

This should cover everything, and the animated output is now disabled at runtime if it detects that it would be inappropriate (must be enabled, have color support, and be an actual terminal on `stdout`), so you don't need to define `ODIN_TEST_FANCY` as false anymore.

This is a small breaking change for anyone using `core:encoding/ansi` as I moved that to the now-more-appropriate `core:terminal/ansi`. I decided to do most of the detection for this with `@(init)`, as we really shouldn't need post-startup detection of color capability or TTY detection for the majority of use cases.

@Kelimion @laytan Does this work well for you in your respective editors and use cases?


Feoramund:
The CI test logs are cleaner now too.

Kelimion:
This works very well with Sublime, thank you. I'll have a read through the patch now.
```
[INFO ] --- [2025-05-20 22:44:14] Starting test runner with 1 thread. Set with -define:ODIN_TEST_THREADS=n.
[INFO ] --- [2025-05-20 22:44:14] The random seed sent to every test is: 1927765687129245. Set with -define:ODIN_TEST_RANDOM_SEED=n.
[INFO ] --- [2025-05-20 22:44:14] Memory tracking is enabled. Tests will log their memory usage if there's an issue.
[INFO ] --- [2025-05-20 22:44:14] < Final Mem/ Total Mem> <  Peak Mem> (#Free/Alloc) :: [package.test_name]

Finished 1 test in 138µs. The test was successful.
```

Kelimion:
image
Does kill the color in the Windows terminal as well.

Feoramund:
> Does kill the color in the Windows terminal as well.

Is this expected? I feel like I remember `COMMAND.COM` supported 16 basic colors as far back as Windows 98 or so, but I'm not sure if it was ANSI-based or something Microsoft-specific. Is there a stanard way to detect it?

Kelimion:
So if we could restore ansi processing on Windows unless the terminal is redirected, that'd be wonderful.

Kelimion:
> > Does kill the color in the Windows terminal as well.
> 
> Is this expected? I feel like I remember `COMMAND.COM` supported 16 basic colors as far back as Windows 98 or so, but I'm not sure if it was ANSI-based or something Microsoft-specific. Is there a standard way to detect it?

I see you just force-pushed, so I'll try again, but here's my answers to the previous version:

You don't need to detect color support on Windows. If it's an actual console, you can enable the virtual terminal processing like you did in `console_ansi_init`, and it'll support ANSI on any modern Windows. The important thing is to restore the original console mode again on exit.

So if I look at `_is_terminal` it just checks whether `GetConsoleMode` is "true". It correctly disables ANSI within Sublime, but this also kills it for the actual proper terminal:tm:. :-(

Feoramund:
> So if we could restore ansi processing on Windows unless the terminal is redirected, that'd be wonderful.

It looks like there are a few ways to do this. `WT_SESSION` might be an environment variable that's set, as well as `TERM_PROGRAM`.

> It also looks like you reverted (or not yet merged) my [last patch](https://github.com/odin-lang/Odin/pull/5186), which fixed the test runner stealing the blinking cursor by restoring the original console mode on exit.

I hadn't yet pulled it onto my master when I made the branch; I've rebased and it should be there now.

Kelimion:
As an aside, I'd appreciate it if you could not force push. We can squash merge at the end if appropriate, but if I have the PR checked out locally, and you force push, git now thinks I have 4 commits to push to your branch and you have 8 new commits, or something like that. Git loses the plot completely.

I have to kill the checked out branch and recreate it against the new head of the branch, or manually patch up `.git/refs/heads/...`.

(I also appreciate the work you're doing, of course.)

Feoramund:
I think that's because I forgot to use `--force-with-lease`.

Kelimion:
> > So if we could restore ansi processing on Windows unless the terminal is redirected, that'd be wonderful.
> 
> It looks like there are a few ways to do this. `WT_SESSION` might be an environment variable that's set, as well as `TERM_PROGRAM`.

Sublime doesn't appear to set `TERM` or `TERM_PROGRAM` on Windows, and on Linux `TERM_PROGRAM=tmux`, which is what it was before I started Sublime. There's no mention of `sublime` in the environment on either platform, other than it's certificate store and its presence in `PATH`.

`WT_SESSION` is set on Windows under Sublime, but it's also set under cmd.exe itself.

Kelimion:
So we could say enable color (virtual terminal process) on Windows by default, and if you want to disable it, `set NO_COLOR` for the duration of your build script if run from your editor, but that's not ideal.

Kelimion:
> > > So if we could restore ansi processing on Windows unless the terminal is redirected, that'd be wonderful.
> > 
> > 
> > It looks like there are a few ways to do this. `WT_SESSION` might be an environment variable that's set, as well as `TERM_PROGRAM`.
> 
> Sublime doesn't appear to set `TERM` or `TERM_PROGRAM` on Windows, and on Linux `TERM_PROGRAM=tmux`, which is what it was before I started Sublime. There's no mention of `sublime` in the environment on either platform, other than it's certificate store and its presence in `PATH`.
> 
> `WT_SESSION` is set on Windows under Sublime, but it's also set under cmd.exe itself.

It gets a bit more complicated. If I run Sublime from the start menu (Windows), `WT_SESSSION` is not set. But if I start it with `subl .` from the command shell, it inherits that environment and it is set. Either way, ``WT_SESSSION``'s absence or presence is not a useful signal.

Feoramund:
I may be misunderstanding the API, but the issue I see with using `GetConsoleMode` for color depth detection is that it is dependent on a stream. Do we just check both `stderr` and `stdout` if one happens to be redirected?

At least https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences indicates that 16 colors are supported, so we have a definitive answer on how many.

Kelimion:
I found a proper detection for `core/terminal/terminal_windows.odin`
Windows terminal: handle: 84, is_tty: true
Sublime's build output: handle: 736, is_tty: false

```odin
package terminal

import "core:os"
import "core:sys/windows"
import "core:fmt"

_is_terminal :: proc(handle: os.Handle) -> bool {
	// mode: windows.DWORD

	is_tty := windows.GetFileType(windows.HANDLE(handle)) == windows.FILE_TYPE_CHAR
	fmt.printfln("handle: %v, is_tty: %v", handle, is_tty)

	return is_tty
	// return bool(windows.GetConsoleMode(windows.HANDLE(handle), &mode))
}
```

Kelimion:
> I may be misunderstanding the API, but the issue I see with using `GetConsoleMode` for color depth detection is that it is dependent on a stream. Do we just check both `stderr` and `stdout` if one happens to be redirected?
> 
> At least https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences indicates that 16 colors are supported, so we have a definitive answer on how many.

Yes, I think just hardcode the color depth to 16 and check whether one or both are redirected with `windows.GetFileType(windows.HANDLE(handle)) == windows.FILE_TYPE_CHAR`. Then if not redirected, switch on the ANSI processing with that `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag.

Kelimion:
Great. :-) That alone didn't bring back the colors or animation in the terminal, however.
image

One small change still needed. At the bottom of `get_color_depth`:
```odin
	when ODIN_OS == .Windows {
		// This is available unconditionally, but you do have to enable it using
		// win32.SetConsoleMode(handle, old_mode | win32.ENABLE_VIRTUAL_TERMINAL_PROCESSING)`,
		// and restore the old mode on exit to restore the blinking cursor
		return .Four_Bit
	}
```
If we add this, we get the colorful animated test runner we know and love, and nice plain text in Sublime's build tab.
image
image

Kelimion:
Sure, you could do something like this in `core:terminal`'s `@(init)`
```odin
old_stdout_mode: u32
new_stdout_mode: u32
if win32.GetConsoleMode(stdout, &old_out_mode) {
			win32.SetConsoleMode(stdout, old_stdout_mode | win32.ENABLE_VIRTUAL_TERMINAL_PROCESSING)
			win32.GetConsoleMode(stdout, &new_stdout_mode) 
			color_mode_enabled = new_stdout_mode & win32.ENABLE_VIRTUAL_TERMINAL_PROCESSING == win32.ENABLE_VIRTUAL_TERMINAL_PROCESSING
}
```

Kelimion:
> The problem I have with hard-coding a color depth for Windows is that someone could use a dumb terminal on Windows or some alternative, which should override that hard-coded value.

I agree, and that's why I put it right at the bottom of the proc, so you could still set `NO_COLOR`. But the test of trying to set virtual terminal processing and checking if it took is better.

Feoramund:
I think this ought to do it. A good upgrade for the 1-year anniversary of the new test runner.

Kelimion:
> I think this ought to do it. A good upgrade for the 1-year anniversary of the new test runner.

I'll give it a bash.

Kelimion:
And also checked on Linux from within `tmux` (colors + animations present), and in Sublime's build output (nice plain text). <3

Merging...

Kelimion:
> The CI test logs are cleaner now too.
You weren't kidding. Very nice. 😁

Some of the test runners almost, but not quite, support ANSI, and those used to leave behind some escape sequences or colored snippets. Clean as a whistle now.

Feoramund:
> > The CI test logs are cleaner now too.
> 
> You weren't kidding. Very nice. 😁

I noticed a `core:log`-style error (`[ERROR] --- `) in a CI log recently that got highlighted by GitHub now that the ANSI codes are gone which I thought was neat.

> Some of the test runners almost, but not quite, support ANSI, and those used to leave behind some escape sequences or colored snippets. Clean as a whistle now.

Some supported color, but it probably got in the way more than anything. All tidy now.

PR #5192, opened 2025-05-21
COM & Audio useful bindings

Badaxis:
PR for some missing bindings useful to work with COM & Audio stuff on Windows.
This is also in preparation for WASAPI bindings, I will update XAudio2 accordingly afterwards.


core/sys/windows/ole32.odin
```diff
@@ -44,14 +44,10 @@ foreign Ole32 {
 	CoInitializeEx :: proc(reserved: rawptr = nil, co_init: COINIT = .APARTMENTTHREADED) -> HRESULT ---
 	CoUninitialize :: proc() ---
 
-	CoCreateInstance :: proc(
```
gingerBill, line 47: Why was this changed?

Badaxis: It is just for formatting consistency, I can revert it if it is unwanted.

PR #5193, opened 2025-05-21
Also use `ENABLE_PROCESSED_OUTPUT` on Windows terminals

Feoramund:
This is specified to be necessary when using `ENABLE_VIRTUAL_TERMINAL_PROCESSING`.


Kelimion:
Wonders never cease. Your round of `core:testing` patches have even fixed the *BSD VM boot loop. ;-)

PR #5202, opened 2025-05-22
Add `-build-only`, `-keep-test-executable`, delete test executable after running

### LINKED ISSUE #2035: allow `-out` with `odin test` for easier test debugging
## Summary
When debugging tests I have to point the debugger at

`{root}/source/.exe` (the default test executable location)

instead of a more sane custom path like

`{root}/test.exe`

Allowing `-out` with `odin test` would make this possible, but I'm not sure if it's in the spirit of `odin test`, seeing that it was disallowed in the original commit (https://github.com/odin-lang/Odin/commit/2bd0fd932a4eb1c2e6bc8702450c230186bb3c44). The fix is to remove `&~ Command_test` from here:
https://github.com/odin-lang/Odin/blob/81e3b64ecda857458b8e0762bdf6156cdaa1cb74/src/main.cpp#L750

Side note, one awkward thing about debugging tests is they get run twice: first when calling `odin test`, which builds and runs the test executable, and again when debugging. To get around this there would need to be a `odin test-build` and `odin test-run` similar to `odin build` and `odin run`. Right now it's not a big deal and there's benefits to having a cleaner CLI with just `odin test`. Anyway, that's a separate matter.

## Report
```
Odin: dev-2022-09:0bc660c7
OS:   Windows 11 Professional (version: 21H2), build 22000.856
CPU:  Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
RAM:  32605 MiB
```

Feoramund:
Closes #2035 (the @Kelimion suggestions; we already have `-out` as requested in the original post.)

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -441,6 +441,7 @@ struct BuildContext {
 	String extra_assembler_flags;
 	String microarch;
 	BuildModeKind build_mode;
+	bool   build_only;
 	bool   generate_docs;
 	bool   custom_optimization_level;
 	i32    optimization_level;
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -442,6 +442,7 @@ struct BuildContext {
 	String microarch;
 	BuildModeKind build_mode;
 	bool   build_only;
+	bool   keep_test_executable;
 	bool   generate_docs;
 	bool   custom_optimization_level;
 	i32    optimization_level;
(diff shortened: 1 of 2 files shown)


Kelimion:
@laytan reminded me I added `odin build normal.odin -file -build-mode:test`, so we now have two ways to build but not run the test executable.

I'm going to back out the newly added `-build-only` so we don't have several ways to do the same thing, but will keep the default cleanup and `-keep-test-executable` option.

Further, I'll add `-build-mode:test` to `odin help build`.

Kelimion:
See https://github.com/odin-lang/Odin/commit/12167bace05915abda62c93881d711e993e062cd.

Kelimion:
I've also just renamed `-keep-test-executable` to `-keep-executable`.

`odin run` and `odin test` now both clean up the generated executable unless directed otherwise.

Issue #5017, opened 2025-04-07
Regex .Multiline consuming \n with $


jist99:
## Current behaviour
This code:
```odin
    str := "foo1\nfoo2\nfoo3\nfoo4"
    it, _ := regex.create_iterator(str, `^foo.$`, {.Multiline})
    for match in regex.match(&it) {
        fmt.println(match)
    }
```
Gives the output:
```
Capture{pos = [[0, 5]], groups = ["foo1\n"]}
Capture{pos = [[9, 15]], groups = ["\nfoo3\n"]}
```
So foo2 and foo4 are getting skipped.

I suspect that this is because .Multiline is causing $ to consume \n meaning we only get 2 of the 4 captures.

## Expected behaviour
I expect that we get an output containing all 4 groups with `foo1, foo2, foo3, foo4`.

## Context
Latest main:
        Odin:    dev-2025-04:2b26c0b39
        OS:      macOS Sequoia 15.3.2 (build 24D81, kernel 24.3.0)
        CPU:     Apple M1
        RAM:     16384 MiB
        Backend: LLVM 14.0.6

Kelimion:
Those aren't supported by the iterator.

jist99:
What isn't supported exactly, and why? This feels like a reasonable thing to do.

Kelimion:
`.Multiline` isn't supported in the iterator at the moment.

It's not about it being a reasonable ask, but that the iterator was added in response a to non-multiline iteration request, and that was simple enough to add and do so in an efficient way. If someone wants to make it work without this caveat you encountered, we'll happily take the patch.

PR #5209, opened 2025-05-24
Fix RegEx iterator, remove `.Global`, make patterns unanchored by default (breaking change)

### LINKED ISSUE #4557: Unexpected behaviour in `core:text/regex` regarding global flag
Over the last few days (advent of code meaning new people or existing people trying the package and language out) we've had a few questions from people regarding unexpected behaviour in `core:text/regex`. To be clear with "unexpected" I mean different from most/all other implementations.

# Problems

## Patterns are anchored by default

Patterns seem to be anchored to the start of the given input by default, and the behaviour people expect from regex is that a pattern matches wherever in the input (which at the moment requires the .Global flag, more on that later).

For example, the following assertion fails:

```odin
rg, err := regex.create_by_user(`/world/`)
assert(err == nil)

_, ok := regex.match(rg, "hello world")
assert(ok) // Fails, what??
```

The expected behaviour is for it to match. And if one wants it anchored they would use `/^world/` to say so.

## The global flag is either unfortunately named, or wrong

The global flag and `/g` currently seem to mean "match wherever in the input" (which should be the default, see previous point). The expected behaviour is for `/g` in regex to mean "do not stop after the first match".

One would expect this:

```odin
rg, err := regex.create_by_user(`/(hello\d)/g`) 
assert(err == nil)

captures, ok := regex.match(rg, "hello0, hello1, hello2")
assert(ok)
fmt.println(captures) // ["hello0, hello1, hello2", "hello0", "hello1", "hello2"]
```

but that is not the behaviour.

# Suggestions

For the first point I think that the entire global flag should not exist and anchors should be done in the pattern, I believe that's how it works everywhere and is what people expect. If it is kept, it should at least be renamed so it doesn't conflict with what people expect the "global" or /g flag to do.

For the other point I would say either make the Global (/g) flag do what was expected or remove the flag entirely and outline that this is out of scope.

Feoramund:
Fixes #4557

After a bit of research, it looks like the `g` flag on some RegEx implementations has more to do with the RegEx being an iterable object in its own right to carry runtime state—not a pattern that matches anywhere (implicit `.*?` at the start). My implementation has all the runtime state on the virtual machine. Removing this flag will end the confusion about how to iterate with RegEx patterns.

I've also fixed a couple infinite loops in the iterator and made it such that you no longer need to check against some expected number of patterns before breaking out. You can iterate just like any normal iterator now.

Removing `.Global` also comes with making all patterns unanchored by default; giving some more use to the `^` symbol and hopefully being more expected of an API.

NOTE: This change is breaking against anyone using bare patterns with no `Global` flag, as previously flagless `abc` would render to `^abc` but now it is `.*?abc`.

---

I will revisit #5017 / #5027 again after this is reviewed.

### DIFF
--- a/core/text/regex/common/common.odin
+++ b/core/text/regex/common/common.odin
@@ -15,8 +15,6 @@ MAX_PROGRAM_SIZE   :: int(max(i16))
 MAX_CLASSES        :: int(max(u8))
 
 Flag :: enum u8 {
-	// Global: try to match the pattern anywhere in the string.
-	Global,
 	// Multiline: treat `^` and `$` as if they also match newlines.
 	Multiline,
 	// Case Insensitive: treat `a-z` as if it was also `A-Z`.
@@ -36,7 +34,6 @@ Flags :: bit_set[Flag; u8]
 
 @(rodata)
 Flag_To_Letter := #sparse[Flag]u8 {
-	.Global            = 'g',
 	.Multiline         = 'm',
 	.Case_Insensitive  = 'i',
 	.Ignore_Whitespace = 'x',
--- a/core/text/regex/compiler/compiler.odin
+++ b/core/text/regex/compiler/compiler.odin
@@ -401,7 +401,7 @@ compile :: proc(tree: Node, flags: common.Flags) -> (code: Program, class_data: 
 
 	pc_open := 0
 
-	add_global: if .Global in flags {
+	optimize_opening: {
 		// Check if the opening to the pattern is predictable.
 		// If so, use one of the optimized Wait opcodes.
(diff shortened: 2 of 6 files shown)


Kelimion:
This is excellent. Thanks for revisiting it, @Feoramund.

Kelimion:
I'm curious how you intend to address `.Multiline` iteration.

I also wonder what would happen if `\n` is part of the needle, to e.g. match `^(foo).*\nbaz` in `foobar\nbarzle\nfoo\nfoobaz\n` to only return matches where a line starts with `foo` and the next line starts with `baz`.

Not that I know why you'd want that necessarily.

PR #5217, opened 2025-05-25
Add multicast support for windows

shiver:
I am working on an application that requires multicast support for Windows, Linux & Darwin platforms, and potentially more down the road.

This PR adds support for the following multicast related Socket_Options: 
- `.Multicast Interface`
- `.Multicast Loop`
- `.Multicast_TTL`
- `.Join_Multicast_Group`
- `.Leave_Multicast_Group`

It also adds support some rudimentary support for `win.getsockopt`, which can be used through the new `net.get_option` function.

What is missing:
- Remaining multicast socket options
- Support for IPv6 multicast options
- `get_option` support for the remaining valid `Socket_Option`s 
- Support for other platforms

There is more to add, but I suspect these changes are enough for what most would want from multicast support, and the remainder wouldn't be too tough to add from here. 

Fairly new to Odin, so happy to take on any feedback you might have. I wanted to use this PR as a bit of a litmus test to see what the appetite for taking on something like this would be. If people are happy with the changes, I can also take a look at adding support for Linux & Darwin platforms too. Alternatively, I'm also happy for this to be used as a proof-of-concept for someone on the team to implement instead.


shiver:
Addressed the issue where the test was trying to run on as yet unsupported platforms.

PR #5245, opened 2025-05-31
Fix odin.js loadCstring to use pointer address correctly.

TheTophatDemon:
### Problem
The `loadCstring` function of the Odin WASM memory interface did not work correctly because it used the `loadPtr` function, which actually loads memory at the given address and uses that as the pointer, even though the real address of the pointer is already given as the parameter.
### Solution
I turned `loadCstringDirect` into `loadCstring`, since `loadCstringDirect` takes the pointer address directly instead of going through `loadPtr`. I would question the need for `loadPtr` at all, but there are currently two other functions that use it: `getSource` and `TransformFeedbackVaryings` from the WebGL interface objects. These functions are tricky to test on their own, as I haven't found any examples of them being used, so I decided to leave them alone for the time being.
### Testing
I discovered this bug while writing the following simple program testing the interop between Odin in WASM. I just had a local web server running with these files:

- odin.js from this repository's /core/sys/wasm/js

- main.odin:
```Odin
package lib

@(export)
get_message :: proc() -> cstring {
    return "Weiner time"
}
```

I compiled the Odin code into a WASM module using this command:
`odin.exe build . -debug -target:js_wasm32 -build-mode:lib -vet -out:game`

- index.html:
```
<html lang="en">
	<head>
		<title>Odin Canvas API test</title>
		<script type="text/javascript" src="odin.js"></script>
		<script type="text/javascript">
			(async () => {
				const memInterface = new odin.WasmMemoryInterface();
				await odin.runWasm('game.wasm', null, null, memInterface);
				const msgPtr = memInterface.exports.get_message();
				const msg = memInterface.loadCstring(msgPtr);
				odinMessage.innerHTML = msg?.toString();
			})();
		</script>
	</head>
	<body>
		<p>Message from Odin: <span id="odinMessage"></span></p>
	</body>
</html>
```

Before my change, a console error like this would appear and the message would be blank:
image

After my change, the message appears and there is no error:
image

### DIFF
--- a/core/sys/wasm/js/odin.js
+++ b/core/sys/wasm/js/odin.js
@@ -110,7 +110,7 @@ class WasmMemoryInterface {
 	}
 
 	loadCstring(ptr) {
-		return this.loadCstringDirect(this.loadPtr(ptr));
+		return this.loadCstringDirect(ptr);
 	}
 
 	loadCstringDirect(start) {
--- a/core/sys/wasm/js/odin.js
+++ b/core/sys/wasm/js/odin.js
@@ -110,16 +110,12 @@ class WasmMemoryInterface {
 	}
 
 	loadCstring(ptr) {
-		return this.loadCstringDirect(ptr);
-	}
-
-	loadCstringDirect(start) {
-		if (start == 0) {
+		if (ptr == 0) {
 			return null;
 		}
 		let len = 0;
-		for (; this.mem.getUint8(start+len) != 0; len += 1) {}
-		return this.loadString(start, len);
+		for (; this.mem.getUint8(ptr+len) != 0; len += 1) {}
+		return this.loadString(ptr, len);
 	}
 
 	storeU8(addr, value)  { this.mem.setUint8  (addr, value); }


TheTophatDemon:
> This was not a bug. The loadPtr was done because if your string is inside a struct that's passed in you offset from the start address of the struct and use loadCstring in which case you do need it to first loadPtr.

I am having trouble understanding this explanation. The parameter to `loadCstring` was named `ptr`, which led me to believe that it was expecting a pointer -- an address into WASM memory -- to be passed in. Surrounding functions like `loadB32`, `loadBytes` appear to treat the `ptr` argument this way. When you say a struct could be being passed in, do you mean something like an Odin string, where the struct has a pointer and a length? I would typically expect a C string to be represented as a plain pointer, and the example function that I wrote in Odin returning a `cstring` returns a memory address on the Javascript side. So, could you explain what scenario is `loadCstring` meant to be used in?

It was not clear to me at first that `loadPtr` was meant to take a 'pointer to a pointer', but I understand now why it is needed.

PR #5246, opened 2025-05-31
Posix: Signal: add SIGWINCH support

RichardFevrier:
Adding SIGWINCH values for posix allows us to grab resize events with ioctl (not included in this commit but i can make another one if you want to add it to the core).

### DIFF
--- a/core/sys/posix/signal.odin
+++ b/core/sys/posix/signal.odin
@@ -50,7 +50,7 @@ foreign lib {
 
 	/*
 	Send a signal to a thread.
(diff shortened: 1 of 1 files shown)


laytan:
SIGWINCH seems to be standard since POSIX 2024 (edition 8) and this package was made against POSIX 2019 (edition 7). It's fine to merge and add this but we may want to look at a full update at some point.

RichardFevrier:
Thank you for the very informative data.

PR #5261, opened 2025-06-02
add a missing nullptr check

spiel0meister:
When I build my package with `odin build .`, the compiler segfaults. The reason was a missing `nullptr` check. After that fix, the package now builds, and the compiler does not segfault anymore (at least in my scenario).

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -263,11 +263,13 @@ gb_internal void check_did_you_mean_scope(String const &name, Scope *scope, char
 
 gb_internal Entity *entity_from_expr(Ast *expr) {
 	expr = unparen_expr(expr);
-	switch (expr->kind) {
-	case Ast_Ident:
-		return expr->Ident.entity;
-	case Ast_SelectorExpr:
-		return entity_from_expr(expr->SelectorExpr.selector);
+	if (expr != nullptr) {
+		switch (expr->kind) {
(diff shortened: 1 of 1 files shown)


Kelimion:
Do you know what about your package it is that caused this?
This nullptr check may actually fix the issue, or be a bandaid over another one.

spiel0meister:
I am sorry, but I don't know what part of my package caused this. That was the best effort to make the compiler actually build the package and it was an easy fix to do.

Kelimion:
Okay, before we merge this, two small issues I hope you can remedy.
1) In the Odin code style, and the Odin compiler as well, `case` is on the same column as `switch`.
2) Instead of putting everything inside of the nullptr check, we tend to return early.

```cpp
gb_internal Entity *entity_from_expr(Ast *expr) {
	expr = unparen_expr(expr);
	if (expr == nullptr) {
		return nullptr;
	}	
	switch (expr->kind) {
	case Ast_Ident:
		return expr->Ident.entity;
	case Ast_SelectorExpr:
		return entity_from_expr(expr->SelectorExpr.selector);
	}
	return nullptr;
}

spiel0meister:
There you go, done! Sorry for the mismatch in code stlye.

Feoramund:
Are you able to get a stack trace at least?

PR #5264, opened 2025-06-02
Fix scope attribute proc grouping (#5243 )

### LINKED ISSUE #5243: Scope attribute @(private) for procedure grouping.
## Context

Let's say I have a package file with the following content:

```odin
// private_reproduction.odin
#+private //Should make everything inside file private to package 'reproduction'
package reproduction


import "core:fmt"


foo :: proc(text : string) {
    fmt.printfln("Text: %v", text)
}


bar :: proc(number : int) {
    fmt.printfln("Number: %v", number)
}


foobar :: proc {
    foo,
    bar,
}
``` 

Now, I have the following file on the same package:

```odin
// reproduction.odin
package reproduction


do_something_with_foobar :: proc() {
    foobar("Helope")
    foobar(2025)
}
```

`foobar` should possibly be used inside package, external usage should be interfaced by `do_something_with_foobar` proc.

Yet, this is still possible:

```odin
// main.odin
package test


import rep "reproduction"


main :: proc() {
    rep.foobar(10)
}
```

Why is that possible?
Note that even if I add `@(private)` to foobar declaration, it's the same behaviour.
Proc groups are only being affected by `@(private="file")`.

* Operating System & Odin Version: Solus OS 4.7 Endurance 64bit / odin version dev-2025-05-nightly

## Expected Behavior

Proc groups inside `#+private` scopes or with `@(private)` annotations should be private to the package and not exposed outside.

## Current Behavior

Proc groups inside `#+private` scopes or with `@(private)` annotations are public.

LeandroLibanio28H:
Procedure Groups were being checked before checking exporting from package.
Fixing checking order.

Removed unreachable block:

```
if (entity->kind == Entity_ProcGroup) { ... }
``` 
fixes #5243

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -5394,37 +5394,6 @@ gb_internal Entity *check_selector(CheckerContext *c, Operand *operand, Ast *nod
 				return entity;
 			}
 			GB_ASSERT_MSG(entity->type != nullptr, "%.*s (%.*s)", LIT(entity->token.string), LIT(entity_strings[entity->kind]));
-
-			if (entity->kind == Entity_ProcGroup) {
-				Array<Entity *> procs = entity->ProcGroup.entities;
-				bool skip = false;
-				for (Entity *p : procs) {
-					Type *t = base_type(p->type);
-					if (t == t_invalid) {
-						continue;
-					}
-
-					Operand x = {};
-					x.mode = Addressing_Value;
-					x.type = t;
-					if (type_hint != nullptr) {
-						if (check_is_assignable_to(c, &x, type_hint)) {
-							entity = p;
-							skip = true;
-							break;
-						}
-					}
-				}
-
-				if (!skip) {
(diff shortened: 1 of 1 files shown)


src/check_expr.cpp
```diff
@@ -1,3 +1,5 @@
+#include <iostream>
```
Kelimion: This was left over from debugging the patch?

LeandroLibanio28H: Yes, It was. I didn't have a debugger set up.  Sorry. Just fixed.

Kelimion: It happens. Although usually it's a stray `gb_printf`. :-)

LeandroLibanio28H: I presume `gb_printf` is specific to Odin code. It's my first time looking at the code base. Next PR I'll make sure to left over a `gb_printf`call out of respect.

PR #5281, opened 2025-06-04
Guard against negative `index` in `inject_at`

### LINKED ISSUE #4595: inject_at writes out of bounds when given a negative index argument
## Context

The following program writes out of bounds. Is this expected behavior? An assert could be added to the `inject_at` family of procedures but I see that there is a `pop` and `pop_safe` which leads me to believe an assert isn't the favored approach.

```
package main

import "core:fmt"

main :: proc() {
    test: [dynamic]int

    inject_at(&test, 0, 50)
    fmt.printf("%v\n", test)
    
    inject_at(&test, 1, 100)
    fmt.printf("%v\n", test)
    
    inject_at(&test, -1, 150)
    fmt.printf("%v\n", test)
}
```

Odin:    dev-2024-12:0a29d36aa
OS:      macOS Sonoma 14.4.1 (build 23E224, kernel 23.4.0)
CPU:     Apple M2
RAM:     24576 MiB
Backend: LLVM 18.1.8

## Expected Behavior

I expected inject_at to not write out of bounds and return false or panic at runtime.

## Current Behavior

inject_at will write out of bounds given a negative index

### Steps to Reproduce

1. Write the above example to a file
2. odin run example.odin -file

### Failure Logs

On my machine, the output is:
[50]
[50, 100]
[105553138418384, 50, 100]

Feoramund:
Fixes #4595

### DIFF
--- a/base/runtime/core_builtin.odin
+++ b/base/runtime/core_builtin.odin
@@ -648,6 +648,9 @@ append_nothing :: proc(array: ^$T/[dynamic]$E, loc := #caller_location) -> (n: i
 
 @builtin
 inject_at_elem :: proc(array: ^$T/[dynamic]$E, #any_int index: int, #no_broadcast arg: E, loc := #caller_location) -> (ok: bool, err: Allocator_Error) #no_bounds_check #optional_allocator_error {
+	when !ODIN_NO_BOUNDS_CHECK {
+		ensure(index >= 0, "Index must be positive.", loc)
+	}
 	if array == nil {
 		return
 	}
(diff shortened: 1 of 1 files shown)


Kelimion:
Just after merging I realized that we might want to use `assert`  instead of `ensure` so it gets compiled out with either `-no-bounds-check` or `-disable-assert`. Not a big deal as release versions will likely use both of these and so bypass `ensure` as well.

Feoramund:
I wasn't 100% sure if I should've used an assert or an ensure behind the bounds-check constant, but I'd say it's your call if you think the other is better.

PR #5284, opened 2025-06-05
[parser.cpp] - packages with `.odin` in the name no longer attempt to parse as odin files

A1029384756:
previously, if the following directory structure existed:
```
├──  my_fancy_package.odin/
└──  main.odin
```
the compiler would attempt to parse the directory as if it were a file, this adds checks to ensure that the files are in fact files before adding them to the parser's task list

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -5794,7 +5794,7 @@ gb_internal AstPackage *try_add_import_path(Parser *p, String path, String const
 	for (FileInfo fi : list) {
 		String name = fi.name;
 		String ext = path_extension(name);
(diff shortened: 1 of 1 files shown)


Kelimion:
Looks different from but related to https://github.com/odin-lang/Odin/pull/5198.

Feoramund:
I can confirm a segfault on Linux when trying to build `.` in a directory, with an `.odin`-suffixed directory inside that directory.

It should not be related to #5198 (as in caused by).

Feoramund:
No more segfault on my end after trying this patch.

PR #5286, opened 2025-06-05
Add `@(no_sanitize_memory)` with additions to `base:sanitizer`

### LINKED ISSUE #3885: Bug in the if with atribution if the variable is declared outside the if.
## Context

 Operating System & Odin Version: Linux OpenSuse  
 My version of Odin is: odin version dev-2024-05:c091b0d06
 Please paste `odin report` output:

```
 make run
./main.exe
==123765==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x4de60d in os.read_entire_file_from_handle /home/joaocarvalho/odin/Odin/core/os/os.odin:117:2
    #1 0x4dd943 in os.read_entire_file_from_filename /home/joaocarvalho/odin/Odin/core/os/os.odin:105:2
    #2 0x4d654a in main.main /home/joaocarvalho/odin_proj/tests/test_read_entire_file.odin:9:5
    #3 0x596df1 in main /home/joaocarvalho/odin/Odin/base/runtime/entry_unix.odin:54:4
    #4 0x7f54bc02a1ef in __libc_start_call_main (/lib64/libc.so.6+0x2a1ef) (BuildId: a2c0942c27fb9483b47886a1b937337a797bbceb)
    #5 0x7f54bc02a2b8 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2a2b8) (BuildId: a2c0942c27fb9483b47886a1b937337a797bbceb)
    #6 0x4360f4 in _start /home/abuild/rpmbuild/BUILD/glibc-2.39/csu/../sysdeps/x86_64/start.S:115

SUMMARY: MemorySanitizer: use-of-uninitialized-value /home/joaocarvalho/odin/Odin/core/os/os.odin:117:2 in os.read_entire_file_from_handle
Exiting
make: *** [Makefile:5: run] Error 1
```

## Expected Behavior

## Current Behavior

What is the current behavior?

If you run the following code compiled with 

```
-sanitize:memory
```

```
package main

import "core:fmt"
import "core:os"

main :: proc ( ) {
    file_path := "test_read_entire_file.odin"

    data, sucess := os.read_entire_file_from_filename( file_path )
}
```

It will give the following warning:

```
==123765==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x4de60d in os.read_entire_file_from_handle /home/joaocarvalho/odin/Odin/core/os/os.odin:117:2
    #1 0x4dd943 in os.read_entire_file_from_filename /home/joaocarvalho/odin/Odin/core/os/os.odin:105:2
    #2 0x4d654a in main.main /home/joaocarvalho/odin_proj/tests/test_read_entire_file.odin:9:5
```

And looking at the code, we can see that this is a problem (length) inicialized outside the if with atribution:

```
read_entire_file_from_handle :: proc(fd: Handle, allocator := context.allocator, loc := #caller_location) -> (data: []byte, success: bool) {
   context.allocator = allocator

   length: i64
   err: Errno
   if length, err = file_size(fd); err != 0 {
       return nil, false
   }

   if length <= 0 {
      return nil, true
   }
```

It is saying that length scope is inside the if and that it is shadoing so the outside lenght is zero or unitialized.

Best regards,
João Carvalho

Feoramund:
Given the extreme limited utility of this feature (marking uninitialized memory that will be filled by a kernel syscall as safe), I've only implemented the bare essentials.

Fixes #3885

The test suite should now pass with `-sanitize:memory` enabled on Linux. Unfortunately, `libc` on FreeBSD triggers MSan when calling `posix.glob`, so it does not pass there too.

### DIFF
--- a/base/sanitizer/memory.odin
+++ b/base/sanitizer/memory.odin
@@ -1,0 +1,74 @@
+#+no-instrumentation
+package sanitizer
+
+@(private="file")
+MSAN_ENABLED :: .Memory in ODIN_SANITIZER_FLAGS
+
+@(private="file")
+@(default_calling_convention="system")
+foreign {
+	__msan_unpoison :: proc(addr: rawptr, size: uint) ---
+}
+
+/*
+Marks a slice as fully initialized.
+
+Code instrumented with `-sanitize:memory` will be permitted to access any
+address within the slice as if it had already been initialized.
+
+When msan is not enabled this procedure does nothing.
+*/
+memory_unpoison_slice :: proc "contextless" (region: $T/[]$E) {
+	when MSAN_ENABLED {
+		__msan_unpoison(raw_data(region),  size_of(E) * len(region))
+	}
+}
+
+/*
+Marks a pointer as fully initialized.
+
+Code instrumented with `-sanitize:memory` will be permitted to access memory
+within the region the pointer points to as if it had already been initialized.
+
+When msan is not enabled this procedure does nothing.
+*/
+memory_unpoison_ptr :: proc "contextless" (ptr: ^$T) {
+	when MSAN_ENABLED {
+		__msan_unpoison(ptr, size_of(T))
+	}
+}
+
+/*
+Marks the region covering `[ptr, ptr+len)` as fully initialized.
+
+Code instrumented with `-sanitize:memory` will be permitted to access memory
+within this range as if it had already been initialized.
+
+When msan is not enabled this procedure does nothing.
+*/
+memory_unpoison_rawptr :: proc "contextless" (ptr: rawptr, len: int) {
+	when MSAN_ENABLED {
+		__msan_unpoison(ptr, uint(len))
+	}
+}
+
+/*
+Marks the region covering `[ptr, ptr+len)` as fully initialized.
+
+Code instrumented with `-sanitize:memory` will be permitted to access memory
+within this range as if it had already been initialized.
+
+When msan is not enabled this procedure does nothing.
+*/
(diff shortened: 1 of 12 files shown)


laytan:
IS that actually safe though? https://github.com/odin-lang/Odin/issues/3885#issuecomment-2313143702

Feoramund:
The only condition under which it would not be safe is if the syscall does not write to every bit of memory accessed outside the procedure.

laytan:
Why would they have added this error if it's considered safe?

I don't know about this one. I would rather take the miniscule overhead of zeroing these values than silence an error that could be valid.

Feoramund:
If I understand your concern correctly, you believe MSan is erroring on uninitialized memory being passed to a function, based on the quote in the comment link, is that right?

```odin
package main

import "core:fmt"
import "core:sys/posix"

main :: proc() {
	st: posix.stat_t = ---
	posix.stat("/dev", &st)
	fmt.printfln("%#v", st)
}
```
`odin run . -sanitize:memory`

This will not raise any error.

To the degree that I understand MSan, it will error only on memory that is accessed without being written to first—except for syscalls; those are opaque to it (same situation for TSan with regards to `munmap`-ed memory being passed around different threads).

That said, I'm not opposed to removing the usage of `= ---` in `core:os` and elsewhere. It likely isn't worth the performance, as you say, to rely on every kernel always filling out every field. Though, there may be uses yet for passing uninitialized memory to the system where it's known without a doubt to be fully initialized upon return.

PR #5291, opened 2025-06-06
Deprecate old @(deprecated) things.

Kelimion:


core/sys/windows/ws2_32.odin
```diff
@@ -242,10 +242,4 @@ foreign ws2_32 {
 	ntohl :: proc(netlong: c_ulong) -> c_ulong ---
 	// [MS-Docs](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-ntohs)
 	ntohs :: proc(netshort: c_ushort) -> c_ushort ---
-	// [MS-Docs](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-htonl)
-	@(deprecated="Use endian specific integers instead, https://odin-lang.org/docs/overview/#basic-types")
```
laytan, line 246: Think these should be kept in, it's not actually deprecated but it's telling the user Odin has a better way. If somebody is translating some C code using this windows api and see it is not a symbol they may think we just haven't added a binding.

Kelimion: Fair. I'll re-add them with a comment instead of `@(deprecated)` so we won't be tempted to remove them again.

PR #5296, opened 2025-06-07
Add `@(default_context)` and `intrinsics.__default_context`

### LINKED ISSUE #5136: Can't control allocator used in procs tagged with `@(init)` and globals
There is no way to make a proc tagged with `@(init)` use a custom allocator. For example, there are packages in core that use `@(init)` to allocate some memory.

Below is an example:
```
package allocator_issue

import "base:runtime"

my_arr: [dynamic]int

// Imagine this happens in some package that is out of my control.
@(init)
run_before_main :: proc() {
	append(&my_arr, 5)
}

main :: proc() {
	// There's no good way to make this apply to `run_before_main`
	context.allocator = runtime.panic_allocator()
	assert(my_arr[0] == 5)
}
```

Similarly, if you have a global written like so:
```
arr := make([dynamic]int)
```
then there is no way to make it use a custom `context.allocator`, since `main` hasn't been run yet.

Another example is global dynamic literals:

```
global_map := map[int]float {
	5 = 3.2,
	7 = 1.2,
}
```

It will initialized with the platform's default allocator. There is no good way to override which one it uses.

## Suggested solutions

There have been multiple ways suggested to implement this, such as:
- Add a compiler command line flag that specifies a procedure to run before `@(init)`, i.e. you name a  "pre-init" proc name.
- Make it possible to provide a custom entry point procedure, so that you can do special things in there. I.e. look at `main` in `base/runtime/entry_windows.odin` etc, and figure out to inject a custom entry procedure. However, that would maybe be platform-dependent.

Feoramund:
The `__default_context` intrinsic calls the procedure marked with`@(default_context)` if it exists and returns true. If it does not exist, it returns false.

The `@(default_context)` is a file-scope procedure attribute that designates a procedure fitting the signature of `proc "contextless" (^runtime.Context)` as the procedure which will initialize all default contexts used throughout the entire program, including `@(init)` initializers, global variables, shared library initializers and shared global variables.

This change gives the programmer full control over the Odin context, and hence allocation, throughout all initialization that happens before `main` is reached.

Currently, `@(default_context)` is only supported in full executable builds, including `-build-mode:test`.

---

Fixes #5136 

Presently, the test suite I added for this is failing only on Windows. ~~I am unable to debug it for lack of the platform, so~~ any help is appreciated.

Given that only the shared global and shared `@(init)` are failing, I suspect there are linkage particulars involved here, possibly order-related. I had to set `__init_context` to be linkage `"weak"` by default and `"strong"` for the executable build in order to link correctly on Linux with a single module build in `-o:speed` mode. I tried setting it to `"internal"` and `"link_once"` for Windows only instead of `"weak"` to exhaust the other linkage options, but neither proved to work on my testing commits against the CI; ~~someone on Windows will have to investigate this, unfortunately.~~

Pinging @karl-zylinski to see if this works for the original problem and @laytan as I've iterated on his original idea.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,7 @@ jobs:
           ./odin test tests/core/normal.odin -file -all-packages -vet -strict-style -disallow-do -define:ODIN_TEST_FANCY=false -define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true
           ./odin test tests/core/speed.odin -file -all-packages -vet -strict-style -disallow-do -o:speed -define:ODIN_TEST_FANCY=false -define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true
           ./odin test tests/vendor -all-packages -vet -strict-style -disallow-do -define:ODIN_TEST_FANCY=false -define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true
+          (cd tests/default_context; ./run.sh)
           (cd tests/issues; ./run.sh)
           ./odin check tests/benchmark -vet -strict-style -no-entry-point
 
@@ -68,6 +69,7 @@ jobs:
           ./odin test tests/core/normal.odin -file -all-packages -vet -strict-style -disallow-do -define:ODIN_TEST_FANCY=false -define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true
           ./odin test tests/core/speed.odin -file -all-packages -vet -strict-style -disallow-do -o:speed -define:ODIN_TEST_FANCY=false -define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true
           ./odin test tests/vendor -all-packages -vet -strict-style -disallow-do -define:ODIN_TEST_FANCY=false -define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true
+          (cd tests/default_context; ./run.sh)
           (cd tests/issues; ./run.sh)
           ./odin check tests/benchmark -vet -strict-style -no-entry-point
   ci:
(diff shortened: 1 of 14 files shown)


Feoramund:
After some investigation, I'm getting the sense that weak linkage either does not have an effect on Windows, or there is something blocking DLLs from accessing the main program if it does work. That's a shame, since this solution entirely relies on it.

The only other solution I can think of would be to remove `DllMain` in favor of an exported `setup` procedure that takes an optional contextless procedure that anything in the DLL can call in order to setup new contexts. My limited understanding of dynamic linking tells me that this would have to be done with runtime linking, because the `setup` symbols would all be the same name on each DLL. This would have to happen in the main program's `__$startup_runtime`, probably before any of the program's initializers, in import order.

karl-zylinski:
Thank for spending time on this. For my limited use cases, this seems to work.

Feoramund:
If nobody cares about `@init` or global variables inside of DLLs allocating, then this patch would be fine as-is, but I can see that as a possible pain point in the future.

There's definitely an issue if a DLL allocates memory in the global initialization phase and passes it across the border for the main program to free, as the heap allocators may not be the same.

karl-zylinski:
@Feoramund I see. I tested it on Windows. And you are correct, `@init` in the main executable works, but not in a DLL. I do think it is something one would want, as it becomes quite inconsistent otherwise.

herohiralal:
This would be great. Personally, I've been hard-coding nil-allocator and all to ensure that the context isn't uninitialised.

gingerBill:
Is this needed any more since `@(init)` must be `proc "contextless" ()`?

PR #5298, opened 2025-06-08
Added missing parameter to `gettimeofday`

slowhei:
The Linux `gettimeofday` syscall takes 2 parmeters (`tv` and `tz`), but the Odin binding only takes 1 (`tv`). This was causing `gettimeofday` to behave inconsistently on my machine. My fix adds a binding for the missing second parameter (`tz`).

The rest of this PR goes into exactly what inconsistent behavior I observed.

### Extra Details You Likely Don't Need to Read
The following code correctly prints seconds since the Epoch in a release build but prints only "Seconds: 0" in a debug build on my machine.

```
Odin:    dev-2025-06-nightly
OS:      Ubuntu 22.04.5 LTS, Linux 6.6.87.1-microsoft-standard-WSL2
CPU:     11th Gen Intel(R) Core(TM) i7-11700F @ 2.50GHz
RAM:     32028 MiB
Backend: LLVM 20.1.6
```

```
package gettimeofday_bug

import "core:fmt"
import "core:sys/linux"

main :: proc() {
    time: linux.Time_Val
    for {
        linux.gettimeofday(&time)
        fmt.printfln("Seconds: %v", time.seconds)
    }
}
```

In a debugger, right before the `syscall` instruction is executed, `RDI` is correctly set to `&time`, but `RSI` contains a non-zero garbage value (`RDI` and `RSI` are parameters 1 and 2 in the Linux syscall ABI, I think). After executing the syscall, `time.seconds` is equal to 0. But if you change the value of `RSI` to 0 before executing the syscall, `time.seconds` is set to seconds since Epoch.

This reliance on a register whose value the compiler assumes doesn't matter would explain why `gettimeofday` behaves inconsistently depending on release or debug build.

### DIFF
--- a/core/sys/linux/sys.odin
+++ b/core/sys/linux/sys.odin
@@ -1412,8 +1412,8 @@ umask :: proc "contextless" (mask: Mode) -> Mode {
 	Get current time.
 	Available since Linux 1.0.
 */
-gettimeofday :: proc "contextless" (tv: ^Time_Val, tz: ^Time_Zone) -> (Errno) {
-	ret := syscall(SYS_gettimeofday, tv, tz)
+gettimeofday :: proc "contextless" (tv: ^Time_Val) -> (Errno) {
+	ret := syscall(SYS_gettimeofday, tv, nil)
 	return Errno(-ret)
 }
 
--- a/core/sys/linux/sys.odin
+++ b/core/sys/linux/sys.odin
@@ -1412,8 +1412,8 @@ umask :: proc "contextless" (mask: Mode) -> Mode {
 	Get current time.
 	Available since Linux 1.0.
 */
-gettimeofday :: proc "contextless" (tv: ^Time_Val) -> (Errno) {
-	ret := syscall(SYS_gettimeofday, tv)
+gettimeofday :: proc "contextless" (tv: ^Time_Val, tz: ^Time_Zone) -> (Errno) {
+	ret := syscall(SYS_gettimeofday, tv, tz)
 	return Errno(-ret)
 }
 
--- a/core/sys/linux/types.odin
+++ b/core/sys/linux/types.odin
@@ -73,14 +73,6 @@ Time_Val :: struct {
 }
 
 /*
-	Represents a time zone.
-*/
-Time_Zone :: struct {
-	minutes_west: int,
-	dst_time:     int,
-}
-
-/*
 	Access and modification times for files
 */
 UTim_Buf :: struct {
--- a/core/sys/linux/types.odin
+++ b/core/sys/linux/types.odin
@@ -73,6 +73,14 @@ Time_Val :: struct {
 }
 
 /*
+	Represents a time zone.
+*/
+Time_Zone :: struct {
+	minutes_west: int,
+	dst_time:     int,
+}
+
+/*
 	Access and modification times for files
 */
 UTim_Buf :: struct {


Feoramund:
> This reliance on a register whose value the compiler assumes doesn't matter would explain why `gettimeofday` behaves inconsistently depending on release or debug build.

These observations are in line with my experience in debugging syscalls both on Linux and FreeBSD. If the signature of the syscall is incorrect, the arguments not passed but expected will be undefined; there is no magical zeroing, it's just whatever happens to be in the register (or on the stack, for the one single syscall on FreeBSD that takes 7 arguments).

Kelimion:
The [man page](https://man7.org/linux/man-pages/man2/gettimeofday.2.html) suggests that the 2nd param is nullable, and even ignored:
```
The kernel accepts NULL for both tv and tz.  The timezone argument
is ignored by glibc and musl, and not passed to/from the kernel.
Android's bionic passes the timezone argument to/from the kernel,
but Android does not update the kernel timezone based on the
device timezone in Settings, so the kernel's timezone is typically
UTC.
```
So if it's ignored but the issue is that the register gets stomped, then perhaps we don't need to pass it on the user side:
```odin
gettimeofday :: proc "contextless" (tv: ^Time_Val) -> (Errno) {
	tz: ^Time_Zone = nil
	ret := syscall(SYS_gettimeofday, tv, tz)
	return Errno(-ret)
}
```
Or even:
```odin
gettimeofday :: proc "contextless" (tv: ^Time_Val) -> (Errno) {
	ret := syscall(SYS_gettimeofday, tv, nil)
	return Errno(-ret)
}
```

Feoramund:
> The [man page](https://man7.org/linux/man-pages/man2/gettimeofday.2.html) suggests that the 2nd param is nullable, and even ignored:
> 

This is the important bit, mentioned earlier in the man page:

```
        The  use  of  the timezone structure is obsolete; the tz argument should
        normally be specified as NULL.  See VERSIONS.
```

Kelimion:
Can you change the patch to just pass `nil`, @slowhei?
```odin
gettimeofday :: proc "contextless" (tv: ^Time_Val) -> (Errno) {
	ret := syscall(SYS_gettimeofday, tv, nil)
	return Errno(-ret)
}
```
No need to define the `Time_Zone` struct.

slowhei:
Thanks for the replies!

> Can you change the patch to just pass `nil`, @slowhei?
> 
> ```odin
> gettimeofday :: proc "contextless" (tv: ^Time_Val) -> (Errno) {
> 	ret := syscall(SYS_gettimeofday, tv, nil)
> 	return Errno(-ret)
> }
> ```
> 
> No need to define the `Time_Zone` struct.

The change has been made.

I prefer to keep the time zone parameter to maintain consistency with the Linux man page and support any obscure use cases of the time zone parameter that may exist. But I will defer such decision making to you.

Kelimion:
Further on in that man page, it said:
```
POSIX.1-2008 marks gettimeofday() as obsolete,
       recommending the use of [clock_gettime(2)](https://man7.org/linux/man-pages/man2/clock_gettime.2.html) instead.
```
Which is what's used by `core:time` and `core:prof\spall`. You're better off using that instead in the end.

Feoramund:
`src/types.cpp(3803): Assertion Failure: is_type_typed(t)`

Well that's an interesting compiler error.

Does the wrapper `syscall` not handle untyped `nil`? It would explain all the usages of `rawptr(nil)` in `sys/linux/sys.odin`. I'm assuming this is because the wrapper has a `where` clause that checks the size of `$T` and that's causing the compiler to break.

Kelimion:
> `src/types.cpp(3803): Assertion Failure: is_type_typed(t)`
> 
> Well that's an interesting compiler error.
> 
> Does the wrapper `syscall` not handle untyped `nil`? It would explain all the usages of `rawptr(nil)` in `sys/linux/sys.odin`. I'm assuming this is because the wrapper has a `where` clause that checks the size of `$T` and that's causing the compiler to break.

No, I got it earlier today on the previous commit on one Ubuntu machine, but not on another with the same LLVM backend. And even when newly cloning the git repo on the affected machine in another directory, it also worked just fine.

And that was just compiling `exampes/demo`. It's unrelated to this patch.

Kelimion:
And now I can't replicate it on 77f4199af. Must be a race condition that got exposed a couple commits back, but that's still intermittent.

Either way, this patch should be good to merge.

slowhei:
I also got the error `src/types.cpp(3803): Assertion Failure: is_type_typed(t)`. It seems to show up randomly on my machine. I implemented the `syscall(SYS_gettimeofday, tv, nil)` change, got the error, changed it to `syscall(SYS_gettimeofday, tv, 0)`, still got the error, made more changes, eventually changed back to `syscall(SYS_gettimeofday, tv, nil)`, and then the error disappeared.

Kelimion:
We'll have to look into that `is_type_typed` race condition.

Feoramund:
I ran `odin check examples/demo -vet` 200 times with no problem as of 00b67831d11b2b0f4db904af9c58be4a2cd4470c, then I pulled master with this commit merged, and I am now getting a coredump every single time I `odin check`.

```
[lldb]$ bt
* thread #1, name = 'odin', stop reason = signal SIGILL: illegal operand
  * frame #0: 0x000055c3c6b461a1 odin`type_align_of_internal(t=0x00007f9f4d64e680, path=0x00007fff554e5270) at types.cpp:3803:3
    frame #1: 0x000055c3c6b450e8 odin`type_align_of(t=0x00007f9f4d64e680) at types.cpp:3786:25
    frame #2: 0x000055c3c6b0f67a odin`check_parsed_files(c=0x00007f9f4f04e510) at checker.cpp:6798:10
    frame #3: 0x000055c3c6af7dd6 odin`main(arg_count=4, arg_ptr=0x00007fff554e6ce8) at main.cpp:3746:2
    frame #4: 0x00007f9f5c8376b5 libc.so.6`__libc_start_call_main(main=(odin`main at main.cpp:3286), argc=4, argv=0x00007fff554e6ce8) at libc_start_call_main.h:58:16
    frame #5: 0x00007f9f5c837769 libc.so.6`__libc_start_main_impl(main=(odin`main at main.cpp:3286), argc=4, argv=0x00007fff554e6ce8, init=<unavailable>, fini=<unavailable>, rtld_fini=<unavailable>, stack_end=0x00007fff554e6cd8) at libc-start.c:360:3
    frame #6: 0x000055c3c6ad5935 odin`_start + 37

[lldb]$ f
frame #0: 0x000055c3c6b461a1 odin`type_align_of_internal(t=0x00007f9f4d64e680, path=0x00007fff554e5270) at types.cpp:3803:3
   3800	
   3801		switch (t->kind) {
   3802		case Type_Basic: {
-> 3803			GB_ASSERT(is_type_typed(t));
    			^
   3804			switch (t->Basic.kind) {
   3805			case Basic_string:  return build_context.int_size;
   3806			case Basic_cstring: return build_context.ptr_size;

[lldb]$ v t -p 1
(Type *) t = 0x00007f9f4d64e680 {
  kind = Type_Basic
   = {
    Basic = {
      kind = Basic_UntypedNil
      flags = 512
      size = 0
      name = (text = "", len = 11)
    }
<...>
```

It's due to an untyped nil. My intuition was correct.

In fact, I can't even build or run anymore as of this patch being merged.

Kelimion:
Strange. I got it on the previous commit.

Feoramund:
Did you investigate the stack trace?

Kelimion:
I pushed a workaround for this `is_type_typed` assertion, @slowhei. That should allow you to use the compiler again while we investigate. It's entirely unrelated to this PR, just a sleeping bug that got triggered.

PR #5304, opened 2025-06-09
Clarify `core:flags` variadic behaviors

### LINKED ISSUE #5302: Undocumented variadic struct tag field behaviour
## Context

I'm using the `flags` package to parse variadic flags into a dynamic string array, without specifying the flag name of the variadic arg in the invocation. I noticed that I can get it working iff the variadic field's name is `varg`, and not otherwise.

```
	Odin:    dev-2025-06:cd1f66e85
	OS:      Arch Linux, Linux 6.14.10-arch1-1
	CPU:     AMD Ryzen 5 7640U w/ Radeon 760M Graphics      
	RAM:     15166 MiB
	Backend: LLVM 19.1.7
```

## Expected Behavior

Consistent behaviour of flag struct tags regardless of field name, or mention in the `flags` package documentation if and how behaviour will be affected by field name.

## Current Behavior

Variadic field not named `varg`, doesn't work.
```
[args] > cat ./args.odin
package main

import "core:flags"
import "core:fmt"
import "core:os"

Args :: struct {
	first:  u8 `args:"required,pos=0"`,
	second: [dynamic]string `args:"variadic,pos=1"`,
}

main :: proc() {
	values: Args
	flags.parse_or_exit(&values, os.args, .Unix)
	fmt.printfln("%#v\n%d", values, len(values.second))
}

[args] > odin run . -- 5 abc def
[Parse_Error.Extra_Positional] Got extra positional argument `def` with nowhere to store it.
Error in `odin` (0,1): exit status 1
```
Variadic field is named `varg`, success.
```
[args] > cat args.odin 
package main

import "core:flags"
import "core:fmt"
import "core:os"

Args :: struct {
	first:  u8 `args:"required,pos=0"`,
	varg: [dynamic]string `args:"variadic,pos=1"`,
}

main :: proc() {
	values: Args
	flags.parse_or_exit(&values, os.args, .Unix)
	fmt.printfln("%#v\n%d", values, len(values.varg))
}

[args] > odin run . -- 5 abc def
Args{
	first = 5,
	varg = [
		"abc",
		"def",
	],
}
2
```

[This](https://github.com/odin-lang/Odin/blob/b8ac7776494eb7c55828b11fa5a431ef2eb44197/core/flags/internal_assignment.odin#L36) is the cause. It checks that the field's name is `varg`. I don't know if this behaviour is intended and not well-documented, or a bug.

### Steps to Reproduce

See above

### Failure Logs

[Parse_Error.Extra_Positional] Got extra positional argument `def` with nowhere to store it.

Feoramund:
After some thought, I decided to rename `variadic` to `manifold` in hopes of removing any connotations of the flags being of the sort to consume all unspecified arguments. It makes me think of folding many arguments at once into a specified flag, which is what it does.

The `varg` field is now explicitly documented, instead of being mentioned in the example, as where unspecified positional arguments go.

I have also forbidden the combination of `pos` with `manifold` as this is likely to generate confusion even if it worked (which it did not). The failure message suggests to use `varg` instead.

Fixes #5302


Kelimion:
One last change incoming. `varg` -> `overflow`, because it collects any arguments not hoovered up by the explicitly specified flags, and a `#config` option to rename `overflow` if it conflicts with a chosen flag name.

PR #5311, opened 2025-06-10
Turn `core:math/bìg` tests into regular `core:testing` tests.

Kelimion:
`core:math/big` has been verified against Python's big integer implementation long enough. Turn it into a regular regression test using the `core:testing` framework, testing against a generated corpus of test vectors.


Kelimion:
And now we're no longer leaning on Python3 ctypes, this means we're now running `core:math/big` tests anywhere we run the optimized core tests. Huzzah.

PR #5315, opened 2025-06-11
Expose getpeername() in core:net package as "peer_endpoint"

peperronii:
Added getpeername api in core:net package as 

`peer_endpoint :: proc(socket: Any_Socket) -> (endpoint: Endpoint, err: TCP_Recv_Error) `

Right now I'm using TCP_Recv_Error as a placeholder as it covers all the cases in darwin's and window's getpeername() errors. For freebsd and linux the TCP_Recv_Error does not cover all cases, lacking : 

EINVAL : freebsd, linux
ENOBUFS : freebsd

Please suggest a change.

### DIFF
--- a/core/net/common.odin
+++ b/core/net/common.odin
@@ -64,6 +64,7 @@ Network_Error :: union #shared_nil {
 	UDP_Recv_Error,
 	Shutdown_Error,
 	Interfaces_Error,
+	Socket_Info_Error,
 	Socket_Option_Error,
 	Set_Blocking_Error,
 	Parse_Endpoint_Error,
--- a/core/net/errors.odin
+++ b/core/net/errors.odin
@@ -246,6 +246,23 @@ Shutdown_Error :: enum i32 {
 	Unknown,
 }
 
+Socket_Info_Error :: enum i32 {
+	None,
+	// No network connection, or the network stack is not initialized.
+	Network_Unreachable,
+	// Not enough space in internal tables/buffers to create a new socket, or an unsupported protocol is given.
+	Insufficient_Resources,
+	// Socket is invalid or not connected, or the manner given is invalid.
+	Invalid_Argument,
+	// The socket is valid, but unsupported by this opperation.
+	Unsupported_Socket,
+	// Connection was closed/aborted/shutdown.
+	Connection_Closed,
+
+	// An error unable to be categorized in above categories, `last_platform_error` may have more info.
+	Unknown,
+}
+
 Socket_Option_Error :: enum i32 {
 	None,
 	// No network connection, or the network stack is not initialized.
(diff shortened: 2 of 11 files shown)


Kelimion:
It occurs to me that [getpeername](https://man7.org/linux/man-pages/man2/getpeername.2.html) and [getsockname](https://man.freebsd.org/cgi/man.cgi?query=getsockname&sektion=2&apropos=0&manpath=FreeBSD+14.3-RELEASE+and+Ports) have the same kinds of error values.

`getsockname` is used internally by `_bound_endpoint`, which returns [Listen_Error](https://github.com/odin-lang/Odin/blob/5bf180d8f9ab3f673f74119c9dbbe239e94ced63/core/net/errors.odin#L98-L115), but it's not an exact fit.

But given they're cousins and `Listen_Error` is used only by `bound_endpoint`, I'd add whatever errors necessary for `getpeername` to the enum and each platform's `_listen_error` errror helper.

`Listen_Error` might also be better renamed as `Socket_Info_Error` if it services only these two calls.

peperronii:
> `Listen_Error` might also be better renamed as `Socket_Info_Error` if it services only these two calls.

Listen_Error is used by _listen_tcp also. Then, may I comply by creating this new Socket_Info_Error that combines both
getpeername's and getsockname's error and replace the current Listen_Error in _bound_endpoint with it, while keeping the original Listen_Error for _listen_tcp?

Kelimion:
> > `Listen_Error` might also be better renamed as `Socket_Info_Error` if it services only these two calls.
> Listen_Error is used by _listen_tcp also.
Ah yes, it's part of `Network_Error`. Easy to miss when grepping for `Listen_Error` to see where it's used.

> Then, may I comply by creating this new Socket_Info_Error that combines both getpeername's and getsockname's error and replace the current Listen_Error in _bound_endpoint with it, while keeping the original Listen_Error for _listen_tcp?

That seems the most sensible way forward to me. Both ask for _information_ about a `Socket`'s `Endpoint`, and a name like `Endpoint_Error` feels too general.

Kelimion:
CC: @colrdavidson + @laytan. Thoughts?

laytan:
Sure, sounds reasonable

peperronii:
Btw. I saw that we'd routed window's [WSAEINPROGRESS](https://learn.microsoft.com/en-us/windows/desktop/WinSock/windows-sockets-error-codes-2) as .Invalid_Arguments, even though by definition it should be .Insufficient_Resource or at least  be .Network_Unreachable. But for this PR, I kept it as .Invalid_Arguments for compatability.

Kelimion:
Looks good in general, but you may have forgotten to add the new enum to the `Network_Error` union. :-)

PR #5317, opened 2025-06-11
Tidy up `core:container/queue`

Feoramund:
- Add more bounds checking and fix some that was already there
- Deprecate `peek_*`; these function equivalently to `*_ptr`
- Add documentation
- Add `shrink`
- Add aliases `enqueue` and `dequeue` as `push_back` and `pop_front`, respectively
- Add tests

Looking for some input from @colrdavidson on this since the `peek_*` didn't make much sense to me in context of `*_ptr`.

### DIFF
--- a/core/container/queue/queue.odin
+++ b/core/container/queue/queue.odin
@@ -122,6 +122,33 @@ reserve :: proc(q: ^$Q/Queue($T), capacity: int) -> runtime.Allocator_Error {
 }
 
 /*
+Shrink a queue's dynamically allocated array.
+
+This has no effect if the queue was initialized with a backing slice.
+*/
+shrink :: proc(q: ^$Q/Queue($T), temp_allocator := context.temp_allocator, loc := #caller_location) {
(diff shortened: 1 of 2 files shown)


colrdavidson:
I don't use the queue for this anymore, I wound up writing a custom thing.
I don't really remember why I had to add peek_*, and looking at my comments when I did it, I'm not tracking my thought process.

At the time, this is what I'd said around the change
```
I need to write a proper peek_back / peek_front.
Those can get slots that are in the reserve space,
rather than getting the user-relevant head/tail

They're (presumably front_ptr / back_ptr?) still useful
if that's what you expect front/back to return,
I suppose, but back_ptr() != get_ptr(q, len(q)-1)
```

I think I was using the queue as both a queue and a stack, but if things have changed since 2022 with how the queue works, these functions are likely no longer really relevant

Feoramund:
Thank you for having a look.

It was more so that they did the same thing, except one wrapped the index in a modulo, but all assignments to the offset were already modulo.

PR #5329, opened 2025-06-12
Jfm fix chan try send

JackMordaunt:
This PR has two goals:

1. `send` should unblock and return false if the channel is closed by another thread (original logic returns true if the channel was closed)
2. `try_send` should _never_ block under any circumstances (original logic blocks for un-buffered channels)

For 1 we add a branch to check if the channel is closed after waking up. If closed we return false.
For 2 we add a branch to check if there is a reader waiting on the other end of the channel. If not we return false. 

`can_recv` and `can_send` return true if the respective calls _won't_ block the caller and the docs have been updated to say as much.

New tests are added to lock in the behaviour described above.

Given the change to `try_send`, some of the old tests have been altered to use blocking channel operations to synchronize instead of relying on heuristic sleeping and yield loops.

Finally `try_select` tests are serialized because they race over some global state (despite the fact that only one of those tests explicitly touch the state, the logic in `try_select` consumes it). 

Refer to each commit for more details.


core/sync/chan/chan.odin
```diff
@@ -633,6 +634,11 @@ send_raw :: proc "contextless" (c: ^Raw_Chan, msg_in: rawptr) -> (ok: bool) {
 			sync.signal(&c.r_cond)
 		}
 		sync.wait(&c.w_cond, &c.mutex)
+
+		if c.closed {
+			return false
```
laytan, line 647: Could this now return false if it sent the message and it was also closed? (if that can even happen)

JackMordaunt: Yes good point, that does seem possible.

If you have three threads: the first calls `send` and the other two call `recv` and `close`, it's possible that `recv` executes (copies the value) and then `close` grabs the lock before `send` can wakeup. In that case the value was actually received but the sender reports false because the close call sneaked in between the recv and the send wakeup.

Good: `send: sleep` -> `recv` -> `send: wake` -> `close`
Bad:  `send: sleep` -> `recv` -> `close` -> `send: wake`

JackMordaunt: One possible solve is to introduce some state, like a boolean `did_recv`. The `recv` proc sets the `did_recv` to true after doing the copy, but before signalling the sender. If the sender wakes up and `did_recv` is true then it can know that it was sent, even if the channel was also closed.

JackMordaunt: Or we can make the `send` proc wait for a reader in the same way that `recv` waits for a writer.

JackMordaunt:
```go
		for !c.closed &&
			c.w_waiting == 0 {
			c.r_waiting += 1
			sync.wait(&c.r_cond, &c.mutex)
			c.r_waiting -= 1
		}
```

JackMordaunt:
CI failure is unrelated to this PR.

PR #5344, opened 2025-06-15
Review `core/mem/allocators.odin`

### LINKED ISSUE #2694: JSON parser memory corruption with scratch allocator
## Context

```
	Odin: dev-2023-08:9453b238
	OS:   Linux Mint 20.3, Linux 5.4.0-155-generic
	CPU:  Intel(R) Core(TM) i3-1005G1 CPU @ 1.20GHz
	RAM:  7753 MiB
```

## Expected Behavior

There should be no memory corruption / segfault

## Current Behavior

Certain json inputs of long arrays of objects result in corrupted memory and segfaults when using the scratch allocator

This memory corruption does not occur with other temporary allocators. My program no longer crashed upon switching from a scratch allocator to an arena allocator.

More than likely this is an issue with the allocator rather than the json parser, however json parsing was what surfaced the bug, so there may be an issue with json parsing as well.

The threshold seems to be json arrays of 16 json objects or more (presumably this triggers a reallocation of the Array)

### Steps to Reproduce
Minimal program to reproduce:

```
package bug

import "core:fmt"
import "core:encoding/json"
import "core:mem"

main :: proc() {
	scratch := new(mem.Scratch_Allocator)
	defer free(scratch)
	if mem.scratch_allocator_init(scratch, 4 * mem.Megabyte) != .None {
		panic("unable to initialize scratch allocator")
	}
	defer mem.scratch_allocator_destroy(scratch)

	into: struct {
		foo: int,
		bar: string,
		things: []json.Object,
	}

	err := json.unmarshal_string(SAMPLE_JSON, &into, allocator = mem.scratch_allocator(scratch))
	fmt.println(err)
	fmt.println(into.things[0])
	fmt.println(into.things[1])
	fmt.println(len(into.things[2]))
	fmt.println(len(into.things[2]["longlist"].(json.Array)))
	fmt.println(into.things[2]) // this will segfault
	fmt.println(into)
}

SAMPLE_JSON :: `
{
	"foo": 123,
	"things": [
		{
			"a": "c",
			"numberlist": [1, 2, 3],
			"asdf": "qwerty"
		},
		{
			"a": "b",
			"shortlist": [
				{
					"type": "blob",
					"x": 123,
					"y": 321
				}
			],
			"c": "asdf"
		},
		{
			"a": "d",
			"longlist": [
				{
					"type": "skeleton",
					"x": 234,
					"y": 432
				},
				{
					"type": "skeleton",
					"x": 5632,
					"y": 4875
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				},
				{
					"type": "skeleton",
					"x": 574,
					"y": 585
				}
			],
			"asdf": "foo"
		}
	],
	"bar": "baz"
}
`

```

### Failure Logs

output of above program:

```
nil
map[a="c", numberlist=[1, 2, 3], asdf="qwerty"]
map[c="asdf", shortlist=[map[x=123, type="blob", y=321]], a="b"]
3
16
Segmentation fault (core dumped)
```

Feoramund:
I originally wanted to just fix #2694, but that snowballed into a total review after noticing more problems.

The problem was due to the Scratch allocator not accounting for alignment when calculating the final size. It tried to align the size forward, which makes no sense without the context of a base pointer to align on. It would then end up overwriting memory of adjacent allocations when they came around. The allocator was also letting any allocation resize in-place, independent of location, which as you can imagine would lead to more problems.

So I tried to fix everything I could find.

Fixes #2694 

---

As an important addendum, I noticed that our `Scratch_Allocator` has a certain log message `"mem.Scratch_Allocator resorted to backup_allocator"`. Back when I used to use `ols` for an LSP, I recalled seeing that a lot in the error logs. So while I was investigating this bug, I went to check on `ols`'s source code, and they are using an old copy (not an import) of our current Scratch allocator with a similar bug.

I will be contacting them about this later with a link here.

---

As a second addendum, after doing some testing of the newly-poisonable allocators, I found that ASan doesn't actually support sub-8-byte ranges: https://github.com/google/sanitizers/wiki/AddressSanitizerAlgorithm#mapping

So, I have disabled ASan everywhere until each allocator can be reviewed whether or not it should actually use ASan. I thought TLSF would be okay, given the recent review by @laytan, but I got use-after-poison failures when I disabled ASan in everything _but_ TLSF. The error was coming from one of the rollback stack allocator (the test runner's allocator) procedures, which I am assuming was hiding this bug because of the layered poison state management.

Allocators should probably not be only tested in the test suite because of this kind of layering that can hide issues, but ideally we would build all applicable general-purpose allocators in their own DLLs and run them up against a worthy test suite or benchmark, running in various debug and production modes, with and without sanitizers, in addition to running specific tests in Odin code.

### DIFF
--- a/base/runtime/core_builtin.odin
+++ b/base/runtime/core_builtin.odin
@@ -67,7 +67,7 @@ init_global_temporary_allocator :: proc(size: int, backup_allocator := context.a
 // Prefer the procedure group `copy`.
 @builtin
 copy_slice :: proc "contextless" (dst, src: $T/[]$E) -> int {
-	n := max(0, min(len(dst), len(src)))
+	n := min(len(dst), len(src))
 	if n > 0 {
 		intrinsics.mem_copy(raw_data(dst), raw_data(src), n*size_of(E))
 	}
@@ -80,7 +80,7 @@ copy_slice :: proc "contextless" (dst, src: $T/[]$E) -> int {
 // Prefer the procedure group `copy`.
 @builtin
 copy_from_string :: proc "contextless" (dst: $T/[]$E/u8, src: $S/string) -> int {
-	n := max(0, min(len(dst), len(src)))
+	n := min(len(dst), len(src))
 	if n > 0 {
 		intrinsics.mem_copy(raw_data(dst), raw_data(src), n)
 	}
--- a/base/runtime/default_temp_allocator_arena.odin
+++ b/base/runtime/default_temp_allocator_arena.odin
@@ -1,7 +1,7 @@
 package runtime
 
 import "base:intrinsics"
-import "base:sanitizer"
+// import "base:sanitizer"
 
 DEFAULT_ARENA_GROWING_MINIMUM_BLOCK_SIZE :: uint(DEFAULT_TEMP_ALLOCATOR_BACKING_SIZE)
 
@@ -44,7 +44,7 @@ memory_block_alloc :: proc(allocator: Allocator, capacity: uint, alignment: uint
 	block.base = ([^]byte)(uintptr(block) + base_offset)
 	block.capacity = uint(end - uintptr(block.base))
 
-	sanitizer.address_poison(block.base, block.capacity)
+	// sanitizer.address_poison(block.base, block.capacity)
 
 	// Should be zeroed
 	assert(block.used == 0)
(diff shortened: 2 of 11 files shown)


nathany:
I posted a follow-up discussion about this: https://github.com/odin-lang/Odin/discussions/7075

Issue #5353, opened 2025-06-17
Parser adds comments on same line as field definitions to the docs of the following field


BradLewis:
## Context

I am currently working on `ols`, adding field docs to the hover information, and the `core:odin/parser` parser currently adds any trailing comments behind a field definition as the documentation for the next field. For example: 

```odin
Foo :: struct {
  my_int: int, // this is my int <-- this comment will be included in the docs for `my_string` in the ast
  my_string: string, // this is my string <-- this won't be included in the field ast as there's no following field
}
```

This seems to be the case for both the `core:odin/parser` parser and the cpp parser.

You can also see this in the documentation on the website, for example here: https://pkg.odin-lang.org/core/bytes/#Reader

```odin
// definition in code
Reader :: struct {
	s:         []byte, // read-only buffer
	i:         i64,    // current reading index
	prev_rune: int,    // previous reading index of rune or < 0
}

// display in the docs
Reader :: struct {
	s:         [][u8](https://pkg.odin-lang.org/base/builtin#u8),
	// read-only buffer
	i:         [i64](https://pkg.odin-lang.org/base/builtin#i64),
	// current reading index
	prev_rune: [int](https://pkg.odin-lang.org/base/builtin#int),
}
```

Edit: It looks like some changes were added to fix this, so comments are now added to struct fields. However there are still 2 issues:

```odin
Foo :: struct {
  my_int: int, // this is my int <-- this comment will still be included in the docs for `my_string` in the ast
  my_string: string // this is my string <-- this won't be included in the field ast as there's no `,` before the comment
}
```

## Expected Behavior

Comments that follow the line of a field definition should be added to the docs of that field.

## Current Behavior

Comments that follow the line of a field definition should be added to the docs of the following field.

PR #5355, opened 2025-06-17
Parse line comments correctly

BradLewis:
Addresses https://github.com/odin-lang/Odin/issues/5353

I noticed that the odin parser is very similar to the go parser (which handles this case correctly). Essentially, rather than parsing just one line for the comment after a field definition, the odin parser would continue scanning until it finished parsing all of the comments, meaning that the line comment would end up being interpreted as a leading comment for the next field.

This PR just makes the parser first parse the comment on the same line for the line comment, and then continue to parsing the lead comment, which is the behaviour of the go parser.

This just updates the odin parser, but if it looks good it should be easy to make the same change on the cpp parser.

### DIFF
--- a/core/odin/parser/parser.odin
+++ b/core/odin/parser/parser.odin
@@ -321,9 +321,6 @@ consume_comment :: proc(p: ^Parser) -> (tok: tokenizer.Token, end_line: int) {
 	}
 
 	_ = next_token0(p)
-	if p.curr_tok.pos.line > tok.pos.line {
-		end_line += 1
-	}
 
 	return
 }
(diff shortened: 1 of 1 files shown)


core/odin/parser/parser.odin
```diff
@@ -354,7 +351,7 @@ consume_comment_groups :: proc(p: ^Parser, prev: tokenizer.Token) {
 
 		if p.curr_tok.pos.line == prev.pos.line {
 			comment, end_line = consume_comment_group(p, 0)
-			if p.curr_tok.pos.line != end_line || p.curr_tok.kind == .EOF {
+			if p.curr_tok.pos.line != end_line || p.curr_tok.kind == .EOF  || p.curr_tok.pos.line == prev.pos.line + 1 {
```
BradLewis, line 354: The equivalent go code for this is here: https://github.com/golang/go/blob/master/src/go/parser/parser.go#L246
`if p.lineFor(p.pos) != endline || p.tok == token.SEMICOLON || p.tok == token.EOF {`
In their case they set the token to be `SEMICOLON` on a newline. Not sure if there's a better equivalent here.

Kelimion:
Don't copy code from somewhere else without clearing this with us first.

BradLewis:
My bad, I didn't realise that would be a problem. What's the appropriate course of action for this then?

Kelimion:
For someone without knowledge of the Go implementation (or your derivative patch) to write it. Sorry. :-)

PR #5356, opened 2025-06-18
Fix `WAVEFORMATEX` struct size

denovodavid:
Was struggling to get WASAPI working, made sanity check in C, found that `WAVEFORMATEX` was `18` bytes in C but `20` bytes in Odin. Added `#packed` and was able to get sound playing. Size is now `18` bytes, and `WAVEFORMATEXTENSIBLE` is now correctly `40` bytes.


denovodavid:
Wait... but the generated CI test fails as it expects the size to be `20` bytes... but why was it `18` bytes on my machine?

I also tried many values for `cbSize` before adding `#packed` and nothing would work when using `WAVEFORMATEXTENSIBLE`.

UPDATE:

From my understanding, `WAVEFORMATEX` is defined in both `mmeapi.h` and `mmreg.h`.
`mmreg.h` appears to byte pack the whole file.
The [WAVEFORMATEX docs](https://learn.microsoft.com/en-us/windows/win32/api/mmeapi/ns-mmeapi-waveformatex) state: "mmeapi.h (include Mmreg.h)"
While Odin only has `#include <mmeapi.h>` in `tests\core\sys\windows\win32gen\win32gen.cpp`.
I'll just add `mmreg.h` to the test file, and that should fix things.

PR #5362, opened 2025-06-18
added readgs{byte,word,dword,qword} intrinsics

SaculRennorb:
These instrinsics allow reading from the G Sector on x86 machines, which is used for windows internals ThreadInfoBlock s. Specifically I am interested in calculating some stack space metrics, which to my knowledge cannot be obtained in a different way. 

I don't love the implementation i added, but it works. There _might_ be a llvm intrinsic that does this in a more elegant way, but the intel assembler version is the one I am familiar with. 
I mostly just copied the implementation from the `BuiltinProc_x86_xgetbv` one above, I am not sure this necessarily requires a function call. 

I also thought about if it was possible to implement the intrinsic in a way that doesn't require multiple of them, but since odin doesn't have have overload resolution based on the return value i don't really see a way to do this more elegantly. 
An option might be to take a type as the first argument, but i feel like this doesn't really improve anything. 

See core/sys/windows/winternals.odin for an application of the function.

### DIFF
--- a/base/intrinsics/intrinsics.odin
+++ b/base/intrinsics/intrinsics.odin
@@ -356,6 +356,12 @@ wasm_memory_atomic_notify32 :: proc(ptr: ^u32, waiters: u32) -> (waiters_woken_u
 x86_cpuid  :: proc(ax, cx: u32) -> (eax, ebx, ecx, edx: u32) ---
 x86_xgetbv :: proc(cx: u32) -> (eax, edx: u32) ---
 
+x86_readgsbyte :: proc(offset : u32) -> u8 ---
+x86_readgsword :: proc(offset : u32) -> u16 ---
+x86_readgsdword :: proc(offset : u32) -> u32 ---
+// Used to implement  NtCurrentTeb() since we cant just link against it (only 32bit version of ntdll exposes it).
+x86_readgsqword :: proc(offset : u32) -> u64 ---
+
 
 // Darwin targets only
 objc_object   :: struct{}
--- a/core/sys/windows/winternal.odin
+++ b/core/sys/windows/winternal.odin
@@ -1,0 +1,39 @@
+#+build windows
+package sys_windows
+
+import "base:intrinsics"
+
+when ODIN_ARCH == .amd64 {
+	NtCurrentTeb :: proc() -> ^TEB {
+		return cast(^TEB) cast(uintptr) intrinsics.x86_readgsqword(cast(u32)offset_of(NT_TIB{}.Self))
+	}
+}
+
+NT_TIB :: struct {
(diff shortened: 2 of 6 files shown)


Feoramund:
> These instrinsics allow reading from the G Sector on x86 machines, which is used for windows internals ThreadInfoBlock s. Specifically I am interested in calculating some stack space metrics, which to my knowledge cannot be obtained in a different way.

You can write an x86 assembly file with an exposed procedure, build and link against that, then call it, if you need arbitrary instructions not otherwise accessible. This could accomplish your goal in less lines and without introducing new intrinsics.

SaculRennorb:
That is correct, but that would also require additional compile and link steps. 
I very much dislike the idea of having to introduce an assembler for essentially one assembly instruction, which is why I proposed this pr. 

Also we already have x86 specific intrinsics.

After thinking about it for a night I came up with a more flexible approach;
```odin

X86_Segment :: enum { CS, DS, SS, ES, FS, GS }

x86_read_from_segment($segment : X86_Segment, $Width : typeid, offset : u32) -> Width where Width == u8 || Width == u16 || Width == u32 || Width == u64 --
```

For most purposes you only ever really need the GS register (on windows, the FS on 32 bit windows), so i didn't originally bother with this, but that approach might be more generally useful to others.
If that is preferable i can sit down again and change it to that signature.

Feoramund:
> That is correct, but that would also require additional compile and link steps. I very much dislike the idea of having to introduce an assembler for essentially one assembly instruction, which is why I proposed this pr.

`procs_windows_amd64.asm` exists and is used when a CRT-less program is built for Windows. There are also files with entry point assembly instructions for the Unix-likes for building without a CRT. This isn't [unheard of](https://odin-lang.org/docs/overview/#foreign-system) in Odin, nor is it difficult. You can `foreign import "foo.asm"`.

SaculRennorb:
Oh, i somehow wasn't aware of that option. 

I will change the implementation accordingly.

Kelimion:
Good luck. Given we can import foreign asm and it'll be assembled and linked for you, I don't see the value in adding it as an intrinsic. There's no getting around `cpuid` or `xgetbv`, but this is on the niche side comparatively speaking.

You'll need to adapt the asm snippet, but this should get you started.
```odin
package main

import "core:fmt"

foreign import lowlevel "read_gs.asm"
foreign lowlevel {
	// Returns offset (as passed in rcx)
	read_gs_u64 :: proc "stdcall" (offset: uintptr) -> u64 ---
}

main :: proc() {
	fmt.println("Hellope!", read_gs_u64(42))
}
```
```asm
bits 64

global read_gs_u64

section .text
read_gs_u64:
	mov rax, rcx
	ret
```

PR #5364, opened 2025-06-19
Added `NtCurrentTeb` to `sys/windows` (on x86-64)

SaculRennorb:
This adds the `NtCurrentTeb()` function from winnt.h, as well as functions to read from the FS / GS segments required to do so. 

The implementation of this function is unfortunately quite platform dependent, so for now I only added the implementation for x64 I currently need.

This was originally #5362 with the intent to add the `__readgs*` functions as compiler intrinsics, but I was made aware that the assembly version should be sufficient. 

This is now only about adding the functions to `sys/windows`.

### DIFF
--- a/base/intrinsics/intrinsics.odin
+++ b/base/intrinsics/intrinsics.odin
@@ -356,12 +356,6 @@ wasm_memory_atomic_notify32 :: proc(ptr: ^u32, waiters: u32) -> (waiters_woken_u
 x86_cpuid  :: proc(ax, cx: u32) -> (eax, ebx, ecx, edx: u32) ---
 x86_xgetbv :: proc(cx: u32) -> (eax, edx: u32) ---
 
-x86_readgsbyte :: proc(offset : u32) -> u8 ---
-x86_readgsword :: proc(offset : u32) -> u16 ---
-x86_readgsdword :: proc(offset : u32) -> u32 ---
-// Used to implement  NtCurrentTeb() since we cant just link against it (only 32bit version of ntdll exposes it).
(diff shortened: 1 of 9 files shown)


Kelimion:
As a PR it's still fatally flawed:
- There is no need for a new file `winternal.odin` when all other struct definitions for the package live in `types.odin`.
- `foreign ass`. How puerile. Yes it's a shortening for assembly, but really?
- Odin doesn't have word, dword, or qword.
- The `__` prefix can be gotten rid of with `@(link_prefix="__")`, so you can call `read_gs_u8` instead of `__read_gs_u8`.

If this works for your personal needs, great. No need for a third take.

SaculRennorb:
Sure, I will just use it in my own project, I just thought to add missing functionality to the library. 

Regarding your issues:
- not all structs are defined in tyes.odin. ntdll.odin also defines its respective types in there. 
- I did not even consider this to be problematic, and it's trivial to change - far from "a fatal flaw". I also used that abbreviation in the previous pr, neither you nor Feoramund commented on the usage there.
- Odin doesn't have those types, but those are the names of the intrinsic functions used by the MS headers. Again, a trivial change to make. 
- Again, those are the names of the original functions. I just adhered to what the MS headers specify. 

I would have understood if it was undesirable to add a partial implementation, as this is only for a single platform, but I find the reasons you've provided rather questionable. All of them are strongly influenced by taste and project structure, both of which i am obviously not deeply familiar with as a first time contributor, and are trivial to adjust.

I also find it a bit crass to immediately close PRs without me even being able to react to any feedback on them, but I do also understand the general open source dilemma. Thank you for spending time on this either way.

PR #5368, opened 2025-06-19
Fix #5331 and `check_shift` improvements

### LINKED ISSUE #5331: LLVM-backend assertion failure when non-constant shifting untyped int larger than infered type
## Context
```
Odin:    dev-2025-04-nightly:d9f990d
OS:      Windows 10 Professional (version: 22H2), build 19045.5487
CPU:     AMD Ryzen 7 6800H with Radeon Graphics
RAM:     15612 MiB
Backend: LLVM 20.1.0
```

## Expected Behavior

Compile error.

## Current Behavior

LLVM-backend assertion failure.

### Steps to Reproduce

On 32-bit targets:

```odin
i : uint = 1
j := 4294967296 << i //2^32 << i
```

On 64-bit targets:

```odin
i : uint = 1
j := 18446744073709551616 << i //2^64 << i
```

### Failure Logs

```
[DEBUG] 18446744073709551616
int -> 8
src/llvm_backend_const.cpp(424): Assertion Failure: `sz >= max_count` max_count: 9, sz: 8, written: 0, type int
Program terminated with signal: SIGILL
Compiler returned: 132

```

Airtz:
Fix #5331.

**Additional `check_shift` improvements**

Given shift expression `x << y`

Fixes :
- `untyped y` isn't sign-checked when `x` is `const`.
e.g. `1 << -1` is valid and evaluates to the default `BigInt` value, `0`.
--> do the sign check.
- `untyped x` cannot convert to `any` when `y` is `value`.
e.g. `y: uint = 1; fmt.print(1 << a)` is invalid.
--> convert `x` to `default(t_untyped_integer)` first.
- `untyped x` can be `untyped bool` which makes little sense.
e.g. `true << 1` is valid.
--> restrict to `t_untyped_integer`-convertible `untyped`s.

Better error messages :
- `Shift amount too large: '%s'` --> `Shift amount '%s' must be <= %u` 
specifying `MAX_BIG_INT_SHIFT`
- `Conversion of shifted operand '%s' to '%s' is not allowed` --> `Shifted operand '%s' cannot convert to non-integer type '%s'`
- `Shift amount cannot be negative: '%s'` --> `Shift amount '%s' must be positive`
since `0` is valid.


gingerBill:
The reason I wrote "cannot be negative" over "must be positive" is because `0` is a weird number which is neither positive nor negative, and thus the specific wording used.

as2te:
Maybe it's because of my math background but to me `0` is both positive and negative. Literature uses "positive" for `>= 0` and "strictly positive" for `> 0`.
Just a convention issue.

Kelimion:
That and `|0| == 0` equivalence renders it positive for practical purposes. With that said, I do prefer the `cannot be negative` phrasing. It still conveys the intended meaning if you consider `0` to be positive.

src/check_expr.cpp
```diff
@@ -3086,126 +3086,106 @@ gb_internal void check_shift(CheckerContext *c, Operand *x, Operand *y, Ast *nod
...
 
-	if (is_type_unsigned(y->type)) {
-
-	} else if (is_type_untyped(y->type)) {
-		convert_to_typed(c, y, t_untyped_integer);
-		if (y->mode == Addressing_Invalid) {
+	if (y->mode == Addressing_Constant) {
+		if (big_int_is_neg(&y->value.value_integer)) {
+			gbString y_str = expr_to_string(y->expr);
+			error(y->expr, "Shift amount '%s' must be positive", y_str);
```
Kelimion: Can you please phrase this as `cannot be negative` as @gingerBill intended? As far as I can tell it's the only thing holding up the patch.

PR #5376, opened 2025-06-20
#5270 Escape Object files properly during linker_stage

### LINKED ISSUE #5270: Odin fails to build inside a directory with a `$` in the name on Linux
Do:
```
cd /tmp
mkdir '$'
cd '$'
printf "package main\nmain::proc(){}\n" > main.odin
odin build .
```

Output:
```
clang: error: no such file or directory: '/tmp/hBcmain-7f905ae5c980.o'
clang: error: no such file or directory: '/tmp/hBcbuiltin-7f905ae5c1e8.o'
clang: error: no such file or directory: '/tmp/hBcruntime-7f905ae5c680.o'
clang: error: no such file or directory: '/tmp/hBcsanitizer-7f905ae5cc80.o'
```

Creativty:
`system` function calls `sh` in the background which in turn introduces string interpolation and `$` expansion...
In order to avoid that and to fix #5270 we use single quotes which are used to express static string literals in `sh`.

### DIFF
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -704,12 +704,12 @@ try_cross_linking:;
 					return result;
 				}
 
-				object_files = gb_string_append_fmt(object_files, "\"%.*s\" ", LIT(android_glue_static_lib));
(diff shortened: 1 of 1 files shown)


Kelimion:
Thank you.

PR #5379, opened 2025-06-20
Push `context` onto stack before evaluating procedure parameters

### LINKED ISSUE #4711: Interal assertion triggered on a bad assignment inside a procedure return type declaration
## Context

Tested on two versions, both produce the same result:
Odin:       dev-2025-01:47030501a  &&  dev-2024-12-nightly:cf53404
OS:          Windows 11 Home Basic (version: 24H2), build 26100.2605
CPU:        Intel(R) Core(TM) Ultra 7 155H
RAM:       65033 MiB
Backend: LLVM 18.1.8

## Expected Behavior

The compiler reports an error specifying why the assignment is invalid

## Current Behavior

The compiler fails on an internal assertion.

Assertion Report:

C:\Projects\odin-lang\src\llvm_backend_utility.cpp(879): Assertion Failure: `pt->Proc.calling_convention != ProcCC_Odin`
exited with code -2147483645

The assertion report is the same in both tested versions, including the line number

### Steps to Reproduce

Attempt to compile the following:

package main;

import "base:runtime";

foo :: proc () -> (allocator : runtime.Allocator = context.allocator) {
	return;
}

main :: proc() -> () {
	foo();
}

Feoramund:
Fixes #4711

### DIFF
--- a/src/llvm_backend_proc.cpp
+++ b/src/llvm_backend_proc.cpp
@@ -546,6 +546,9 @@ gb_internal void lb_begin_procedure_body(lbProcedure *p) {
 	GB_ASSERT(p->type != nullptr);
 
 	lb_ensure_abi_function_type(p->module, p);
+	if (p->type->Proc.calling_convention == ProcCC_Odin) {
+		lb_push_context_onto_stack_from_implicit_parameter(p);
+	}
 	{
 		lbFunctionType *ft = p->abi_function_type;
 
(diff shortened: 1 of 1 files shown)


laytan:
Shouldn't this be a compiler error, I am surprised if default values are supposed to be a thing in return values, but maybe they are?

Feoramund:
`core/io/conv.odin`. We use them everywhere in there. They're assigned at the start of the procedure.

Kelimion:
> Shouldn't this be a compiler error, I am surprised if default values are supposed to be a thing in return values, but maybe they are?

I was also surprised, but it works `foo :: proc() -> (i := 42) { }`. Print it and you get 42. Can override it with `i = 13`.

Kelimion:
(Different values are forbidden in examples, naturally.)

laytan:
Ah, I never knew

Feoramund:
Looks like I have something new to add to the overview. It's not mentioned.

PR #5404, opened 2025-06-24
Disable two tests under core/sys/posix.

Kelimion:


laytan:
IMO I don't like just commenting out tests, nothing will be done with them if it's not accompanied by an issue to fix it, or just fix or remove it right away.

Kelimion:
I left it commented out so someone could revisit it in the future and rethink it, as I needed to move on to something else. You're right that I could've made an issue for it.

It may be better to remove it, because the only permissions we can count on for `#file` is that at least one of `ugw` is set so the compiler can read and compile the file, which is self-evidently true if the test runs, and so it's not a very interesting test.

PR #5412, opened 2025-06-26
[source-code-locations] - added options to show, obfuscate, trim, and hide source code locations

A1029384756:
adds the following modification options to source code locations:
- normal (unchanged)
- obfuscated (previously `-obfuscate-source-code-locations`), `fnv64a` hash
- filename
- none (zero-valued)

also deprecates (and adds a message for) `-obfuscate-source-code-locations` in favor of `-source-code-locations:obfuscated`


laytan:
The only thing none (and filename in a lesser extent) break is the value being a "unique" thing, to use as a `map` key for example.

I do not recall us relying on this in base/core/vendor, and it's opt-in, so I don't think it's a big deal, but worth the mention.

laytan:
Could you put the `switch` and `case`'s on the same indentation level to follow the convention of other switch statements?

Feoramund:
> I do not recall us relying on this in base/core/vendor, and it's opt-in, so I don't think it's a big deal, but worth the mention.

`testing.expect_assert_from` relies on it, but if you're building tests with obfuscated source code locations, you're probably doing something that'll void your warranty anyway.

src/main.cpp
```diff
@@ -1422,7 +1424,23 @@ gb_internal bool parse_build_flags(Array<String> args) {
...
+							if (str_eq_ignore_case(value.value_string, str_lit("normal"))) {
+								build_context.source_code_location_info = SourceCodeLocationInfo_Normal;
+							} else if (str_eq_ignore_case(value.value_string, str_lit("obfuscated"))) {
+								build_context.source_code_location_info = SourceCodeLocationInfo_Obfuscated;
+							} else if (str_eq_ignore_case(value.value_string, str_lit("filename"))) {
+								build_context.source_code_location_info = SourceCodeLocationInfo_Filename;
+							} else if (str_eq_ignore_case(value.value_string, str_lit("none"))) {
+								build_context.source_code_location_info = SourceCodeLocationInfo_None;
+							} else {
+								gb_printf_err("-source-code-location-info:<string> options are 'normal', 'obfuscated', 'filename', and 'none'\n");
```
laytan:
```suggestion
								gb_printf_err("-source-code-locations:<string> options are 'normal', 'obfuscated', 'filename', and 'none'\n");
```

PR #5415, opened 2025-06-26
Use temp allocator in load_shaders_file() and load_compute_file().

Byron-Hodel:
Modified `load_shaders_file()` and `load_compute_file()` to use the temp allocator instead of the normal allocator memory management. This is because the memory allocated was freed at the end of the function using `defer delete(...)`.

### DIFF
--- a/vendor/OpenGL/helpers.odin
+++ b/vendor/OpenGL/helpers.odin
@@ -149,8 +149,7 @@ create_and_link_program :: proc(shader_ids: []u32, binary_retrievable := false) 
 }
 
 load_compute_file :: proc(filename: string, binary_retrievable := false) -> (program_id: u32, ok: bool) {
-	cs_data := os.read_entire_file(filename) or_return
-	defer delete(cs_data)
+	cs_data := os.read_entire_file(filename, context.temp_allocator) or_return
 
 	// Create the shaders
 	compute_shader_id := compile_shader_from_source(string(cs_data), Shader_Type(COMPUTE_SHADER)) or_return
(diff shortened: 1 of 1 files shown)


Kelimion:
It was written as intended, and doesn't seem to need its allocation strategy changed to work.

Yes, it calls `defer delete(vs_data)` and `defer delete(fs_data)`, but that won't be called until the procedure returns and is done using it. And `load_shaders_source` performs the same trick with `defer DeleteShader(vertex_shader_id)` that's fired up after `create_and_link_program` has used these inputs and returned. So it's not like it needs changing because of a use-after-free.

If you're switching to `temp_allocator` only because this memory gets freed again and you think the `temp_allocator` is a better fit, please also roll it back using `runtime.DEFAULT_TEMP_ALLOCATOR_TEMP_GUARD()` at the top of `load_shaders_file` + `load_compute_file`.

Because unless the user knows the temp allocator was used and they call `free_all(context.temp_allocator)` on occasion, this patch steadily eats up memory these procedures used to clean up.

Byron-Hodel:
@Kelimion You're right. When I was writing this, I was under the impression that one of the purposes of the temp allocator was for use as scratch space inside of a function. Upon inspecting how other parts of the core library used it, it is clear that this is not really the case.

Kelimion:
> @Kelimion You're right. When I was writing this, I was under the impression that one of the purposes of the temp allocator was for use as scratch space inside of a function.

The old implementation of the temp allocator was a cyclic buffer that would automatically overwrite older allocations, with a total of around 4 MiB of scratch space. It would've been fine for this purpose, but was less predictable in its use when you call a bunch of nested procedures that each need to make temporary allocations.

The current one is an arena that grows as necessary, but that does mean you need to call `free_all` occasionally so it doesn't grow unbounded. Or you can checkpoint it using `runtime.DEFAULT_TEMP_ALLOCATOR_TEMP_GUARD()`, which resets the watermark to what it was on exit.

PR #5417, opened 2025-06-27
posix: add RTLD_DEEPBIND

xoxorwr:
https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/dlfcn.h

Btw, what architectures does Odin officially support? this posix package seems to only care about X86/ARM? on MIPS the values for dlfcn are different

This value is also only supported on Linux, what should be done to update the `RTLD_Flag_Bits`? since odin's `dlopen` requires it


laytan:
This is not a POSIX standard value and thus won't be added to sys/POSIX (https://pubs.opengroup.org/onlinepubs/9799919799/).

PR #5425, opened 2025-07-01
wasm: Set stack first and increase stack size to 1mb

xoxorwr:
The stack size is 64 KB by default, which is too small to be useful and leads to difficult-to-guess problems.

I recommend setting the stack size to 1 MB, which is precisely what emscripten, [zig](https://github.com/ziglang/zig/issues/3735), and rust are doing.

And the rationale for `--stack-first`: https://github.com/ziglang/zig/issues/4496

### DIFF
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -72,7 +72,7 @@ gb_internal i32 linker_stage(LinkerData *gen) {
 		inputs = gb_string_append_fmt(inputs, "\"%.*s.o\"", LIT(output_filename));
 
 
-		gb_string_appendc(inputs, "--stack-first -z stack-size=1048576")
+		inputs = gb_string_appendc(inputs, " --stack-first -z stack-size=1048576");
 
 		for (Entity *e : gen->foreign_libraries) {
 			GB_ASSERT(e->kind == Entity_LibraryName);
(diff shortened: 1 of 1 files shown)


xoxorwr:
I forgot the semicolon (I have been writting too much Odin lol)

But doesn't seem to work, the params are not being passed to wasm-ld, the command must be missing something:

xoxorwr:
Now it's working

laytan:
A concern I have here is that you probably can't change the stack size anymore if we merge this. You used to be able to add -extra-linker-flags:"" with any flags for wasm-ld, but if we already give the stack size flag does it complain about duplicates?

xoxorwr:
You can still set what ever stack size you want, it'll override the previous flags, you can test it by doing:

 `-extra-linker-flags:'-z stack-size=5048576'` 

And then checking with: `wasm-objdump -x bin/test.wasm | grep -i stack_pointer`

It'll print:

```
- global[0] i32 mutable=1 <__stack_pointer> - init i32=5048576
```

laytan:
Sorry for the delay, then I agree we can add the flags. Your are adding them to `inputs` though, and these aren't inputs.

xoxorwr:
I'm not sure where exactly it should go, do you mind taking over this PR please?

PR #5438, opened 2025-07-04
Timezone API: Drop region_load("local"), move to region_load_local()

colrdavidson:


core/time/timezone/tz_unix.odin
```diff
@@ -7,61 +7,65 @@ import "core:strings"
...
-	if !ok {
-		orig_localtime_path := "/etc/localtime"
-		path, err := os.absolute_path_from_relative(orig_localtime_path, allocator)
-		if err != nil {
-			// If we can't find /etc/localtime, fallback to UTC
-			if err == .ENOENT {
-				str, err2 := strings.clone("UTC", allocator)
-				if err2 != nil { return }
-				return str, true
+local_tz_name :: proc(check_env: bool, allocator := context.allocator) -> (name: string, success: bool) {
```
laytan, line 10: Is there a reason this `check_env` was added, I don't see a real use case (when would you not check env?) and am not a fan of bool flags (because reading `local_tz_name(true)` is hard, what does `true` mean)

Kelimion:
This PR is based on the old `timezone` package that used "old os" and conflicts with the current version of the package that was ported to "new os".

It's going to be easier to forward port the actual changes than fix the conflicts.

PR #5455, opened 2025-07-11
Add `type_enum_is_contiguous` intrinsic (Closes #5395)

### LINKED ISSUE #5395: [FEATURE REQUEST] Add `intrinsics.type_enum_is_contiguous`
I would like to ask for an `intrinsics.type_enum_is_contiguous(E)` proc, where this would at compile time tell whether an enum range is
contiguous. This would return `true` iff all enum constants have values that either increment or decrement with 1, starting from a base value, with no gaps between them.

An example use-case would be to implement a fast-path for checking whether an int is a valid enum constant:

```go
int_is_valid_enum :: proc(i: int, $E: typeid) -> bool where intrinsics.type_is_enum(E) {
    when intrinsics.type_enum_is_contiguous(E) {
        return E(i) >= min(E) && E(i) <= max(E)
    } else {
        for e in E do if e == E(i) {
            return true
        }
        return false
    }
}
```

I would imagine the logic for checking whether an enum is sparse is already in place, so this ideally wouldn't be too hard to implement?
Additionally an enum is also sparse when it has constants that share the same value (where this proc would instead return `false`).

Examples:

```
type_enum_is_contiguous(enum { A=0, B=1, C=2 }) // -> true
type_enum_is_contiguous(enum { A=3, B=4, C=5 }) // -> true
type_enum_is_contiguous(enum { A=5, B=5, C=6 }) // -> false
type_enum_is_contiguous(enum { A=3, B=2, C=1 }) // -> true
type_enum_is_contiguous(enum { A=1, B=0, C=-1 }) // -> true
type_enum_is_contiguous(enum { A=0, B=0, C=-1 }) // -> false
```

Whether this could also prove its usefulness in the `reflect.Type_Info_Enum` struct is also a question that could be asked

FourteenBrush:
This adds an intrinsic that indicates whether an enum has a contiguous range.
In this context, "contiguous" means that the enum range, when sorted, have differences of either 0 or 1 between consecutive values, it also allows duplicates.

This could not be done with `max`/`len` fiddling due to duplicate values.

The following holds true:
```go
    c :: intrinsics.type_enum_is_contiguous
    #assert(c(enum { A=-1, B=-1, C=0 }))
    #assert(c(enum { A=3, B=1*1, C=2 }))
    #assert(c(enum { A=-1, B=0, C=1 }))
    #assert(c(enum { A=-4 }))
    #assert(c(enum { }))
    #assert(c(enum { A=-5 }))
    #assert(!c(enum { A=-3, B=-5 }))
    #assert(!c(enum { A=0, B=2 }))
```

Internally this sorts the `Type.Enum.fields` in place. Although I haven't found any issues with this, I am not sure whether the checker expects the exact declaration order of those fields in certain places?

### DIFF
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -6941,8 +6941,8 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 				return false;
 			}
 			
-			// sort enum fields in place in ascending order
-			Array<Entity *> enum_constants = type->Enum.fields;
+			auto enum_constants = array_make<Entity *>(temporary_allocator(), type->Enum.fields.count);
+			array_copy(&enum_constants, type->Enum.fields, 0);
 			array_sort(enum_constants, enum_constant_entity_cmp);
 			
 			BigInt minus_one = big_int_make_i64(-1);
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -6946,6 +6946,7 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 			array_sort(enum_constants, enum_constant_entity_cmp);
 			
 			BigInt minus_one = big_int_make_i64(-1);
+			defer (big_int_dealloc(&minus_one));
 			BigInt diff = {};
 			
 			bool contiguous = true;
@@ -6956,6 +6957,7 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 				BigInt curr = enum_constants[i]->Constant.value.value_integer;
 				BigInt next = enum_constants[i + 1]->Constant.value.value_integer;
 				big_int_sub(&diff, &curr, &next);
+				defer (big_int_dealloc(&diff));
 				
 				if (!big_int_is_zero(&diff) && big_int_cmp(&diff, &minus_one) != 0) {
 					contiguous = false;
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -1,9 +1,12 @@
 typedef bool (BuiltinTypeIsProc)(Type *t);
 
 gb_internal int enum_constant_entity_cmp(void const* a, void const* b) {
-	BigInt bi = (*cast(Entity const **)a)->Constant.value.value_integer;
-	BigInt bj = (*cast(Entity const **)b)->Constant.value.value_integer;
-	return big_int_cmp(&bi, &bj);
+	Entity const *ea = *(cast(Entity const **)a);
+	Entity const *eb = *(cast(Entity const **)b);
+	GB_ASSERT(ea->kind == Entity_Constant && eb->kind == Entity_Constant);
+	GB_ASSERT(ea->Constant.value.kind == ExactValue_Integer && eb->Constant.value.kind == ExactValue_Integer);
+	
+	return big_int_cmp(&ea->Constant.value.value_integer, &eb->Constant.value.value_integer);
 }
 
 gb_global BuiltinTypeIsProc *builtin_type_is_procs[BuiltinProc__type_simple_boolean_end - BuiltinProc__type_simple_boolean_begin] = {
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -1,5 +1,11 @@
 typedef bool (BuiltinTypeIsProc)(Type *t);
 
+gb_internal int enum_constant_entity_cmp(void const* a, void const* b) {
+	BigInt bi = (*cast(Entity const **)a)->Constant.value.value_integer;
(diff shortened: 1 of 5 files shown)


Kelimion:
Patch looks good to me, but as this touches on language semantics, let's let @gingerBill review it.

src/check_builtin.cpp
```diff
@@ -6919,6 +6925,48 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
...
+			}
+			if (!is_type_enum(type)) {
+				gbString t = type_to_string(type);
+				error(ce->args[0], "Expected an enum type for '%.*s', got %s", LIT(builtin_name), t);
+				gb_string_free(t);
+				return false;
+			}
+			
+			// sort enum fields in place in ascending order
+			Array<Entity *> enum_constants = type->Enum.fields;
```
gingerBill: I'd recommend making a temporary array and copy to that, and then sort.

```cpp
auto enum_constants = array_make<Entity *>(temporary_allocator(), type->Enum.fields.count);
array_copy(&enum_constants, type->Enum.fields);
array_sort(&enum_constants, enum_constant_entity_cmp);
```

FourteenBrush: @gingerBill Done

Feoramund:
Could we have the asserts listed in your original post as a test?

src/check_builtin.cpp
```diff
@@ -1,5 +1,11 @@
 typedef bool (BuiltinTypeIsProc)(Type *t);
 
+gb_internal int enum_constant_entity_cmp(void const* a, void const* b) {
+	BigInt bi = (*cast(Entity const **)a)->Constant.value.value_integer;
```
laytan: Could you add some
`GB_ASSERT(a->kind == Entity_Constant);` for safety/to make sure

src/check_builtin.cpp
```diff
@@ -6919,6 +6925,48 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
...
+				gb_string_free(t);
+				return false;
+			}
+			
+			auto enum_constants = array_make<Entity *>(temporary_allocator(), type->Enum.fields.count);
+			array_copy(&enum_constants, type->Enum.fields, 0);
+			array_sort(enum_constants, enum_constant_entity_cmp);
+			
+			BigInt minus_one = big_int_make_i64(-1);
+			BigInt diff = {};
```
laytan: I believe you are leaking these

src/check_builtin.cpp
laytan: same with the exact value kind

laytan:
The issue this closes has a comment https://github.com/odin-lang/Odin/issues/5395#issuecomment-2996350639 that says this is already possible, is that not true?

FourteenBrush:
> The issue this closes has a comment [#5395 (comment)](https://github.com/odin-lang/Odin/issues/5395#issuecomment-2996350639) that says this is already possible, is that not true?

No: consider enum { A=1, B=2, C=2, D=3, E=4 }, or { 2, 3, 3, 4 }, which would return true here, but wouldn't be detected with min/max trickery.

So actually `!type_is_contiguous(E)` implies `#sparse` is needed when using said enum as "key type" of an enumerated array

Feoramund:
> The issue this closes has a comment [#5395 (comment)](https://github.com/odin-lang/Odin/issues/5395#issuecomment-2996350639) that says this is already possible, is that not true?

I went back and checked, and my calculation using the primitives does not account for same-value duplicate enums on account of `len` returning the number of values total and without regard to uniqueness.

```odin
package main

import "base:runtime"

Greek :: enum {
	Alpha = -2,
	Beta = -1,
	Gamma = 0,
	// Delta = 0, // This will cause a failure if uncommented.
}

main :: proc() {
	when int(max(Greek)-min(Greek)) == len(Greek)-1 {
		runtime.println_any("Yes")
	} else {
		runtime.println_any("No")
	}
}
```

However, without further thought, I don't know if an enum can be said to be contiguous if it contains duplicates.

FourteenBrush:
> However, without further thought, I don't know if an enum can be said to be contiguous if it contains duplicates.

I ended up modelling the meaning of contiguous to what the compiler already considers it to be:

```
main.odin(17:10) Error: Non-contiguous enumeration used as an index in an enumerated array
    X :: [enum { A=1, B=3 }]int
```

So it considers an enum as having "no gaps", and also allows duplicates, the exact opposite of sparse.

src/check_builtin.cpp
FourteenBrush: Done

src/check_builtin.cpp
FourteenBrush: Fixed

gingerBill (approved):

PR #5457, opened 2025-07-12
Added cfmakeraw()

peperronii:
Not exactly POSIX compliant but present in every build target


laytan:
This package is only meant to have bindings to POSIX specified symbols. Besides setting a terminal to raw mode is already possible with what's provided, see https://github.com/odin-lang/examples/tree/master/console/raw_console

peperronii:
Wow I didn’t know those exist. Thank you for pointing it out.

PR #5462, opened 2025-07-14
Add iOS & iPhoneSimulator subtargets for Darwin target

harold-b:
Updated PR to provides build support for iOS & iPhoneSimulator

This does not provide bundling support, but I plan to add it as a separate PR later. The bare minimum packaging is simple and there's an example in my frameworks repo.

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -172,6 +172,7 @@ struct TargetMetrics {
 enum Subtarget : u32 {
 	Subtarget_Default,
 	Subtarget_iOS,
+	Subtarget_iPhoneSimulator,
 	Subtarget_Android,
 
 	Subtarget_COUNT,
@@ -180,6 +181,7 @@ enum Subtarget : u32 {
 gb_global String subtarget_strings[Subtarget_COUNT] = {
 	str_lit(""),
 	str_lit("ios"),
+	str_lit("iphonesimulator"),
 	str_lit("android"),
 };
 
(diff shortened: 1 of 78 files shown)


src/linker.cpp
```diff
@@ -772,10 +772,54 @@ try_cross_linking:;
...
+
+				const char* original_clang_path = clang_path;
+
+				gbString darwin_find_sdk_cmd = gb_string_make(temporary_allocator(), "");
+				darwin_find_sdk_cmd = gb_string_append_fmt(darwin_find_sdk_cmd, "xcrun --sdk %s --show-sdk-path", darwin_xcrun_sdk_name);
+
+				if (!system_exec_command_line_app_output(darwin_find_sdk_cmd, &darwin_sdk_path)) {
+
+					// Fallback to default clang, since `xcrun --sdk` did not work.
+					clang_path = original_clang_path;
```
laytan, line 808: It doesn't look like this is doing anything? After line 800 where you set original_clang_path, clang_path is never changed. Should line 800 be higher up?

src/build_settings.cpp
laytan, line 1: Does a minimum macos version of 11 make sense on ios/simulator? (Line 1910 is where that default is set).

src/linker.cpp
```diff
@@ -772,10 +772,54 @@ try_cross_linking:;
...
+					darwin_platform_name  = "iPhoneOS";
+					darwin_xcrun_sdk_name = "iphoneos";
+					darwin_min_version_id = "ios";
+					clang_path            = "xcrun --sdk iphoneos clang";
+					break;
+				case Subtarget_iPhoneSimulator:
+					darwin_platform_name  = "iPhoneSimulator";
+					darwin_xcrun_sdk_name = "iphonesimulator";
+					darwin_min_version_id = "ios-simulator";
+					clang_path            = "xcrun --sdk iphonesimulator clang";
```
laytan, line 796: The `ODIN_CLANG_PATH` env variable should probably still take precedence if somebody needs a different one?

harold-b: You're right. I'll fix along with the "no platform load command found" issue

src/build_settings.cpp
harold-b: I am currently writing the fix there, and I defaulted it to 17.4 in order to have os_sync_wait_on_address available to avoid ulock_* usage by default as the private API might cause issue in the App Store. But the user can override it by manually specifying a version still.

laytan: Good idea

src/linker.cpp
harold-b: Not sure I understand this one. `clang_path` will be used to build the final `link_command_line` below so it get's restored to the original value upon `xcrun --sdk` failure. Or did I misunderstand?

harold-b: I see what you mean now

harold-b: Fixed in https://github.com/odin-lang/Odin/pull/5465

Issue #5314, opened 2025-06-11
set_thread_name(^Thread, string)


Kelimion:
### Discussed in https://github.com/odin-lang/Odin/discussions/5313

<div type='discussions-op-text'>

<sup>Originally posted by @GloriousPtr June 11, 2025</sup>
Would be great to have a way to set the thread name without using os specific API calls.</div>

GloriousPtr:
along with this it would be great to have `get_thread_name :: proc(^Thread) -> string` as well

peperronii:
After an attempt at this issue, I found out that most OSs that implements pthread have their own take at getthreadname and setthreadname here's an exhautive list of the things I've found:

note for Windows : they have SetThreadDescription and GetThreadDescription so lucky them

pthread users that have pthread get/set name extention includes : MacOS, [Linux](https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html), [OpenBSD](https://man.openbsd.org/pthreads.3#Non_Portable_Extensions), [FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=pthread_set_name_np), [NetBSD](https://man.netbsd.org/pthread_setname_np.3)

MacOS : Not documented, but I've tested it on Xcode. it does have pthread_getname_np and pthread_setname_np. Returns errors as int. Though MacOS's setthreadname cannot refer to a particular thread using pthread_t. It can only change the name of the thread it is called on. Includes pthread.h

Linux : pthread_setname_np / pthread_getname_np. Returns int. can be set/get by specifying pthread_t. Uses feature_test_macros and includes pthread.h

OpenBSD : pthread_set_name_np / pthread_get_name_np. Returns void, all errors are silently ignored. Can be set/get by specifying pthread_t. includes both pthread.h and pthread_np.h

FreeBSD : Has both Linux's and OpenBSD's implementations. Includes pthread_np.h

NetBSD: pthread_setname_np / pthread_getname_np. Returns int, _Fanciest_ of them all, getname same as Linux but setname will take a [printf](https://man.netbsd.org/printf.3) formated string and a pointer to the argument. includes pthread.h

~~I've never used thread names, but for now, if you really want cross-compability among pthreads, I'd recommend storing the names in a pthreadkey~~ - @flysand7 pointed out that the thread name will get read by other programs in the system too

Extra notes:

[...]

PR #5464, opened 2025-07-14
core:thread get_name/set_name

peperronii:
feature request [#5314](https://github.com/odin-lang/Odin/issues/5314)

tested on arm_darwin


core/thread/thread.odin
```diff
@@ -155,6 +155,28 @@ yield :: proc() {
...
+
+/*
+Set thread's name/description.
+
+If thread is nil the procedure will set the name of the calling thread.
+
+MacOS: only support changing the name of the calling thread.
+if thread is not nil the procedure will do nothing.
+*/
+set_name :: proc(thread: ^Thread, name: string) {
```
FourteenBrush: Perhaps add a note on that the name will be truncated to their platform specific max length?

Also what must the lifetime of this string be?

core/thread/thread_windows.odin
```diff
@@ -152,3 +155,39 @@ _yield :: proc() {
...
+
+	copy(buf, buf_8[:])
+
+	name = transmute(string)buf
+
+	return
+}
+
+_set_name :: proc(thread: ^Thread, name: string) {
+	t_handle : win32.HANDLE
```
FourteenBrush: `t_handle: win32.HANDLE`

core/thread/thread_windows.odin
```diff
@@ -152,3 +155,39 @@ _yield :: proc() {
...
+	if thread == nil {
+		t_handle = win32.GetCurrentThread()
+	} else {
+		t_handle = t.win32_thread
+	}
+
+	buf : [_THREAD_DESCRIPTION_LENGTH]u16
```
FourteenBrush: `buf: [_THREAD_DESCRIPTION_LENGTH]u16`

core/thread/thread_windows.odin
```diff
@@ -152,3 +155,39 @@ _yield :: proc() {
...
+_get_name :: proc(thread: ^Thread, allocator: runtime.Allocator, loc : runtime.Source_Code_Location) -> (name:string, err:runtime.Allocator_Error) {
+	t_handle : win32.HANDLE
+	if thread == nil {
+		t_handle = win32.GetCurrentThread()
+	} else {
+		t_handle = t.win32_thread
+	}
+	
+	buf_8 : [_THREAD_DESCRIPTION_LENGTH * 2]u8
+	buf_16 : [_THREAD_DESCRIPTION_LENGTH]u16
```
FourteenBrush: Also get rid of that extra space

core/thread/thread_unix.odin
```diff
@@ -182,3 +182,63 @@ _terminate :: proc(t: ^Thread, exit_code: int) {
...
 	posix.sched_yield()
 }
+
+_get_name :: proc(thread: ^Thread, allocator: runtime.Allocator, loc: runtime.Source_Code_Location) -> (name:string, err:runtime.Allocator_Error) {
+	// Haiku doesn't have pthread_getname yet
+	when ODIN_OS == .Haiku {
+		unimplemented("core:thread get_name for haiku is not yet supported")
+	}
+
+	tid : posix.pthread_t
```
FourteenBrush: Style: `tid: posix.pthread_t`

core/thread/thread_unix.odin
```diff
@@ -182,3 +182,63 @@ _terminate :: proc(t: ^Thread, exit_code: int) {
...
+	} else {
+		tid: posix.pthread_t
+		if thread == nil {
+			tid = posix.pthread_self()
+		} else {
+			tid = t.unix_thread
+		}
+	}
+
+	buf : [_MAX_PTHREAD_NAME_LENGTH]u8
```
FourteenBrush: `buf: [_MAX_PTHREAD_NAME_LENGTH]u8`

core/thread/thread.odin
```diff
@@ -155,6 +155,28 @@ yield :: proc() {
...
+get_name :: proc(thread: ^Thread, allocator := context.allocator, loc := #caller_location) -> (string, runtime.Allocator_Error) #optional_allocator_error {
+	return _get_name(thread, allocator, loc)
+}
+
+/*
+Set thread's name/description.
+
+If thread is nil the procedure will set the name of the calling thread.
+
+MacOS: only support changing the name of the calling thread.
```
FourteenBrush: Either `MacOS only supports..` or `MacOS: only support for..`

core/sys/windows/kernel32.odin
```diff
@@ -168,7 +168,7 @@ foreign kernel32 {
 	ResumeThread :: proc(thread: HANDLE) -> DWORD ---
 	GetThreadPriority :: proc(thread: HANDLE) -> c_int ---
 	SetThreadPriority :: proc(thread: HANDLE, priority: c_int) -> BOOL ---
-	GetThreadDescription :: proc(hThread: HANDLE, ppszThreadDescription: ^PCWSTR) -> HRESULT ---
+	GetThreadDescription :: proc(hThread: HANDLE, ppszThreadDescription: PWSTR) -> HRESULT ---
```
blob1807: Should be a `^PWSTR`. The function returns a pointer to a null terminated buffer to you through `ppszThreadDescription`.

core/sys/windows/kernel32.odin
peperronii, line 1: Actually I’m not sure about this API call. per documentation it might be ^PWSTR but that doesn’t make sense cause it’s a pointer to a pointer.

blob1807: It's a pointer to a pointer because it's writing the buffer's pointer at the one we give it. It doesn't copy the whole buffer back to you. It just returns a pointer to it.

core/thread/thread_windows.odin
```diff
@@ -152,3 +155,39 @@ _yield :: proc() {
...
+	if thread == nil {
+		t_handle = win32.GetCurrentThread()
+	} else {
+		t_handle = thread.win32_thread
+	}
+	
+	buf_8: [_THREAD_DESCRIPTION_LENGTH * 2]u8
+	buf_16: [_THREAD_DESCRIPTION_LENGTH]u16
+	win32.GetThreadDescription(t_handle, raw_data(buf_16[:]))
+	n := utf16.decode_to_utf8(buf_8[:], buf_16[:])
```
blob1807: `core:sys/windows` provides functionally for doing uft16 to utf8 `win32.wstring_to_utf8_alloc(ptr, -1)`

core/thread/thread_windows.odin
```diff
@@ -152,3 +155,39 @@ _yield :: proc() {
...
+_set_name :: proc(thread: ^Thread, name: string) {
+	t_handle: win32.HANDLE
+	if thread == nil {
+		t_handle = win32.GetCurrentThread()
+	} else {
+		t_handle = thread.win32_thread
+	}
+
+	buf: [_THREAD_DESCRIPTION_LENGTH]u16
+	utf16.encode_string(buf[:], name)
```
blob1807: `core:sys/windows` provides functionally for doing utf8 to uft16 `win32.utf8_to_wstring_buf()`. though it's functionally the same.

laytan (commented):
With macos not supporting setting the name of another thread, should we just not support it anywhere, so we have a consistent API?

core/thread/thread.odin
```diff
@@ -155,6 +155,31 @@ yield :: proc() {
...
 }
 
+/*
+Get thread's name/description.
+
+If thread is nil the procedure will get the name of the calling thread.
+
+allocates memory for the returned string using provided allocator.
+*/
+get_name :: proc(thread: ^Thread, allocator := context.allocator, loc := #caller_location) -> (string, runtime.Allocator_Error) #optional_allocator_error {
```
laytan: We've been moving away from optional errors in core
```suggestion
get_name :: proc(thread: ^Thread, allocator := context.allocator, loc := #caller_location) -> (string, runtime.Allocator_Error) {
```

core/thread/thread_unix.odin
```diff
@@ -182,3 +182,65 @@ _terminate :: proc(t: ^Thread, exit_code: int) {
...
+	
+	buf := make([]u8, _MAX_PTHREAD_NAME_LENGTH, allocator, loc) or_return
+
+	when ODIN_OS == .Darwin || ODIN_OS == .Linux || ODIN_OS == .FreeBSD || ODIN_OS == .NetBSD {
+		pthread_getname_np(tid, raw_data(buf), len(buf))
+	} else when ODIN_OS == .OpenBSD {
+		pthread_get_name_np(tid, raw_data(buf), len(buf))
+	}
+
+	name = transmute(string)buf
```
laytan: This string will be of length _MAX_PTHRAD_NAME_LENGTH, not the length of the actual name

core/thread/thread_unix.odin
```diff
@@ -7,7 +7,7 @@ import "core:sync"
 import "core:sys/posix"
 
 _IS_SUPPORTED :: true
-
+_MAX_PTHREAD_NAME_LENGTH :: 16
```
laytan: Is this an actual limit or something made up for this package?

core/thread/thread_unix.odin
```diff
@@ -182,3 +182,65 @@ _terminate :: proc(t: ^Thread, exit_code: int) {
 _yield :: proc() {
 	posix.sched_yield()
 }
+
+_get_name :: proc(thread: ^Thread, allocator: runtime.Allocator, loc: runtime.Source_Code_Location) -> (name:string, err:runtime.Allocator_Error) {
```
laytan:
```suggestion
_get_name :: proc(thread: ^Thread, allocator: runtime.Allocator, loc: runtime.Source_Code_Location) -> (name: string, err: runtime.Allocator_Error) {
```

core/thread/thread_unix.odin
```diff
@@ -182,3 +182,65 @@ _terminate :: proc(t: ^Thread, exit_code: int) {
...
+	} else when ODIN_OS == .OpenBSD {
+		pthread_get_name_np(tid, raw_data(buf), len(buf))
+	}
+
+	name = transmute(string)buf
+
+	return
+}
+
+_set_name :: proc(thread: ^Thread, name:string) {
```
laytan:
```suggestion
_set_name :: proc(thread: ^Thread, name: string) {
```

core/thread/thread_unix.odin
```diff
@@ -182,3 +182,65 @@ _terminate :: proc(t: ^Thread, exit_code: int) {
...
+		tid: posix.pthread_t
+		if thread == nil {
+			tid = posix.pthread_self()
+		} else {
+			tid = thread.unix_thread
+		}
+	}
+
+	buf: [_MAX_PTHREAD_NAME_LENGTH]u8
+	copy_from_string(buf[:], name)
```
laytan:
```suggestion
	copy(buf[:], name)
```

core/thread/thread_unix.odin
```diff
@@ -182,3 +182,65 @@ _terminate :: proc(t: ^Thread, exit_code: int) {
...
+	// _MAX_PTHREAD_NAME_LENGTH includes terminating null
+	buf[len(buf) - 1] = 0
+
+	when ODIN_OS == .Darwin {
+		pthread_setname_np(raw_data(buf[:]))
+	} else when ODIN_OS == .OpenBSD {
+		pthread_set_name_np(tid, raw_data(buf[:]))
+	} else when ODIN_OS == .NetBSD {
+		format := []u8{'%','s', 0}
+		pthread_setname_np(tid, raw_data(format), raw_data(buf[:]))
```
laytan: You can make the binding take a cstring and just write "%s"

core/thread/thread_windows.odin
```diff
@@ -152,3 +155,39 @@ _yield :: proc() {
 	win32.SwitchToThread()
 }
 
+_get_name :: proc(thread: ^Thread, allocator: runtime.Allocator, loc : runtime.Source_Code_Location) -> (name:string, err:runtime.Allocator_Error) {
```
laytan:
```suggestion
_get_name :: proc(thread: ^Thread, allocator: runtime.Allocator, loc: runtime.Source_Code_Location) -> (name: string, err: runtime.Allocator_Error) {
```

core/thread/thread_unix.odin
FourteenBrush: On Linux this is 16 characters, as per https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html, however I cannot find any length constraints in the BSD docs myself

FourteenBrush:
> With macos not supporting setting the name of another thread, should we just not support it anywhere, so we have a consistent API?

I'm wondering whether adding a possibility to `thread.create_*`/`thread.start` to add a name for the newly created thread would be an idea, instead of adding a `thread.set_name` proc?

Similar to how Java does it with `new Thread(Runnable, name)`, this way thread creation/starting could set the name upfront, as this name is merely a "debugging" thing and I don't see a reason why someone would arbitrarily change one.

laytan:
> > With macos not supporting setting the name of another thread, should we just not support it anywhere, so we have a consistent API?
> 
> I'm wondering whether adding a possibility to `thread.create_*`/`thread.start` to add a name for the newly created thread would be an idea, instead of adding a `thread.set_name` proc?
> 
> Similar to how Java does it with `new Thread(Runnable, name)`, this way thread creation/starting could set the name upfront, as this name is merely a "debugging" thing and I don't see a reason why someone would arbitrarily change one.

Good idea!

core/sys/windows/kernel32.odin
peperronii: Do we have to still manually allocate a buffer for the entire string or we’ll be fine just using the pointer?
edit:I forgot that we’d have to allocate for the returned string anyway, but I’m still curious how it will manage the buffer after the call.

peperronii:
> With macos not supporting setting the name of another thread, should we just not support it anywhere, so we have a consistent API?

Yeah, I was also hesitant about this but It feels weird to shun others capability just to make this one feels consistent. I think the solution to have all the OS refer to the calling thread is pretty neat though. but, I would gladly comply if it’s somehow idiomatic. 

> I'm wondering whether adding a possibility to `thread.create_*`/`thread.start` to add a name for the newly created thread would be an idea, instead of adding a `thread.set_name` proc?
> 
> Similar to how Java does it with `new Thread(Runnable, name)`, this way thread creation/starting could set the name upfront, as this name is merely a "debugging" thing and I don't see a reason why someone would arbitrarily change one.

I also think this is a good idea. but, I’m not sure about not adding thread.set_name at all though. even though I don’t personally use it.

@GloriousPtr thoughts?

core/thread/thread_unix.odin
peperronii: Brilliant

core/thread/thread_windows.odin
peperronii: `utf16.encode_string` will write a truncated version if the buffer is shorter than the string while `win32.utf8_to_wstring_buf()` returns nil.

core/thread/thread_windows.odin
peperronii: this is very useful. thank you.

FourteenBrush:
I would still get rid of the set_name, and instead move this name to thread.create, where something like __unix_thread_entry_proc and similar would set the name on the running thread itself

FourteenBrush:
@laytan I just looked at how Zig does it, and it simply returns an `error.Unsupported`, perhaps we should just also do that? :slightly_smiling_face: 

https://ziglang.org/documentation/master/std/#std.Thread.setName

Seems like a reasonable tradeoff right?, additionally still keep the idea of an initial thread name param.

peperronii:
If we’re going down the thread.create_with_name* route, I think we don’t need to have a get_name. we have to make a struct field (thread.Thread) for the name to make it callable from inside the thread, so if we want to retrieve the name, we can just t.name or something like that.

FourteenBrush:
>  I think we don’t need to have a get_name. we have to make a struct field (thread.Thread) for the name to make it callable from inside the thread

I disagree, I don't see a reason to add this field to a Thread, the kernel already represents the single point of truth for this value.
This isn't a hot spot either so in my opinion nothing wrong with performing a syscall here to obtain the value.

My idea was to do something like this:

```go
thread.create :: proc(procedure: Thread_Proc, priority: Thread_Priority = Thread_Priority.Normal, name: Maybe(string) = nil) -> ^Thread
```
(Similarly for all those other variants)
Which then calls the respective _set_name proc in the threads __entry_proc, as this would get rid of the issue that darwin does not support changing arbitrary thread's names

peperronii:
> I disagree, I don't see a reason to add this field to a Thread, the kernel already represents the single point of truth for this value. This isn't a hot spot either so in my opinion nothing wrong with performing a syscall here to obtain the value.

yeah, I suppose you’re right on not getting rid of get_thread. But we do need to add a field in thread.Thread in order to pass it inside __unix_thread_entry_proc and __windows_thread_entry_proc though… right?

FourteenBrush:
> But we do need to add a field in thread.Thread in order to pass it inside __unix_thread_entry_proc and __windows_thread_entry_proc though… right?

Looks like it, can't use context._internal here because the windows thread entry proc may not be called immediately

core/thread/thread.odin
```diff
@@ -155,6 +158,17 @@ yield :: proc() {
 	_yield()
 }
 
+/*
+Get thread's name/description.
+
+If thread is nil the procedure will get the name of the calling thread.
```
flysand7: The line that describes what the procedure does normally should be included regardless of whether or not it duplicates the first line. Also added the note for linux/BSD, because it seems relevant as "thread name" is a weird concept on those platforms.

```suggestion
This procedure returns the name of the given thread. If `thread` is `nil`, this procedure returns the name of the calling thread.

**Note(linux, bsd)**: Because the thread name is stored in as the `cmdline`, if the thread name was not set, the command that has been used to create the process as the name of the thread.
```

flysand7: Also I seem to be unable to access microsoft documentation for GetThreadDescription so not sure what happens if it's not set, is it an empty string or something else? I think should be described here as well.

core/thread/thread_other.odin
```diff
@@ -45,3 +46,10 @@ _yield :: proc() {
 	unimplemented("core:thread procedure not supported on target")
 }
 
+_get_name :: proc(thread: ^Thread, allocator : runtime.Allocator, loc : runtime.Source_Code_Location) -> (string, runtime.Allocator_Error) {
+	unimplemented("core:thread procedure not supported on target")
```
flysand7:
```suggestion
	unimplemented("core:thread procedure not supported on this target")
```

core/thread/thread_other.odin
```diff
@@ -45,3 +46,10 @@ _yield :: proc() {
...
+}
+
+_set_name :: proc(thread: ^Thread, name:string) {
+	unimplemented("core:thread procedure not supported on target")
```
flysand7:
```suggestion
	unimplemented("core:thread procedure not supported on this target")
```

core/thread/thread_name_freebsd.odin
```diff
@@ -0,0 +1,13 @@
...
+#+private
+package thread
+
+import "core:sys/posix"
+import "core:c"
+
+foreign import pthread "system:pthread"
+
+foreign pthread {
+	pthread_getname_np :: proc(thread: posix.pthread_t, name: [^]u8, len: c.size_t) -> posix.Errno ---
```
flysand7, line 13: Instead of creating 5 files for each of the platforms, I think you should just add these bindings to `sys/posix/pthread.odin`. They don't seem to be platform specific and there's no point in duplicating the files when you can bind these procedures in one place.

flysand7 (commented):
Please add the note for all procedures that take in a `name` for creating a thread/setting the thread name that `name` parameter is truncated to 16 characters on unix and 64 characters on windows.

core/thread/thread_name_freebsd.odin
peperronii: 1) it is actually platform specific, the input arguments for the foreign functions are not the same. I actually wanted to put it in the same file but I’m not sure if we can do something like a `WHEN ODIN_OS == .smth` inside a foreign block though, so I went with this.

2) These functions are not listed in the POSIX standard. I’ve been informed that the core:sys/posix adheres to it very strictly.

flysand7: Okay, then it's good. I'm not familiar with posix much, I assumed it would have been part of pthreads

FourteenBrush:
Looks fine to me, is it tested ~~on all platforms~~?

peperronii (commented):
reviewed and tested on macos_arm using thread.create and thread.get_name.

core/thread/thread_windows.odin
```diff
@@ -6,8 +6,11 @@ import "base:intrinsics"
 import "base:runtime"
 import "core:sync"
 import win32 "core:sys/windows"
+import "core:unicode/utf16"
 
 _IS_SUPPORTED :: true
+//NOTE(peperronii): not sure about the exact length but there must be a limit
+_THREAD_DESCRIPTION_LENGTH :: 64
```
FourteenBrush: Perhaps you could make this slightly longer, as I've seen thread descriptions get quite long on Windows, they aren't really names but rather actual "descriptions". I've tried setting a description of 1024 characters and this works perfectly fine, maybe set this to 128 or something?

GloriousPtr:
Is this still in review?

Kelimion:
It slipped through the cracks. We'll have to test it on the various platforms and possibly make some changes.

core/thread/thread_unix.odin
```diff
@@ -184,3 +190,60 @@ _terminate :: proc(t: ^Thread, exit_code: int) {
...
+	}
+
+	name, err = strings.clone_from_cstring(cstring(raw_data(buf[:])), allocator, loc)
+	return
+}
+
+_set_name :: proc(thread: ^Thread) {
+	when ODIN_OS == .Haiku {
+		return
+	}
```
peperronii: Considered calling unimplemented() but it seems to not terminate properly when calling from the child thread.

core/thread/thread_windows.odin
```diff
@@ -6,8 +6,11 @@ import "base:intrinsics"
 import "base:runtime"
 import "core:sync"
 import win32 "core:sys/windows"
+import "core:unicode/utf16"
 
 _IS_SUPPORTED :: true
+//NOTE(peperronii): not sure about the exact length but there must be a limit
+_THREAD_DESCRIPTION_LENGTH :: 128
```
FourteenBrush, line 13: While we are busy.. 😂 

Image

Source: [NodeJS worker_threads lib](https://nodejs.org/api/worker_threads.html#new-workerfilename-options)

peperronii:
I've double check each OS's name limit from the source and also added a test to the test_core_thread.odin. This should be ready for a final review.

core/thread/thread_unix.odin
```diff
@@ -184,3 +189,59 @@ _terminate :: proc(t: ^Thread, exit_code: int) {
 _yield :: proc() {
 	posix.sched_yield()
 }
+
+_get_name :: proc(thread: ^Thread, allocator: runtime.Allocator, loc: runtime.Source_Code_Location) -> (name: string, err: runtime.Allocator_Error) {
+	// Haiku doesn't have pthread_getname yet
+	when ODIN_OS == .Haiku {
+		unimplemented("core:thread get_name for haiku is not yet supported")
```
FourteenBrush: Haiku support was removed, so I don't think this compiles, check other places as well

Kelimion: I have had this PR open in a tab to remind me. Will be adopting for cleanup and merger next week.

Kelimion:
Barring the unforeseen, this'll go into the August release in a week and change.

Kelimion:
Thanks, @peperronii. Sorry for the long wait.

PR #5467, opened 2025-07-15
Rename `iOS` subtarget to `iPhone`, add `ODIN_PLATFORM_SUBTARGET_IOS`

harold-b:


base/runtime/core.odin
```diff
@@ -566,6 +566,8 @@ Odin_Platform_Subtarget_Type :: type_of(ODIN_PLATFORM_SUBTARGET)
 
 Odin_Platform_Subtarget_Types :: bit_set[Odin_Platform_Subtarget_Type]
 
+@(builtin)
+ODIN_PLATFORM_SUBTARGET_IOS :: ODIN_PLATFORM_SUBTARGET == .iPhone || ODIN_PLATFORM_SUBTARGET == .iPhoneSimulator
```
harold-b, line 570: I am unsure if this is the correct location for this. Or if the `@(builtin)` is needed. Please let me know and I'll update it.

laytan: Looks like it works right? Seems like a fine place and `@(builtin)` seems appropriate with the other `ODIN_` constants we have.

PR #5470, opened 2025-07-15
core:add thread cleanup proc

tralph3:
This PR adds an optional parameter to `pool_init` that allows you to define a procedure to be called on thread exit for each of the threads in the pool.

I think this won't be widely used, but for some situations it may come in handy.

For instance, in my case, I'm using SDL3 in my project, and I use a thread pool to load some images with SDL Image. SDL allocates some data per thread for error strings and some other stuff.

To ensure that gets cleaned up, you must call `CleanupTLS` per thread, but using a thread pool, there's no easy way to do that.

This PR would make that trivial.

### DIFF
--- a/core/thread/thread_pool.odin
+++ b/core/thread/thread_pool.odin
@@ -12,6 +12,7 @@ import "core:mem"
 import "core:container/queue"
 
 Task_Proc :: #type proc(task: Task)
+ThreadCleanup_Proc :: #type proc(pool: ^Pool)
 
 Task :: struct {
 	procedure:  Task_Proc,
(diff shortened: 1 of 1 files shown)


laytan:
Is https://pkg.odin-lang.org/base/runtime/#add_thread_local_cleaner not enough for your uses?

That would be for any thread, not just your pool, but you can often check if it's a thread you are interested in, in a variety of ways.

tralph3:
I'm not sure how would I use that. The pool doesn't let me run any code on arbitrary threads. The only way to run code on the threads is through tasks, which will not be deterministically asigned to threads (by design, that's fine).

laytan:
That functions allows you to add a function that is called on every thread before it is exited out of, so you don't need to run code on the thread to initialize it.

So your callback/cleanup proc would be called on every thread that exits, which is why I said you would need a way to check if this is a thread you are interested in, but that's usually doable.

PR #5472, opened 2025-07-16
Allow objective-c procedures to have their `@objc_name` attribute inferred

harold-b:
The `@objc_name` is automatically inferred if it is not specified and the procedure name is prefixed with type name specified in `@objc_type`, followed by an `_`. What 
followed the `_` is interpreted as the `@objc_name`.

Note that the commit changes appear large, but this is due to indentation change as a branch was eliminated. The body of the branch itself was not changed. Mainly initial check simplification/removals and inference check in line 1013.

### DIFF
--- a/src/check_decl.cpp
+++ b/src/check_decl.cpp
@@ -1001,119 +1001,127 @@ gb_internal String handle_link_name(CheckerContext *ctx, Token token, String lin
 
 
 gb_internal void check_objc_methods(CheckerContext *ctx, Entity *e, AttributeContext &ac) {
-	if (!(ac.objc_name.len || ac.objc_is_class_method || ac.objc_type)) {
+	if (!ac.objc_type) {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


gingerBill:
My main gut issue with this is that it does mean it is a little harder to read and know what the Obj-C name is, and relies on this convention. This convention does hold for pretty much all of Apple's code, but this could fail for other people's code.

harold-b:
The explicit `@obj_name` is always respected above all if the user still specifies it. The inference attempt only happens in the case it's not specified. It's mainly to avoid typing repetitiveness and visual noise that happens if you are implementing Objective-C classes in Odin as you also have to specify the selector separately in many cases. Such as when implementing delegates, will be off by 1 character most of the time, since they take a parameter:
image

VS with inferred name:
image

gingerBill:
Yeah it does reduce quite a bit of repetition.

I'm just worried about abuse, but I guess the people will use it, are probably not going to abuse it in the first place.

PR #5473, opened 2025-07-16
Check for invalid subtargets and add "ios" pseudo subtarget

harold-b:
This PR is supposed to be a single commit (5e9d6f9cd41e606ce4140323ad64fcaf9b56bb35) and is supposed to go after https://github.com/odin-lang/Odin/pull/5467, which includes the other commits that sneaked in this one.

The other one should be merged before this, at which point this one should show up as a single-commit, single-scope PR.

Details:

Check for invalid subtargets.
- Add 'ios' pseudo-subtarget which triggets with either iPhone or iPhoneSimulator subtargets.
- Treat an explicit 'default' subtarget as exclusive only to the default subtarget, not an other platform-compatible subtargets.
- 'generic' continues to resolve to true for any platform-compatible subtarget as it names appears to imply such behavior.


harold-b:
Now that https://github.com/odin-lang/Odin/pull/5467 has been merged, I've force-pushed the single commit to this branch on top master for a a clean merge.

PR #5484, opened 2025-07-19
crypto/hash: hash_bytes_to_buffer slice result to digest size

laytan:
I wrote this and was surprised `hashed` contained unused parts of the input buffer, does this make sense @Yawning ?

```odin
_buf: [hash.MAX_DIGEST_SIZE]byte
hashed  := hash.hash_bytes_to_buffer(.SHA256, id, _buf[:])
```


Yawning:
Sure, following the principle of least surprise this seems fine to me.

PR #5501, opened 2025-07-26
vendor:Vulkan correct SHADER_UNUSED_KHR value

RoboMage:
VK_SHADER_UNUSED_KHR was 0, it should be ~u32(0)

https://registry.khronos.org/vulkan/specs/latest/man/html/VK_SHADER_UNUSED_KHR.html

Validation layer result when it was 0:
```
Validation Layer: Validation Error: [ VUID-VkRayTracingShaderGroupCreateInfoKHR-type-03475 ] | MessageID = 0x38a68524 | vkCreateRayTracingPipelinesKHR(): pCreateInfos[0].pGroups[0] anyHitShader is 0, closestHitShader is 0, intersectionShader is 0.
The Vulkan spec states: If type is VK_RAY_TRACING_SHADER_GROUP_TYPE_GENERAL_KHR then closestHitShader, anyHitShader, and intersectionShader must be VK_SHADER_UNUSED_KHR (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-VkRayTracingShaderGroupCreateInfoKHR-type-03475)
```

### DIFF
--- a/vendor/vulkan/core.odin
+++ b/vendor/vulkan/core.odin
@@ -690,7 +690,7 @@ NV_SHADING_RATE_IMAGE_EXTENSION_NAME                      :: "VK_NV_shading_rate
 NV_ray_tracing                                            :: 1
 NV_RAY_TRACING_SPEC_VERSION                               :: 3
 NV_RAY_TRACING_EXTENSION_NAME                             :: "VK_NV_ray_tracing"
-SHADER_UNUSED_KHR                                         :: 0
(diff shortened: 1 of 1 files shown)


gingerBill:
Note we generate the vulkan code with a python script. Fix that python script, and not the Odin code.

The top of that file even states that:
```
//
// Vulkan wrapper generated from "https://raw.githubusercontent.com/KhronosGroup/Vulkan-Headers/master/include/vulkan/vulkan_core.h"
//
```

PR #5503, opened 2025-07-26
Vulkan generator - Support bitwise NOT for constants

RoboMage:
Generator was previously discarding the ~ operator in constants so values like ~(0U) were being interpreted as just 0. 

This change ensures such expressions are handled correctly.

Here's the value I focused on as reference:
https://registry.khronos.org/vulkan/specs/latest/man/html/VK_SHADER_UNUSED_KHR.html

### DIFF
--- a/vendor/vulkan/_gen/create_vulkan_odin_wrapper.py
+++ b/vendor/vulkan/_gen/create_vulkan_odin_wrapper.py
@@ -312,7 +312,11 @@ def parse_constants(f):
         value = remove_prefix(value, 'VK_')
         v = number_suffix_re.findall(value)
         if v:
-            value = v[0]
+            if '~' in value:
+                value = "~u32(" + v[0] + ")"
+            else:
+                value = v[0]
+
         f.write("{}{} :: {}\n".format(name, "".rjust(max_len-len(name)), value))
     f.write("\n")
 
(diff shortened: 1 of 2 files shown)


vendor/vulkan/_gen/create_vulkan_odin_wrapper.py
```diff
@@ -312,7 +312,11 @@ def parse_constants(f):
         value = remove_prefix(value, 'VK_')
         v = number_suffix_re.findall(value)
         if v:
-            value = v[0]
+            if '~' in value:
+                value = "~u32(" + v[0] + ")"
```
gingerBill, line 316: Are they always `u32` in size?

RoboMage: I couldn't find anything in the Vulkan spec saying constants need to be a certain size but Vulkan uses uint32_t for most values (Seems like pretty much anything not memory related).
Some other defines even seem to use uint32_t directly.
`#define VK_API_VERSION_MAJOR(version) (((uint32_t)(version) >> 22U) & 0x7FU)`
Also the generated headers already use the format else where:
https://github.com/odin-lang/Odin/blob/a93d3503b1bb647f3d1b4ec168ed6138d90ef1e3/vendor/vulkan/core.odin#L33

Seems like a safe bet to use u32 to me

Neirokan: Sorry, I accidentally noticed this PR, but it looks like there is one such constant that uses `u64`:
`#define VK_WHOLE_SIZE (~0ULL)`

RoboMage:
Hey, I was hoping we could get this merged. If there's anything I can do to help, please let me know.

wrapperup:
I'd also like to see this merged. What's blocking it?

gingerBill:
This was corrected properly much later because there was a `u64` case too. I am closing this stale PR now.

PR #5504, opened 2025-07-26
Add Dawn backend for `vendor:wgpu`

zeozeozeo:
This PR adds the [Dawn](https://dawn.googlesource.com/dawn) backend for the `vendor:wgpu` package. It can be activated with the `WGPU_USE_DAWN` #config variable.

It works with the shared build of Dawn (.dll) but I couldn't get it working with the static build (DAWN_BUILD_MONOLITHIC_LIBRARY=STATIC). It seems to be an issue on their side:
```
$ odin run .
webgpu_dawn.lib(UtilsD3D11.obj) : error LNK2019: unresolved external symbol WKPDID_D3DDebugObjectName referenced in function "void __cdecl dawn::native::d3d11::SetDebugName(class dawn::native::d3d11::Device *,struct ID3D11DeviceChild *,char const *,class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >)" (?SetDebugName@d3d11@native@dawn@@YAXPEAVDevice@123@PEAUID3D11DeviceChild@@PEBDV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@Z)
webgpu_dawn.lib(UtilsD3D12.obj) : error LNK2001: unresolved external symbol WKPDID_D3DDebugObjectName
webgpu_dawn.lib(SharedFenceD3D11.obj) : error LNK2019: unresolved external symbol __imp_CompareObjectHandles referenced in function "public: static class dawn::Result<class dawn::Ref<class dawn::native::d3d11::SharedFence>,class dawn::native::ErrorData> __cdecl dawn::native::d3d11::SharedFence::Create(class dawn::native::d3d11::Device *,struct dawn::native::StringView,struct dawn::native::SharedFenceDXGISharedHandleDescriptor const *)" (?Create@SharedFence@d3d11@native@dawn@@SA?AV?$Result@V?$Ref@VSharedFence@d3d11@native@dawn@@@dawn@@VErrorData@native@2@@4@PEAVDevice@234@UStringView@34@PEBUSharedFenceDXGISharedHandleDescriptor@34@@Z)
C:\Users\zeo\Desktop\Code\Odin\vendor\wgpu\wgpu.exe : fatal error LNK1120: 2 unresolved externals
```

TODO: wgpu_native.odin doesn't seem to trigger the linker. haven't actually checked whether Dawn supports all of the wgpu-rs extensions. Atleast it does seem to support multidraw.

TODO: add static library on windows; add and test darwin & linux support


vendor/wgpu/wgpu.odin
```diff
@@ -2,36 +2,52 @@ package wgpu
...
 	@(private) EXT  :: ".dll.lib" when WGPU_SHARED else ".lib"
-	@(private) LIB  :: "lib/wgpu-windows-" + ARCH + "-msvc-" + TYPE + "/lib/wgpu_native" + EXT
+	@(private) LIB_TYPE :: "dawn" when WGPU_USE_DAWN else "wgpu"
+	@(private) LIB_FILE :: "webgpu_dawn" when WGPU_USE_DAWN else "wgpu_native"
+	@(private) LIB  :: "lib/" + LIB_TYPE + "-windows-" + ARCH + "-msvc-" + TYPE + "/lib/" + LIB_FILE + EXT
 
 	when !#exists(LIB) {
-		#panic("Could not find the compiled WGPU Native library at '" + #directory + LIB + "', these can be downloaded from https://github.com/gfx-rs/wgpu-native/releases/tag/v25.0.2.1, make sure to read the README at '" + #directory + "README.md'")
+		when WGPU_USE_DAWN {
+			#panic("Could not find the compiled Dawn library at '" + #directory + LIB + "', these can be downloaded from https://github.com/mmozeiko/build-dawn/releases, make sure to read the README at '" + #directory + "README.md'")
```
zeozeozeo, line 20: https://github.com/mmozeiko/build-dawn/releases only provides prebuilt shared libraries, so IDK if we should even redirect people to there.
Maybe we should just tell them to build Dawn from source :trollface:

wrapperup:
Dawn should probably be a separate package (`vendor:dawn`). Usually "wgpu" specifically means the Rust implementation (which this is a package for), it doesn't really make sense to add a backend to it.

laytan:
wgpu and dawn both adhere to https://github.com/webgpu-native/webgpu-headers so making it one package isn't the worst idea, possibly with a rename to `webgpu`. It's the native extensions they add that cause problems though. If a separation is better we could do `vendor:webgpu/wgpu` and `vendor:webgpu/dawn` instead. But I do wonder if it's worth the maintenance.

When writing the package I was hoping we could just pick one backend and everyone would be happy. I have seen people say dawn is better since, but often without many arguments (it is the chrome backend, which could be why). On the other hand wgpu is smaller, easier to build, provides prebuilt libraries, and is slightly more permissively licensed.

zeozeozeo:
There are some differences between dawn and wgpu even in the non-native code. For example, Dawn expects you to pass some `callbackType` in the functions that accept it, while wgpu seems to just figure it out if you zero it out. Not ideal

PR #5508, opened 2025-07-28
Flush accumulated errors before printing assert

Paul-Andre:
In order to make debugging easier, when assert is called, "flush" all the accumulated error messages before printing the assert message. (The previous behavior was that the accumulated error messages would never get printed.)

(And tell me, what is the preferred way in this case to forward-declare the functions? I just did the lazy thing.)


gingerBill:
This is a bit of a bodge but it does work.

PR #5510, opened 2025-07-28
add cgltf filter type and wrap mode enums

taylormck:
The cgltf library has added enums for the sampler filter types and wrap modes: https://github.com/jkuhlmann/cgltf/pull/258/files. This replaces the bare ints that were being used before.

Even though the vendor library isn't using a version of the header file with these enums, they can still be added to the binding without issue or needing to update.

This PR implements those new enums.

### DIFF
--- a/vendor/cgltf/cgltf.odin
+++ b/vendor/cgltf/cgltf.odin
@@ -263,19 +263,19 @@ image :: struct {
 }
 
 filter_type :: enum c.int {
-	undefined = 0,
-	nearest = 9728,
-	linear = 9729,
+	undefined              = 0,
+	nearest                = 9728,
+	linear                 = 9729,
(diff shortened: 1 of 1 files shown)


vendor/cgltf/cgltf.odin
```diff
@@ -262,12 +262,28 @@ image :: struct {
...
+	linear_mipmap_nearest = 9985,
+	nearest_mipmap_linear = 9986,
+	linear_mipmap_linear = 9987,
+}
+
+wrap_mode :: enum c.int {
+	clamp_to_edge = 33071,
+	mirrored_repeat = 33648,
+	repeat = 10497,
+}
```
gingerBill: I'd add align the `=` bits together to make it look like the rest of the code.

PR #5516, opened 2025-07-29
Additional obj-c methods for darwin Foundation

cardboardguru76:
Added support for NSBitmapImageRep class.
Added ability to set contents to a CALayer.
I needed these to support a port of Handmade Hero, but they are of general use.


laytan:
Please use tabs instead of spaces

gingerBill (approved):

PR #5525, opened 2025-07-31
`core:hash/xxhash`: Static SIMD Support for XXH3

Barinzaya:
This adds a SIMD implementation of the kernel functions for XXH3, and uses compile-time build features to decide how large of a SIMD vector to use for the default implementation.

There are currently checks for `amd64`/`i386` to size the vectors accordingly for SSE2/AVX2/AVX512, and for other platforms will just use `runtime.HAS_HARDWARE_SIMD` to select between 128-bit vectors or scalar code.

There are also some other minor changes/improvements that were made to the code along the way, but which generally shouldn't have much tangible impact (see [the "minor changes" commit](https://github.com/odin-lang/Odin/pull/5525/commits/2f8b390c1922f096daf56c2400b212457fd37dca) for some details).

For performance comparison, here are the results of the existing XXH3_128 benchmark on 1 MiB of data in `tests/benchmark/hash` on two different `amd64` systems (my desktop PC and a Steam Deck):

| System | `master` | `xxh3-simd` (this PR) | Reference (Clang) | Reference (GCC) |
|---|---|---|---|---|
| Desktop (`x86-64-v2`/SSE, default) | 16,925 MiB/s | 46,764 MiB/s | 49,170 MiB/s | 44,093 MiB/s |
| Desktop (`x86-64-v3`/AVX2) | --- | 59,850 MiB/s | 65,987 MiB/s | 94,399 MiB/s |
| Desktop (`x86-64-v4`/AVX512) | --- | 88,772 MiB/s | 71,141 MiB/s | 148,134 MiB/s \* |
| Steam Deck (`x86-64-v2`/SSE, default) | 6,314 MiB/s | 16,006 MiB/s | 22,977 MiB/s | 21,424 MiB/s |
| Steam Deck (`x86-64-v3`/AVX2) | --- | 30,300 MiB/s | 48,212 MiB/s | 48,015 MiB/s |

\* This one actually hits almost 190 GiB/s for sizes below 1 MiB (!), but falls off at 1 MiB, probably due to running into the limits of the cache size.

I've also included runs of [the reference implementation's benchmark](https://github.com/Cyan4973/xxHash/tree/dev/tests/bench) for comparison. The changes in this PR have narrowed the gap, but there's definitely room for improvement.

Unfortunately, I don't really have any systems of other arches to test on, but I would welcome comparisons on other systems too, `amd64` or otherwise, to ensure there aren't regressions on some hardware. There were notes in some of the comments here hinting at the possibility of instruction cache sizes being a possible factor in performance here, and getting good SIMD performance has seemingly required removing some `optimization_mode="favor_size"`.

The SIMD kernel procs are also generalized to allow them to be used with the SIMD size as a parameter, which is intended to lay some groundwork for a multi-versioned implementation as well (i.e. to include SSE/AVX2/AVX512 implementations and choose the implementation at run-time using `core:sys/info`). I was hoping to include that here, but that has proven difficult to get the up-featured code to run well, so I'm going to revisit that in a possible future PR.

### DIFF
--- a/core/hash/xxhash/xxhash_3.odin
+++ b/core/hash/xxhash/xxhash_3.odin
@@ -63,10 +63,11 @@ XXH3_INTERNAL_BUFFER_SIZE :: 256
 	Streaming state.
 
 	IMPORTANT: This structure has a strict alignment requirement of 64 bytes!! **
-	Do not allocate this with `make()` or `new`, it will not be sufficiently aligned.
-	Use`XXH3_create_state` and `XXH3_destroy_state, or stack allocation.
+	Default allocators will align it correctly if created via `new`, as will
+	placing this struct on the cache, but if using a custom allocator make sure
+	that it handles the alignment correctly!
 */
-XXH3_state :: struct {
+XXH3_state :: struct #align(64) {
 	acc:               [8]u64,
 	custom_secret:     [XXH_SECRET_DEFAULT_SIZE]u8,
 	buffer:            [XXH3_INTERNAL_BUFFER_SIZE]u8,
@@ -478,7 +479,7 @@ XXH3_128bits_internal :: #force_inline proc(
 /* ===   Public XXH128 API   === */
 @(optimization_mode="favor_size")
 XXH3_128_default :: proc(input: []u8) -> (hash: XXH3_128_hash) {
-	return XXH3_128bits_internal(input, 0, XXH3_kSecret[:], XXH3_hashLong_128b_withSeed)
+	return XXH3_128bits_internal(input, 0, XXH3_kSecret[:], XXH3_hashLong_128b_default)
 }
 
 @(optimization_mode="favor_size")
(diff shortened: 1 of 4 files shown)


core/hash/xxhash/xxhash_3.odin
```diff
@@ -62,10 +63,11 @@ XXH3_INTERNAL_BUFFER_SIZE :: 256
 	Streaming state.
 
 	IMPORTANT: This structure has a strict alignment requirement of 64 bytes!! **
-	Do not allocate this with `make()` or `new`, it will not be sufficiently aligned.
-	Use`XXH3_create_state` and `XXH3_destroy_state, or stack allocation.
+	Default allocators will align it correctly if created via `new`, as will
+	placing this struct on the cache, but if using a custom allocator make sure
```
Kelimion: Placing this struct on the "cache"?

Barinzaya: Oops, meant to be "on the stack". No idea how that slipped!

Kelimion:
Looks good in general. Let us know when you think it's ready.

laytan:
There is a possibility that #5526 speeds this up

Barinzaya:
Fixed typo, found some bounds checking in the hot loop that was slowing things down a little bit (helps SSE and AVX512 slightly, AVX2 seems to be unaffected).

I noticed also that I get somewhat better results when running the benchmark through `odin test . -o:speed`, rather than `odin build . -o:speed` and then running the binary, probably due to the binary still being cached. I've switched to results from running the separate `build`/run, as that's a more realistic usecase.

I also noticed that the results for the reference benchmark are somewhat different when compiled on Clang instead of GCC, which was used by default. Added Clang results to the table.

I think I can call this ready now.

gingerBill:
LGTM

PR #5526, opened 2025-07-31
amd64 abi fixes regarding vectors

### LINKED ISSUE #5429: core/crypto/deoxysii: llvm20 causes massive performance regression
Arch Linux (btw) Edition finally updated the llvm package to llvm 20.1.6.

With everything else equal (only thing I did different was set LLVM_CONFIG and rebuild the compiler), I observe a massive performance regression between llvm 19.1.7 and llvm 20.1.6 like thus:

LLVM 19.1.7:
```
|    Deoxys-II-256|   64|    99ns| 615.423 MiB/s|
|    Deoxys-II-256| 1024|   764ns|1277.798 MiB/s|
|    Deoxys-II-256|65536|47.657µs|1311.440 MiB/s|
```

LLVM 20.1.6:
```
|    Deoxys-II-256|   64|    217ns| 280.764 MiB/s|
|    Deoxys-II-256| 1024|  2.618µs| 372.903 MiB/s|
|    Deoxys-II-256|65536|161.844µs| 386.174 MiB/s|
```

This is with both whatever I had when I last pulled, and master as of 2025/07/03, using `tests/benchmark/crypto` with `-o:speed` on an amd64 target.

laytan:
- Fixes the code so SSEUp is grouped/skipped over properly (Fixes #5429)
- Fixes f16 vectors using garbage widths, because it would call LLVMGetIntTypeWidth and an f16 is not an int so doesn't have that function (opted to support `half` properly instead of using i16 too)

### DIFF
--- a/src/llvm_abi.cpp
+++ b/src/llvm_abi.cpp
@@ -1028,7 +1028,7 @@ namespace lbAbiAmd64SysV {
 			unify(cls, ix + off/8, RegClass_Int);
 			break;
 		case LLVMHalfTypeKind:
-			unify(cls, ix + off/8, (off%8 == 2) ? RegClass_SSEHv : RegClass_SSEHs);
+			unify(cls, ix + off/8, (off%8 == 6) ? RegClass_SSEHv : RegClass_SSEHs);
 			break;
 		case LLVMFloatTypeKind:
 			unify(cls, ix + off/8, (off%8 == 4) ? RegClass_SSEFv : RegClass_SSEFs);


Yawning:
Nice work!  Thanks for looking into this.

gingerBill:
The `half` as `i16` thing was because LLVM didn't properly support them in earlier versions, which is why I did what I did. I am not sure if I made a note as to which versions did it or not, yet.

laytan:
> The `half` as `i16` thing was because LLVM didn't properly support them in earlier versions, which is why I did what I did. I am not sure if I made a note as to which versions did it or not, yet.

Ah, let me try it on llvm 14 in a bit

Barinzaya:
I think this did fix the issues I was having trying to get multi-versioning to work in the XXH3 implementation, thank you!

PR #5530, opened 2025-08-02
UTF-16 string types: `string16` & `cstring16`

gingerBill:


blob1807:
Is `intrinsics.constant_utf16_cstring` going to be deprecated or changed to emit a `cstring16`?

gingerBill:
@blob1807 It will be changed to emit a `cstring16` and then be deprecated, assuming this PR is merged at all.

PR #5533, opened 2025-08-03
add missing xlib functions for getting and setting text properties

weskerfoot:
Adding these two XLib functions which I found were missing.

I tested it and it works as expected with this example

```
get_active_window_name :: proc(display: ^xlib.Display, xid: xlib.XID) -> cstring {
  props : xlib.XTextProperty
  active_window_atom := xlib.InternAtom(display, "_NET_WM_NAME", false)
  xlib.GetTextProperty(display, xid, &props, active_window_atom)
  return cast(cstring)props.value
}
```

https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/XGetTextProperty.html
https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/XSetTextProperty.html

Let me know if there's anything else I need to add here, thanks!

### DIFF
--- a/vendor/x11/xlib/xlib_procs.odin
+++ b/vendor/x11/xlib/xlib_procs.odin
@@ -389,6 +389,18 @@ foreign xlib {
 		requestor:   Window,
 		time:        Time,
 		) ---
+	GetTextProperty :: proc(
+		display:            ^Display,
+		window:             Window,
+		text_prop_return:   ^XTextProperty,
+		property:           Atom,
+		) -> Status ---
+	SetTextProperty :: proc(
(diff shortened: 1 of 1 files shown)


weskerfoot:
I looked more closely at this and it seems like the "Status" `XGetTextProperty` returns isn't the same as the enum in other functions, should this just return i32 instead?

> If it was able to read and store the data in the [XTextProperty](https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/converting-string-lists.html) structure, XGetTextProperty() returns a nonzero status; otherwise, it returns a zero status.

PR #5535, opened 2025-08-03
Amend help run text

cardboardguru76:
Just a small change to point out that the executable is deleted after a successful odin run.
```
run		Same as 'build', but also then runs the newly compiled executable.
		If the executable exits successfully, then it is deleted.
		Append an empty flag and then the args, '-- <args>', to specify args for the output.
		Examples:
			odin run .                     Builds and runs package in current directory.
			odin run <dir>                 Builds and runs package in <dir>.
			odin run filename.odin -file   Builds and runs single-file package, must contain entry point.
```


gingerBill:
You've got two PRs mixed together here. I won't accept it.

PR #5537, opened 2025-08-04
[ci:nightly] added linux arm build

A1029384756:
- adds a linux arm version to the nightly builds
- swaps the `setup-alpine` action with a docker container bind mounted to `PWD`

the arm build uses the partner runner images offered [here](https://github.com/actions/partner-runner-images). these are images that are not "official github images" but are still offered by them and enable much faster arm builds (since any sort of QEMU translation can be skipped)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -47,20 +47,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: jirutka/setup-alpine@v1
-        with:
-          branch: edge
-      - name: (Linux) Download LLVM
+      - name: (Linux) Download LLVM and Build Odin
         run: |
-          apk add --no-cache \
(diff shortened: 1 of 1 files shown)


laytan:
Probably want to add this to the testing ci too, if we are going to provide builds we want to make sure they work. And I know that there are tests that currently fail on musl/alpine.

A1029384756:
@laytan is this something that would retroactively apply to the x86_64 builds that are done using alpine as well? the changes that were made to the build step just replace an alpine chroot with a docker image and the executable is tested immediately after with the `Odin run` step in the pipeline; no major structural changes to how the compiler is built were made. as far as the failing alpine/musl tests, that isn't something that is impacted by *building* on alpine. alpine is only used here to get a statically linked compiler binary, the behavior of the code generated by that binary on different platforms is unrelated to where the compiler was built

laytan:
Ah right, I wasn't thinking clearly. Of course this is the compiler being built, not something in Odin. Nevermind my comment.

PR #5545, opened 2025-08-06
add quaternion marshal and unmarshal to `core:encoding/json`

nico-barbas:
Add JSON marshal/unmarshal support for quaternion types in `core:encoding/json`.

- **Marshal**: Quaternions are serialized as JSON arrays with four float components in [x, y, z, w] order
- **Unmarshal**: JSON arrays of exactly 4 numeric values are parsed back into quaternion types in [x, y, z, w] order
- **Tests**: Add a corresponding test `quaternion` to the test suits `tests/core/encoding/json`

### DIFF
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -150,7 +150,23 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err: 
 		io.write_byte(w, ']')    or_return
 
 	case runtime.Type_Info_Quaternion:
-		return .Unsupported_Type
+		i, j, k, r: f64
+		switch q in a {
+		case quaternion64: i, j, k, r =  imag(q), jmag(q), kmag(q), real(q)
+		case quaternion128: i, j, k, r =  imag(q), jmag(q), kmag(q), real(q)
+		case quaternion256: i, j, k, r =  imag(q), jmag(q), kmag(q), real(q)
+		}
+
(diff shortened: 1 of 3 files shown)


gingerBill:
We don't support BECAUSE the order might be different for foreign software. There isn't a good way to make this work properly because quaternions in other places could be xyzw or wxyz.

gingerBill:
Maybe the best option is to allow us to specify the order in the struct tags. e.g. 

```odin
field_a: quaternion256 `json:"field_a,xyzw"`,
field_b: quaternion256 `json:"field_b,wxyz"`,
```

nico-barbas:
> We don't support BECAUSE the order might be different for foreign software. There isn't a good way to make this work properly because quaternions in other places could be xyzw or wxyz.

Yes, that's what I initially thought when I looked at `core:encoding/json` and had some order issues in my json file. 

However, I asked around and looked at the `core:encoding/cbor`. In the cbor marshal and unmarshal, it is always decoded as xyzw. I thought it was worth doing the same for the json one, and also consistent with Odin (quaternion are xyzw if I'm not mistaken) as well as with the core library.

> Maybe the best option is to allow us to specify the order in the struct tags. e.g.
> 
> ```odin
> field_a: quaternion256 `json:"field_a,xyzw"`,
> field_b: quaternion256 `json:"field_b,wxyz"`,
> ```

I can change my PR to take that into account if you see this as a valid and more robust way to handle it.

PR #5547, opened 2025-08-06
Implements the Apple block ABI.2010.3.16 natively via `objc_intrinsic`

harold-b:
Implements the Apple block ABI.2010.3.16 natively via `objc_intrinsic` and the `Objc_Block` builtin type.

This is meant to replace the `NS.Block` runtime type with native Block support on Darwin platforms.

It supports stack and global blocks, where global blocks are used when no captures happen. Stack blocks are used when any capture happens, including implicit context capture.

See: https://clang.llvm.org/docs/Block-ABI-Apple.html

Details on usage and design follows


harold-b:
The `objc_block()` intrinsic has the signature: 

```odin
objc_block(capture_args*, handler_proc) -> ^Objc_Block(type_of(handler_proc))
```

Example:

```odin
// No captures
objc_block(proc(){ fmt.println("Hello World") })

// Capture of a constant
objc_block(7, proc( n: int ){ fmt.printfln("Number: %v", n) })

// Capture of a local
foo := "foo"
objc_block(foo, proc( f: string ){ fmt.printfln("Foo: %v", f) })

// Capture of multiple locals
bar := "bar"
objc_block(foo, bar, proc( f, b: string ){ fmt.printfln("Foo: %v | Bar: %v", f, b) })
```


A call to `objc_block` resolves to a pointer to new builtin type representing Objective-C blocks:

```odin
Objc_Block :: struct($T: typeid) where intrinsics.type_is_proc(T) { using _: intrinsics.objc_object }
```

This exists to have a staitcally-typed representation of the block procedure meant to ensure type coherency with callsites where an API expects a specific type of block.

Example:
```objective-c
- (void) addCompletedHandler:(MTLCommandBufferHandler) block;`
```

Where `MTLCommandBufferHandler` is:
```objective-c
`typedef void (^)(id<MTLCommandBuffer>) MTLCommandBufferHandler;`
```

Which translates to in Odin:
```odin
CommandBufferHandler :: ^Objc_Block(proc "c" (_: ^CommandBuffer))`
```

And it would be used as such:
```odin
cmd_buffer->addCompletedHandler(objc_block(proc "c" ( buf: ^MTL.CommandBuffer) {
    /// ...
}))
```


### Capturing values
Values can be captured explicitly by passing them as arguments before the final argument, which is the handler proc argument. These values are captured by copy and stored in the Block literal as specified in the ABI.

If specified handler proc uses the Odin calling convention, then the current scope's context is captured implicitly and stored in the Block literal. This context is then set as the current context from the block's invoke procedure before calling the handler procedure.

If there is no context in the current scope and, the Odin calling convention is used, then it will yield a compiler error.

Captured arguments are received in the handler proc as parameters which appear after any parameters (if any) required in the block's type signature.


### Global and Local blocks

Both global and local blocks are supported.

Global blocks are used when there are no captures and are initialized as a global constant struct.

Local blocks are used whenever there is a capture, including an implicit `context` capture.

Local blocks are, as per ABI, allocated on the stack. The receiver of the block is responsible for calling `copy` on it to convert it to a malloc block. ([See source reference here](https://github.com/apple-oss-distributions/libclosure/blob/3668b0837f47be3cc1c404fb5e360f4ff178ca13/runtime.cpp#L882)).


The previous example has no captures, and uses a "c" calling convention (which means no implicit context is captured), this yields a global block.

The emitted IR looks like this:
```llvm
; Declarations
@"__$objc_block_desc_2" = global %"__$ObjC_Block_Descriptor_2" { i64 0, i64 40 }
@"__$objc_block_literal_2" = global %"__$ObjC_Block_Literal_2" { ptr @_NSConcreteGlobalBlock, i32 268435456, i32 0, ptr @"__$objc_block_invoker_2", ptr @"__$objc_block_desc_2" }


; Call site
call void @"darwodin_Metal::CommandBuffer_addCompletedHandler"(ptr %26, ptr @"__$objc_block_literal_2"), !dbg !6803


; Invoker procedure
define private void @"__$objc_block_invoker_2"(ptr %0, ptr %1) {
decls:
  br label %entry

  call void @"rock::[gpu_darwin.odin]::_gpu_frame_present_impl$anon-10"(ptr %1)
  ret void
}

```

If it is changed so that it uses the Odin calling convention:
```odin
cmd_buffer->addCompletedHandler(objc_block(proc ( buf: ^MTL.CommandBuffer) {
    // ...
}
```

Then the `context` will be implicitly captured and a stack block will ne used:
```llvm
; Decl
%"__$ObjC_Block_Descriptor_2" = type { i64, i64 }
%"__$ObjC_Block_Literal_2" = type { ptr, i32, i32, ptr, %"__$ObjC_Block_Descriptor_2", %"runtime::Context" }


; Call site
%"__$objc_block_literal_2" = alloca %"__$ObjC_Block_Literal_2", align 8, !dbg !6817
...
%isa = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %"__$objc_block_literal_2", i32 0, i32 0, !dbg !6817
%flags = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %"__$objc_block_literal_2", i32 0, i32 1, !dbg !6817
%reserved = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %"__$objc_block_literal_2", i32 0, i32 2, !dbg !6817
%invoke = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %"__$objc_block_literal_2", i32 0, i32 3, !dbg !6817
%descriptor = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %"__$objc_block_literal_2", i32 0, i32 4, !dbg !6817
store ptr @_NSConcreteStackBlock, ptr %isa, align 8, !dbg !6817
store i32 0, ptr %flags, align 4, !dbg !6817
store i32 0, ptr %reserved, align 4, !dbg !6817
store ptr @"__$objc_block_invoker_2", ptr %invoke, align 8, !dbg !6817
store ptr @"__$objc_block_desc_2", ptr %descriptor, align 8, !dbg !6817
%context = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %"__$objc_block_literal_2", i32 0, i32 5, !dbg !6817
call void @llvm.memcpy.p0.p0.i64(ptr align 1 %context, ptr align 1 %__.context_ptr, i64 112, i1 false), !dbg !6817
call void @"darwodin_Metal::CommandBuffer_addCompletedHandler"(ptr %26, ptr %"__$objc_block_literal_2"), !dbg !6817


; Invoker
define private void @"__$objc_block_invoker_2"(ptr %0, ptr %1) {
decls:
  br label %entry

  %context = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %0, i32 0, i32 5
  call void @"rock::[gpu_darwin.odin]::_gpu_frame_present_impl$anon-10"(ptr %1, ptr %context)
  ret void
}
```

Even if an odin calling convention is used in the proc, the type signature of the block returned by `objc_block()` will use the "c" calling convention to remain type-compatible with what the target expects. This is fine since we always use an intermediate `invoke` proc that is of C calling convention and does the capture forwarding/context setup.


### ^NSObject Captures

When there are captured arguments which are Objective-C objects, the block will be generated with copy and dispose helpers, as specified by the ABI, which are called to ensure correct retain cycles on those objects when the block is copied or disposed.

Changing the previous example to capture an Objective-C object:

```odin
cmd_buffer->addCompletedHandler(objc_block(device, proc( buf: ^MTL.CommandBuffer, device: ^MTL.Device ) {
    // ...
}))
```

Generates the following block descriptor and literal types, along with the copy/dispose helper procedures:

```llvm
; Decl
@"__$objc_block_desc_2" = global %"__$ObjC_Block_Descriptor_2" { i64 0, i64 176, ptr @"__$objc_block_copy_helper_2", ptr @"__$objc_block_dispose_helper_2" }
%"__$ObjC_Block_Literal_2" = type { ptr, i32, i32, ptr, %"__$ObjC_Block_Descriptor_2", %"runtime::Context", ptr }


; Helpers
define private void @"__$objc_block_copy_helper_2"(ptr %0, ptr %1, i32 %2) {
decls:
  br label %entry

  %3 = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %0, i32 0, i32 6
  %4 = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %1, i32 0, i32 6
  %5 = load ptr, ptr %4, align 8
  call void @_Block_object_assign(ptr %3, ptr %5, i32 3)
  ret void
}

define private void @"__$objc_block_dispose_helper_2"(ptr %0, i32 %1) {
decls:
  br label %entry

  %2 = getelementptr inbounds nuw %"__$ObjC_Block_Literal_2", ptr %0, i32 0, i32 6
  %3 = load ptr, ptr %2, align 8
  call void @_Block_object_dispose(ptr %3, i32 3)
  ret void
}
```

harold-b:
I'll fix the build issues

gingerBill:
So I love this. It is adding closures to Odin for darwin-only platforms through the backdoor. But it's also kind of necessary. We did have the current approach already, but it wasn't type safe in the slightest and a bit of annoying to type out too.

PR #5553, opened 2025-08-07
Fix ICE on missing procedure in base:runtime

flysand7:
## Context

Returned to working on my odin-os project, found out it doesn't compile because `base:runtime` just got updated and the compiler is crashing with my runtime implementation.

## Detail

When a required built-in procedure is missing from the base:runtime package, an assert should be triggered. However this does not happen and instead the compiler crashes silently. The cause is the null-dereference after scope_lookup_current returns nullptr.

This adds an assertion that the runtime procedure is found, before proceeding to check it's type and performing further lookups.


flysand7:
Oops, that's not the fix. I messed up my git commands and pushed the wrong thing

Kelimion:
I figured it wasn't a fix so much as getting it to fail early at a more informative position. 

And it does match your explanation: "This adds an assertion that the runtime procedure is found". That's why I merged it anyway.

PR #5554, opened 2025-08-07
Print the name of missing runtime procedure

flysand7:
This continues the fix made at 46b7abee9fdd339bac31fe7bc79551f2b3881f46, by asserting on missing entity early and printing it's name if its not found.


flysand7:
Okay now the assertion triggers properly with all the information:

```
src/llvm_backend_proc.cpp(988): Assertion Failure: `e != nullptr` Runtime procedure not found: string_eq
```

I will now undraft this

PR #5555, opened 2025-08-07
Update runtime doc file

flysand7:
This updates and reformats the doc file for the runtime package. The list of things that need to be implemented for a compiler runtime to work has been updated in accordance with recent changes to runtime (notably `cstring16` and `string16` types and `#load_directory`). Some copy-paste errors around string-comparison procedures have been fixed.

The file has been renamed to `doc.odin` as per convention, the comment has been moved to the top of the package declaration and the comment is now markdown-formatted.

Also wrote down a little map to maybe slightly help with understanding navigation within package runtime's files and added a few more explanations as to why that list of declarations exists and how to use it.

The list of declarations has been detailed to some degree, explaining under what conditions the declarations are required.


flysand7:
nice number

flysand7:
Noticing some errors right away, I'll get back to it tomorrow

PR #5556, opened 2025-08-08
Define Integer Division By Zero

gingerBill:
LLVM does not define what happens with integer division by zero:
> https://llvm.org/docs/LangRef.html#id124
> 
> Division by zero is undefined behavior. For vectors, if any element of the divisor is zero, the operation has undefined behavior.

As Odin strives to define as much behaviour as possible, we have to define this behaviour directly. Luckily, integer division by zero is low-hanging fruit of which can be trivially defined.

For the PR, the current default behaviour has been set to TRAP (subject to change).

Possible options:

* `-integer-division-by-zero:trap`
  * Any division/modulo/remainder by zero will result in a trap
* `-integer-division-by-zero:zero`
  * `a/0  == 0`
  * `a%0  == a`
    * `r = a - b*trunc(a/b)`
  * `a%%0 == a`
    * `r = a - b*floor(a/b)`
* `-integer-division-by-zero:self`
  * `a/0  == a`
  * `a%0  == 0`
    * `r = a - b*trunc(a/b)`
  * `a%%0 == 0`
    * `r = a - b*floor(a/b)`

And can be changed on a per-file basis with `#+feature` tags:

* `#+feature integer-division-by-zero:trap`
* `#+feature integer-division-by-zero:zero`
* `#+feature integer-division-by-zero:self`

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2583,7 +2583,7 @@ gb_internal int print_show_help(String const arg0, String command, String option
 	}
 
 	if (check) {
-		if (print_flag("-integer-division-by-zero:<strng>")) {
+		if (print_flag("-integer-division-by-zero:<string>")) {
 			print_usage_line(2, "Specifies the default behaviour for integer division by zero.");
 			print_usage_line(2, "Available Options:");
 			print_usage_line(3, "-integer-division-by-zero:trap        Trap on division/modulo/remainder by zero");


src/main.cpp
```diff
@@ -2561,7 +2580,19 @@ gb_internal int print_show_help(String const arg0, String command, String option
 		if (print_flag("-ignore-warnings")) {
 			print_usage_line(2, "Ignores warning messages.");
 		}
+	}
 
+	if (check) {
+		if (print_flag("-integer-division-by-zero:<strng>")) {
```
itziksn:
```suggestion
		if (print_flag("-integer-division-by-zero:<string>")) {
```

edubart:
You could also add an option to reproduce the same semantics from RISC-V ISA, which doesn't trap on integer division, here is an excerpt from the spec:

image

gingerBill:
@edubart I could add this behaviour quite easily, but I never thought that all-bits-set was actually that useful in practice. But adding it should be pretty easy to do.

gingerBill:
@edubart I've added `-integer-division-by-zero:all-bits` which matches this behaviour for when the divisor is `0`.

PR #5581, opened 2025-08-14
Cleanup in `math/rand` and `runtime/random_generator`

Xotchkass:


core/math/rand/rand.odin
```diff
@@ -508,22 +505,12 @@ Possible Output:
...
-		if pos == 0 {
-			val = int63(gen)
-			pos = 7
-		}
-		p[n] = byte(val)
-		val >>= 8
-		pos -= 1
-	}
-	return
+	if !runtime.random_generator_read_bytes(gen, p) {return 0}
```
Xotchkass, line 508: Alternatively, assert it?

Xotchkass: Or even better would be change signature to `read :: proc(p: []byte, gen := context.random_generator) -> (ok: bool)` and just propagate return value of `random_generator_read_bytes` as the only reason why `n` wouldn't be the same as `len(p)` is if rng is broken, in which case it's 0 anyway.
But it would be a backward incompatible change, and I don't know if it worth it.

gingerBill:
```odin
return runtime.random_generator_read_bytes(gen, p)
```

Xotchkass: Well, like I said above it would require changing signature, as `read` return int and `runtime.random_generator_read_bytes` returns bool. And changing signature would break for example the `encoding/uuid`
https://github.com/odin-lang/Odin/blob/71e992d5bf0e3169ddaa90347e6cd264d5394de4/core/encoding/uuid/generation.odin#L43
https://github.com/odin-lang/Odin/blob/71e992d5bf0e3169ddaa90347e6cd264d5394de4/core/encoding/uuid/generation.odin#L66
https://github.com/odin-lang/Odin/blob/71e992d5bf0e3169ddaa90347e6cd264d5394de4/core/encoding/uuid/generation.odin#L108
...

gingerBill (approved):

PR #5586, opened 2025-08-14
Use Maybe in sdl.RenderGeometry

thetarnav:
In [sdl docs](https://wiki.libsdl.org/SDL3/SDL_RenderGeometry) the texture parameter for `sdl.RenderGeometry` is optional,
which was not obvious from the bindings alone.
`Maybe` should help expressing that.

### DIFF
--- a/vendor/sdl3/sdl3_render.odin
+++ b/vendor/sdl3/sdl3_render.odin
@@ -212,8 +212,8 @@ foreign lib {
 	RenderTextureAffine              :: proc(renderer: ^Renderer, texture: ^Texture, srcrect: Maybe(^FRect), origin, right, down: Maybe(^FPoint)) -> bool ---
 	RenderTextureTiled               :: proc(renderer: ^Renderer, texture: ^Texture, srcrect: Maybe(^FRect), scale: f32, dstrect: Maybe(^FRect)) -> bool ---
 	RenderTexture9Grid               :: proc(renderer: ^Renderer, texture: ^Texture, srcrect: Maybe(^FRect), left_width, right_width, top_height, bottom_height: f32, scale: f32, dstrect: Maybe(^FRect)) -> bool ---
-	RenderGeometry                   :: proc(renderer: ^Renderer, texture: Maybe(^Texture), vertices: [^]Vertex, num_vertices: c.int, indices: Maybe([^]c.int), num_indices: c.int) -> bool ---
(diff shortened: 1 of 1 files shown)


flysand7:
This seems to violate the ABI. Unlike Rust, in Odin there's no Maybe optimizations for pointers. `Maybe(^T)` expresses something that is either an address, nil, or an absence of any (nil included) kind of address.

Thus the `Maybe(^T)` type in this case is still something like 12 bytes with 4 bytes for a boolean indicating presence and 8 bytes for the pointer value.

---

That said, I'm kinda confused. I'm assuming you are working on something, and having implemented this change, your code still works? In which case I must be wrong about the size maybe of a pointer.

Regardless of whether it works or not, I'd prefer to not mark nil-able pointers as `Maybe(^T)`. This is because in all bindings that vendor collection does, this kind of pattern is never used, so it would stand out at the very least.

laytan:
> This seems to violate the ABI. Unlike Rust, in Odin there's no Maybe optimizations for pointers. `Maybe(^T)` expresses something that is either an address, nil, or an absence of any (nil included) kind of address.
> 
> Thus the `Maybe(^T)` type in this case is still something like 12 bytes with 4 bytes for a boolean indicating presence and 8 bytes for the pointer value.
> 
> That said, I'm kinda confused. I'm assuming you are working on something, and having implemented this change, your code still works? In which case I must be wrong about the size maybe of a pointer.
> 
> Regardless of whether it works or not, I'd prefer to not mark nil-able pointers as `Maybe(^T)`. This is because in all bindings that vendor collection does, this kind of pattern is never used, so it would stand out at the very least.

That's not true, in Odin a Maybe of a pointer is the same size as the pointer, it's either `nil` or "present". `size_of(Maybe(^T)) == size_of(T)`

thetarnav:
> Regardless of whether it works or not, I'd prefer to not mark nil-able pointers as Maybe(^T). This is because in all bindings that vendor collection does, this kind of pattern is never used, so it would stand out at the very least.

The only reason I'm proposing this change is because `Maybe` is already being used to express that `nil` is a valid value.
For example in `RenderFillRect`: https://github.com/odin-lang/Odin/blob/1215f9587ab6f59b18028400cbbda5152cdf71ca/vendor/sdl3/sdl3_render.odin#L206-L209
https://wiki.libsdl.org/SDL3/SDL_RenderFillRect

gingerBill:
I've had to add this manually because how old the PR is.

PR #5591, opened 2025-08-15
Update DXC binaries to 1.8.2505.32

connnnal:
https://www.nuget.org/api/v2/package/Microsoft.Direct3D.DXC/1.8.2505.32

I diffed against `1.7.2212.23` and didn't see any API changes. I've not added bindings for the new `Support/ErrorCodes.h`, which contains a dozen dxc-specific `HRESULT` definitions.


gingerBill:
We won't merge this PR because we cannot accept random binaries from other than trusted core contributors.

PR #5593, opened 2025-08-16
Add `fill_slice` proc to `math/rand`

Xotchkass:


FourteenBrush:
Can't you just change `rand.read` to allow any slice of ints? Also I'm wondering why results cannot be negative?

Xotchkass:
> Also I'm wondering why results cannot be negative?

Well, all `i*` functions in rand return positives, so for consistency.

> Can't you just change rand.read to allow any slice of ints?

Mostly because `read` return number of bytes written, which doesn't make much sense when different type is passed. And changing it's signature may break other things (like here https://github.com/odin-lang/Odin/pull/5581#discussion_r2276669438).

gingerBill:
No. Other than being a pooly named procedure `fill_slice` but only allows integers, the logic is very specific too as in it only returns unsigned numbers. 

If it meets your needs, that's great, but this is for the core library and I cannot see a need for this.

PR #5595, opened 2025-08-16
Fix rect layout for stb_truetype

fendevel:
The intended rect memory layout differs from the one imported from stb/rect_pack, causing all sorts of problems when reading from or writing to them. This replaces the import with an exact match.

### DIFF
--- a/vendor/stb/truetype/stb_truetype.odin
+++ b/vendor/stb/truetype/stb_truetype.odin
@@ -1,7 +1,6 @@
 package stb_truetype
 
 import c "core:c"
-import stbrp "vendor:stb/rect_pack"
 
 @(private)
 LIB :: (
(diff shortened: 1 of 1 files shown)


fendevel:
Oh, whoops!

gingerBill:
Erm... this isn't a proper fix. The proper fix is to make sure to build `stb_truetype.h` with `STB_RECT_PACK_VERSION` enabled thus with compiling with it. I will not accept this PR.

gingerBill:
I have fixed this in this commit
: https://github.com/odin-lang/Odin/commit/a0f6084b48e86295e3e6f4c444cbd936651fbdb0

fendevel:
Thanks for looking into it!

PR #5600, opened 2025-08-19
fixing `miniaudio` build issues for `wasm` target

ekliot:
Matching the `#+build` directives to those in `core:sys/posix`

### DIFF
--- a/vendor/miniaudio/common_threading.odin
+++ b/vendor/miniaudio/common_threading.odin
@@ -1,0 +1,12 @@
+#+build !linux
+#+build !darwin
+#+build !netbsd
+#+build !openbsd
+#+build !freebsd
+#+build !haiku
+package miniaudio
+
+thread    :: distinct rawptr
+mutex     :: distinct rawptr
+event     :: distinct rawptr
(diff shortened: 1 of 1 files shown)


gingerBill:
Why not just exclude `wasm` on those targets rather than explicitly adding everything else?

ekliot:
@gingerBill I'm not clear what you're referring to with excluding `wasm`, but my reasoning for the changes:
- `common_unix.odin` imports `posix`, so it ought to only build on `posix` targets (rather than all of `!windows`)
- that leaves non-posix targets that need those typedefs (which are used elsewhere in the package), so I renamed `common_windows.odin` and added those exclusions so that `windows`, `wasm`, and other targets get the expected types (this seemed the most correct, if least elegant, option)

happy to adjust to another way to organize this.

perhaps to clarify the issue I'm trying to address: when building for `wasm` target, the `posix` and `libc` stuff gets included but is unavailable. so we need the `wasm` target to get those types when building, without missing or double-defining for other targets

ekliot:
@gingerBill any word?

PR #5619, opened 2025-08-26
Add SDL_shadercross bindings to vendor

thetarnav:
Adds bindings for this little library: https://github.com/libsdl-org/SDL_shadercross

There are no releases, but the windows binaries *could* be taken from github action artifacts:
https://github.com/libsdl-org/SDL_shadercross/actions/runs/17222511959

It requires other libraries though, that are not bundled in, so I'm not sure if it's worth adding them all:
```
dxcompiler.dll      libspirv-cross-c-shared.dll  SDL3.dll
dxil.dll            libstdc++-6.dll              SDL3_shadercross.dll
libgcc_s_seh-1.dll  libwinpthread-1.dll          shadercross.exe*
```

### DIFF
--- a/vendor/sdl3/shadercross/LICENSE.txt
+++ b/vendor/sdl3/shadercross/LICENSE.txt
@@ -1,0 +1,18 @@
+Copyright (C) 2024 Sam Lantinga <slouken@libsdl.org>
+  
+This software is provided 'as-is', without any express or implied
+warranty.  In no event will the authors be held liable for any damages
+arising from the use of this software.
+
+Permission is granted to anyone to use this software for any purpose,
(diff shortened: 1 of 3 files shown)


gingerBill:
My biggest worry is the amount of other libraries it requires. It's not a simple library to bind to because of that.

My gut is saying it's not a good fit for `vendor`.

thetarnav:
That's fine too.
My goal is just sharing odin bindings.
I'm working on linux so I don't benefit much from binaries in vendor anyway.

thetarnav:
Moved the bindings to a gist: https://gist.github.com/thetarnav/3754d24410a5a238796ad6be29e3783b

Issue #5446, opened 2025-07-09
Procedure group with compile-time constant parameter fails in an unambiguous case.


mgoncerz:
I have a use case where I would like to combine in-lined when clauses with variable number of input arguments (composing different numbers of transformations with normalization after arbitrary amount of steps compiled in). While experimenting I found that including a compile-time constant as a parameter overlapping with other non-constant ones, causes the compilation to fail.

```
package main 

import "core:fmt"

astruct :: struct {
    a, b: f32
}

compile_fails :: proc {test1, test3,}
compiles_fine :: proc {test2, test3,}


test1 :: proc(i1: int, $i2: int) {

    when i2 == 1 {
        fmt.println("1")
    } else {
        fmt.println("not 1")
    }
}

test2 :: proc(i1: int, i2: int) {

    if i2 == 1 {
        fmt.println("1")
    } else {
        fmt.println("not 1")
    }
}

test3 :: proc(i1: int, s: astruct) {

    fmt.println(s.a, s.b)
}

main :: proc() {
    
    s := astruct{1,2}

    compiles_fine(1, 2)
    compiles_fine(1, s)

    compile_fails(1, 2)
    compile_fails(1, s)
}
```

The group with procedure using normal argument and if statement works perfectly fine, the one with constant-value integer causes the following compiler error:

```
PS E:\projects\odin\other> odin build .
E:/projects/odin/other/test.odin(44:22) Error: Expected a constant value for this polymorphic name parameter, got s
        compile_fails(1, s)
```

However, it works fine if I make the structure constant in the code, so I suppose having a constant parameter forces every other in that place to be one as well. Unfortunately, this is not ideal for most use cases and I think something that could be handled, because the selection of correct procedure is not ambiguous (at least in this simple example).

mgoncerz:
I'll try to illustrate the usefulness with a more real life code example. All of the procedures below have distinct signatures, however, if you combine them in a single procedure group, it will generate a cascade of compiler errors. This is because having a procedure with (A, B, C: ^Quaternion($T), $normalizeEvery: int, $normalizeAtEnd: bool) signature, causes the compiler/parser to always expect a constant value at the 4th position, which clashes with (A, B, C, D: ^Quaternion($T) ... ) and so on.

Of course, this can be (partially) mitigated by using a fixed size array or splitting everything into separate procedure groups for each number of multiplicants, but it would be nice to be able to gather them into a single compose group. Especially that it works fine with non-constant parameters.

```
@(private)
[...]

PR #5627, opened 2025-08-28
skip errors on polymorphic procs when in a proc group with other options

A1029384756:
This resolves an issue where polymorphic procedures would immediately fail a procedure group, even when there were more procedures to check. For example:
```odin
foo :: proc { bar, baz }
bar :: proc(text: string) {}
baz :: proc($text: string) {}

foo("foo")
bar := "bar"
foo(bar)
```

Would ostensibly use `baz` in the case of the constant string while it would use `bar` in the case of the variable string. This *previously* has worked but was removed in this commit: [Add error message for non-constant polymorphic name parameters](https://github.com/odin-lang/Odin/commit/5877017d30999388c832f6467116336733b607e2). That was seemingly done to prevent the following code:
```odin
foo :: proc($text: string) {}

bar := "bar"
foo(bar)
```
but had impacts on the proc group system. As of now, this fix just special cases the situation where the user is in a proc group that has multiple resolution candidates. In testing, this still produces sane error messages in the case where no variant can be satisfied and doesn't mess up cases where there is only a single proc in the group.

Fixes: #5446

Addendum:
It should be noted that the first example *does* compile and doesn't complain about the proc group until the `foo(bar)` line. If the behavior added by this PR isn't intended, there should be an error message for "duplicate" procs in the group in that case.

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -7126,6 +7126,7 @@ gb_internal CallArgumentData check_call_arguments_proc_group(CheckerContext *c, 
 	gbString expr_name = expr_to_string(operand->expr);
 	defer (gb_string_free(expr_name));
 
+	c->in_proc_group = true;
 	for_array(i, procs) {
 		Entity *p = procs[i];
 		if (p->flags & EntityFlag_Disabled) {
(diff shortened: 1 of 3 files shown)


gingerBill:
This is a very good question. I think what you've done is correct.

PR #5629, opened 2025-08-29
Fix rbtree.find_or_insert

### LINKED ISSUE #5628: `rbtree.find_or_insert` doesn't update existing node
## Context

Odin:    dev-2025-08:2f0b018db
OS:      Fedora Linux 42 (KDE Plasma Desktop Edition), Linux 6.15.10-200.fc42.x86_64
CPU:     AMD Ryzen 7 5700X 8-Core Processor
RAM:     32008 MiB
Backend: LLVM 20.1.8

## Expected Behavior

`rbtree.find_or_insert` to update the existing node if the key is already present as per the documentation:
  
_find_or_insert attempts to insert the value into the tree, and returns the node, a boolean indicating if the value was inserted, and the node allocator error if relevant. **If the value is already present, the existing node is updated.**_

I would also expect the documentation to read: _…If the **key** is already present…_

## Current Behavior

If the key is already present, the existing node **isn't** updated, the procedure just returns:

```odin
...
case .Equal:
	return
}
...
```

Should be a simple fix if the expected behavior is correct (otherwise it's worth updating the documentation).

`avl` has a `find_or_insert` procedure that behaves in the same way (API is meant to mimick that), but in that case it aligns with the documentation: _...If the value is already present, the existing node is returned un-altered._

### Steps to Reproduce

Here is a simple standalone file to showcase the behavior:

```odin
package main

import "core:container/rbtree"
import "core:fmt"

main :: proc() {
	rbt: rbtree.Tree(int, int)
	rbtree.init_ordered(&rbt)
	defer rbtree.destroy(&rbt)

	rbtree.find_or_insert(&rbt, 5, 50)
	rbtree.find_or_insert(&rbt, 5, 100)

	fmt.println("Current value:", rbtree.find(&rbt, 5).value)
}
```

### Failure Logs

The file produces the following output: **Current value: 50**

smercer10:
Update the existing node's value if the key is already present and align documentation with behavior.

Fixes #5628

### DIFF
--- a/core/container/rbtree/rbtree.odin
+++ b/core/container/rbtree/rbtree.odin
@@ -128,9 +128,9 @@ find_value :: proc(t: ^$T/Tree($Key, $Value), key: Key) -> (value: Value, ok: bo
 	return
 }
 
-// find_or_insert attempts to insert the value into the tree, and returns
-// the node, a boolean indicating if the value was inserted, and the
-// node allocator error if relevant.  If the value is already present, the existing node is updated.
+// find_or_insert attempts to insert the key-value pair into the tree, and returns
+// the node, a boolean indicating if a new node was inserted, and the
+// node allocator error if relevant. If the key is already present, the existing node is updated and returned.
 find_or_insert :: proc(t: ^$T/Tree($Key, $Value), key: Key, value: Value) -> (n: ^Node(Key, Value), inserted: bool, err: runtime.Allocator_Error) {
 	n_ptr := &t._root
 	for n_ptr^ != nil {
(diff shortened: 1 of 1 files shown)


smercer10:
`insert_or_update` may also be a more suitable name for this behavior — it will diverge from the `avl` API, but I'm not sure if that's a bad thing, considering the fundamental difference (`rbtree` is associative and `avl` isn't). 

Alternatively, `rbtree.find_or_update` can retain the current behavior and the documentation just needs to be amended, but it's quite clunky/inefficient if you want to update a value.

PR #5632, opened 2025-08-29
windows i386 support

kalsprite:
base/runtime/entry_windows.odin - Correct entry point generation
base/runtime/core_builtin.odin - TLS disabled for temp allocator
core/testing/signal_handler_libc.odin - TLS disabled for test indices
src/llvm_backend.cpp - Entry point handling for test builds
src/linker.cpp - Skip /ENTRY: directive for i386 with CRT
src/main.cpp - Warning message about TLS limitations

I wasn't sure if there is a better location to put the TLS limit warning.

### DIFF
--- a/core/testing/signal_handler_libc.odin
+++ b/core/testing/signal_handler_libc.odin
@@ -24,10 +24,18 @@ import "core:terminal/ansi"
 @(private="file") stop_test_passed: libc.sig_atomic_t
 @(private="file") stop_test_alert:  libc.sig_atomic_t
 
-@(private="file", thread_local)
-local_test_index: libc.sig_atomic_t
-@(private="file", thread_local)
-local_test_index_set: bool
+when ODIN_ARCH == .i386 && ODIN_OS == .Windows {
+	// Thread-local storage is problematic on Windows i386
+	@(private="file")
+	local_test_index: libc.sig_atomic_t
+	@(private="file")
+	local_test_index_set: bool
+} else {
+	@(private="file", thread_local)
+	local_test_index: libc.sig_atomic_t
+	@(private="file", thread_local)
+	local_test_index_set: bool
+}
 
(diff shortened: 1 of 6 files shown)


core/testing/signal_handler_libc.odin
```diff
@@ -24,10 +24,18 @@ import "core:terminal/ansi"
 @(private="file") stop_test_passed: libc.sig_atomic_t
 @(private="file") stop_test_alert:  libc.sig_atomic_t
 
-@(private="file", thread_local)
-local_test_index: libc.sig_atomic_t
-@(private="file", thread_local)
-local_test_index_set: bool
+when ODIN_ARCH == .i386 && ODIN_OS == .Windows {
+	// Thread-local storage is problematic on Windows i386
```
gingerBill, line 28: But why?

kalsprite: For my use case, I just need single-threaded x32 windows, and updating the compiler to do proper TLS on x32 windows seems like a substantial work statement, that very few people will ever use.

  As I understand it, you get a static 64 slots, cross-dll boundary problems, and it needs special handling via
  __declspec(thread). The main issue is that DLLs with static TLS can't be loaded dynamically with LoadLibrary - they crash on pre-Vista and have reliability issues even on newer versions.

  Rather than implementing a complex workaround for a niche platform that's essentially legacy at this point, I just disabled TLS for the few places that needed it (temp allocator, test harness) and made them use global variables instead.

  example issue from raymond chen: https://devblogs.microsoft.com/oldnewthing/20101122-00/?p=12233
  MS x32 TLS limitations: https://learn.microsoft.com/en-us/cpp/parallel/thread-local-storage-tls?view=msvc-170
  
I think if TLS support is desired, it could be done as a separate patch. This gets us working, and that makes it back to full featured.

PR #5646, opened 2025-09-02
fix(sys/linux): incorrect epoll_wait & EPoll_Event on  arm64 & riscv64 

xb-bx:
previous code uses epoll_pwait for arm64 & riscv64 and epoll_wait for other archs, but it uses it incorrectly.
The syscall expects timeout in milliseconds but the code passes Time_Spec
new code calls epoll_pwait for all archs with NULL sigmask so it will act as epoll_wait

updated:
also fixes EPoll_Event struct always being packed, it should be packed only for x86/x86_64


xb-bx:
force-pushed to fix the typo in the commit message

PR #5654, opened 2025-09-05
Fix compiler segfault when trying to use proc at type level

### LINKED ISSUE #5620: Compiler Segfault when trying to use proc at type level
## Context

* Operating System & Odin Version: Windows 11, dev-2025-08-nightly:accdd7c
* `odin report` output:
        Odin:    dev-2025-08-nightly:accdd7c
        OS:      Windows 11 Professional (version: 23H2), build 22631.5768
        CPU:     12th Gen Intel(R) Core(TM) i7-12700H
        RAM:     32492 MiB
        Backend: LLVM 20.1.0

## Expected Behavior

The compiler should show an error message that explains why the code is not valid

## Current Behavior

The compiler segfaults

-

### Steps to Reproduce

Make a file `bug.odin` with the following contents:
```odin
package main

S :: struct(T: typeid) {}

f :: proc() -> S(f()) {}
```
Then run
```bash
odin build bug.odin -file
```

### Failure Logs

Segmentation fault

smoke-y:
Fixes #5620

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -7947,10 +7947,12 @@ gb_internal CallArgumentError check_polymorphic_record_type(CheckerContext *c, O
 				s = gb_string_append_fmt(s, "$%.*s", LIT(name));
 
 				if (v->kind == Entity_TypeName) {
-					if (v->type->kind != Type_Generic) {
-						s = gb_string_append_fmt(s, "=");
-						s = write_type_to_string(s, v->type, false);
-					}
+					if(v->type){
+						if (v->type->kind != Type_Generic) {
+							s = gb_string_append_fmt(s, "=");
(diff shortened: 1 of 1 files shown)


src/check_expr.cpp
```diff
@@ -7947,10 +7947,12 @@ gb_internal CallArgumentError check_polymorphic_record_type(CheckerContext *c, O
 				s = gb_string_append_fmt(s, "$%.*s", LIT(name));
 
 				if (v->kind == Entity_TypeName) {
-					if (v->type->kind != Type_Generic) {
```
gingerBill, line 7950: I would have just done: `if (v->type != nullptr && v->type->kind != Type_Generic) {`

PR #5660, opened 2025-09-08
[core:image/png] use .Image_Dimensions_Too_Large

leecommamichael:
PNG was sending .Invalid_Image_Dimensions for both the zero case and exceeding the MAX_DIMENSIONS.


Kelimion:
`.Image_Dimensions_Too_Large` must've been added to the enum at a later date, and splitting out this case was overlooked.

PR #5662, opened 2025-09-08
[vendor:stb_truetype] amend PackFontRanges types

leecommamichael:
These parts of the API need to process an array for a range of characters, so I changed them to multi-pointers.


vendor/stb/truetype/stb_truetype.odin
```diff
@@ -108,9 +108,9 @@ packedchar :: struct {
 pack_range :: struct {
 	font_size:                        f32,
 	first_unicode_codepoint_in_range: c.int,
-	array_of_unicode_codepoints:      [^]rune,
+	array_of_unicode_codepoints:      [^]rune, // optional
```
gingerBill: `Maybe([^]rune)` could be added? Not saying it should be though.

leecommamichael: My mistake, the comment is erroneous since either `first_unicode_codepoint_in_range` or `array_of_unicode_codepoints` is expected. The `first_codepoint...` is ignored if the array is present. I can't think of a better Odin binding that isn't a wrapper. I could do two `Maybe` parameters, but then it seems that neither are needed 😅

By the way, would you mind if I converted some of these `c.int` returns to `b32`? Sean returns only 1 or 0 in many of these functions. Thanks.

EDIT: Check my math, but; I _think_ in practice Odin only supports ILP32/LP64 things, so `c.int` can safely be typed as `i32` (and thus b32) in all cases, so there'd be no ABI fuss.

leecommamichael:
I've not worked with forks very much, not sure why it's including my (already merged) PNG change from earlier today... I guess I've got to rebase/squash more to convince it.

leecommamichael:
The failed test on BSD is the test_core_net DNS resolution.

Issue #5604, opened 2025-08-20
[target:js_wasm32] tick_now() stops growing after ~2 sec of app execution


greenya:
## Context

        Odin:    dev-2025-08-nightly:accdd7c
        OS:      Windows 11 Home Basic (version: 24H2), build 26100.4946
        CPU:     AMD Ryzen 7 5700X 8-Core Processor
        RAM:     32688 MiB
        Backend: LLVM 20.1.0

## Expected Behavior

`tick_now()` keep growing as long as app is running.

## Current Behavior

Using `tick_now()` when building for JS (`-target:js_wasm32`) will stop growing after ~2 sec, due to conversion from f32 to i64 (nanoseconds).

### Steps to Reproduce

Consider following code:
```odin
fmt.printfln("now=%v, tick_now=%v", time.now()._nsec, time.tick_now()._nsec)
```
The desktop build (windows):
<img width="463" height="301" alt="Image" src="https://github.com/user-attachments/assets/fc11d708-ff49-4656-81b5-2cce8f2cad1c" />

The JS build (Google Chrome console):
<img width="322" height="300" alt="Image" src="https://github.com/user-attachments/assets/01f540d5-b677-4e03-8c50-a3fa8a6fd452" />

### Investigation

After looking into https://github.com/odin-lang/Odin/blob/master/core/time/time_js.odin we see the code:
```odin
_tick_now :: proc "contextless" () -> Tick {
	foreign odin_env {
		tick_now :: proc "contextless" () -> f32 ---
	}
	return Tick{i64(tick_now()*1e6)}
}
```

Some ways to fix it, maybe:
- Changing foreign return type to `f64` seems to be working correct.
- Converting to i64 earlier, e.g. `return Tick{i64(tick_now())*1e6}` also works, but has low precision (4 non-zero digits only)
- Changing code to be equivalent to `_now()` is also working, and precision is high.

leecommamichael:
Can this be closed?

greenya:
I guess so.

PR #5663, opened 2025-09-08
[core:time] time_js: tick_now(): Use f64 (was f32) as a return type of odin_env.tick_now()

greenya:
Fix for issue #5604.

The change is tiny, but please review it.

I changed return type of used `odin_env.tick_now()`. Tested, and produced web build works (time ticks). If the fix is incorrect, please reject and fix the issue in correct way. Please do not leave the issue open, as I think it is quite severe one, because currently in a web build the core:time.tick_now() doesn't work (doesn't tick after 2 sec).

### DIFF
--- a/core/time/time_js.odin
+++ b/core/time/time_js.odin
@@ -24,7 +24,7 @@ _sleep :: proc "contextless" (d: Duration) {
 
 _tick_now :: proc "contextless" () -> Tick {
 	foreign odin_env {
-		tick_now :: proc "contextless" () -> f32 ---
+		tick_now :: proc "contextless" () -> f64 ---
 	}
 	return Tick{i64(tick_now()*1e6)}
 }


leecommamichael:
https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

Javascript "high precision" timers have up to 5 microsecond resolution when you're running an "isolated" app. The vast majority of sites will only have 100 microsecond resolution. I think the author's intent is to make delta-times more precise, but that won't necessarily be an outcome. Delta-times usually have a 1-digit integer-part, which leaves enough precision for any delta-time JS can emit even with a 32-bit float.

That being said, I don't think it's _wrong_ to make this change.

leecommamichael:
> Please do not leave the issue open, as I think it is quite severe one, because currently in a web build the core:time.tick_now() doesn't work (doesn't tick after 2 sec).

Can you paste some code to reproduce the issue? I've run a fair amount of code in the js_wasm targets and haven't encountered this. Also what does "doesn't tick" mean?

greenya:
> > Please do not leave the issue open, as I think it is quite severe one, because currently in a web build the core:time.tick_now() doesn't work (doesn't tick after 2 sec).
> 
> Can you paste some code to reproduce the issue? I've run a fair amount of code in the js_wasm targets and haven't encountered this. Also what does "doesn't tick" mean?

I have described the issue in #5604, please check, it has basically single line of code. But you need to build it for web to see the issue. The issue only with time.tick_now(), and "doesn't tick" means after about 2 sec of app execution, the return value of time.tick_now() is frozen (doesn't change). Which is a big issue for anything that uses it as a time measurement (game clock, frame time, animation time etc).

leecommamichael:
> I have described the issue in #5604, please check, it has basically single line of code. But you need to build it for web to see the issue. The issue only with time.tick_now(), and "doesn't tick" means after about 2 sec of app execution, the return value of time.tick_now() is frozen (doesn't change). Which is a big issue for anything that uses it as a time measurement (game clock, frame time, animation time etc).
 
Ah apologies. Multiple oversights on my part. Didn't see you linked that issue, and that I already have a similar local change 🤦

PR #5675, opened 2025-09-10
Fix incorrect json encoding for control characters < 32

### LINKED ISSUE #5609: `encoding/json` encodes invalid escape character in string for nil bytes
```odin
package main

import "core:encoding/json"
import "core:fmt"

main :: proc() {
	v := transmute(string)([]byte{0, 1, 2, 3})

	data, write_err := json.marshal(v)
	fmt.println(string(data), write_err) // "\x00\x01\x02\x03" nil

	read_err := json.unmarshal(data, &v)
	fmt.println(read_err) // Invalid_Data
}
```

this is not valid json:
```json
"\x00\x01\x02\x03"
```

it's odd to have unmarshal error on the same data that was just written with marshal.

The same happens with other bytes `< 32`.
On all available json specifications.

odin report
```
	Odin:    dev-2025-08:accdd7c2a
	OS:      Ubuntu 24.04.3 LTS, Linux 6.8.0-59-generic
	RAM:     15692 MiB
	Backend: LLVM 18.1.3
```

jfcode4:
Hi, I would like to contribute to Odin. This fixes #5609 where json.marshal gives invalid json when there are control characters in a string:
```json
"\x00\x01\x02\x03"
```
should instead give
```json
"\u0000\u0001\u0002\u0003"
```

### DIFF
--- a/core/io/util.odin
+++ b/core/io/util.odin
@@ -189,6 +189,23 @@ write_escaped_rune :: proc(w: Writer, r: rune, quote: byte, html_safe := false, 
 		write_encoded_rune(w, r, false, &n) or_return
 		return
 	}
+	if r < 32 && for_json {
+		switch r {
+		case '\b': write_string(w, `\b`, &n) or_return
+		case '\f': write_string(w, `\f`, &n) or_return
+		case '\n': write_string(w, `\n`, &n) or_return
+		case '\r': write_string(w, `\r`, &n) or_return
+		case '\t': write_string(w, `\t`, &n) or_return
+		case:
+			write_byte(w, '\\', &n) or_return
+			write_byte(w, 'u', &n)  or_return
+			write_byte(w, '0', &n)  or_return
+			write_byte(w, '0', &n)  or_return
+			write_byte(w, DIGITS_LOWER[r>>4 & 0xf], &n) or_return
+			write_byte(w, DIGITS_LOWER[r    & 0xf], &n) or_return
+		}
+		return
+	}
 	switch r {
 	case '\a': write_string(w, `\a`, &n) or_return
 	case '\b': write_string(w, `\b`, &n) or_return


core/io/util.odin
```diff
@@ -189,6 +189,23 @@ write_escaped_rune :: proc(w: Writer, r: rune, quote: byte, html_safe := false,
 		write_encoded_rune(w, r, false, &n) or_return
 		return
 	}
+	if r < 32 && for_json {
```
Kelimion, line 192: There's a case for `< ' '` (ascii 32) at original line 203.
```odin
	switch r {
	case '\a': write_string(w, `\a`, &n) or_return
	case '\b': write_string(w, `\b`, &n) or_return
	case '\e': write_string(w, `\e`, &n) or_return
	case '\f': write_string(w, `\f`, &n) or_return
	case '\n': write_string(w, `\n`, &n) or_return
	case '\r': write_string(w, `\r`, &n) or_return
	case '\t': write_string(w, `\t`, &n) or_return
	case '\v': write_string(w, `\v`, &n) or_return
	case:
		switch c := r; {
		case c < ' ': // <--- here
			write_byte(w, '\\', &n)                      or_return
			write_byte(w, 'x', &n)                       or_return
			write_byte(w, DIGITS_LOWER[byte(c)>>4], &n)  or_return
			write_byte(w, DIGITS_LOWER[byte(c)&0xf], &n) or_return
```
Could the `for_json` condition not be tested here, writing either `\x<c>` or `\u00<c>` as appropriate?
Or is it that JSON requires `\a`, `\e` and `\v` to also be escaped instead of written as literals in the cases above this default branch?

jfcode4: Yes but `\a`, `\e` and `\v` are indeed incorrect JSON according to the spec.

PR #5689, opened 2025-09-16
Improve type inferencing of literals when calling proc groups

slowhei:
Hi there,

This PR is not a bugfix, but a slight improvement to the Odin compiler's type inferencing that only minimally changes the type checking code.

I will begin with a really simple example demonstrating the current limitations of Odin's type inferencing:
```
package example

Enum :: enum{A, B, C}

group :: proc{proc_nil, proc_enum}
proc_nil  :: proc() { }
proc_enum :: proc(Enum) { }

main :: proc() {
    group(.A) // <-- does not compile. error message is "Cannot determine type for implicit selector expression '.A' "
}
```
It turns out that Odin's type inferencing for implicitly typed values never succeeds if there exists a proc in the proc group with fewer parameters than the one you are trying to call. The reason why is apparent in the compiler code, but more on that later.

But why is this a problem? If I just do `group(Enum.A)`, everything will be fine. Odin is not supposed to be a language that optimizes for characters typed, but for clarity and simplicity. Well, here is one more motivating example that reflects my personal experience:
```
package example

GPU_Buffer :: struct{
    // ...
}
GPU_Buffer_Usage :: bit_set[enum{Uniform_Buffer, Storage_Buffer}]

create_gpu_buffer :: proc{
    create_gpu_buffer_from_size,
    create_gpu_buffer_from_type,
}
create_gpu_buffer_from_size :: proc(size: int, usage: GPU_Buffer_Usage) -> (GPU_Buffer, []byte) {
    // ...
    return GPU_Buffer{}, nil
}
create_gpu_buffer_from_type :: proc($T: typeid, usage: GPU_Buffer_Usage) -> (GPU_Buffer, ^T) {
    // ...
    return GPU_Buffer{}, nil
}

main :: proc() {
    Uniform_Buffer_Data :: struct{
        // ...
    }
    ub, ub_data := create_gpu_buffer(Uniform_Buffer_Data, {.Uniform_Buffer})
    sb, sb_data := create_gpu_buffer(1_024_000, {.Storage_Buffer})
}
```
So this code actually compiles fine—Odin does in fact allow type inference in proc groups. But let's suppose you've been using `create_gpu_buffer` throughout your codebase quite a lot and have 100 usages of it. And now suppose you want to add `create_gpu_buffer_from_buffer :: proc(GPU_Buffer) -> (GPU_Buffer, []byte)` to the procedure group. All of a sudden, all of your existing usages of `create_gpu_buffer` will fail to compile. You now have 100 scattered lines of code to fix and wonder whether depending on type inference in proc groups is an anti-pattern in Odin. The negative effects would cascade worse if you were writing a library and broke a ton of user's code in a new update.

Additionally, if you tried to add `create_gpu_buffer_from_type_and_count :: proc($T: typeid, count: int, usage: GPU_Buffer_Usage) -> (GPU_Buffer, []T)` to the proc group, you might get annoyed that `create_gpu_buffer(Uniform_Buffer_Data, 4, {.Uniform_Buffer})` fails to compile.

---

My proposed code change is to filter out procs in the proc group that do not have enough parameters to fit the number of args passed to the proc group. In my first example, my code change filters out `proc_nil` when `group(.A)` is called. By filtering out `proc_nil`, the code that builds type hints (https://github.com/slowhei/Odin/blob/5e71ba44562630c50fd04fb50513c5c7758e16f5/src/check_expr.cpp#L7018-L7087) computes a `proc_arg_count` of 1 instead of 0, which allows a type hint of `GPU_Buffer_Usage` for `.A` to be added to `lhs`.

Since this code change only aims to make type inference more flexible than before, existing Odin codebases should continue to compile successfully. I have also added a test file showing all of the new type inference on literals that would become possible.

What do the maintainers think? I submit completely to your authority.

### DIFF
--- a/tests/internal/test_proc_group_type_inference.odin
+++ b/tests/internal/test_proc_group_type_inference.odin
@@ -65,13 +65,20 @@ test_type_inference_on_literals_for_various_types :: proc(t: ^testing.T) {
 	testing.expect_value(t, len(group_slice_u8({})),              0)
 	testing.expect_value(t, group_slice_u8(nil) == nil,           true)
 
-	proc_dynamic_array :: proc(a: [dynamic]u8) -> [dynamic]u8 { return a }
+	proc_dynamic_array :: proc(t: ^testing.T, array: [dynamic]u8, expected_len: int) {
+		if expected_len < 0 {
+			testing.expect_value(t, array == nil, true)
+		} else {
+			testing.expect_value(t, len(array), expected_len)
+		}
+		delete(array)
+	}
 	group_dynamic_array :: proc{proc_nil, proc_dynamic_array}
-	testing.expect_value(t, len(group_dynamic_array([dynamic]u8{1, 2, 3})), 3)
-	testing.expect_value(t, len(group_dynamic_array({1, 2, 3})),            3)
-	testing.expect_value(t, len(group_dynamic_array({0=1, 1=2, 2=3})),      3)
-	testing.expect_value(t, len(group_dynamic_array({})),                   0)
-	testing.expect_value(t, group_dynamic_array(nil) == nil,                true)
+	group_dynamic_array(t, [dynamic]u8{1, 2, 3}, 3)
+	group_dynamic_array(t, {1, 2, 3},            3)
+	group_dynamic_array(t, {0=1, 1=2, 2=3},      3)
+	group_dynamic_array(t, {},                   0)
+	group_dynamic_array(t, nil,                  -1)
 
 	Enum :: enum{A, B, C}
 	proc_enum :: proc(a: Enum) -> Enum { return a }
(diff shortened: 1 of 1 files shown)


slowhei:
There are a few more interesting behaviors I found with the compiler not relevant to the PR that I will put here:

```
// My code changes to do not make type inferencing possible here. This will still fail

Bit_Set :: bit_set[enum{A, B, C}]
group :: proc{proc_1, proc_2}
proc_1 :: proc(int, Bit_Set) { }
proc_2 :: proc(Bit_Set, int) { }

group(9, {.A})
group({.A}, 9)
```

```
// Type casting of unions seems inconsistent

proc_nil :: proc() { }
proc_union :: proc(a: Union) -> Union { return a }
group_union :: proc{proc_nil, proc_union}

group_union(9) // compiles
Union(9)       // doesn't compile
```

```
// This is pretty wild
package example

import "core:fmt"

main :: proc() {
    Bit_Set :: bit_set[enum{A, B, C}]
    group :: proc{proc_1, proc_2}
    proc_1 :: proc(bit_set_: Bit_Set, args: ..any)        -> int { return 1 }
    proc_2 :: proc(args: ..any, bit_set_: Bit_Set = {.A}) -> int { return 2 }

    fmt.printfln("%v", group(Bit_Set{}, Bit_Set{})) // 2 gets printed
    fmt.printfln("%v", group(Bit_Set{}))            // 2 gets printed
}
```

JesseRMeyer:
Thanks -- this looks good but could you address the failing tests?

slowhei:
Oh! I am so sorry! I thought my tests were passing, but clearly they were not. That's embarrassing. They have now been fixed.

Now, the CI/CD pipeline for my PR is still failing, but I notice that it is also failing right now on the official Odin repo's master branch, and for the same reasons. Perhaps we should wait for the compiler to become stable again before merging this PR. Right now, I can't effectively test `tests/core` and `tests/vendor`.

PR #5691, opened 2025-09-16
Fix out-of-band allocations in dynamic arenas

### LINKED ISSUE #5693: Out-of-band allocations in Dynamic_Arenas don't work as intended
## Context

This is an issue that corresponds to my pull request. I'm making an issue for this b/c it seems like pull requests aren't looked at as often, and I want to make sure this gets seen.

Basically, allocating arrays out of a `Dynamic_Arena` will start failing once the array size is larger than the block size. This shouldn't happen unless the out-of-band size is larger than the block size.

Odin report:
```
        Odin:    dev-2025-09:3bbf3641a
        OS:      Windows 10 Professional (version: 22H2), build 19045.6332
        CPU:     AMD Ryzen 7 3700X 8-Core Processor
        RAM:     16297 MiB
        Backend: LLVM 20.1.0
```

## Expected Behavior

Out-of-band allocations work correctly and this test passes:

```odin
// tests/core/mem/test_mem_dynamic_pool.odin


// Modified on my branch to accept a block size
expect_pool_allocation_out_of_band :: proc(t: ^testing.T, num_bytes, block_size, out_band_size: int) {
	testing.expect(t, num_bytes >= out_band_size, "Sanity check failed, your test call is flawed! Make sure that num_bytes >= out_band_size!")

	pool: mem.Dynamic_Pool
	mem.dynamic_pool_init(&pool, block_size = block_size, out_band_size = out_band_size)
	pool_allocator := mem.dynamic_pool_allocator(&pool)

	element, err := mem.alloc(num_bytes, allocator = pool_allocator)
	testing.expect(t, err == .None)
	testing.expect(t, element != nil)
	testing.expectf(t, pool.out_band_allocations != nil,
		"Allocated data with size %v bytes, which is >= out_of_band_size and it should be in pool.out_band_allocations, but isn't!",
	)

	mem.dynamic_pool_destroy(&pool)
	testing.expect(t, pool.out_band_allocations == nil)
}

@(test)
test_dynamic_pool_alloc_out_of_band :: proc(t: ^testing.T) {
	expect_pool_allocation_out_of_band(t, num_bytes = 128, block_size = 512, out_band_size = 128)
	expect_pool_allocation_out_of_band(t, num_bytes = 129, block_size = 512, out_band_size = 128)
	// This is the new test
	expect_pool_allocation_out_of_band(t, num_bytes = 513, block_size = 512, out_band_size = 128)
}

```

## Current Behavior

This test fails

### Steps to Reproduce

1. Modify `expect_pool_allocation_out_of_band` to match my code
2. Add the new test line `expect_pool_allocation_out_of_band(t, num_bytes = 513, block_size = 512, out_band_size = 128)` to `test_dynamic_pool_alloc_out_of_band`
3. Run `odin test tests/core/mem/test_mem_dynamic_pool.odin -file` and see that the test fails

rationalcoder:
I noticed while using a `Dynamic_Arena` that dynamic arrays would start failing to grow for some reason. Turns out, the out-of-band allocation logic was wrong for two reasons:

1. It wasn't allocating out-of-band allocations out of the correct allocator
2. Out-of-band allocations would always either allocate a new block or fail if the allocation size was larger than the block size

Luckily, the fix to the code and the changes to the test code seem pretty trivial.

Fixes #5693

### DIFF
--- a/core/mem/allocators.odin
+++ b/core/mem/allocators.odin
@@ -1819,17 +1819,17 @@ memory region.
 */
 @(require_results)
 dynamic_arena_alloc_bytes_non_zeroed :: proc(a: ^Dynamic_Arena, size: int, loc := #caller_location) -> ([]byte, Allocator_Error) {
-	n := align_formula(size, a.alignment)
-	if n > a.block_size {
-		return nil, .Invalid_Argument
-	}
-	if n >= a.out_band_size {
-		assert(a.block_allocator.procedure != nil, "Backing block allocator must be initialized", loc=loc)
-		memory, err := alloc_bytes_non_zeroed(a.block_size, a.alignment, a.block_allocator, loc)
+	if size >= a.out_band_size {
+		assert(a.out_band_allocations.allocator.procedure != nil, "Backing array allocator must be initialized", loc=loc)
+		memory, err := alloc_bytes_non_zeroed(size, a.alignment, a.out_band_allocations.allocator, loc)
 		if memory != nil {
 			append(&a.out_band_allocations, raw_data(memory), loc = loc)
 		}
 		return memory, err
+	}
+	n := align_formula(size, a.alignment)
+	if n > a.block_size {
+		return nil, .Invalid_Argument
 	}
 	if a.bytes_left < n {
 		err := _dynamic_arena_cycle_new_block(a, loc)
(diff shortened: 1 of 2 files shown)


rationalcoder:
Currently removing debug code that I accidentally committed...

PR #5705, opened 2025-09-23
Implement more Linux syscalls

bc-universe:
While looking for some io_uring specific syscalls around Odin, I ended up extending them with those and some more.

I was able to test most of the syscalls with success, even the io_uring ones, but there still might be something lurking around.
For testing I used a version of Odin with these changes and observed the results with strace.

This was my first go at Odin, please let me know if something feels off in terms of syntax.

### DIFF
--- a/core/sys/linux/sys.odin
+++ b/core/sys/linux/sys.odin
@@ -2136,7 +2136,8 @@ pivot_root :: proc "contextless" (new_root: cstring, old_root: cstring) -> (Errn
 	Operations on a process or thread
 	Available since Linux 2.1.57
 */
-prctl :: proc "contextless" (op: i32, args: ..u64) -> (Errno) {
+prctl :: proc "contextless" (op: i32, args: ..uint) -> (Errno) {
+	assert_contextless(len(args) <= 4)
 	ret := syscall(SYS_prctl, op, args[0], args[1], args[2], args[3])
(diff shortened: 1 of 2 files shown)


laytan:
I have Io uring bindings in a pr that I'm working on, may be best to keep those out of this pr for now, also because I see flags, opcode, etc. being untyped here + I've written against those bindings of mine.

bc-universe:
Hey @laytan that sounds very nice!

I tried to keep everything simple and since I intended only to write the bare syscalls I did not go further with types and everything else.

If you feel like these do not add value I could remove the io_uring related code.
If the rest is useful it could be kept, if not I wouldn't mind closing this PR.

laytan:
Hey sorry this was left so long, if you want to resolve the conflicts I can merge this. My IO Uring additions have been merged since then.

bc-universe:
I will take a look this weekend, thank you for the heads up.
I need to go through your changes and strip out what is already there and I will then merge my changes.

laytan:
I was going through the changes and had to fix a lot of it up. I pushed them here directly instead of commenting on everything. Do the changes look ok to you / is this ready to go?

bc-universe:
Had a look through your changes and those do look good, thank you for taking the time and correcting them.

PR #5717, opened 2025-09-26
[core:sys/darwin]: Add ptrace syscall wrapper to xnu_system_call_wrappers

eligrubb:
Adds ptrace to the set of odin wrappers for xnu syscalls.


Platin21:
Issue with this entire thing is that technically you can't pass Apples validations using any of these calls. So to a part thats also the reason why it wasn't build out anymore.

eligrubb:
Gotcha, TIL. I'll close this out, it's been pretty easy to maintain locally for the one project I use it on anyway. Testament to Odin's setup. Thanks!

PR #5727, opened 2025-09-29
Add intrinsics.objc_super and Automatically emit objc_msgSend calls

harold-b:
Implements the `objc_super()` intrinsic which, when used by itself simply returns the superclass-typed version of an Objective-C object pointer. However when combined with a call to an Objective-C method, it `emits objc_msgSendSuper2` allowing us to easily perform typical initialization routines in this manner:

```odin
if objc_super(self)->init() != nil {
    // ...
}
```

For this to work, automatic `objc_msgSend` codegen was required when calling Objective-C methods. Therefore the following editions were made:

- 1. Allow import of bodyless Objective-C methods within foreign blocks.

- 2. Unconditionally call Objective-C methods implemented in odin (through `@obj_implement`) via the Objective-C runtime. No direct dispatch.

Edit #2 also makes it more explicit that the `context` in Odin calling convention Objective-C method implementations **always** comes from the specified context provider, as it is called through the Objective-C runtime. -->

Importing Objective-C is done in this way:

```odin
@(require, export) foreign import foundation "system:Foundation.framework"

@(objc_class="NSArray", objc_superclass=Object)
Array :: struct { using _: Object, 
    using _: Copying,
    using _: MutableCopying,
    using _: SecureCoding,
    using _: FastEnumeration,
}

foreign lib {
    @(objc_type=Array, objc_selector="objectAtIndex:", objc_name="objectAtIndex")
    Array_objectAtIndex :: proc(self: ^Array, index: UInteger) -> id ---

    @(objc_type=Array, objc_selector="init", objc_name="init")
    Array_init :: proc(self: ^Array) -> instancetype ---

    @(objc_type=Array, objc_selector="initWithObjects:count:", objc_name="initWithObjects_count")
    Array_initWithObjects_count :: proc(self: ^Array, objects: ^id, cnt: UInteger) -> instancetype ---

    @(objc_type=Array, objc_selector="initWithCoder:", objc_name="initWithCoder")
    Array_initWithCoder :: proc(self: ^Array, coder: ^Coder) -> instancetype ---

    @(objc_type=Array, objc_selector="count", objc_name="count")
    Array_count :: proc(self: ^Array) -> UInteger ---

    // ...
}
```

Where the `@objc_superclass` attribute is now allowed on non-implementation classes in order for `objc_super` to statically determine the resulting Type of the intrinsic.

The `@objc_selector` attribute is required on foreign-imported Objectice-C methods.

In addition, a dummy type, `objc_instancetype`, which is simply a distinct alias to `objc_id`, to mimic the Objective-C `instancetype*` returns to mimic its behavior in Objective-C where the return type is typed to the left hand value of the selector, or selector call expression.

This facilitates calls like `alloc` and `init` to return the expected type instead of the subtype, without haivng to create dummy sub procs.

This PR includes some ABI fixes to objc_implement and objc_block method wrappers as well.


harold-b:
Working on fixing errors identified by CI

harold-b:
Foreign import frameworks available [here https://github.com/harold-b/darwodin/tree/foreign-imports](https://github.com/harold-b/darwodin/tree/foreign-imports) until relevant portions are brought in the official repo

PR #5729, opened 2025-09-30
Relax `signature_parameter_similar_enough` on struct params and fix various foreign signatures

harold-b:
When comparing 2 structs in`signature_parameter_similar_enough()`, it simply checks whether they are the same size and alignment

### DIFF
--- a/src/check_decl.cpp
+++ b/src/check_decl.cpp
@@ -851,6 +851,12 @@ gb_internal bool signature_parameter_similar_enough(Type *x, Type *y) {
 		}
 	}
 
+	Type *x_base = base_type(x);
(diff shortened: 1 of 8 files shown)


src/check_decl.cpp
```diff
@@ -851,6 +851,12 @@ gb_internal bool signature_parameter_similar_enough(Type *x, Type *y) {
 		}
 	}
 
+	Type *x_base = base_type(x);
+	Type *y_base = base_type(y);
+	if (x_base->kind == y_base->kind && x_base->kind == Type_Struct) {
```
gingerBill, line 856: I'll accept this but then make it more "correct" if that makes sense.

Issue #5737, opened 2025-10-01
LLVM Codegen Failure with SIMD transmute.


Chamberlain91:
## Context

Have been experimenting and learning SIMD stuff via trying to implement a software renderer. Currently trying to figure out how to sample the image from 4 offsets, and encountered an LLVM codegen error.

```
Odin:    dev-2025-09:ea7e9bc11
OS:      Windows 11 Home Basic (version: 24H2), build 26100.6584
CPU:     AMD Ryzen 9 9900X 12-Core Processor
RAM:     63109 MiB
Backend: LLVM 20.1.0
```

## Expected Behavior

I would expect the program to compile OR a standard diagnostic informing me of an invalid operation.

## Current Behavior

LLVM Code generation failure.

### Steps to Reproduce

```odin
package simd_bug

main :: proc() {
    data_addrs: #simd[4]u64
    ptrs := transmute(#simd[4]rawptr)data_addrs
}
```

### Failure Logs

```
LLVM CODE GEN FAILED FOR PROCEDURE: simd_bug::main
define internal void @"simd_bug::main"(ptr noalias nocapture nonnull %__.context_ptr) {
decls:
  %data_addrs = alloca <4 x i64>, align 32
  %ptrs = alloca <4 x ptr>, align 32
  br label %entry

entry:                                            ; preds = %decls
  store <4 x i64> zeroinitializer, ptr %data_addrs, align 32
  %0 = load <4 x i64>, ptr %data_addrs, align 32
  %1 = bitcast <4 x i64> %0 to <4 x ptr>
  store <4 x ptr> %1, ptr %ptrs, align 32
  ret void
}




Invalid bitcast
  %1 = bitcast <4 x i64> %0 to <4 x ptr>
```

alsakandari:
This should be fixed on master after #5738

Chamberlain91:
Seems like it works now, at least in the minimal example I posted here.

PR #5738, opened 2025-10-01
[LLVM backend] SIMD vector transmute only uses bitcast when the elements aren't pointers

jakubtomsu:
This should fix issue #5737.

This crash happened because LLVM Bitcast isn't valid for pointer values. This change handles SIMD vectors of pointers (e.g. `#simd[N]rawptr` for doing a gather) the same way as arrays, where the transmute happens through a pointer.


jakubtomsu:
not sure why the FreeBSD build is failing due to timeout, is that something I should be concerned with?

PR #5742, opened 2025-10-02
feat: added rtoi & utoi to core/strconv/strconv.odin

samwega:
rtoi & utoi are two procs for converting a rune and a u8 character respectively to int

### DIFF
--- a/core/strconv/strconv.odin
+++ b/core/strconv/strconv.odin
@@ -1664,18 +1664,18 @@ Casts a rune to int
 
 Example:
 
-    import "core:strconv"
-    rune_to_int :: proc () {
-        my_int, ok := strconv.rtoi('5')
-        if ok {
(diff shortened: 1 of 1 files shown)


Tetralux:
I have some concerns with this, as reflected in a Discord conversation, which here's the highlights of for posterity:
1. I wonder if we should even have `atoi` and co as they simply ignore the error from `parse_int` etc which more than one person has been bitten by.
2. Names like `digit_to_int` are better in essentially all ways.
  - More self documenting
  - Less needs for a doc comment
  - Easier to read the code that uses it
3. Not sure we need both `rtoi` and `utoi` since `'0'..='9'` is always ASCII and thus `u8`, which is one cast away.
4. I would not use `#optional_ok`. You DO in fact care about the error.

PR #5744, opened 2025-10-02
Fix srgb <-> linear rgb conversion functions in `core:math/linalg`

Znarf64:
Uses the formulas from [wikipedia](https://en.wikipedia.org/wiki/SRGB#Transfer_function_(%22gamma%22)) and also adds implementations for vector3

### DIFF
--- a/core/math/linalg/specific.odin
+++ b/core/math/linalg/specific.odin
@@ -164,25 +164,25 @@ orthogonal :: proc{vector2_orthogonal, vector3_orthogonal}
 
 @(require_results)
 vector4_srgb_to_linear_f16 :: proc "contextless" (col: Vector4f16) -> Vector4f16 {
-	r := math.pow(col.x, 2.2)
-	g := math.pow(col.y, 2.2)
-	b := math.pow(col.z, 2.2)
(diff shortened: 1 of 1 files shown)


core/math/linalg/specific.odin
```diff
@@ -192,70 +192,55 @@ vector4_srgb_to_linear :: proc{
...
+	vector3_srgb_to_linear_f64,
+}
 
-	x = (x * (a * x + b)) / (x * (c * x + d) + e)
-	y = (y * (a * y + b)) / (y * (c * y + d) + e)
-	z = (z * (a * z + b)) / (z * (c * z + d) + e)
 
-	x = math.pow(clamp(x, 0, 1), 1.0 / 2.2)
-	y = math.pow(clamp(y, 0, 1), 1.0 / 2.2)
-	z = math.pow(clamp(z, 0, 1), 1.0 / 2.2)
```
gingerBill, line 214: This was correct logic, no?

Znarf64: Not to my understanding of the linked wikipedia article, no. What the old code was doing was an ACES tone map + gamma correction, which is not what I would expect a function called `srgb_to_linear` to do

PR #5749, opened 2025-10-03
Deprecate c procs

samwega:
and a short C name alias. From discussion on discord.

### DIFF
--- a/core/strconv/strconv.odin
+++ b/core/strconv/strconv.odin
@@ -1547,35 +1547,6 @@ write_u128 :: proc(buf: []byte, u: u128, base: int) -> string {
 }
 
 /*
-`itoa` C name deprecated, use `int_to_string` instead (same procedure)
-
-Converts an integer value to a string and stores it in the given buffer
-
-**Inputs**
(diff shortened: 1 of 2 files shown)


core/strconv/strconv.odin
```diff
@@ -1547,85 +1547,8 @@ write_u128 :: proc(buf: []byte, u: u128, base: int) -> string {
 }
 
 /*
-Converts an integer value to a string and stores it in the given buffer
+`ftoa` C name deprecated, use `write_float` instead (same procedure)
```
alsakandari, line 1550: I don't think this should be in the documentation of write_float

samwega: ftoa was only an alias to write_float which was deprecated, I thought it might be useful to point out that the procedure did not change besides the name, so people don't have to rewrite their code, but if it's not useful then it can easily be removed 👍

alsakandari: The pull request got merged, so kelimion doesn't see any problem with that, I am fine too

PR #5753, opened 2025-10-04
[ci.yml] matrix strategy for llvm versions on macos and ubuntu

A1029384756:
expands testing to all supported llvm versions on linux and macos. this could be expanded to the other, lesser-used targets (the BSDs) in the future but was left off for now

notes:
- ubuntu was downgraded to 22.04 vs 24.04 as 24.04 doesn't package llvm 14 in its repositories. 22.04 is still in active support until april 2027
- this is expected to fail with llvm 14 on ubuntu
- github actions are fully free with no limits on public repositories
- thse jobs run in parallel so ci time is not increased

### DIFF
--- a/core/unicode/fold.odin
+++ b/core/unicode/fold.odin
@@ -14,7 +14,7 @@ package unicode
 // 	simple_fold('\u212a') == 'k'
 // 	simple_fold(-3)       == -3
 @(require_results)
-simple_fold :: proc(r: rune) -> rune {
+simple_fold :: proc(r: rune) -> rune #no_bounds_check {
 	Fold_Pair :: struct {
 		from: u16,
 		to:   u16,
--- a/core/unicode/letter.odin
+++ b/core/unicode/letter.odin
@@ -1,5 +1,7 @@
 package unicode
 
+import "base:runtime"
+
 MAX_RUNE         :: '\U00010fff' // Maximum valid unicode code point
 REPLACEMENT_CHAR :: '\ufffd'     // Represented an invalid code point
 MAX_ASCII        :: '\u007f'     // Maximum ASCII value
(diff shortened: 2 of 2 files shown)


gingerBill:
This probably shouldn't be ran all the time. It's going to be too costly for the CI and more likely for random failures.

A1029384756:
out of curiosity, why is this too costly for the CI? are there CI limits i am not aware of? when i looked at githubs policy for github action quotas, public repositories were listed as having unlimited usage when using standard runners (which we do):
> GitHub Actions usage is free for self-hosted runners and for public repositories that use standard GitHub-hosted runners. See [Choosing the runner for a job](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories).

as mentioned earlier as well, these jobs also *do* run in parallel so there's not an end-to-end CI time increase either

as for random failure detection, what would you recommend then? a nightly/weekly job or on merges to `master`? as it stands, there have been multiple releases (including the most recent one) where "officially supported" llvm versions have been broken for one reason or another

gingerBill:
Called on nightly-releases is probably the better option rather than each PR.

laytan:
That would mean broken code can be merged and we'd only see it later in the nightly logs. I don't love that personally. You would have to find a way to reach the person that broke it to have them fix it, or fix it ourselves when it happens.

I do agree we will see more random failures, because of more runs.

gingerBill:
@laytan I understand that concern which is why it's quite hard to know. I just wish GitHub allowed for a lot more control in this regard. Maybe it does allow you to have an optional CI run when pressed but I am not sure.

A1029384756:
the closest thing that could be implemented would probably use this <https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request>

there could be a custom label (i.e. `llvm_backend`) that could be applied by maintainers when the llvm backend is changed. this would then kick off a build for the CI and get decently close to the desired end result. i'm willing to make those changes and test them if that's a valid potential path forwards. this could also just be more generalized to a `ready_to_merge` label that could be added to run stricter checks

it *is* possible to run workflows manually: <https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow> but i would be lying if i said i thought that was a good process for reviewers to have to go through

(also when testing my llvm 14 fixes last night, i did find that there will be some changes to this action anyways to get asan working with llvm 14)

PR #5759, opened 2025-10-05
Fix #5225 wasm odin.js undefined this.mem

### LINKED ISSUE #5225: Odinjs 'get_element_value_string' function throwing undefined error on `this.mem`
## Context

Not sure if I'm using this wrong, but `this` doesn't seem to be bound to anything in the function `get_element_value_string` in core/sys/wasm/js/odin.js [link](https://github.com/odin-lang/Odin/blob/master/core/sys/wasm/js/odin.js#L1871C7-L1871C78)

I'm using apple silicon and chrome. Also on latest master.

I managed to fix this quite easily by changing the call to `wasmMemoryInterface.loadBytes(...)`. I naively tried to make a PR for it which ofc didn't work 😆 but this is what it looks like:

```odin
get_element_value_string: (id_ptr, id_len, buf_ptr, buf_len) => {
	let id = wasmMemoryInterface.loadString(id_ptr, id_len);
	let element = getElement(id);
	if (element) {
		let str = element.value;
		if (buf_len > 0 && buf_ptr) {
			let n = Math.min(buf_len, str.length);
			str = str.substring(0, n);

                         // this line here
			wasmMemoryInterface.loadBytes(buf_ptr, buf_len).set(new TextEncoder().encode(str))
			return n;
		}
	}
	return 0;
}
```

## Expected Behavior

Function should not throw console error

## Current Behavior

Console error is thrown

Image

### Steps to Reproduce

1. Export an odin function that uses `get_element_value_string`
2. Build and load into js
3. Call function from js

Creativty:
Fix for #5225, Change `get_element_value_string` and `get_element_key_string` from accessing the memory interface via `this.mem` to `wasmMemoryInterface` like all other methods.

### DIFF
--- a/core/sys/wasm/js/odin.js
+++ b/core/sys/wasm/js/odin.js
@@ -1937,7 +1937,7 @@ function odinSetupDefaultImports(wasmMemoryInterface, consoleElement, memory) {
 					if (buf_len > 0 && buf_ptr) {
 						let n = Math.min(buf_len, str.length);
 						str = str.substring(0, n);
-						this.mem.loadBytes(buf_ptr, buf_len).set(new TextEncoder().encode(str))
+						wasmMemoryInterface.loadBytes(buf_ptr, buf_len).set(new TextEncoder().encode(str))
 						return n;
 					}
 				}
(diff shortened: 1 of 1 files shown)


alsakandari:
That worked because you linked it in the commit, which closed the issue after it was merged

PR #5773, opened 2025-10-08
Fix fontstash crash with .TTC files

karl-zylinski:
Fix for fontstash crash because it didn't fetch the offset of the first font correctly. The old setup didn't work with TTC files that contain multiple fonts.

### DIFF
--- a/vendor/fontstash/fontstash.odin
+++ b/vendor/fontstash/fontstash.odin
@@ -323,11 +323,15 @@ __AtlasAddWhiteRect :: proc(ctx: ^FontContext, w, h: int) -> bool {
 
 // push a font to the font stack
 // optionally init with ascii characters at a wanted size
+//
+// 'fontIndex' controls which font you want to load within a multi-font format such
+// as TTC. Leave it as zero if you are loading a single-font format such as TTF.
 AddFontMem :: proc(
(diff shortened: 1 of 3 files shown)


vendor/fontstash/fontstash.odin
```diff
@@ -335,7 +335,10 @@ AddFontMem :: proc(
 	res.freeLoadedData = freeLoadedData
 	res.name           = strings.clone(name)
 
-	stbtt.InitFont(&res.info, &res.loadedData[0], 0)
+	// Get offset of first font (if the font is a TTC then it can contain multiple fonts)
+	// Note: There is currently no support for specifying any other font than first one.
+	font_offset := stbtt.GetFontOffsetForIndex(raw_data(res.loadedData), 0)
```
Kelimion: Instead of hardcoding, I'd  give `AddFontMem` and `AddFontFile` a default `ttc_index := 0` param, with a little comment that that param determines which ttf file in a `.ttc` to load, defaulting to the first (or only for `.ttf`).

karl-zylinski: Thanks. Reasonable. I'll fix it tomorrow

karl-zylinski: Fixed!

PR #5794, opened 2025-10-11
Improve target features flag by allowing the user to disable a target feature

alsakandari:
This is already supported by LLVM, and by llc, but not by the Odin compiler for some reason


alsakandari:
Using this flag I was able to build and run https://github.com/flysand7/odin-os without the enable_sse function, by adding `-target-features:-sse,-sse2,-avx`, but required copying base/runtime/procs.odin into there

PR #5803, opened 2025-10-13
Fix hang, courtesy of cloin.

Kelimion:


Kelimion:
Found by @colrdavidson, with thanks.

PR #5810, opened 2025-10-15
Fix: #load(<string-path>, <type>) produces unaligned accesses

### LINKED ISSUE #5771: #load(<string-path>, <type>) produces unaligned reads and writes
## Context

```
        Odin:    dev-2025-10:f564f66c0
        OS:      Windows 10 Professional (version: 22H2), build 19045.6332
        CPU:     Intel(R) Core(TM) i5-9600KF CPU @ 3.70GHz
        RAM:     16319 MiB
        Backend: LLVM 20.1.0
```

## Expected vs Current Behavior

Given the following program, I'd expect a load directive with type "[]u32" to be 4 byte aligned, thus I'd expect the assert to never fail, but it does fail (on my particular configuration, so in general it can fail). This was reproduced on my end by simply running on debug mode: `odin run test.odin -file`

```odin
package main

main :: proc()
{
    s := "ab"
    data := #load("test1", []u32)
    assert(uintptr(raw_data(data)) % 4 == 0)
}
```

`C:/work/odin_bug_report/test.odin(9:5) runtime assertion: uintptr(raw_data(data)) % 4 == 0`

LeonardoTemperanza:
Fixes #5771.


Kelimion:
Confirmed that it fixes the bug, thank you. Merged.

PR #5814, opened 2025-10-15
vendor/wgpu: update to 27.0.2.0

laytan:


src/llvm_backend_proc.cpp
```diff
@@ -284,6 +282,7 @@ gb_internal lbProcedure *lb_create_procedure(lbModule *m, Entity *entity, bool i
 		LLVMSetLinkage(p->value, LLVMExternalLinkage);
 	}
 
+	lb_set_linkage_from_entity_flags(p->module, p->value, entity->flags);
```
laytan, line 285: Moved down because linkage wasn't being applied properly on foreign procs, because the `if (ignore_body)` above would always overwrite it to `LLVMExternalLinkage`.

PR #5816, opened 2025-10-16
[core:sys/windows] Sync Barriers, macros for buffered rawinput

jakubtomsu:
I noticed Ryan is using native win32 barriers in raddebugger, so here are the bindings. It might be worth it replacing `core:sync`'s `Barrier` with it? The native barrier generally spins in userspace for a while and then the kernel can efficiently wake it up once all threads are finished.

https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-entersynchronizationbarrier

And also utility macros for buffered rawinput.

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getrawinputbuffer


Creativty:
It would be a good idea to have a native implementation of barriers in `core:sync`.

core/sys/windows/user32.odin
```diff
@@ -398,6 +398,14 @@ MAKEINTRESOURCEW :: #force_inline proc "contextless" (#any_int i: int) -> LPWSTR
 	return cast(LPWSTR)uintptr(WORD(i))
 }
 
+RAWINPUT_ALIGN :: proc(x: uintptr) -> uintptr {
+	return (x + size_of(uintptr) - 1) & ~uintptr(size_of(uintptr) - 1)
+}
+
+NEXTRAWINPUTBLOCK :: proc(ptr: ^RAWINPUT) -> ^RAWINPUT {
+	return cast(^RAWINPUT)RAWINPUT_ALIGN(uintptr(ptr) + uintptr(ptr.header.dwSize))
```
gingerBill: I'd probably mark these as `proc "c" (...)`

PR #5817, opened 2025-10-16
On wrong append call, help users pass-by-pointer

duchainer:
## WHY?

First Odin users that try to use append often forget to pass a pointer, and pass-by-value.

This is a first attempt at helping with that, but a better version would be a generic way to check when any proc values should be pass-by-pointer instead of pass-by-value, as discussed at https://forum.odin-lang.org/t/least-parsable-most-common-error-message/1241/3

image

## Better takes on it
1. If we keep the gb_string_is_equal, we should have the append_proc_name declaration defined once or at compile time, instead of on all check_call_arguments_proc_group with non-empty arguments.
2. The generic approach for all missed pass-by-pointer arguments:
      a. Check, efficiently, if any argument match a pointer type of the possible procs
      b. If so, have the pass-by-pointer message for that/those argument(s).
      
## Why not have done those things already
I need more practice with C++, not yet familiar enough to make it good and generic yet.
So I might come back to it later.

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -7283,6 +7283,10 @@ gb_internal CallArgumentData check_call_arguments_proc_group(CheckerContext *c, 
 			error_line("\tNo given arguments\n");
 		} else {
 			print_argument_types();
+			const gbString append_proc_name = gb_string_make(temporary_allocator(), "append");
+			if(gb_string_are_equal(expr_name, append_proc_name)){
+				error_line("Suggestion: Make sure that you passed, as the first argument, a dynamic array by pointer, like so `append(&some_dynamic_array, ...)`.\n");
+			}
 		}
 
 		if (procs.count == 0) {


Creativty:
I think a better solution would be to whenever a procedure takes a pointer, and you pass the underlying type of the pointer, the suggestion comes up. Not just hard-coding it for `append`.
Although I can see this being a pain in the ass to handle due to procedure groups.

duchainer:
> I think a better solution would be to whenever a procedure takes a pointer, and you pass the underlying type of the pointer, the suggestion comes up. Not just hard-coding it for `append`. Although I can see this being a pain in the ass to handle due to procedure groups.

Yeah I agree that this is a naive way and that the generic approach would be way better.

I also have an open PR for the more generic non-proc-group polymorphic proc case : https://github.com/odin-lang/Odin/pull/5818

I'll see to also have the "Suggestion:" for non-polymorphic procs for now too.

And maybe I'll find a not-too-bad generic way for the proc-group, but that looks like something that someone with more C++ would have an easier time. We'll see if I can get some time later for that. :)

Thanks for the comment/review :100:

gingerBill:
This is a bodge and I will not accept it. Please don't hardcode things like this.

duchainer:
> This is a bodge and I will not accept it. Please don't hardcode things like this.

Gotcha, sorry that this PR is not good. I'll do better.

Would https://github.com/odin-lang/Odin/pull/5818 make sense, for non-proc-group, but polymorphic procedures?
Or would it still be a bodge because we only check for the the first argument being a pointer, and we should cover cases of any parameters count?

PR #5822, opened 2025-10-17
Add hint when only missing '&' in polymorphic parameters

Creativty:
Provides useful suggestion in cases where applicable such as:
main.odin
```odin
package main

main :: proc() {
    arr := make([dynamic]int)
    append_elem(arr, 3)
}
```
```sh
$ odin check .
.../main.odin(5:14) Error: Cannot determine polymorphic type from parameter: '[dynamic]int' to '^$T/[dynamic]$E'
	append_elem(arr, 3)
	            ^~^
	Suggestion: Did you mean '&arr'
```
This commits also add the ctags file `tags` to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -308,3 +308,6 @@ build/
 cmake-build*/
 CMakeLists.txt
 sandbox/
+
+# Ctags
+tags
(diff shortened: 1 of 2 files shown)


Creativty:
The commit 43529dd has implemented the proposed solution.

PR #5824, opened 2025-10-18
Correct parser end_pos for multiline strings

BradLewis:
In `ols` we had an issue reported with the formatter here https://github.com/DanielGavin/ols/issues/1100. After a brief investigation, I found that the odin parser doesn't handle multi-line strings correctly and would report the end position on the same line as the start. This just updates to parsing to behave the same as multi-line comments.


Kelimion:
Don't mind the spurious CI error. A test failed DNS resolution, which is obviously nothing to do with this patch.

PR #5859, opened 2025-10-28
Change Odin's LICENSE to zlib from BSD 3-clause

gingerBill:
This change was made in order to allow things produced with Odin and using Odin's core library, to not require the LICENSE to also be distributed alongside the binary form.

### DIFF
--- a/core/mem/tlsf/tlsf.odin
+++ b/core/mem/tlsf/tlsf.odin
@@ -3,7 +3,7 @@ package mem_tlsf
 
 /*
 	Copyright 2024 Jeroen van Rijn <nom@duclavier.com>.
-	Made available under Odin's license.
+	Made available under tlsf's LICENSE (BSD 3-clause).
 
 	List of contributors:
 		Matt Conte:      Original C implementation, see LICENSE file in this package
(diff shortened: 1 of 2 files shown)


core/hash/xxhash/streaming.odin
```diff
@@ -4,7 +4,7 @@ package xxhash
 	An implementation of Yann Collet's [xxhash Fast Hash Algorithm](https://cyan4973.github.io/xxHash/).
 	Copyright 2021 Jeroen van Rijn <nom@duclavier.com>.
 
-	Made available under Odin's BSD-3 license, based on the original C code.
+	Made available under Odin's license, based on the original C code.
```
hasenj, line 7: Can you just _change_ the license from BSD2 to zlib?

https://github.com/Cyan4973/xxHash/blob/dev/LICENSE

Not sure how this stuff works, but given the original is licensed under BSD2, it requires:

> Redistributions of source code must retain the above copyright notice, this
  list of conditions and the following disclaimer.

Kelimion: Normally you try to release a source port under the same license, but Yann clearly doesn't mind if they're under a different license: https://xxhash.com, as he links to different implementations under a variety of licenses himself, which don't reproduce the original license text.

I've spotted [MIT](https://github.com/cespare/xxhash/blob/main/LICENSE.txt), [Boost](https://github.com/DoumanAsh/xxhash-rust/blob/master/LICENSE), and more. I'll also note that BSD-2, like zlib, doesn't have the clause attribution in binary releases. They're functionally equivalent.

Issue #5603, opened 2025-08-20
Compiler panic - nesting same struct both with & without `using`?


samwega:
## Context

* Operating System & Odin Version:
Windows 11 LTSC & dev-2025-08-nightly:accdd7c
 
* Please paste `odin report` output:
        Odin:    dev-2025-08-nightly:accdd7c
        OS:      Windows 11 Unknown Edition (000000bf) (version: 24H2), build 26100.4946
        CPU:     AMD Ryzen 7 6800HS Creator Edition
        RAM:     14020 MiB
        Backend: LLVM 20.1.0

Another person in community reproduced.

## Current Behavior

Compiler panic. You can remove the `using`, or add the missing `using`, and it works either way. You can also remove either of the structs and it also works.. so there's no error in either side, just in both trying to make use of the same Person struct with `using` and without `using`.

### Steps to Reproduce

```odin
package main
import "core:fmt"

Person_Stats :: struct {
        health: int,
        age: int,
}

Person :: struct {
        stats: Person_Stats, //no `using` 
        name: string,
}

PersonU :: struct {
        using stats: Person_Stats, // `using`
        name : string,
}

main :: proc() {

        p := Person {
                stats = {
                        health = 12,
                        age = 42,
                },
                name = "John Doe",
        }

        fmt.println(p)

        q := PersonU {
                stats = {
                        health = 23,
                        age = 12,
                },
                name = "UsingDrugz",
        }

        fmt.println(q)
}
```

### Failure Logs

```error
D:\a\Odin\Odin\src\checker.cpp(6967): Panic: struct {stats: Person_Stats, name: string} (struct{stats:main::Person_Stats,name:string}) 17194212857420152691 vs struct {stats: Person_Stats, name: string} (struct{stats:main::Person_Stats,name:string}) 17194212857420152691
```

Type_ids are different if moving the structs inside/outside main proc().

Kelimion:
Fixed by https://github.com/odin-lang/Odin/pull/5871.

PR #5871, opened 2025-11-02
Add #subtype using to name canonicalization

### LINKED ISSUE #5865: Type hash colission compiler panic in checker
## Context

        Odin:    dev-2025-10:d062203ff
        OS:      Windows 10 Enterprise N (version: 22H2), build 19045.5371
        CPU:     Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
        RAM:     65470 MiB
        Backend: LLVM 20.1.0

## Expected Behavior

No compiler panic.

## Current Behavior

C:\Users\Rennorb\source\external\Odin\src\checker.cpp(7535): Panic: struct {value: FileView, __ref_count: u32} (struct{value:core::FileView,__ref_count:u32}) 895794444630130322 vs struct {value: FileView, __ref_count: u32} (struct{value:core::FileView,__ref_count:u32}) 895794444630130322

### Steps to Reproduce

I was not able to create a minimal reproduction case, but i can give access to a private repo that reliably hits the panic. 

This seems related to #5603, and maybe also #5743 since i'm doing some wired multi-builds (the more i think about it the less likely the later one is).

Kelimion:
Fixes #5865 + #5603.

### DIFF
--- a/src/name_canonicalization.cpp
+++ b/src/name_canonicalization.cpp
@@ -758,11 +758,11 @@ gb_internal void write_type_to_canonical_string(TypeWriter *w, Type *type) {
 			}
 
 			if (f->flags & EntityFlags_IsSubtype) {
-				type_writer_appendc(w, "#subtype");
+				type_writer_appendc(w, "#subtype ");
 			}
 
 			if (f->flags & EntityFlag_Using) {
-				type_writer_appendc(w, "using");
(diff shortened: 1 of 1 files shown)


Kelimion:
I've checked this against #5871 + #5603 and it fixes both, but @gingerBill may want to look at it. It may well be considered a bodge.

PR #5882, opened 2025-11-04
encoding/json: custom json (un)marshalling

### LINKED ISSUE #4257: Json should allow a way to specify how to marshal/unmarshal an object (e.g. encoding/uuid)
# PLEASE POST THIS IN THE DISCUSSION TAB UNDER "PROPOSALS" OR "IDEAS/REQUESTS"

**Is your feature request related to a problem? Please describe.**
uuids are being used more and more often as id and when writing or reading from json, it's much better to use the `string` format of the uuid which is a string of the format 8-4-4-4-12. However due to implementation detail, `uuid.Identifier` is basically `distinct [16]u8` which when printing or marshalling, it's basically an array of bytes that is not only long but also difficult to discern.
Contrast `[16]u8`
`[38,101,146,205,150,13,64,145,152,28,140,37,196,75,16,24]`
with `string`
`266592cd-960d-4091-981c-8c25c44b1018`

**Describe the solution you'd like**
A generic solution would be something similar to how `fmt` allows you to use either `fmt:"s"` in the struct and then overwrite behavior with `User_Formatter`. It would be great if json has/reuses the similar ideas.

Otherwise, it would be great if we can just get uuid to be able to marshal/unmarshal to `json` and printing to `fmt` using the string format while still maintaining maybe a binary format such as `distinct [16]u8` internally.

**Additional context**
I'm using uuid as the id field of multiple structs and I'm also writing a rest api which uses the id to specify which resource to start/stop etc.

IllusionMan1212:
Closes #4257 

This allows passing an optional map of `typeid` to `User_Marshaller`/`User_Unmarshaller` to `json.marshal()` and `json.unmarshal()` respectively.

I will admit I primarily focused on making this work for my usecase and hadn't considered much else. Let me know if I missed something or if something needs changing.

### DIFF
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -62,10 +62,79 @@ Marshal_Options :: struct {
 	mjson_skipped_first_braces_end: bool,
 }
 
-User_Marshaller :: #type proc(w: io.Writer, v: any) -> Marshal_Error
-User_Marshaller_Map :: map[typeid]User_Marshaller
+User_Marshaler :: #type proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> Marshal_Error
 
-marshal :: proc(v: any, opt: Marshal_Options = {}, user_marshallers: User_Marshaller_Map = nil, allocator := context.allocator, loc := #caller_location) -> (data: []byte, err: Marshal_Error) {
+Register_User_Marshaler_Error :: enum {
+	None,
+	No_User_Marshaler,
+	Marshaler_Previously_Found,
+}
+
+// Example User Marshaler:
+// Custom Marshaler for `int`
+// Some_Marshaler :: proc(w: io.Writer, v: any, opt: ^json.Marshal_Options) -> json.Marshal_Error {
+// 	io.write_string(w, fmt.tprintf("%b", v))
+// 	return json.Marshal_Data_Error.None
+// }
+//
+// main :: proc() {
+//	// Ensure the json._user_marshaler map is initialized
+//	json.set_user_marshalers(new(map[typeid]json.User_Marshaler))
+//	reg_err := json.register_user_marshaler(type_info_of(int).id, Some_Marshaler)
+//	assert(reg_err == .None)
+//
+//
+// 	// Use the custom marshaler
+// 	SomeType :: struct {
+// 		value: int,
+// 	}
+//
+// 	x := SomeType{42}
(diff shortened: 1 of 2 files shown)


core/encoding/json/marshal.odin
```diff
@@ -62,7 +62,10 @@ Marshal_Options :: struct {
 	mjson_skipped_first_braces_end: bool,
 }
 
-marshal :: proc(v: any, opt: Marshal_Options = {}, allocator := context.allocator, loc := #caller_location) -> (data: []byte, err: Marshal_Error) {
+User_Marshaller :: #type proc(w: io.Writer, v: any) -> Marshal_Error
+User_Marshaller_Map :: map[typeid]User_Marshaller
+
+marshal :: proc(v: any, opt: Marshal_Options = {}, user_marshallers: User_Marshaller_Map = nil, allocator := context.allocator, loc := #caller_location) -> (data: []byte, err: Marshal_Error) {
```
Neirokan: It's probably better to place `user_marshallers` inside `Marshal_Options`.

core/encoding/json/marshal.odin
```diff
@@ -62,7 +62,10 @@ Marshal_Options :: struct {
 	mjson_skipped_first_braces_end: bool,
 }
 
-marshal :: proc(v: any, opt: Marshal_Options = {}, allocator := context.allocator, loc := #caller_location) -> (data: []byte, err: Marshal_Error) {
+User_Marshaller :: #type proc(w: io.Writer, v: any) -> Marshal_Error
```
Neirokan: I think that `User_Marshaller` needs `^Marshaller_Options` too. If marshaller produces a json object, then there should be a way to respect all those options.
Also makes it possible to just call `json.marshal_to_writer(w, transformed_data, opt)`

Kelimion:
I'll be making a few changes to bring it further in line with the way `core:fmt` does its custom formatters, if you don't mind, @IllusionMan1212.

IllusionMan1212:
Go ahead and thank you for taking the time.

Kelimion:
@gingerBill, can you have a gander at this as well?

It works the same way as the user formatters in `core:fmt`, but because there's a bidirectional case to be made here, I wonder if we should keep one `map` each for the marshalers and unmarshalers, or have 1 map serving both?

Kelimion:
Previous force-push erased the changes I made, including the examples, so I had to force-push to put those back.

PR #5892, opened 2025-11-05
container/small_array: update procs according to docs

NicknEma:
Make the get() and get_ptr() procedures actually do what the documentation says. Replace the word "crash" with "panic" in the comment.

### DIFF
--- a/core/container/small_array/small_array.odin
+++ b/core/container/small_array/small_array.odin
@@ -105,8 +105,8 @@ This operation assumes that the small-array is large enough.
 
 This will result in:
 	- the value if 0 <= index < len
-	- the zero value of the type if len < index < capacity
-	- 'crash' if capacity < index or index < 0
+	- the zero value of the type if len <= index < capacity
+	- 'panic' if capacity < index or index < 0
 
 **Inputs**
 - `a`: The small-array
@@ -116,6 +116,9 @@ This will result in:
 - the element at the specified position
 */
 get :: proc "contextless" (a: $A/Small_Array($N, $T), index: int) -> T {
+	if a.len <= index && index < builtin.len(a.data) {
+		return {}
+	}
 	return a.data[index]
 }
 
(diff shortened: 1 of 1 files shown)


NicknEma:
I don't know if it should be the other way around, and I should've changed the docs to match the code...

Neirokan:
I would argue that based on git blame it's a docs (added 7 months ago) issue, not a code (added 4 years ago) issue. Especially when there are `*_safe` versions of those functions below.
Also I don't think that nil is "the pointer to the zero value".

Kelimion:
I'll leave this tab open and review the code tomorrow so see which way around the docs ought to be.

Kelimion:
I pushed https://github.com/odin-lang/Odin/commit/91409cb37ee075a380eacdcccf26a37bf93687e0, which documents what `get` and `get_ptr` do.

The point of these two is that they don't perform (their own) bounds checks and return a safe value, as this PR introduces to match the old documentation. Rather, I've updated the documentation to match the implementation and its intent.

We can use `get_safe` and `get_ptr_safe` to perform a bounds check and return `T`'s nil value.
You could of course add your own `pop_back_safe` and `pop_front_safe` that clear the orphaned slot.

e.g.
```odin
	a: small_array.Small_Array(5, int)
	small_array.push(&a, 0, 1, 2)

	fmt.println("BEFORE:", small_array.slice(&a)) // [0, 1, 2]
	small_array.pop_back(&a)
	fmt.println("AFTER: ", small_array.slice(&a)) // [0, 1]
	fmt.printfln("a[2]: %v", small_array.get(a, 2)) // a[2]: 2 (as left behind from the original `push`, not erased by `pop_back`.
	p := small_array.get_ptr(&a, 2) // pointer to the original 2
	fmt.printfln("&a[2]: %p -> %v", p, p^) // &a[2]: 0xAD8853F750 -> 2

	fmt.printfln("a[2]: %v %v", small_array.get_safe(a, 2)) // a[2]: 0 false
	p2, p2_ok := small_array.get_ptr_safe(&a, 2)
	fmt.printfln("&a[2]: %p -> %v", p2, p2_ok) // &a[2]: 0x0 -> false
```

PR #5895, opened 2025-11-06
encoding/json: clone unquoted strings

### LINKED ISSUE #5815: SJSON parser crashes with "pointer being freed was not allocated" when unquoted value contains parentheses
## Context

Operating System & Odin Version:
 - macOS Sequoia 15.7.1 (build 24G231, kernel 24.6.0)
 - Odin: dev-2025-02:584fdc0d4
 - CPU: Apple M1 Pro
 - RAM: 32768 MiB
 - Backend: LLVM 19.1.7

## Expected Behavior

Parsing SJSON with an unquoted value containing parentheses should not crash.  
It should either parse as a bare identifier token or produce a normal parse error.

## Current Behavior

The process aborts with a malloc error: `pointer being freed was not allocated` when the unquoted value contains parentheses.

### Steps to Reproduce

1. Create a minimal Odin file:
 ```odin
package bugreport

import "core:encoding/json"
import "core:fmt"
import "core:strings"

main :: proc() {
	b: strings.Builder
	strings.builder_init_none(&b)
	defer strings.builder_destroy(&b)

	strings.write_string(&b, `{ compute: compute() }`)
	v, err := json.parse(b.buf[:], json.Specification.SJSON)

	fmt.println("unreacheable code")
}

```
2. Run:
```sh
odin run .
```

3. Observe the crash.

Failure Logs

```sh
-> % odin run libs/bugreport
bugreport(17655,0x20732e140) malloc: *** error for object 0x203a657475706d6f: pointer being freed was not allocated
bugreport(17655,0x20732e140) malloc: *** set a breakpoint in malloc_error_break to debug
[1]    17642 abort      odin run libs/bugreport
```

Additional Notes
 - Reproducible 100% of the time with the provided input.
 - The crash occurs before json.parse returns an error.
 - The issue appears specific to json.Specification.SJSON when an unquoted value contains parentheses.

Neirokan:
Fixes #5815

### DIFF
--- a/core/encoding/json/unmarshal.odin
+++ b/core/encoding/json/unmarshal.odin
@@ -188,8 +188,18 @@ assign_float :: proc(val: any, f: $T) -> bool {
 
 
 @(private)
-unmarshal_string_token :: proc(p: ^Parser, val: any, str: string, ti: ^reflect.Type_Info) -> (ok: bool, err: Error) {
-	val := val
+unmarshal_string_token :: proc(p: ^Parser, val: any, token: Token, ti: ^reflect.Type_Info) -> (ok: bool, err: Error) {
+	str: string
+	switch {
+	case token.kind == .String:
+		str = unquote_string(token, p.spec, p.allocator) or_return
+	case:
+		str = clone_string(token.text, p.allocator) or_return
+	}
+	defer if !ok || (val.id != string && val.id != cstring) {
+		delete(str, p.allocator)
+	}
+
 	switch &dst in val {
 	case string:
 		dst = str
@@ -339,7 +349,7 @@ unmarshal_value :: proc(p: ^Parser, v: any) -> (err: Unmarshal_Error) {
 	case .Ident:
 		advance_token(p)
 		if p.spec == .MJSON {
-			if unmarshal_string_token(p, any{v.data, ti.id}, token.text, ti) or_return {
+			if unmarshal_string_token(p, any{v.data, ti.id}, token, ti) or_return {
 				return nil
 			}
 		}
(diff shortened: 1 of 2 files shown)


core/encoding/json/unmarshal.odin
```diff
@@ -188,8 +188,18 @@ assign_float :: proc(val: any, f: $T) -> bool {
...
-	val := val
+unmarshal_string_token :: proc(p: ^Parser, val: any, token: Token, ti: ^reflect.Type_Info) -> (ok: bool, err: Error) {
+	str: string
+	switch {
+	case token.kind == .String:
+		str = unquote_string(token, p.spec, p.allocator) or_return
+	case:
+		str = clone_string(token.text, p.allocator) or_return
+	}
+	defer if !ok || (val.id != string && val.id != cstring) {
```
Kelimion, line 199: I would perform this `val.id` check at the top, and if the target is not a `string` or `cstring`, exit early.
No need to first unquote or clone, only to delete it again, or is there?

Neirokan: It can't exit early because this function tries to parse string as enum/int/float below.
It also needs to unquote string for those cases to remove quotes and unescape (at least unicode sequences, I don't expect anything else to affect enum name/int/float).
It can use `str = token.str` when it's not quoted, and initially I had a separate case for this, but I realized that I messed up defer condition and removed it to make this a bit simpler.
Should I return it with fixed condition?

Kelimion: That's fair. Damn diffs and their myopic context. :D

PR #5897, opened 2025-11-06
Add "contextless" to procs in slice package

thetarnav:
Looks like a lot of procedures in the `slice` package could use a `"contextless"` calling convention as most don't require context. Some procedures there already were contextless, but it was applied inconsistently—why would `to_bytes` be contextless, but not `to_type`.
Let me know if some of these should require context by design or something.

### DIFF
--- a/core/slice/slice.odin
+++ b/core/slice/slice.odin
@@ -41,7 +41,7 @@ to_bytes :: proc "contextless" (s: []$T) -> []byte {
 	Turns a byte slice into a type.
 */
 @(require_results)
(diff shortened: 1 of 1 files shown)


Barinzaya:
The main consequence of this is that anything that can fail a bounds check (e.g. `swap`, `split_first`) will no longer be able to use `context.assertion_failure_proc` to allow for the panic behavior to be overridden (e.g. to "catch" the panic, or to log it).

gingerBill:
I'm going to close this PR and make the adjustments myself as not everything should have `"contextless"`

PR #5905, opened 2025-11-08
Remove `do`

dozn:
Also removed the `-disallow-do` flag, and apparently a bunch of unnecessary spaces at the end of lines (automatically by VSCode).

Here for posterity, dreaming of a simpler world.
Do as you wish with this, of course.

### DIFF
--- a/tests/issues/test_issue_4584.odin
+++ b/tests/issues/test_issue_4584.odin
@@ -193,6 +193,6 @@ expect_float_matrix_value :: proc(t: ^testing.T, value, expected: $M/matrix[$N, 
 			}
 		}
 	}
-	if !ok do log.errorf("expected %v to be %v, got %v", value_expr, expected, value, location=loc)
+	if !ok { log.errorf("expected %v to be %v, got %v", value_expr, expected, value, location=loc) }
 	return ok
 }


Apis035:
What am I gonna do with my code that uses a bunch of `do`'s 😔

dozn:
> What am I gonna do with my code that uses a bunch of `do`'s 😔

You must atone for your sins, and complete your penance.

9thGluon:
> You must atone for your sins, and complete your penance.

Forgive me, Father of Modules, for I have sinned — I used the `do` keyword in Odin.
And I shall continue to `do` so for all eternity, indulging in high dosage.

dozn:
Reasoning:

- doesn't belong in a language which has minimized decision-making
- makes the language more complex for minimal benefit, with three different usages:
- - `if true do thing()` (creating a one-liner)
- - `for x in xs do if true {` (adding a condition on the same line as a loop declaration)
- - `if true do switch thing {` (avoiding indentation)
- makes code harder to scan/read since a single line is now doing multiple different things
- - also harder to visually differentiate letters from symbols
- at worst, costs two characters `_do_` vs `{}`
- - at best, saves twelve characters `for x in xs do if is_ok {` vs `for x in xs {\n\tif !is_ok { continue }`
- - - although this also reverses an early-exit loop condition, adding confusion
- requires a minor refactor as soon as more code is added, messing with diffs
- has split the community
- is a keyword we can remove
- we won't be able to right this wrong post-release, better to rip off the bandage early

korvahkh:
As an individual who micro`do`ses `do`, I `do` not oppose this change.

antisaling:
time to fork the copiler. stop being a dumbass. this is really useful for multidimensional math stuff (stuffing for loopings over Ndim). if you can't read it you are just not qualified for programming.

Issue #5881, opened 2025-11-04
default random generator not uniform on 256 bit interval


SOBEX:
## Context

constructing 4d points and counting in which orthant they fall has a significant bias
confirmed on octobers release dev-2025-10-nightly:7237747

```odin
random_generator_period_quat256::proc(iterations:=100_000_000,gen:=context.random_generator){
   counts:[2][2][2][2]int
   for _ in 0..<iterations{
      quat:quaternion256
      quat.x=rand.float64_range(-10,10,gen)
      quat.y=rand.float64_range(-10,10,gen)
      quat.z=rand.float64_range(-10,10,gen)
      quat.w=rand.float64_range(-10,10,gen)
      counts[quat.w>=0?1:0][quat.z>=0?1:0][quat.y>=0?1:0][quat.x>=0?1:0]+=1
   }
   fmt.println(counts)
}
```

## Expected Behavior

uniform distribution over 256 bit interval, see run with crypto.random_generator
```
[[[[6249432, 6247440], [6254894, 6246002]], [[6249059, 6250686], [6246800, 6245386]]], [[[6252428, 6252212], [6252157, 6246725]], [[6249742, 6250116], [6253940, 6252981]]]]
```

## Current Behavior

```
[[[[6211079, 6217687], [7032779, 7032439]], [[5512272, 5520713], [6234779, 6233805]]], [[[6235810, 6236637], [7089622, 7077973]], [[5466182, 5474363], [6216283, 6207577]]]]
```

## Failure Information (for bugs)

this bias can also be observed using [default_random_generator_proc](https://github.com/odin-lang/Odin/blob/a414061998010fb71268dbdb45576ed6367a6a7f/base/runtime/random_generator.odin#L53-L59) directly
specifically for every four u64 generated the 2nd, 65th and 66th, 129th and 130th, 194th bits show bias which impacts floats significantly
additionally the lower half of each u64 also shows biases on most bits

### Steps to Reproduce

1. run the snippet in the above context

Yawning:
This is fixed by #5963 (chacha8rand), and we also support xoshiro256** + splitmix64 as well.

PR #5907, opened 2025-11-09
runtime: replace default RNG with xoshiro256** + splitmix64; SOLVE: #5881

bogwi:
runtime: replace default RNG with xoshiro256** + splitmix64; deterministic seeding

- Replace custom RNG with xoshiro256** (rotl(s1*5,7)*9) and standard state transition for high-quality, unbiased 64-bit outputs.
- Seed via splitmix64 expansion into 4×u64; `.Reset(seed)` is deterministic for all seeds, including 0.
- Use cycle-counter seeding only on first use when state is all-zero and no explicit `.Reset` was called.

BREAKING CHANGE: Default RNG sequences for a given seed have changed.
Refs: #5881

---

# Odin default RNG bias

## Background: original RNG and observed bias

The previous default RNG in `Odin/base/runtime/random_generator.odin` implemented a custom 64‑bit generator that:
- updated a 64‑bit internal `state` with a linear recurrence,
- constructed the output by combining bitwise operations (including a dependency on the highest bits of `state`), and
- used that same top‑bit region to parameterize a rotation of the mixed value.

Symptoms reported by @SOBEX (issue #5881):
- A simple 4D sign test (counting in which orthant uniformly sampled points fall) showed large deviations from uniformity with the default RNG.
- The bias was visible across specific bit positions within each group of generated 64‑bit outputs (e.g., the 2nd, 65th/66th, etc.), impacting floating‑point generation significantly.
- Substituting `core:crypto.random_generator()` (OS entropy) produced nearly uniform results in the same test.

Example reproduction (20,000,000 iterations; expected per cell ≈ 1,250,000):

Default RNG (old):
```
[[[[6211079, 6217687], [7032779, 7032439]], [[5512272, 5520713], [6234779, 6233805]]],
 [[[6235810, 6236637], [7089622, 7077973]], [[5466182, 5474363], [6216283, 6207577]]]]
```

Crypto RNG (OS entropy):
```
[[[[6249432, 6247440], [6254894, 6246002]], [[6249059, 6250686], [6246800, 6245386]]],
 [[[6252428, 6252212], [6252157, 6246725]], [[6249742, 6250116], [6253940, 6252981]]]]
```

The default RNG’s orthant counts (old) show pronounced skew not seen under the crypto RNG.

---

## Root cause

Two interacting issues in the original algorithm caused structural bias:
1) Non‑standard mixing and rotation:
   - The output transform used a rotation amount derived from the same high bits that also guided other parts of the mixing function.
   - Reusing the same bit slice for both "how much to rotate" and "what to rotate" creates correlations that systematically favor or disfavor particular bit positions in the output word.
2) Weak low‑order statistics carried into float sampling:
   - Odin’s `float64()` draws from integer output and maps to the unit interval (with rejection for bounds), assuming the integer generator is uniform over its bit width.
   - Any per‑bit skew in the generator leaks into sign tests and floating‑point sampling, manifesting as quadrant/orthant bias in simple geometric tests.

Together these produced the observed, reproducible deviations from uniformity and bit balance.

---

### Core generator: xoshiro256**

I replaced the previous custom transform with xoshiro256**:
- State: 256 bits (4 × u64)
- Output function ("star‑star"): `rotl(s1 * 5, 7) * 9`
- Transition: xorshifts and fixed rotations on the state words

This generator is widely used and extremely fast. 

### Seeding: splitmix64 expansion

Seed the 256‑bit xoshiro state by expanding a single 64‑bit seed with splitmix64:
- On `.Reset(seed)`, we run splitmix64 four times to fill `[4]u64`. This is fully deterministic for any 64‑bit seed value, including `0`.
- If no seed is explicitly provided by the user (i.e., the generator is used without a prior `.Reset` and the internal state is all‑zero), we fallback to a platform cycle counter to produce a non‑zero state via splitmix64.
- We defensively guard against the extremely unlikely all‑zero state.

Splitmix64 is fast, simple, and purpose‑built for turning a small seed into a well‑distributed stream suitable for seeding other PRNGs.

---

## Testing and validation

I added a targeted test (see `Odin/tests/issues/test_issue_5881.odin`) that check:
- Determinism: two RNGs reset with the same seed produce the same first value.
- Flags: `.Uniform` and `.Resettable` are reported by the default RNG.
- Bit balance: for 200,000 samples of `uint64`, each bit’s count of ones is within a 6σ bound around 50%.
- 4D sign uniformity: we sample signs of 4 independent `float64_range(-10, 10)` draws, bin into 16 orthants, and check the chi‑square statistic against a generous threshold. The fixed generator passes; the old version did not.

Example runs (20,000,000 iterations; expected per cell ≈ 1,250,000) after the fix:
```
default random generator:
[[[[1250162, 1250442], [1250820, 1251092]], [[1250518, 1250322], [1248821, 1249426]]],
 [[[1250280, 1250451], [1249173, 1248935]], [[1249642, 1250723], [1249094, 1250099]]]]

crypto random generator:
[[[[1250161, 1249435], [1252261, 1249551]], [[1249452, 1252756], [1249809, 1249990]]],
 [[[1247632, 1250246], [1248428, 1250507]], [[1250048, 1247982], [1251448, 1250294]]]]
```

Both sets are close to uniform. The new default generator exhibits no visible skew in this test.

---

## Compatibility

- API compatibility: unchanged. The default RNG is still exposed through the same `Random_Generator` surface, with `.Reset` accepting a `u64` seed (now expanded via splitmix64).
- Sequence compatibility: changed. The bit sequences for a given seed will differ from prior releases due to the new algorithm and seeding method. This is expected. Users relying on exact historical sequences should pin to an older compiler version or provide their own PRNG implementation.
- Cryptography: unchanged guidance. 

---

## Acknowledgements

This fix draws on research and practice around the xoshiro family (Sebastiano Vigna, David Blackman) and splitmix64.

---

## Reproduction snippet (4D sign orthant)

```odin
import "core:math/rand"
import "core:fmt"

random_generator_period_quat256 :: proc(iterations := 100_000_000, gen := context.random_generator) {
    counts: [2][2][2][2]int
    for _ in 0..<iterations {
        x := rand.float64_range(-10, 10, gen)
        y := rand.float64_range(-10, 10, gen)
        z := rand.float64_range(-10, 10, gen)
        w := rand.float64_range(-10, 10, gen)
        counts[w >= 0 ? 1 : 0][z >= 0 ? 1 : 0][y >= 0 ? 1 : 0][x >= 0 ? 1 : 0] += 1
    }
    fmt.println(counts)
}
```

## How it can break in theory (general stuff)

  - Deterministic sequences change: Any code that assert exact numbers from a given seed will fail.
  - Procedural generation reproducibility: Worlds/levels/assets keyed by a seed won’t reproduce identically across compiler versions.
  - Replays/snapshots: Simulations or game replays that depend on "same seed => same evolution" across versions will diverge.
  - Serialized/checkpointed RNG state: If only the seed (not full internal state) is saved, resumed runs will diverge from pre-change behavior.

- What won’t break
  - Tests checking properties/statistics (distribution, bounds, determinism per run) rather than exact values.
  - Code that only needs determinism within the same version while providing its own seed.
  - Users who don’t rely on cross-version reproducibility.

- Additional to be aware of
  - Seed 0 is now a valid deterministic seed. If anyone passed 0 expecting "randomize" behavior, they’ll now get a stable sequence. Randomized seeding still happens but only when no Reset occurred and the state is all-zero.

CI building fine is expected (I hope): this is a behavioral change, not an API/ABI change.

## Why changing in theory

A biased default RNG might cause subtle errors that would be hard to diagnose.

- **Statistical correctness**: Monte Carlo, simulations, p-values, and estimators get skewed; results can be systematically uncorrect rather than just noisy.
- **Procedural generation**: Symmetry-breaking artifacts, uneven distributions, and reproducibility that "works" but yields the wrong distribution.
- **Security-adjacent** (non-crypto): More predictable patterns and worse behavior in randomized algorithms (hash seeding, randomized quicksort pivoting), increasing worst-case risks.

Even small biases become visible with large samples and propagate widely. Fixing the default (as done or with any other PR) is the right call.

---

@gingerBill , what do you think?

### DIFF
--- a/tests/issues/test_issue_5881.odin
+++ b/tests/issues/test_issue_5881.odin
@@ -1,0 +1,93 @@
+// Tests issue #5881 https://github.com/odin-lang/Odin/issues/5881
+package test_issues
+
+import "core:math"
+import "core:math/rand"
+import "core:testing"
+import "base:runtime"
+
+// Helper: compute chi-square statistic for counts vs equal-expected across k bins
+chi_square_equal :: proc(counts: []int) -> f64 {
+	n := 0
+	for c in counts {
+		n += c
+	}
+	if n == 0 {
+		return 0
+	}
+	k := len(counts)
+	exp := f64(n) / f64(k)
+	stat := f64(0)
+	for c in counts {
+		d := f64(c) - exp
+		stat += (d * d) / exp
+	}
+	return stat
+}
+
+// Helper: check bit balance on u64 across many samples
+expect_u64_bit_balance :: proc(t: ^testing.T, gen: rand.Generator, samples: int, sigma_k: f64 = 6) {
+	ones: [64]int
+	for i := 0; i < samples; i += 1 {
+		v := rand.uint64(gen)
+		for b := 0; b < 64; b += 1 {
+			ones[b] += int((v >> u64(b)) & 1)
+		}
+	}
+	mu := f64(samples) * 0.5
+	sigma := math.sqrt(f64(samples) * 0.25)
+	limit := sigma_k * sigma
+	for b := 0; b < 64; b += 1 {
+		diff := math.abs(f64(ones[b]) - mu)
+		if diff > limit {
+			testing.expectf(t, false, "u64 bit %d imbalance: ones=%d samples=%d diff=%.1f limit=%.1f", b, ones[b], samples, diff, limit)
+			return
+		}
+	}
+}
+
+// Uniformity sanity via 4D sign orthant chi-square with modest sample size.
+expect_quaternion_sign_uniformity :: proc(t: ^testing.T, gen: rand.Generator, iterations: int) {
+	counts: [16]int
+	for _ in 0..<iterations {
+		// Map 4D signs to 0..15 index
+		x := rand.float64_range(-10, 10, gen)
+		y := rand.float64_range(-10, 10, gen)
+		z := rand.float64_range(-10, 10, gen)
+		w := rand.float64_range(-10, 10, gen)
+		idx := 0
+		if x >= 0 { idx |= 1 }
+		if y >= 0 { idx |= 2 }
+		if z >= 0 { idx |= 4 }
+		if w >= 0 { idx |= 8 }
+		counts[idx] += 1
+	}
+	// df = 15. For a modest sample size, use a generous cutoff to reduce flakiness.
+	// Chi-square critical values (df=15): p=0.001 -> ~37.7, p=0.0001 -> ~43.8
+	// We accept < 55 as a conservative stability bound across platforms.
+	chi := chi_square_equal(counts[:])
+	testing.expectf(t, chi < 55.0, "4D sign chi-square too high: %.3f (counts=%v)", chi, counts)
+}
+
+@test
+test_runtime_default_rng_properties :: proc(t: ^testing.T) {
+	// Determinism with same seed
+	state1 := rand.create(123456789)
+	state2 := rand.create(123456789)
+	gen1 := rand.default_random_generator(&state1)
+	gen2 := rand.default_random_generator(&state2)
+	a1 := rand.uint64(gen1)
+	a2 := rand.uint64(gen2)
+	testing.expect(t, a1 == a2, "default RNG not deterministic with same seed")
+
+	// Info flags
+	info := rand.query_info(gen1)
+	testing.expect(t, runtime.Random_Generator_Query_Info_Flag.Uniform in info, "default RNG must be Uniform")
+	testing.expect(t, runtime.Random_Generator_Query_Info_Flag.Resettable in info, "default RNG must be Resettable")
+
+	// Bit balance and sign uniformity (modest samples to keep CI fast)
+	expect_u64_bit_balance(t, gen1, 200_000)
+	expect_quaternion_sign_uniformity(t, gen1, 200_000)
+}
+
+
--- a/base/runtime/random_generator.odin
+++ b/base/runtime/random_generator.odin
@@ -44,33 +44,53 @@ random_generator_reset_u64 :: proc(rg: Random_Generator, p: u64) {
 
 
 Default_Random_State :: struct {
-	state: u64,
-	inc:   u64,
+	s: [4]u64,
 }
 
 default_random_generator_proc :: proc(data: rawptr, mode: Random_Generator_Mode, p: []byte) {
 	@(require_results)
 	read_u64 :: proc "contextless" (r: ^Default_Random_State) -> u64 {
-		old_state := r.state
-		r.state = old_state * 6364136223846793005 + (r.inc|1)
-		xor_shifted := (((old_state >> 59) + 5) ~ old_state) * 12605985483714917081
-		rot := (old_state >> 59)
-		return (xor_shifted >> rot) | (xor_shifted << ((-rot) & 63))
+		// xoshiro256** output function and state transition
+		rotl :: proc "contextless" (x: u64, k: u64) -> u64 {
+			return (x << k) | (x >> ((-k) & 63))
+		}
+
+		result := rotl(r.s[1] * 5, 7) * 9
+		t := r.s[1] << 17
+
+		r.s[2] = r.s[2] ~ r.s[0]
+		r.s[3] = r.s[3] ~ r.s[1]
+		r.s[1] = r.s[1] ~ r.s[2]
+		r.s[0] = r.s[0] ~ r.s[3]
+		r.s[2] = r.s[2] ~ t
+		r.s[3] = rotl(r.s[3], 45)
+
+		return result
 	}
 
 	@(thread_local)
 	global_rand_seed: Default_Random_State
 
 	init :: proc "contextless" (r: ^Default_Random_State, seed: u64) {
-		seed := seed
-		if seed == 0 {
-			seed = u64(intrinsics.read_cycle_counter())
+		// splitmix64 to expand a 64-bit seed into 256 bits of state
+		sm64_next :: proc "contextless" (s: ^u64) -> u64 {
+			s^ += 0x9E3779B97F4A7C15
+			z := s^
+			z = (z ~ (z >> 30)) * 0xBF58476D1CE4E5B9
+			z = (z ~ (z >> 27)) * 0x94D049BB133111EB
+			return z ~ (z >> 31)
 		}
-		r.state = 0
-		r.inc = (seed << 1) | 1
-		_ = read_u64(r)
-		r.state += seed
-		_ = read_u64(r)
+
+		local_seed := seed
+		r.s[0] = sm64_next(&local_seed)
+		r.s[1] = sm64_next(&local_seed)
+		r.s[2] = sm64_next(&local_seed)
+		r.s[3] = sm64_next(&local_seed)
+		// Extremely unlikely all zero; ensure non-zero state
+		if (r.s[0] | r.s[1] | r.s[2] | r.s[3]) == 0 {
+			// force a minimal non-zero tweak
+			r.s[0] = 1
+		}
 	}
 
 	r: ^Default_Random_State = ---
@@ -82,8 +102,8 @@ default_random_generator_proc :: proc(data: rawptr, mode: Random_Generator_Mode,
 
 	switch mode {
 	case .Read:
-		if r.state == 0 && r.inc == 0 {
-			init(r, 0)
+		if (r.s[0] | r.s[1] | r.s[2] | r.s[3]) == 0 {
+			init(r, u64(intrinsics.read_cycle_counter()))
 		}
 
 		switch len(p) {
@@ -106,7 +126,7 @@ default_random_generator_proc :: proc(data: rawptr, mode: Random_Generator_Mode,
 		}
 
 	case .Reset:
-		seed: u64
+		seed: u64 = 0
 		mem_copy_non_overlapping(&seed, raw_data(p), min(size_of(seed), len(p)))
 		init(r, seed)
 
--- a/tests/issues/run.sh
+++ b/tests/issues/run.sh
@@ -22,6 +22,7 @@ $ODIN test ../test_issue_3435.odin $COMMON
 $ODIN test ../test_issue_4210.odin $COMMON
 $ODIN test ../test_issue_4364.odin $COMMON
 $ODIN test ../test_issue_4584.odin $COMMON
+$ODIN test ../test_issue_5881.odin $COMMON
 if [[ $($ODIN build ../test_issue_2395.odin $COMMON 2>&1 >/dev/null | grep -c "Error:") -eq 2 ]] ; then
 	echo "SUCCESSFUL 1/1"
 else


bogwi:
`+ ../../../odin test ../test_issue_4584.odin -define:ODIN_TEST_FANCY=false -file -vet -strict-style
./run.sh: line 24:  5547 Segmentation fault      (core dumped) $ODIN test ../test_issue_4584.odin $COMMON
Error: Process completed with exit code 139.`

This one is weird. 

It doesn’t rely on the default RNG.
No rand import, no context.random_generator, no RNG calls.
Purely deterministic linear algebra checks with fixed matrices and a 1e-6 tolerance.

Any CI failure on Ubuntu ARM for this test is unrelated to random_generator.odin.  Or I missed something?

Kelimion:
> Any CI failure on Ubuntu ARM for this test is unrelated to random_generator.odin. Or I missed something?

A spurious CI failure. It happens from time to time.

Kelimion:
One thing that would be nice to see is the preservation of old PRNG somewhere in `core:math/rand`, for those who have a previous reliance on it.

e.g.
```odin
// Use the old default generator
pcg_state: rand.PCG_Random_State 
context.random_generator = rand.pcg_random(&pcg_state)
```

Kelimion:
@Yawning, what's your take on this PR? You were proposing chacha8.

tests/issues/test_issue_5881.odin
```diff
@@ -0,0 +1,93 @@
...
+	// Chi-square critical values (df=15): p=0.001 -> ~37.7, p=0.0001 -> ~43.8
+	// We accept < 55 as a conservative stability bound across platforms.
+	chi := chi_square_equal(counts[:])
+	testing.expectf(t, chi < 55.0, "4D sign chi-square too high: %.3f (counts=%v)", chi, counts)
+}
+
+@test
+test_runtime_default_rng_properties :: proc(t: ^testing.T) {
+	// Determinism with same seed
+	state1 := rand.create(123456789)
```
SOBEX, line 75: should this be nonstatic, like `t.seed`?

SOBEX:
passed my uniformity tests as well and i see no noticeable difference in execution times compared to the previous implementation

Yawning:
> @Yawning, what's your take on this PR? You were proposing chacha8.

Since we promise no guarantee on consistency here, this is fine in the interim imo.  In the long run I do want to replace the default RNG with chacha8rand (will cause a negligible performance regression), but we can put this one into `math/rand` if/when that happens.

bogwi:
> > @Yawning, what's your take on this PR? You were proposing chacha8.
> 
> Since we promise no guarantee on consistency here, this is fine in the interim imo. In the long run I do want to replace the default RNG with chacha8rand (will cause a negligible performance regression), but we can put this one into `math/rand` if/when that happens.

Nice.

How about this:

I can

1. place `xoshiro256.odin` in here `Odin/core/math/rand/`

2. Add some xoshiro256 lightweight uniformity testing in here `Odin/tests/core/math/rand/test_core_math_rand.odin`

3. Unless you insist on doing it yourself, I can replace the default generator with chacha8:

<details>
  <summary>ChaCha8</summary>

  ```odin
default_random_generator_proc :: proc(data: rawptr, mode: Random_Generator_Mode, p: []byte) {
	@(require_results)
	read_u64 :: proc "contextless" (r: ^Default_Random_State) -> u64 {
		rotl32 :: proc "contextless" (x: u32, k: u32) -> u32 {
			return (x << k) | (x >> (32 - k))
		}

		// Generate next ChaCha8 block into r.buf when exhausted
		generate_block :: proc "contextless" (r: ^Default_Random_State) {
			// Constants
			const0: u32 = 0x61707865
			const1: u32 = 0x3320646e
			const2: u32 = 0x79622d32
			const3: u32 = 0x6b206574

			state: [16]u32
			working: [16]u32

			state[0] = const0
			state[1] = const1
			state[2] = const2
			state[3] = const3
			state[4] = r.key[0]
			state[5] = r.key[1]
			state[6] = r.key[2]
			state[7] = r.key[3]
			state[8] = r.key[4]
			state[9] = r.key[5]
			state[10] = r.key[6]
			state[11] = r.key[7]
			state[12] = u32(r.counter & 0xffffffff)
			state[13] = u32(r.counter >> 32)
			state[14] = r.nonce[0]
			state[15] = r.nonce[1]

			// Copy to working
			for i in 0..<16 {
				working[i] = state[i]
			}

			// 8 rounds = 4 double-rounds
			for _ in 0..<4 {
				// Column round
				working[0]  += working[4];  working[12] ~= working[0];  working[12] = rotl32(working[12], 16)
				working[8]  += working[12]; working[4]  ~= working[8];  working[4]  = rotl32(working[4], 12)
				working[0]  += working[4];  working[12] ~= working[0];  working[12] = rotl32(working[12], 8)
				working[8]  += working[12]; working[4]  ~= working[8];  working[4]  = rotl32(working[4], 7)

				working[1]  += working[5];  working[13] ~= working[1];  working[13] = rotl32(working[13], 16)
				working[9]  += working[13]; working[5]  ~= working[9];  working[5]  = rotl32(working[5], 12)
				working[1]  += working[5];  working[13] ~= working[1];  working[13] = rotl32(working[13], 8)
				working[9]  += working[13]; working[5]  ~= working[9];  working[5]  = rotl32(working[5], 7)

				working[2]  += working[6];  working[14] ~= working[2];  working[14] = rotl32(working[14], 16)
				working[10] += working[14]; working[6]  ~= working[10]; working[6]  = rotl32(working[6], 12)
				working[2]  += working[6];  working[14] ~= working[2];  working[14] = rotl32(working[14], 8)
				working[10] += working[14]; working[6]  ~= working[10]; working[6]  = rotl32(working[6], 7)

				working[3]  += working[7];  working[15] ~= working[3];  working[15] = rotl32(working[15], 16)
				working[11] += working[15]; working[7]  ~= working[11]; working[7]  = rotl32(working[7], 12)
				working[3]  += working[7];  working[15] ~= working[3];  working[15] = rotl32(working[15], 8)
				working[11] += working[15]; working[7]  ~= working[11]; working[7]  = rotl32(working[7], 7)

				// Diagonal round
				working[0]  += working[5];  working[15] ~= working[0];  working[15] = rotl32(working[15], 16)
				working[10] += working[15]; working[5]  ~= working[10]; working[5]  = rotl32(working[5], 12)
				working[0]  += working[5];  working[15] ~= working[0];  working[15] = rotl32(working[15], 8)
				working[10] += working[15]; working[5]  ~= working[10]; working[5]  = rotl32(working[5], 7)

				working[1]  += working[6];  working[12] ~= working[1];  working[12] = rotl32(working[12], 16)
				working[11] += working[12]; working[6]  ~= working[11]; working[6]  = rotl32(working[6], 12)
				working[1]  += working[6];  working[12] ~= working[1];  working[12] = rotl32(working[12], 8)
				working[11] += working[12]; working[6]  ~= working[11]; working[6]  = rotl32(working[6], 7)

				working[2]  += working[7];  working[13] ~= working[2];  working[13] = rotl32(working[13], 16)
				working[8]  += working[13]; working[7]  ~= working[8];  working[7]  = rotl32(working[7], 12)
				working[2]  += working[7];  working[13] ~= working[2];  working[13] = rotl32(working[13], 8)
				working[8]  += working[13]; working[7]  ~= working[8];  working[7]  = rotl32(working[7], 7)

				working[3]  += working[4];  working[14] ~= working[3];  working[14] = rotl32(working[14], 16)
				working[9]  += working[14]; working[4]  ~= working[9];  working[4]  = rotl32(working[4], 12)
				working[3]  += working[4];  working[14] ~= working[3];  working[14] = rotl32(working[14], 8)
				working[9]  += working[14]; working[4]  ~= working[9];  working[4]  = rotl32(working[4], 7)
			}

			for i in 0..<16 {
				working[i] += state[i]
			}

			// Serialize to bytes (little-endian)
			for i in 0..<16 {
				w := working[i]
				r.buf[i*4 + 0] = byte(w)
				r.buf[i*4 + 1] = byte(w >> 8)
				r.buf[i*4 + 2] = byte(w >> 16)
				r.buf[i*4 + 3] = byte(w >> 24)
			}

			// Increment block counter
			r.counter += 1
			r.buf_pos = 0
		}

		// Serve 8 bytes from buffer
		if r.buf_pos > (64 - 8) {
			generate_block(r)
		}
		if r.buf_pos == 0 && !r.initialized {
			// Ensure initialized before first use; this branch will be superseded in .Read prologue
			generate_block(r)
		}
		i := int(r.buf_pos)
		v := u64(r.buf[i+0]) |
		     (u64(r.buf[i+1]) << 8) |
		     (u64(r.buf[i+2]) << 16) |
		     (u64(r.buf[i+3]) << 24) |
		     (u64(r.buf[i+4]) << 32) |
		     (u64(r.buf[i+5]) << 40) |
		     (u64(r.buf[i+6]) << 48) |
		     (u64(r.buf[i+7]) << 56)
		r.buf_pos += 8
		return v
	}

	@(thread_local)
	global_rand_seed: Default_Random_State

	init :: proc "contextless" (r: ^Default_Random_State, seed: u64) {
		// Use splitmix64 to derive key and nonce deterministically from the seed
		sm64_next :: proc "contextless" (s: ^u64) -> u64 {
			s^ += 0x9E3779B97F4A7C15
			z := s^
			z = (z ~ (z >> 30)) * 0xBF58476D1CE4E5B9
			z = (z ~ (z >> 27)) * 0x94D049BB133111EB
			return z ~ (z >> 31)
		}
		local_seed := seed
		for i in 0..<4 {
			v := sm64_next(&local_seed)
			r.key[i*2+0] = u32(v & 0xffffffff)
			r.key[i*2+1] = u32(v >> 32)
		}
		// Derive nonce (96-bit)
		vn := sm64_next(&local_seed)
		r.nonce[0] = u32(vn & 0xffffffff)
		r.nonce[1] = u32(vn >> 32)
		vn2 := sm64_next(&local_seed)
		r.nonce[2] = u32(vn2 & 0xffffffff)

		r.counter = 0
		r.buf_pos = 64 // force regeneration on first read
		r.initialized = true
	}

	r: ^Default_Random_State = ---
	if data == nil {
		r = &global_rand_seed
	} else {
		r = cast(^Default_Random_State)data
	}

	switch mode {
	case .Read:
		if !r.initialized {
			init(r, u64(intrinsics.read_cycle_counter()))
		}

		switch len(p) {
		case size_of(u64):
			// Fast path for a 64-bit destination.
			intrinsics.unaligned_store((^u64)(raw_data(p)), read_u64(r))
		case:
			// All other cases.
			pos := i8(0)
			val := u64(0)
			for &v in p {
				if pos == 0 {
					val = read_u64(r)
					pos = 8
				}
				v = byte(val)
				val >>= 8
				pos -= 1
			}
		}

	case .Reset:
		seed: u64 = 0
		mem_copy_non_overlapping(&seed, raw_data(p), min(size_of(seed), len(p)))
		init(r, seed)

	case .Query_Info:
		if len(p) != size_of(Random_Generator_Query_Info) {
			return
		}
		info := (^Random_Generator_Query_Info)(raw_data(p))
		info^ += {.Uniform, .Resettable}
	}
}
```
</details>

(3.1 - Optional)  Keep the old default as @Kelimion has proposed:
```odin
// Use the old default generator
pcg_state: rand.PCG_Random_State 
context.random_generator = rand.pcg_random(&pcg_state)
```

4. Open a new PR for your review

?

tests/issues/test_issue_5881.odin
bogwi: Actually it shouldn't.

- The test’s goal at L75 is to assert determinism ("same seed -> same first value") and then run distribution checks. A fixed literal seed is ideal here: it makes the test fully reproducible across machines and runs, minimizing CI flakiness.

- Using a dynamic per-run seed like t.seed wouldn’t strengthen the determinism assertion; both generators would still get the same seed, so the check is equivalent. But it would add run-to-run variance to the subsequent statistical checks, slightly increasing the chance of rare tail failures even with generous thresholds.

- For broader coverage, a better approach than t.seed is to iterate a small set of fixed seeds (e.g., 3–5 known constants). That preserves reproducibility while sampling different trajectories.

Kelimion:
Your call, @Yawning.

Yawning:
A few notes:
 * If we do use chacha8rand it should be initialized via the system cryptographic entropy source with 256-bits of high quality entropy.
 * There is a bit more to the construct than "just read from the keystream" (see: https://github.com/C2SP/C2SP/blob/main/chacha8rand.md for a description, rationale, and test vectors)
 * It really needs SIMD for good performance.
 * I would be ok with `xoshiro256` being the default and chacha8rand living in `core:crypto`, but with the 3 other things on my list, we would have a CSPRNG by default, which is "nice".

bogwi:
Cool. @Yawning 

Agree on:
- It should follow the C2SP chacha8rand construct:
- - 256-bit OS entropy seeding only,
- - compliance with the construction (not just keystream),
- - SIMD path (with a portable fallback),
- - conformance tests using published vectors.

That scope is independent of this PR and larger in surface area.

Now, if you want you can merge this PR to resolve #5881 with a solid non-crypto default.

In due time I can open a follow-up PR to implement core:crypto/chacha8rand (C2SP-compliant, OS-seeded, SIMD, vectors), and when that lands and perf looks good, we can wire it as the default in base:runtime and move xoshiro256 into `Odin/core/math/rand/`.

Yawning:
> Cool. @Yawning
>
> In due time I can open a follow-up PR to implement core:crypto/chacha8rand (C2SP-compliant, OS-seeded, SIMD, vectors), and when that lands and perf looks good, we can wire it as the default in base:runtime and move xoshiro256 into `Odin/core/math/rand/`.

Unless you plan on doing this within the next few days, I can probably just do this within a week or so.

bogwi:
> > Cool. @Yawning
> 
> >
> 
> > In due time I can open a follow-up PR to implement core:crypto/chacha8rand (C2SP-compliant, OS-seeded, SIMD, vectors), and when that lands and perf looks good, we can wire it as the default in base:runtime and move xoshiro256 into `Odin/core/math/rand/`.
> 
> 
> 
> Unless you plan on doing this within the next few days, I can probably just do this within a week or so.

No. I was planning to do it over the weekend, or better the weekend after.

I'll close the PR. Thanks for a good talk.

And thanks to @Kelimion for participating and @SOBEX for discovering the issue.

Yawning:
Just to follow up on this, I have a standalone implementation working (portable/128-bit simd), written in a way that will allow easy integration into `base`.  I need to add a special case to the 128-bit simd code for non-amd64 (more registers = more faster), but the current code is 2ns slower than the current default generator, which is "plenty fast".

https://github.com/Yawning/odin-chacha8rand

PR #5914, opened 2025-11-11
introduce nix flake

Yappaholic:
This PR introduces nix flake for usage in development and installation of Odin nightly. 

Some notable features include:

* Flake will track the GitHub source directly, so instead of waiting for package maintainers to update versions one can get Odin directly from the latest or pinned commit (convenient on NixOS, since downloading nightly binary might require ELF patching).
* Programs that rely on nightly Odin version (e.g. `ols`) can reference Odin repository as an input in their nix flakes.

Currently the only downside is that on every new commit users will need to compile Odin by themselves (but since the language is small in size, it can take a couple of minutes, depending on hardware).


gingerBill:
I will **NOT** accept this PR for a few reasons:

* It appears to be rely on numerous bodges to "correct" things within the Odin code base rather than actually correctly do it.
  * fix/improve the actual code, don't do fragile git patches which will need "fixing" later on
  * It's only applies "some fixes" to things like raylib and not any other packages. You are most likely to not maintain this especially when this is your first PR.
* I hate the entitlement that people do when trying to add build recipes to a source repo.
  * The old shell.nix was allowed because of how small it was. This is just giant.
  * The new 
  * Please stop trying to push your effective techno-religion. It doesn't solve any problems you actually think it does. And don't repeat the Nix meme that is fixes anything, it's not true: it just causes more problems in the long run. Get a better religion, please.

Yappaholic:
> I will **NOT** accept this PR for a few reasons:
> 
> * It appears to be rely on numerous bodges to "correct" things within the Odin code base rather than actually correctly do it.
>   
>   * fix/improve the actual code, don't do fragile git patches which will need "fixing" later on
>   * It's only applies "some fixes" to things like raylib and not any other packages. You are most likely to not maintain this especially when this is your first PR.
> * I hate the entitlement that people do when trying to add build recipes to a source repo.
>   
>   * The old shell.nix was allowed because of how small it was. This is just giant.
>   * The new
>   * Please stop trying to push your effective techno-religion. It doesn't solve any problems you actually think it does. And don't repeat the Nix meme that is fixes anything, it's not true: it just causes more problems in the long run. Get a better religion, please.

 I understand. Thank you for not merging this PR though, I did some stupid mistakes there, which would make  it harder to maintain and use

PR #5920, opened 2025-11-12
Clarify function with enum

RobinsAviary:
Heyo! Thanks for the great language and bindings :) I noticed that `rlgl` should use an enum for `Begin()` to match Odin customs better and be more clear to the end-user. Thanks!


gingerBill:
I wanted to keep to the raylib API as close as possible. An enum might make some sense here but it's actually just mapping to the OpenGL constants so it makes more sense to keep to it.

Also there are a few issues with your PR: the first of which is the `enum` is wrong. It should be `enum c.int` to specify the backing integer size, and second the formatting on the comments is not consistent with the rest of the file.

PR #5934, opened 2025-11-15
Fix out of bounds access when parsing end_pos

BradLewis:
We got an error reported in `ols`:
```sh
".../ols/ols"  "stderr"        ".../odin/core/odin/parser/parser.odin(103:14) Invalid slice indices 0:2 is out of range 0..<1\n"
```

My guess is we have a string with the tokenizer text only being 1 character, though I'm not really sure how that happens.

Edit: I changed it to just explicitly verify the length as I'm not even sure how it got into the state of having this error.

### DIFF
--- a/core/odin/parser/parser.odin
+++ b/core/odin/parser/parser.odin
@@ -99,19 +99,15 @@ end_pos :: proc(tok: tokenizer.Token) -> tokenizer.Pos {
 	pos := tok.pos
 	pos.offset += len(tok.text)
 
-	if tok.kind == .Comment || tok.kind == .String {
-		if tok.text[:2] == "/*" || tok.text[:1] == "`" {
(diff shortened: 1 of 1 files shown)


laytan:
I am guessing it happens on this?
```odin
foo :: ``
```
Hmm on second look, I think not

BradLewis:
That was my guess but it seems fine for me, and that should have length 2 for the two quote characters. I also tried with just
```odin
foo :: ` // also ' and "
```
but that seemed to be fine as well

laytan:
Probably an unterminated string (at the end of the file?) actually, where you do get a syntax error but we return an incomplete string token.

BradLewis:
Ah yeah I'm able to reproduce it with the very minimal program now

```odin
package main

foo :: "
```

It just doesn't cause any issues with `ols` but will crash when manually parsing the file.

PR #5935, opened 2025-11-16
Strengthen binary_search return properties.

dalchen:
Prior to this commit, the guarantee is that if the element does not exist, then a uniquely determined position is returned (which is where the element *would* go) according to the documentation and tests.

However, if the element exists, the documentation states that there is no uniquely determined position that gets returned.

1. We can make a stronger statement on the returned value when there are multiple matches given the existing implementation.

2. We can also make our statement more consistent where both cases of finding and not finding the desired element results in a uniquely determined return position.

The binary_search and binary_search_by implementations are basically equivalent to what `std::lower_bound` does in C++. We can examine the code for the case where the range `[left, right)` is looking at a range of equal numbers and see that we set `right = mid` which moves the `right)` interval towards the left (aka an earlier match). The loop terminates without `[left` ever moving and with `right)` shifting all the way to equal `[left`, which is at the first match.

I doubt changing the statement from returning anything within a range to a specific element would break a lot of existing code for a binary search implementation, but this change would result in clearer understanding of the behavior and more consistency within the documentation.

### DIFF
--- a/core/slice/slice.odin
+++ b/core/slice/slice.odin
@@ -255,7 +255,7 @@ Searches the given slice for the given element.
 If the slice is not sorted, the returned index is unspecified and meaningless.
 
 If the value is found then the returned int is the index of the matching element.
-If there are multiple matches, then any one of the matches could be returned.
+If there are multiple matches, then the returned int is the index of the first matching element.
 
 If the value is not found then the returned int is the index where a matching
 element could be inserted while maintaining sorted order.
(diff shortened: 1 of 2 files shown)


gingerBill:
What the heck is this PR? You've just changed a few comments and a few asserts and not actually changed anything about the code itself for this 'strengthening' property. How are you strengthening anything by changing a comment of all things?

PR #5941, opened 2025-11-18
Use SIP hash as name canonicalization hash

gingerBill:
Use SIP hash as name canonicalization hash. This replaces the current fnv64a hash.


Yawning:
This looks good to me.  There are other alternative hashes that are equally collision resistant but, this allows for extra tinfoil hat shenanigans as needed later.

PR #5942, opened 2025-11-18
Adding `Combination_Iterator` to 'core:slice'

antonio-petrillo:
I've added a `Combination_Iterator` to `core:slice`, it let the use generate all the combination of k elements from a slice of n elements.  
The implementation mimics the one in `Permutation_Iterator` also found in `core:slice`.

The API is slightly different from the `Permutation_Iterator` mainly because it need a slice to store the `combination`, on each call to `combine` the new combination is found in the `iter.combination` field, where `iter` is an instance of `Combination_Iterator`.  
An example on how to use the iterator can be found in the dedicated test.

The algorithm is based on "The Art of Computer Programming, Volume 4a", in particular the **Algorithm T** which generate the combinations in lexicographical order, trying to minimize the change between to consecutive combinations.

I hope it could be useful, if the implementation, documentation or anything else doesn't meet the standard for the language I'll be happy to improve them.

Have a nice day.

### DIFF
--- a/tests/core/slice/test_core_slice.odin
+++ b/tests/core/slice/test_core_slice.odin
@@ -295,8 +295,8 @@ test_combination_iterator :: proc(t: ^testing.T) {
             unique_combination[combination_hash] = true
 
             clear(&seen)
-            clear(&unique_combination)
         }
+        clear(&unique_combination)
         expected := n_choose_k(len(s), k)
 
         // Using 'k' there are exactly 'n! / (k! * (n - k)!)'


gingerBill:
As silly as this sounds but who is this even for? What problem does it actually solve that you think it needs to be in the core library?

Other than the lack of care taken to keeping with the Odin coding convention too (using spaces rather than tabs, weird error handling, etc), I don't think this is good enough for the core library to be accepted.

antonio-petrillo:
Sorry, I didn't meant to waste your time, I had thought it could've been a good companion to the `Permutation_Iterator` already present in the `core:slice` library, regarding the weird error handling and conventions I've no excuses, by bad.

PR #5951, opened 2025-11-22
Increase base64 decoding table size to 256, preventing out of bounds reads

Znarf64:
Previously this could have been used to leak up to 128 * size_of(int) = 1KB of data from the data section when the input string contained non-ascii characters, given that bounds checks are disabled here for performance reasons. The bigger table fixes this without introducing overhead from checking and makes the base64 package resemble the base32 package more closely.


Znarf64:
Actually the exploit is somewhat less potent than I thought, it is possible to leak some data but only about every one out of every eight bytes following the decoding table in memory, still it should be fixed

PR #5954, opened 2025-11-22
Added libretro API bindings to vendor

andreas-jonsson:
_"The Libretro API is a lightweight C-based Application Programming Interface (API) that exposes generic audio, video, and input callbacks."_

There are two sides to Libretro development:

- **frontends** (such as [RetroArch](https://www.retroarch.com/)) are programs that can run libretro-compatible cores.
- **cores** are program (such as a game, [emulator](https://docs.libretro.com/development/libretro-overview/#), or media player) that has been ported to the Libretro API so that it can be executed by Libretro frontends.

---

This is a generated [binding](https://github.com/karl-zylinski/odin-c-bindgen/) for the [Libretro API](https://docs.libretro.com/development/libretro-overview/).
These bindings has been tested in an x86 emulator that has a fair amount of API coverage. Since this is just a API definition it does not contain any libraries and works on all platforms. (including freestanding)

### DIFF
--- a/vendor/libretro/LICENSE
+++ b/vendor/libretro/LICENSE
@@ -1,9 +1,0 @@
-MIT License
-
-Copyright (c) 2025 phix
-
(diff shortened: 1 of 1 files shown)


andreas-jonsson:
I don't think the failing OSX test has anything to do with this code..?

Kelimion:
> I don't think the failing OSX test has anything to do with this code..?

It doesn't. macOS Intel tests fail in general at the moment. It's something to look into.

gingerBill:
I don't think this is a good idea for the `vendor` library collection. Cool library though.

PR #5956, opened 2025-11-22
add math.sign and math.sign_bit overloads for int types

thePHTest:


core/math/math.odin
```diff
@@ -468,11 +477,25 @@ gain :: proc "contextless" (t, g: $T) -> T where intrinsics.type_is_float(T) {
 @(require_results) sign_f64le :: proc "contextless" (x: f64le) -> f64le { return f64le(int(0 < x) - int(x < 0)) }
 @(require_results) sign_f64be :: proc "contextless" (x: f64be) -> f64be { return f64be(int(0 < x) - int(x < 0)) }
 sign :: proc{
+	sign_int,
+	sign_i16, sign_i16le, sign_i16be,
+	sign_i32, sign_i32le, sign_i32be,
+	sign_i64, sign_i64le, sign_i64be,
```
gingerBill, line 486: Move these to "end" of this list since the float ones are more likely to be hit. This is a "micro-optimization" but it'll help in the long run.

thePHTest: done. If you wanted to micro-optimize further, would it make even more sense to put all the non-endian specific types first? i.e. f16, f32, f64, int, i16, i32, i64, f16le, etc.

PR #5963, opened 2025-11-24
runtime: Use chacha8rand as the default RNG (BREAKING)

Yawning:
This switches the default random number generator to the [chacha8rand](https://github.com/C2SP/C2SP/blob/main/chacha8rand.md) CSPRNG seeded from system entropy.

This is a breaking change as the output for a given seed will be different than the previous PCG64 generator.  The `orca` target currently lacks a secure entropy source, so existing applications will need to explicitly initialize the RNG (or set it to one of the insecure `math/rand` generators).

Changes:
- Relocate the old PCG64 RXS-M-XS generator to `core:math/rand` (Backward compatibility)
- Add xoshiro256** + splitmix64 from #5907 to `core:math/rand`
- Updated the test runner/demo to use xoshiro256** (So they always run even without a system entropy source)
- Moved the underlying implementation of `core:crypto.rand_bytes` to `base:runtime`
- Test cases

Known issues:
- LLVM appears to attempt to use AVX-2 for the xorshiro256** function, which results in a performance regression when   a microarchitecture with AVX-2 support is specified, which leads to the amusing result where `-microarch:native` makes `chacha8rand` the faster of the two.
- As mentioned the `orca` target currently does not have a cryptographic entropy source.

Uplifting the CPUID code to runtime would allow for dynamic detection of AVX-2, but the non-bulk performance difference is not significant, so this is defered till needed.

### DIFF
--- a/tests/benchmark/all.odin
+++ b/tests/benchmark/all.odin
@@ -3,5 +3,6 @@ package benchmarks
 @(require) import "bytes"
 @(require) import "crypto"
 @(require) import "hash"
+@(require) import "math"
 @(require) import "text/regex"
 @(require) import "strings"
--- a/tests/benchmark/math/benchmark_rand.odin
+++ b/tests/benchmark/math/benchmark_rand.odin
@@ -1,0 +1,130 @@
+package benchmark_core_math
+
+import "base:runtime"
+
+import "core:fmt"
+import "core:math/rand"
+import "core:log"
+import "core:strings"
+import "core:testing"
+import "core:text/table"
+import "core:time"
+
+@(private = "file")
+ITERS :: 10000000
+@(private = "file")
+ITERS_BULK :: 1000
+
+@(private = "file")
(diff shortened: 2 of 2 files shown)


Yawning:
> One thing we may need to consider though, is deprecating the random reset/seed with `u64` procs. Since you indicate it SHOULD get 32 bytes.

I thought about this and am in favor of keeping them around for a few reasons:
- PCG uses a 64-bit seed (backward compatibility)
- Our xorshiro256** + splitmix64 also takes a 64-bit seed (though this could be altered to allow for 256-bit seeds as well).
- chacha8rand degrades gracefully (when seeded with a 64-bit seed with zero padding), the security is "guess the seed"

Separately from this, separating the notion of the seed and the RNG state (for checkpoint/restore purposes) can be done.  The space required for each generator would be:

- PCG -> 128-bits
- xorshiro256** -> 256-bits
- chacha8rand -> 264-bits

But checkpoint/restore in my opinion is orthogonal to initialization, and the default chacha8rand behavior of "use system entropy" should be sufficient for the majority of use-cases.

PR #5972, opened 2025-11-28
`math/rand`: Add range-based number generation procedures

Xotchkass:
Added:
- `uint32_range :: proc(lo, hi: u32, gen := context.random_generator) -> (val: u32)`
- `uint64_range :: proc(lo, hi: u64, gen := context.random_generator) -> (val: u64)`
- `uint128_range :: proc(lo, hi: u128, gen := context.random_generator) -> (val: u128)`
- `uint_range :: proc(lo, hi: uint, gen := context.random_generator) -> (val: uint)`
- `int32_range :: proc(lo, hi: i32, gen := context.random_generator) -> (val: i32)`
- `int64_range :: proc(lo, hi: i64, gen := context.random_generator) -> (val: i64)`
- `int128_range :: proc(lo, hi: i128, gen := context.random_generator) -> (val: i128)`
- `int_range :: proc(lo, hi: int, gen := context.random_generator) -> (val: int)`

### DIFF
--- a/core/math/rand/rand.odin
+++ b/core/math/rand/rand.odin
@@ -546,9 +546,7 @@ Possible Output:
 
 */
 uint32_range :: proc(lo, hi: u32, gen := context.random_generator) -> (val: u32) {
-	if lo >= hi {
-		panic("Invalid arguments to uint32_range: lo must be less than hi")
-	}
+	assert(lo < hi, "Invalid arguments to uint32_range: lo must be less than hi")
 	range := hi - lo
 	if (range & (range - 1)) == 0 {
 		return lo + (uint32(gen) & (range - 1))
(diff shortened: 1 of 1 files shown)


core/math/rand/rand.odin
```diff
@@ -519,6 +519,334 @@ uint_max :: proc(n: uint, gen := context.random_generator) -> (val: uint) {
...
+Possible Output:
+
+	6
+	13
+
+*/
+uint32_range :: proc(lo, hi: u32, gen := context.random_generator) -> (val: u32) {
+	if lo >= hi {
+		panic("Invalid arguments to uint32_range: lo must be less than hi")
+	}
```
Kelimion: This could be `assert(lo < hi, "Invalid arguments to uint32_range: lo must be less than hi")`.

Xotchkass: Sure, it was mostly to match `*_max` procs

Kelimion: They're a bit consistent as-is. `float64_range` for example also uses `assert(low <= high, "msg")`.
We should probably perform a consistency pass on all of these at some point.

PR #5974, opened 2025-11-28
Fix skipping newlines on lines with backslash

krnowak:
…and a drive-by typo fix.

I noticed that the code like this compiles, and I'm not sure if it should:

```
package p

// Use backslash to avoid magical semicolon insertion
// and pick up the f32 for matrix element type.
M_OK :: matrix[2,3] \
f32

// Same, but over many empty lines.
M_OK2 :: matrix[2,3] \
\
\
f32

// Same again, but with a single backslash.
M_QUESTIONABLE :: matrix[2,3] \


f32

main :: proc() {}
```

I think that the `M_QUESTIONABLE` case should be a compiler error, but maybe I'm missing something. And yes, it's a contrived case. I spotted it when figuring out the semicolon insertion for my tree-sitter grammar for odin.

### DIFF
--- a/src/tokenizer.cpp
+++ b/src/tokenizer.cpp
@@ -309,7 +309,7 @@ struct Tokenizer {
 	i32 error_count;
 
 	bool insert_semicolon;
-	
+	bool skip_one_newline;
 	LoadedFile loaded_file;
 };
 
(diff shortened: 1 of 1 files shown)


gingerBill:
This isn't a bug. `\` just consumes a single newline, it's not meant to behave like in C.

krnowak:
> This isn't a bug. `\` just consumes a single newline, it's not meant to behave like in C.

It doesn't consume a single newline, but all of them if the newlines are only separated by whitespace. Which is what the `M_QUESTIONABLE` type shows - there are three newlines between `]` and `f32` and all of them are consumed by a single backslash. I tried to fix the tokenizer, so the backlash actually consumes one newline.

I don't understand a comment about the C behavior.

PR #5979, opened 2025-11-30
Fix #5978: make choice_bit_set respect bit_set domain

### LINKED ISSUE #5978: math/rand.choice_bit_set returns wrong values if the set starts from a number different from 0
## Context

`choice_bit_set` from `math/rand` package returns wrong results if the provided bit_set starts from any number except zero. It seems that the resulting value is adjusted to the difference between starting number and zero.

```odin
s := bit_set[1 ..= 5]{1, 5}

for _ in 0 ..= 100 {
	fmt.println(rand.choice_bit_set(s))

}
```

That code outputs 100 times either `0 true` (which is out of the bit_set range) or `4 true` (which is not in the actual values list).
If you change bit_set to `[2 ..= 5]{2,5}`, then it outputs 0 and 3. 
If you change it to `[-3 .. = 5]{1,5}`, then it outputs 4 and 8.

It seems like the `[final value] = [actual_value] - [bit_set_start]`

### Operating System & Odin Version:
Odin:    dev-2025-11:e5153a937
OS:      EndeavourOS, Linux 6.17.9-arch1-1
CPU:     AMD Ryzen 7 7700 8-Core Processor              
RAM:     31696 MiB
Backend: LLVM 21.1.5

## Expected Behavior

`rand.choice_bit_set(bit_set[1..=5]{1,5})` should return either 1 or 5.
`rand.choice_bit_set(bit_set[2..=5]{2,5})` should return either 2 or 5.
`rand.choice_bit_set(bit_set[-3..=5]{1,5})` should return either 1 or 5.

## Current Behavior

`rand.choice_bit_set(bit_set[1..=5]{1,5})` returns either 0 or 4.
`rand.choice_bit_set(bit_set[2..=5]{2,5})` returns either 0 or 3.
`rand.choice_bit_set(bit_set[-3..=5]{1,5})` returns either 4 or 8.

astenmark:
Fixes #5978.

Previously, choice_bit_set treated the underlying bit index as the element value, which is incorrect for bit_sets whose domain does not start at 0 (e.g. bit_set[1..=5] or bit_set[-3..=5]). This change iterates over the values in the set, and returns the N:th element, where N is the generated random choice index. It uses similar semantics as previously, and the random properties should still be the same.

A new test has been added in tests/core/math/rand/test_core_math_rand.odin, covering the case reported in the issue.

### DIFF
--- a/tests/core/math/rand/test_core_math_rand.odin
+++ b/tests/core/math/rand/test_core_math_rand.odin
@@ -87,11 +87,11 @@ test_issue_5978 :: proc(t:^testing.T) {
 	cases := []struct {
 		seed: u64,
 		expected: int,
-	}{ {13, 1}, {27, 5} };
+	}{ {13, 1}, {27, 5} }
 
 	for c in cases {
-		rand.reset(c.seed);
-		i, _ := rand.choice_bit_set(s);
+		rand.reset(c.seed)
+		i, _ := rand.choice_bit_set(s)
 		testing.expectf(t, i == c.expected, "choice_bit_set returned %v with seed %v, expected %v", i, c.seed, c.expected)
 	}
 }


Kelimion:
Fix looks good, and the added test is a nice addition. If you can remove those stray semicolons, I think we can merge this.

Does `test_issue_5978` take into account that the default RNG was changed to chacha8 the other day? If it relies on the previous PCG generator, the seeds and corresponding expected output won't pass your new test.

astenmark:
Thanks, I've removed the semicolons. The test only verifies exactly what was in the issue description. It's not a full test of choice_bit_set. I did the fix a cpl hours ago, so I believe it does take chacha8 into account.

Kelimion:
Looks like it does, as it passes your test. Thanks for the fix.

PR #5985, opened 2025-12-02
Fix #5957 by using single quotes for run_args

### LINKED ISSUE #5957: Posix Run Arguments Not Passed Directly to the Application
## Context
latest Odin `master` commit. this has persisted for quite some time

## Issue Description

given the following program:
```odin
package demo

import "core:fmt"
import os "core:os/os2"

main :: proc() {
	fmt.println(os.args)
}
```
`odin run . -- 'print("foo")'` should output `["executable_path", "print("\foo\")"]`
(printing `os.args[1]` outputs `print("foo")`)

this is seemingly caused by run args on posix being processed with `run_args_string = string_join_and_quote(heap_allocator(), run_args);`. ideally, there would be zero processing done by the compiler here and the run args would be completely transparently passed to the running application

Creativty:


A1029384756:
can you check if this works in the opposite case (`"print('foo')"` being a run argument)? if i remember correctly from my testing, making this change broke that case. what probably needs to be done is to have the args passed through directly with none of the sanitation the compiler does. this will likely involve changing the function that runs the output binary from a variadic function to one that takes an array

#5957 (linking to issue)

Creativty:
Can you give a more detailed description of said opposite case?

> can you check if this works in the opposite case? in my testing, making this change broke the inverse case iirc

src/main.cpp
```diff
@@ -4059,7 +4059,7 @@ end_of_code_gen:;
 		String exe_name = path_to_string(heap_allocator(), build_context.build_paths[BuildPath_Output]);
 		defer (gb_free(heap_allocator(), exe_name.text));
 
-		system_must_exec_command_line_app("odin run", "\"%.*s\" %.*s", LIT(exe_name), LIT(run_args_string));
+		system_must_exec_command_line_app("odin run", "\'%.*s\' %.*s", LIT(exe_name), LIT(run_args_string));
```
A1029384756, line 4062: these quoting changes will have no effect since that's just quoting the exe, not the run arguments

Creativty:
> this will likely involve changing the function that runs the output binary from a variadic function to one that takes an array

I agree on principle, but maybe it can be solved with minimal changes to the current code as it is, without moving away from the variadic signature.
The way I see it, is that `system` is the main culprit, when an `exec` (+ a potential `fork`) is supposed to be the . But to solve that would require adding complex process management to the codebase.

src/main.cpp
Creativty: There have been cases where the executable contained a '$' in the name, and that ended up expanding in the subsequent `system` call, thus the need for single quote escaping the `exe_name`

Creativty:
I'm closing this, since I don't deem it to be a fitting solution given the problem situation.

PR #5987, opened 2025-12-03
os2: Use libc for linux environment if not no-crt.

### LINKED ISSUE #5873: `core:os/os2` cannot get environment variables in a shared library on linux
## Context
I'm using Gentoo Linux and the latest commit from the master branch.

```
        Odin:    dev-2025-11:91d923bd8
        OS:      Gentoo Linux, Linux 6.12.47-gentoo-dist
        RAM:     31964 MiB
        Backend: LLVM 20.1.8
```

## Expected Behavior

A shared odin library should be able to access the environment variables at runtime.

## Current Behavior

When trying to access any environment variable with os2, the program crashes on a assertion in `env_linux.odin::_get_original_env` because it tries to get the environment variables from `runtime::args__` which never gets initialized for a shared library.

## Steps to Reproduce

1. Use a Linux system
2. Copy the code to reproduce the issue provided below into two packages.
3. `odin build shared_library -build-mode:dll -debug`
4. `odin build loader -debug`
5. Run `loader.bin` in a debugger and see the assertion failure at `_get_original_env`

### Code to reproduce

`loader/loader.odin`
```odin
package loader

import "base:runtime"

import "core:fmt"
import "core:dynlib"

Symbols :: struct {
    __handle: dynlib.Library,
    lookup_shell: proc() -> string,
}

main :: proc() {
    syms: Symbols
    if _, ok := dynlib.initialize_symbols(&syms, "shared_library." + dynlib.LIBRARY_FILE_EXTENSION); !ok { panic("Could not load library.") }

    fmt.println(runtime.args__)
    fmt.println(syms.lookup_shell())
}
```

`shared_library/shared_library.odin`
```odin
package shared_library

import "base:runtime"

import "core:fmt"

import os "core:os/os2"

@export
lookup_shell :: proc() -> string {
    fmt.println(runtime.args__)
    return os.get_env_alloc("SHELL", context.allocator)
}
```

### Failure Logs

```
["/home/<...>/loader.bin"]
[]

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff78c1e74 in os2::[env_linux.odin]::_get_original_env () at /usr/lib/odin/core/os/os2/env_linux.odin:227
227             assert(env[0] == nil)
```

## Workaround
I currently just pass a correct `runtime::args__` to the library which then initializes its own `runtime::args__` with that instead.

## Potential Solution
The environment variables could be taken from the `environ` global variable, but I don't think that that would work with `-no-crt`.

jasonKercher:
I basically copied and pasted posix_env and put it behind a big `when` statement.  This fixes #5873,

### DIFF
--- a/core/os/os2/env_linux.odin
+++ b/core/os/os2/env_linux.odin
@@ -7,242 +7,371 @@ import "base:intrinsics"
 import "core:sync"
 import "core:slice"
 import "core:strings"
+import "core:sys/linux"
 
-// TODO: IF NO_CRT:
-//         Override the libc environment functions' weak linkage to
-//         allow us to interact with 3rd party code that DOES link
-//         to libc. Otherwise, our environment can be out of sync.
-//       ELSE:
-//         Just use the libc.
+_ :: sync
+_ :: slice
+_ :: linux
 
-NOT_FOUND :: -1
+when ODIN_NO_CRT {
+	// TODO: Override the libc environment functions' weak linkage to
+	//       allow us to interact with 3rd party code that DOES link
+	//       to libc. Otherwise, our environment can be out of sync.
 
-// the environment is a 0 delimited list of <key>=<value> strings
-_env: [dynamic]string
+	NOT_FOUND :: -1
 
-_env_mutex: sync.Recursive_Mutex
+	// the environment is a 0 delimited list of <key>=<value> strings
+	_env: [dynamic]string
 
-// We need to be able to figure out if the environment variable
-// is contained in the original environment or not. This also
-// serves as a flag to determine if we have built _env.
-_org_env_begin: uintptr // atomic
-_org_env_end:   uintptr // guarded by _env_mutex
+	_env_mutex: sync.Recursive_Mutex
 
-// Returns value + index location into _env
-// or -1 if not found
-_lookup :: proc(key: string) -> (value: string, idx: int) {
-	sync.guard(&_env_mutex)
+	// We need to be able to figure out if the environment variable
+	// is contained in the original environment or not. This also
+	// serves as a flag to determine if we have built _env.
+	_org_env_begin: uintptr // atomic
+	_org_env_end:   uintptr // guarded by _env_mutex
 
-	for entry, i in _env {
-		if k, v := _kv_from_entry(entry); k == key {
-			return v, i
-		}
-	}
-	return "", -1
-}
+	// Returns value + index location into _env
+	// or -1 if not found
+	_lookup :: proc(key: string) -> (value: string, idx: int) {
+		sync.guard(&_env_mutex)
 
-_lookup_env_alloc :: proc(key: string, allocator: runtime.Allocator) -> (value: string, found: bool) {
-	if intrinsics.atomic_load_explicit(&_org_env_begin, .Acquire) == 0 {
-		_build_env()
-	}
-
-	if v, idx := _lookup(key); idx != -1 {
-		found = true
-		value, _ = clone_string(v, allocator)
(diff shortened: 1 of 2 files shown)


core/os/os2/env_linux.odin
```diff
@@ -7,242 +7,371 @@ import "base:intrinsics"
...
+		addr := uintptr(raw_data(env_data))
+		return addr >= intrinsics.atomic_load_explicit(&_org_env_begin, .Acquire) && addr < _org_env_end
 	}
-	sync.guard(&_env_mutex)
 
-	// NOTE: already terminated by nil pointer via + 1
-	env := make([]cstring, len(_env) + 1, allocator)
+} else {
+	// We are linking with libc, so use libc env functions
+	foreign import libc "system:c"
```
laytan, line 255: Why not use `core:sys/posix` here?

core/os/os2/env_posix.odin
```diff
@@ -80,7 +80,7 @@ _unset_env :: proc(key: string) -> (ok: bool) {
 // NOTE(laytan): clearing the env is weird, why would you ever do that?
 
 _clear_env :: proc() {
-	for i, entry := 0, posix.environ[0]; entry != nil; i, entry = i+1, posix.environ[i] {
+	for entry := libc_environ[0]; entry != nil; entry = libc_environ[0] {
```
laytan, line 83: Does this compile? Where is `libc_environ` coming from?

core/os/os2/env_linux.odin
jasonKercher: That would be ideal.  Last time I tried to do something like that with no-crt, it would try to link things in libc, and we cannot conditionally import on that.  That could have been a bug, or it might be unavoidable.  It's been about a year since then, and I forget the details.

core/os/os2/env_posix.odin
jasonKercher: It's in the foreign block.

laytan: I only see a foreign block in the Linux file, am I missing something?

core/os/os2/env_linux.odin
laytan: So the way it is supposed to work is that if you don't use anything out of the package it would not link libc. So if all usages are in the when block you have here it should be fine.

Kelimion: Right.
`#+build !no-crt` gives me:
```
Syntax Error: Invalid build tag platform: no 
	#+build !no-crt
```
It would be great if `#+build` could switch on any of the `ODIN_`, including `ODIN_NO_CRT`. @gingerBill?

core/os/os2/env_posix.odin
jasonKercher: Oh shit, you're right!  That's a copy paste error.  I wasn't looking at the file name.

Kelimion: If you could make a patch, that'd be great.

jasonKercher: Yes, and I can switch it to use `sys/posix`.

@laytan I just tried it with `sys/posix`, and it worked just fine.  Maybe that was a bug that has been fixed.

core/os/os2/env_linux.odin
jasonKercher: ^ This would be ideal here to get rid of the big ugly `when`.

core/os/os2/env_posix.odin
laytan: Great!

PR #5993, opened 2025-12-04
Adding missing build flags in jpeg loader to avoid load proc collision

jimhub:
This brings the jpeg loader in-line with the other core:image/* loaders, resolving the following compilation error when targeting js/wasm:

> /Odin/core/image/jpeg/jpeg_os.odin(5:1) Error: Redeclaration of 'load' in this scope 
        at /Odin/core/image/jpeg/jpeg_js.odin(3:1) 
        load :: proc{load_from_file, load_from_bytes, load_from_context}

### DIFF
--- a/core/image/jpeg/jpeg_js.odin
+++ b/core/image/jpeg/jpeg_js.odin
@@ -1,4 +1,3 @@
-#+build js
 package jpeg
 
 load :: proc{load_from_bytes, load_from_context}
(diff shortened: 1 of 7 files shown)


core/image/jpeg/jpeg_js.odin
```diff
@@ -1,3 +1,4 @@
+#+build js
```
gingerBill: This one isn't needed since there is already a `_js.odin` suffix.

jimhub: This is true. I only added it because the others include it, and I wasn't sure if we were erring on the side of explicitness. I'm happy to remove it from here, as well as from the others for the sake of consistency.

PR #5995, opened 2025-12-05
(SDL2) Use multi-pointer so you can pass array

RobinsAviary:
It looks like this field is supposed to be a multi-pointer, since that's how you can pass multiple buttons.


RobinsAviary:
Looking at the SDL3 bindings, it's definitely supposed to be a multi-pointer :) I hope this update will be helpful to people still using SDL2 (as SDL3 doesn't play nice with raylib yet...)

PR #5999, opened 2025-12-07
Fix `core:unicode/utf8.decode_grapheme_iterate` returning wrong text

blob1807:
The current behavior uses the monospace width of a grapheme, which isn't the same as the data width.

This PR also fixes a typo & adds documentation to `Grapheme.width`, to help people not assume it's the data width. Though it should be renamed to something more explicate like `monospace_width` or `cell_count`, but I couldn't decided which is better.

Example of current behavior:
```odin
package __test_bug
 
import "core:fmt"
import "core:unicode/utf8"

main :: proc() { 
    base := "子猫"
    iter := utf8.decode_grapheme_iterator_make(base)
    for str, g in utf8.decode_grapheme_iterate(&iter) {
        fmt.println(str, transmute([]u8)str, g)
    }
}
```
This should give:
```
子 [229, 173, 144] Grapheme{byte_index = 0, rune_index = 0, width = 2}
猫 [231, 140, 171] Grapheme{byte_index = 3, rune_index = 1, width = 2}
```
but instead gives (output of `str` may be nothing depending on your term):
```
σ¡ [229, 173] Grapheme{byte_index = 0, rune_index = 0, width = 2}
τî [231, 140] Grapheme{byte_index = 3, rune_index = 1, width = 2}
```

### DIFF
--- a/core/unicode/utf8/grapheme.odin
+++ b/core/unicode/utf8/grapheme.odin
@@ -23,7 +23,7 @@ normalized_east_asian_width       :: unicode.normalized_east_asian_width
 Grapheme :: struct {
 	byte_index: int,
 	rune_index: int,
-	width:      int,
+	width:      int, // The width of the Grapheme in number of monospace cells
 }
 
 
@@ -40,7 +40,7 @@ Grapheme_Iterator :: struct {
 
 	grapheme_count: int, // The number of graphemes in the string
 	rune_count:     int, // The number of runes in the string
-	width:          int, // The widrth of the string in number of monospace cells
+	width:          int, // The width of the string in number of monospace cells
 
 	last_rune:                  rune,
 	last_rune_breaks_forward:   bool,
@@ -152,7 +152,7 @@ decode_grapheme_iterate :: proc(it: ^Grapheme_Iterator) -> (text: string, graphe
 					it.rune_count,
 					it.width - it.last_width,
 				}
-				text = it.str[byte_index:][:grapheme.width]
+				text = it.str[byte_index:it.curr_offset]
 				ok = true
 
 


core/unicode/utf8/grapheme.odin
```diff
@@ -23,7 +23,7 @@ normalized_east_asian_width       :: unicode.normalized_east_asian_width
 Grapheme :: struct {
 	byte_index: int,
 	rune_index: int,
-	width:      int,
+	width:      int, // The width of the Grapheme in number of monospace cells
```
Kelimion, line 26: I'd go with `cell_count: int`.

blob1807: Should I also do the same for the other uses of `width` or are those clear enough?

Kelimion:
Good find. Can you also add your example test case to `tests/core/unicode/test_core_unicode.odin`?

It has a number of string -> expected grapheme tests already, but it appears their "width" and "width" must've been the same for this not to be caught.

core/unicode/utf8/grapheme.odin
Kelimion: On second thought, let's stick with `width` with the added comment. It's clear with that added context, and with that clarification, there is no need to break existing code.

So let's just add the comment to `Grapheme.width` and fix `Grapheme_Iterator`'s typo for the same.

blob1807:
Sure I can do that.

Kelimion:
Thanks!

blob1807:
I'm going to close this PR, as I assumed `text` was meant to be the raw text of the grapheme. 
However during making the test, I realized the algorithm doesn't fully decode the current grapheme. Instead it decodes the last one & returns the start of the next. 
So it could never be the raw text of the grapheme, but the just start of the current one. Which is weird to me.

I just lucked out in the example I gave.

Kelimion:
Understood. And now we know. :-)

Issue #5857, opened 2025-10-27
Warning for hardcoded Timezone Database path on Unix Systems


TESTMECS:
## Context
Hi, I was working with the "core:time/timezone" package from the standard library on NixOS.

Procedure:
```Odin
region_load :: proc(reg: string, allocator := context.allocator) -> (out_reg: ^datetime.TZ_Region, ok: bool) {…}
```
Looks up the Timezone database at a hard-coded path `/usr/share/zoneinfo`. 

However on NixOS, my database is at `/etc/zoneinfo`, so I got confused why the Procedure worked with the string "UTC" but not "America/New_York"

## Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System
[NixOS WSL](https://github.com/TESTMECS/nixconfig/blob/trunk/hosts/testme.nix) 

system.version="24.11", 

time.timeZone = "America/New_York"

wsl.enable=true

## Odin Version:
odin version dev-2025-09

## Please paste `odin report` output:
Useful information to add to a bug report:
        Odin:    dev-2025-09
        OS:      NixOS 25.11 (Xantusia), Linux 6.6.87.2-microsoft-standard-WSL2
        CPU:     AMD Ryzen 5 5600X 6-Core Processor
        RAM:     15955 MiB
        Backend: LLVM 18.1.8
## Expected Behavior.

- Timezone database is looked up or a warning for "Timezone database not found at `usr/share/zoneinfo`

### Please describe the behavior you are expecting.

- A warning that the db file is not found or a reason why it is found for "UTC" but not "America/New_York" without having to go to the source to see the lookup path. 

## Current Behavior

What is the current behavior?

`timezone.region_load(tz:string)` looks up the timezone database at `/usr/share/zoneinfo`, and will fail if that string is not "UTC".

in Odin/core/time/timezone/tz_unix.odin

```odin
        db_path := "/usr/share/zoneinfo"
	region_path := filepath.join({db_path, reg_str}, allocator)
	defer delete(region_path, allocator)

	return load_tzif_file(region_path, reg_str, allocator)
```

## Proposal
```odin
db_path := "/usr/share/zoneinfo"

	if !os.exists(db_path)  {
			/// ..check for other paths here, modify `db_path` variable
			if db_path == "/usr/share/zoneinfo" {
					fmt.eprintln("WARNING: Timezone database not found at /usr/share/zoneinfo")
[...]

PR #6012, opened 2025-12-12
[tz_unix] added additional search paths to match musl

A1029384756:
adds additional timezone search paths to bring things in line with musl [source link](https://git.musl-libc.org/cgit/musl/tree/src/time/__tz.c?id=0ccaf0572e9cccda2cced0f7ee659af4c1c6679a#n131). this should improve distribution support across the various linux/unix flavors


laytan:
Fixes #5857

Issue #6022, opened 2025-12-15
Weak linkage on Windows


Xaryen:
Logging this as an issue here as well.

Following the discord convo, It doesn't seem clear what exactly linkage="weak" should do on Windows.
(and if it should be allowed?)

But either way it's currently preventing glfw from working on Windows

The following will either fail to compile or give a linker error, it seems a bit inconsistent.
In the project that was giving the linker error it also needed to compile with address sanitizer, but when
I went to make the repro it straight up doesn't compile regardless of flags.

```odin
package repro

import "core:fmt"
import "vendor:glfw"

main :: proc() {

    fmt.println(glfw.GetPlatform())
}
```
Example of the linker error:
`odin-daxa-main-22f33519ef0.obj : fatal error LNK1227: conflicting weak extern definition for 'glfwGetPlatform'.  New default '.weak.glfwGetPlatform.default.main::result' conflicts with old default '.weak.glfwGetPlatform.default' in odin-daxa-glfw_bindings-22f3351dcd0.obj`

```
Odin:    dev-2025-12:7c4b8a9af 
OS:      Windows 11 Professional (version: 23H2), build 22631.5624 
CPU:     AMD Ryzen 9 7950X 16-Core Processor 
RAM:     98006 MiB 
Backend: LLVM 20.1.0
```

PR #6023, opened 2025-12-15
glfw bindings - remove weak linkage on windows

Xaryen:
Proposed (temporary?) fix for the problem mentioned in #6022


gingerBill:
I won't accept this as it needs a better approach. Either ignore `weak` on Windows or allow for `""` to be the default, which means you can do `@(linkage = "" when ODIN_OS == .Windows else "weak")`

Issue #6018, opened 2025-12-14
[`base:runtime`]: `print_i64` uses an OOB index when `min(i64)` is given.


blob1807:
## Context
```
Odin:    dev-2025-12:b277d66e0
OS:      Windows 11 Home Basic (version: 24H2), build 26100.7171
CPU:     AMD Ryzen 7 6800HS with Radeon Graphics
RAM:     15608 MiB
Backend: LLVM 20.1.0
```

## Expected Behavior

The logic shouldn't give an OOB index.

## Current Behavior

When `runtime.print_i64` is given `min(i64)`. It'll index `_INTEGER_DIGITS_VAR` by `18446744073709551608` (`-8`). 
It's a result of `abs(min(i64)) == min(i64)`. Which breaks the modulo logic. Giving the massive index, `min(i64) % 10 == -8`.
On Windows at least, this doesn't cause the OS to give an OOB.

I came across this though messing about & manually set a slice's length, and getting `-⌡`, `-i` & `-e` as the length.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Copy Paste
2. Run
3. If your OS doesn't OOB you
4. Remove the `#no_bounds_check` from `print_i64`
5. Run again
6. Get an `odin/base/runtime/print.odin(214:37) Index -8 is out of range 0..<36`

```odin
package print_i64_oob

import "base:runtime"
import "core:fmt"

main :: proc() { 
    v := min(i64)
    fmt.println(v) // sanity check
    runtime.print_i64(v)
    runtime.print_byte('\n')
}
```

PR #6024, opened 2025-12-15
[`base:runtime`] fix `print_i64` using an OOB index when `min(i64)` is given.

blob1807:
Fixes https://github.com/odin-lang/Odin/issues/6018 by wrapping the `abs` in a cast. As by itself `abs(min(T)) == min(T)`. This also mirrors how `core:strconv` does it. 
See linked issue for more information.

This PR also does some minor formatting to `print_u64` to keep it consistent with `print_i64`.

### DIFF
--- a/base/runtime/print.odin
+++ b/base/runtime/print.odin
@@ -184,10 +184,11 @@ print_rune :: #force_no_inline proc "contextless" (r: rune) -> int #no_bounds_ch
 
 
 print_u64 :: #force_no_inline proc "contextless" (x: u64) #no_bounds_check {
+	b :: u64(10)
+	u := x
+
 	a: [129]byte
 	i := len(a)
-	b := u64(10)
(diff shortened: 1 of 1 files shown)


base/runtime/print.odin
```diff
@@ -184,10 +184,11 @@ print_rune :: #force_no_inline proc "contextless" (r: rune) -> int #no_bounds_ch
 
 
 print_u64 :: #force_no_inline proc "contextless" (x: u64) #no_bounds_check {
+	b :: u64(10)
+	u := x
```
gingerBill, line 188: Any reason this was changed too?

blob1807: As already stated, To a have consistent look between the two is all. There's no reason beyond that.
I do understand random formatting changes are weird to do. So I can revert it if you wish.

PR #6030, opened 2025-12-18
Add `mem.buffer_from_slice_panicking` proc

Xotchkass:
Panicking on appending to a full buffer can often help to catch logical errors early. I set the parameter to `false` by default, so it wouldn't cause issues in existing code. I believe there shouldn't be any downsides to having it.


gingerBill:
Just make a new procedure instead of modifying an established one.

Xotchkass:
Ok

PR #6038, opened 2025-12-19
Fix darwin/Foundation/NSScreen type property bindings

JacobEvelyn:
This commit updates the `darwin/Foundation` package's `NSScreen` bindings to treat Objective-C "type properties" as class methods rather than instance methods.

As a result, we can now access these properties like this:

```odin
import NS "core:sys/darwin/Foundation"

main :: proc() {
  NS.Screen.mainScreen()
}
```

instead of this:

```odin
import NS "core:sys/darwin/Foundation"

main :: proc() {
  (^NS.Screen)(nil).mainScreen()
}
```

This commit also adds a binding for `NSScreen`'s `screensHaveSeparateSpaces` type property, which was previously missing from the `darwin/Foundation` package.

### DIFF
--- a/core/sys/darwin/Foundation/NSScreen.odin
+++ b/core/sys/darwin/Foundation/NSScreen.odin
@@ -3,17 +3,21 @@ package objc_Foundation
 @(objc_class="NSScreen")
 Screen :: struct {using _: Object}
 
-@(objc_type=Screen, objc_name="mainScreen")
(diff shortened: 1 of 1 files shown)


JacobEvelyn:
I've tested manually that each of these procs works as expected. Let me know if there are any tests or documentation that should be updated as well.

PR #6044, opened 2025-12-20
os/os2: fix stale errors on windows when reading from console

### LINKED ISSUE #6016: `core:os/os2`: read from stdin on windows returns last platform error even when the read was successful
## Context

```
        Odin:    dev-2025-12-nightly:ab9a05e
        OS:      Windows 11 Professional (version: 24H2), build 26100.3476
        CPU:     AMD Ryzen 5 5600X 6-Core Processor
        RAM:     32693 MiB
        Backend: LLVM 20.1.0
```

## Expected Behavior

Example program should print `2 nil`.

## Current Behavior

Example program prints `2 %!(BAD ENUM VALUE=536870912)`

### Steps to Reproduce

1. Compile and run the program
2. Press Enter without writing anything

```odin
package test

import "core:fmt"
import "core:os/os2"
import "core:sys/windows"

main :: proc() {
	buf: [256]u8
	windows.SetLastError(1<<29) // from docs: Bit 29 is reserved for application-defined error codes
	n, err := os2.read(os2.stdin, buf[:])
	fmt.println(n, err)
}
```

### Possible solution
It looks like `_read_internal` can unconditionally return right after reading from console?
https://github.com/odin-lang/Odin/blob/ab9a05ed2d2e0eb76ef53709f91c54534aee543a/core/os/os2/file_windows.odin#L372-L396

laytan:
Fixes #6016

### DIFF
--- a/core/os/os2/file_windows.odin
+++ b/core/os/os2/file_windows.odin
@@ -378,6 +378,7 @@ _read_internal :: proc(f: ^File_Impl, p: []byte) -> (n: i64, err: Error) {
 			if cerr != nil {
 				return i64(total_read), cerr
 			}
+			ok = true
 		} else {
 			ok = win32.ReadFile(handle, &p[total_read], to_read, &single_read_length, nil)
 		}


Neirokan:
This would lead to `err = .EOF` on every console read, no? Since read from console doesn't update `single_read_length`.

laytan:
Indeed! I tested this with read_entire_file which swallows the EOF error.

PR #6048, opened 2025-12-21
Make `using` as a statement opt-in with `#+feature using-stmt`

gingerBill:


dengzo:
Like this idea! I think its better for beginners not being able to do this by default.

PR #6053, opened 2025-12-22
[`core:testing`] Use Windows API for SIG handling

blob1807:
On Windows `libc.signal` can't handle `SIGSEGV`. Instead you have to set up an `ExceptionHandler/Filter`. 
This PR does that, along with using the appropriate endpoints for the other `SIG*`s. Excluding `SIGTERM` as it has no equivalent.

The runner was tested locally on the below system, using the CI's tests with no issues.
```
Odin:    dev-2025-12:c9fff456c
OS:      Windows 10 Home Basic (version: 22H2), build 19045.6466     
CPU:     AMD Ryzen 9 5900X 12-Core Processor
RAM:     32687 MiB
Backend: LLVM 20.1.0
```

I was unsure if we wanted to handle `CTRL_BREAK_EVENT` & `CTRL_CLOSE_EVENT` as well. So `stop_runner_callback` currently only handles `CTRL_C_EVENT`.

### DIFF
--- a/core/testing/signal_handler_windows.odin
+++ b/core/testing/signal_handler_windows.odin
@@ -165,7 +165,7 @@ _setup_signal_handler :: proc() {
 	// - Asserts and panics;
 	// - Arithmetic errors; and
 	// - Segmentation faults (illegal memory access).
-	win32.AddVectoredExceptionHandler(1, stop_test_callback)
+	win32.AddVectoredExceptionHandler(0, stop_test_callback)
 }
 
 _setup_task_signal_handler :: proc(test_index: int) {


blob1807:
I've tested locally on the below system, on the same tests with on issues.
```
Odin:    dev-2025-12:4eeb509fe
OS:      Windows 11 Home Basic (version: 24H2), build 26100.7462
CPU:     AMD Ryzen 7 6800HS with Radeon Graphics
RAM:     15608 MiB
Backend: LLVM 20.1.0
```

blob1807:
That last commit is having issues with ASAN. If I had to guess, it's setting up its own `ExceptionHandler` as the first. By making ours first, it's messing with it. Causing an Access Violation when trying to setup `context` in our handler, as far as I can tell.
So I've reverted it.

PR #6054, opened 2025-12-22
Add proc to check NSWindow occlusion state

antinaho:
This PR adds a new procedure to NSWindow to check its occlusion state.

Also added helper proc to see whether window is visible or not.


core/sys/darwin/Foundation/NSWindow.odin
```diff
@@ -109,6 +109,9 @@ WindowTabbingMode :: enum Integer {
 	Disallowed = 2,
 }
 
+WindowOcclusionStateVisibleBit :: 0x2
```
laytan: Can this be a bit_set?

PR #6057, opened 2025-12-23
Adds static linking to non-windows platforms close #3818

### LINKED ISSUE #3818: linker: implement `-build-mode:static` on non-windows targets
Should either:
- build an object file that is ready to put in a static library with `ar` and the likes
- or also actually call `ar` and put the one object file in it

Fabbboy:
The static linkage path on non-windows platforms now uses `ar` to "link" or rather archive the object files rather than crashing.


evaporei:
Thank you so much for this!!

Fabbboy:
> Thank you so much for this!!

You're welcome :)

PR #6069, opened 2025-12-26
WebGL binding additions

karl-zylinski:
These are the things I added to the webGL bindings to get the webGL version of my 2D game library "Karl2D" to work.

One pattern you see in this is that I have both "Alloc" and "Buf" variants of procs that need some amount of memory. I've made sure that the odin.js implementation can return the needed length when the buffer is not set. This way the "Alloc" version can first ask for the needed size, and then alloc the buffer and re-run the JS function.

### DIFF
--- a/core/sys/wasm/js/dom.odin
+++ b/core/sys/wasm/js/dom.odin
@@ -22,6 +22,8 @@ foreign dom_lib {
 	window_set_scroll :: proc(x, y: f64) ---
 
 	set_element_style :: proc(id: string, key: string, value: string) ---
+
+	set_document_title :: proc(title: string) ---
 }
 
 get_element_value_string :: proc "contextless" (id: string, buf: []byte) -> string {
--- a/core/sys/wasm/js/odin.js
+++ b/core/sys/wasm/js/odin.js
@@ -2050,6 +2050,11 @@ function odinSetupDefaultImports(wasmMemoryInterface, consoleElement, memory) {
 				}
 			},
 
+			set_document_title: (title_ptr, title_len) => {
+				let title = wasmMemoryInterface.loadString(title_ptr, title_len);
+				document.title = title;
+			},
+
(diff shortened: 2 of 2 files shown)


karl-zylinski:
This is now ready for review!

vendor/wasm/WebGL/webgl2.odin
```diff
@@ -103,21 +104,50 @@ foreign webgl2 {
 	GetUniformBlockIndex      :: proc(program: Program, uniformBlockName: string) -> i32 ---
 	UniformBlockBinding       :: proc(program: Program, uniformBlockIndex: i32, uniformBlockBinding: i32) ---
 	
+	// if `pname` is `UNIFORM_BLOCK_ACTIVE_UNIFORM_INDICES` then an array will be written at
+	// `params`, in that case the length `params` need to have is given first querying using `pname`
+	// `UNIFORM_BLOCK_ACTIVE_UNIFORMS`. 
+	GetActiveUniformBlockParameter :: proc(program: Program, uniformBlockIndex: i32, pname: Enum, params: [^]i32) ---
```
karl-zylinski, line 110: I tried to think of a way to make this binding more "idiomatic". But `params` can be a bool, enum or an array of values. So it's a bit hard to make something that isn't just a multipointer.

karl-zylinski: For those interested, this is the JS function: https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/getActiveUniformBlockParameter

It returns differently sized things depending on the parameter `pname`.

I think to make an idiomatic binding I would need to make a `GetActiveUniformBlockParameterBindings` variant for the "Uint32Array" case that can both take a buf slice or an allocator. The rest can be covered by returning an int.

However, that means adding new procs to the bindings that don't even exist in the webGL world, so perhaps it's a bit too much.

gingerBill (approved):

PR #6071, opened 2025-12-27
Use shared partial return memory

gingerBill:
To minimize the stack usage for partial return values requiring `alloca`, try to reuse it where possible.

**Note:** this will need a lot of proofing to make sure there are no aliasing issues with this. If aliasing issues do exist, they will definitely be subtle.


gingerBill:
I am going to close this because I could not verify it enough to see if it was correct.

PR #6080, opened 2025-12-29
Fix NetBSD CI

krnowak:
`10.0_2025Q2` seems to be gone from https://cdn.netbsd.org/pub/pkgsrc/packages/NetBSD/x86_64/. I can only see `10.0_2025Q3` and `10.0_2025Q4`, so went with the latter here.


krnowak:
Hmph, now FreeBSD failed with some segfault. But it's unrelated to this change.

PR #6083, opened 2025-12-29
Implement more Linux syscalls.

Znarf64:


core/sys/linux/bits.odin
```diff
@@ -593,6 +593,16 @@ Mem_Protection_Bits :: enum {
...
+
+Sched_Flag_Bits :: enum u32 {
+	RESET_ON_FORK  = 0,
+	RECLAIM        = 1,
+	DL_OVERRUN     = 2,
+	UTIL_CLAMP_MIN = 5,
+	UTIL_CLAMP_MAX = 6,
+}
+
+Sched_Attr_Flag_Bits :: enum {}
```
laytan: What is this doing here?

core/sys/linux/bits.odin
```diff
@@ -592,11 +592,6 @@ Mem_Protection_Bits :: enum {
 	GROWSUP   = 25,
 }
 
-/*
-	Bits for Map_Flags
-
-	See `constants.odin` for `MAP_SHARED_VALIDATE` and `MAP_HUGE_16KB`, et al.
-*/
```
laytan, line 599: Why is this removed? :)

Znarf64: oops

Issue #6060, opened 2025-12-23
encoding/json incorrect marshal of \x00 character (null terminator)


AntonGorov195:
## Context

        Odin:    dev-2025-08:fb4641307
        OS:      Linux Mint 22.1, Linux 6.8.0-90-generic
        CPU:     Intel(R) Core(TM) i5-9400F CPU @ 2.90GHz
        RAM:     15929 MiB
        Backend: LLVM 20.1.8

\x00 is not a valid character in a json string and it should be \u0000.
```
package null_rune

import "core:encoding/json"
import "core:fmt"

main :: proc() {
	val, err := json.marshal("\x00")
	if err != nil {
		unreachable()
	}
	fmt.println("\"\\u0000\" ==", string(val))
}
```
## Expected Behavior

`"\u0000" == "\u0000"`

## Current Behavior

`"\u0000" == "\x00"`

## Steps to Reproduce

Run the segment code above with `odin run .`

Issue #6075, opened 2025-12-29
Incorrect Unicode `rune`-handling in `core:encoding/json`


eagercoffee:
## Context
```
        Odin:    dev-2025-12-nightly:6ef91e2
        OS:      Windows 11 Home Basic (version: 25H2), build 26200.7462
        CPU:     AMD Ryzen 7 7735HS with Radeon Graphics
        RAM:     15676 MiB
        Backend: LLVM 20.1.0
```
## Expected Behavior

Marshalling an emoji as a `rune` into JSON should yield the corresponding UTF-16 encoded JSON string.

## Current Behavior

Marshalling an emoji as a `rune` into JSON yields the Odin escape sequence for the emoji, which isn't valid JSON.

### Steps to Reproduce

```odin
package bug_report

import "core:fmt"
import "core:encoding/json"

main :: proc() {
    {
        b, _ := json.marshal('\U0001F642')
        fmt.println(string(b)) // => "\U0001f642" (incorrect)
    }
    {
        R :: struct { v: rune }
        b, _ := json.marshal(R { '\U0001F642' })
        fmt.println(string(b)) // => {"v":"\U0001f642"} (incorrect)
    }
    {
        b, _ := json.marshal("Hi \U0001F642")
        fmt.println(string(b)) // => "Hi \ud83d\ude42" (correct)
    }
}

```

### Failure Logs

```
"\U0001f642"
{"v":"\U0001f642"}
"Hi \ud83d\ude42"
```

AntonGorov195:
The entire JSON package is bad. marshal of the null character is '\x00' instead of '\u0000'. float infinity is written as "Inf" instead of "Infinity" in JSON5 and SJSON specs. Unmarshal doesn't work with runes at all.

PR #6085, opened 2025-12-31
[encoding/json] Fix Unicode Output When Marshalling

dozn:
Fixes https://github.com/odin-lang/Odin/issues/6060 and https://github.com/odin-lang/Odin/issues/6075 
The boolean "true" was setting `html_safe` instead of `for_json`.
Chalk one up for the "use enums instead of booleans" crowd.


laytan:
Are you sure this fixes#6060, the example there uses a string and I only see runes being fixed?

dozn:
So it looks like it was never an issue as a `string`, only as a `rune`.
Do me a favour and confirm for me that #6060 can be closed, would ya?
Cheers!

laytan:
I think strings were never a problem and it was a bad report.

PR #6088, opened 2025-12-31
Remove All Extraneous Whitespace (My Personal Spacegate)

dozn:
Every time I go to create a patch, I have to stage chunks manually because there's random whitespace littered about which my editor fixes.
This removes all whitespace after the last character on a line, and extra newlines at EOF if there were multiple, otherwise they remain untouched.
- Probably easiest to review this with `git diff` by passing `--ignore-all-space --ignore-space-change --ignore-blank-lines --ignore-space-at-eol`
- I was going through each file manually (opening -> hitting save (I sometimes get into a mood where I want to do something repetitive, don't ask =b))
- - There were only ¿three? instances of _clearly_ incorrect whitespace I removed manually (such as the one in `core/sys/wasm/js/dom.odin`)
- I left all the double/triple+ newline gaps alone
- Third-party code has been left untouched


gingerBill:
This will literally add your name everywhere across the codebase. So no.

PR #6090, opened 2026-01-02
Prefer LPCSTR/LPCWSTR types in DirectX bindings

doanamo:
- Replaces [^]u16 parameter types because they are not directly compatible with result of `win32.utf8_to_wstring()`
- Replaces some string/wstring types for consistency, similarly as other packages do to match original C declarations and documentation


doanamo:
The first point is the main reason for the change - 3 instances of `[^]u16` that were not compatible with `win32.utf8_to_wstring()` result, while many other WinApi/DirectX function are. Remaining changes to use LPCSTR/LPCWSTR are for the consistency with C declarations and other Windows bindings in Odin. I've double checked with MSDN to make sure they are correct.

PR #6091, opened 2026-01-02
[vendor/sdl3] update to sdl 3.4.0

A1029384756:
updates the bindings to SDL 3.4.0. passes `odin check` and builds on linux. needs the following testing:
- [x] validate `.lib` file (generated via `lib` in wine environment, will test on real windows)

### DIFF
--- a/vendor/sdl3/sdl3_video.odin
+++ b/vendor/sdl3/sdl3_video.odin
@@ -135,12 +135,12 @@ FlashOperation :: enum c.int {
 }
 
 ProgressState :: enum c.int {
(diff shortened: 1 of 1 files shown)


A1029384756:
tested this on windows and the stuff i tested worked. there were some changes to `SDL_haptic.h` that i *didn't* bind because the bindings apparently already went and made the changes SDL just made. the one caveat is that the naming slightly differs: `HapticType` in the bindings and `HapticEffectType` in SDL itself. additionally, `SDL_dlopennote.h` is a new file that i didn't bind because it is purely a macro api

PR #6093, opened 2026-01-03
Add more win32 bindings

bplu4t2f:
Adds:

 * [Filled Shape Functions](https://learn.microsoft.com/en-us/windows/win32/gdi/filled-shape-functions) (gdi32, user32)
 * Buffered painting functions (uxtheme)
 * GetWindow (user32)
 * GetWindowTheme (uxtheme)
 * Associated constants and types


bplu4t2f:
For the record that CI failure seems unrelated to my commit. Either that's a CI pipeline issue, or a sporadic (and therefore scary) memory bug :S

PR #6094, opened 2026-01-03
In make_map_cap(...) assign map allocator

Faker-09:
In make_map_cap(...) assign allocator to the map allocator not the context allocator


laytan:
Why?

Faker-09:
I thought it is initializing a map and setting the map's allocator for future allocations. Similar to make_map(...) and the other procs in the make proc group.
From make_map_cap(...) comment:
// `make_map_cap` initializes a map with an allocator and allocates space using `capacity`.

Faker-09:
Since map_dynamic_reserve(...) will set the map allocator to the context.allocator if the map allocator procedure is nil there is no functional change with this patch.  It is just a bit more explicit about setting the map allocator.

PR #6106, opened 2026-01-06
Fix typo in the doc comment

bbogdan-ov:
Fixed a small typo in the documentation comment of the `assert` function :)

### DIFF
--- a/base/runtime/core_builtin.odin
+++ b/base/runtime/core_builtin.odin
@@ -1095,7 +1095,7 @@ card :: proc "contextless" (s: $S/bit_set[$E; $U]) -> int {
 
 
 
-// Evaluates the condition and panics the program iff the condition is false.
+// Evaluates the condition and panics the program if the condition is false.
 // This uses the `context.assertion_failure_procedure` to assert.
 //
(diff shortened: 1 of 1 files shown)


gingerBill:
This isn't a typo. Nor are you the first to try and do this.

https://en.wikipedia.org/wiki/If_and_only_if

bbogdan-ov:
oh ok, didn't know that, sorry

dozn:
We _could_ rewrite it so an esoteric abbreviation is unnecessary:

> Panics if `condition` evaluates to `false`.

gingerBill:
@dozn It isn't "esoteric". It's extremely common in mathematics and the sciences.

dozn:
> @dozn It isn't "esoteric". It's extremely common in mathematics and the sciences.

Yes, it's common only with a couple groups of people—that's what "esoteric" attempted to convey =b

Regardless, the shortened version seems more clear (and there's of course other ways of writing it as well, but that was the shortest I thought of).

gingerBill:
That isn't a small group, but sure. Whatever.

Issue #6112, opened 2026-01-08
core:enocding/xml not correctly handling comments before first element and comments end with --


StudebakerGuy:
## Context

	Odin:    dev-2026-01:94cc5d9e6
	OS:      OpenMandriva Lx 25.90 (Nickel) Cooker, Linux 6.18.1-desktop-1omv2590
	CPU:     AMD Ryzen Threadripper PRO 5955WX 16-Cores
	RAM:     96430 MiB
	Backend: LLVM 21.1.8

## Expected Behavior

Comments before the first element should be placed in xml.Document.comments.

Comments should not include the -- from the closing -->

## Current Behavior

Comments before the first element are bing placed in xml.Document.elements

Comments are include the -- from the closing -->

## Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Download [xml.xml](https://github.com/user-attachments/files/24485025/xml.xml) and  [testxml.txt](https://github.com/user-attachments/files/24485035/testxml.txt) into a new directory.
2. Change the name of `testxml.txt` to `testxml.odin`
3. run `odin test . ` 

### Failure Logs

[INFO ] --- [2026-01-08 00:32:40] Starting test runner with 5 threads. Set with -define:ODIN_TEST_THREADS=n.
[INFO ] --- [2026-01-08 00:32:40] The random seed sent to every test is: 2531986737747280. Set with -define:ODIN_TEST_RANDOM_SEED=n.
[INFO ] --- [2026-01-08 00:32:40] Memory tracking is enabled. Tests will log their memory usage if there's an issue.
[INFO ] --- [2026-01-08 00:32:40] < Final Mem/ Total Mem> <  Peak Mem> (#Free/Alloc) :: [package.test_name]
[ERROR] --- [2026-01-08 00:32:40] [testxml.odin:26:test_comment_before_root()] expected len(doc.comments) to be 1, got 0
testxml  [|||||                   ]    1/   5 :: test_comment_before_root
5 threads                              1/   5 :: total
/home/studebakerguy/src/testxml/testxml.odin(34:42) Index 0 is out of range 0..<0
[ERROR] --- [2026-01-08 00:32:40] [testxml.odin:33:test_comment_before_root_value()] expected len(doc.comments) to be 1, got 0
[...]

PR #6113, opened 2026-01-08
core:encoding/xml comments not handled correctly before first element and include --

StudebakerGuy:
This pull request is to address [issue 6112](https://github.com/odin-lang/Odin/issues/6112). This addresses comments not being placed in xml.Document.comments and comments including `--` from the closing `-->` of a comment.

### DIFF
--- a/tests/core/encoding/xml/test_core_xml.odin
+++ b/tests/core/encoding/xml/test_core_xml.odin
@@ -142,7 +142,7 @@ xml_test_entities_unbox_decode :: proc(t: ^testing.T) {
 			},
 			expected_doctype = "html",
 		},
-		crc32     = 0x6e45a697,
+		crc32     = 0xda1ac384,
 	})
 }
 


StudebakerGuy:
Given the failed tests, I'm going to look further into this and will resubmit.

StudebakerGuy:
After examing the failed tests, the isssues reported appear be valid. The issues appear to be untested or uncaught by the tests currently beign performed. As an exmple the file `tests/core/assets/XML/entities.html` contains the comment `<!-- foo attribute should be empty but present -->`. One would exptect the comment to be printed as ` [COMMENT] [" foo attribute should be empty but present "]` but it is printed as  `[COMMENT] [" foo attribute should be empty but present --"]` with the current master branch of core:encoding/xml. 

If it is aggreed that the issue does exist and this patch addresses it, I can move forward with working on a patch to `test_core_xml.odin` if desired.

PR #6121, opened 2026-01-10
implement #override_foreign_library

hgranthorner:
I've attempted to implement the feature requested and described in this discussion: https://github.com/odin-lang/Odin/discussions/1390

Essentially, someone should be able to custom override a foreign import declaration. For example, if a library (`otherlib`) sets up a foreign import like this:
`foreign import lib "system:mylib"`

You should then be able to override that import like so:
```
import "otherlib"

#override_foreign_library otherlib.lib "system:adifferentlib"
```

This is slightly different than the way suggested by @gingerBill in the above discussion. I found that depending on whether or not I was linking with a static library, I needed to include other libraries/frameworks as well, so just swapping out a single library definition (i.e. "system:somelib" -> "myownlib.a") wasn't flexible enough.

I've tested this locally (M series MacOS) and it seems to work for my purposes. I have a few tests cases written to ensure that it works with compound literals, exported libs, and aliasing the library (i.e. `import o "otherlib"`), but I don't know how to integrate those tests into the existing test system. If you'd like I can add another commit including the tests and the little bash script I wrote to run them. I have not tested this feature on Windows or Linux at all.

There's two "design" decisions of note:
1. In the case where a library is overridden multiple times I've chosen to fail compilation
2. The directive optionally allows for `()` calling syntax: `#override_foreign_library(a, "b")` == `#override_foreign_library a "b"`. I wasn't sure which way to go - most directives use parentheses, but not requiring parentheses makes it match `foreign import` syntax more closely and matches the original design in the linked discussion.

I've never contributed to Odin before, so feel free to chuck the code in the bin if it sucks, or just take little bits of it and implement it yourself in an actually good way. Or just put it on the back burner.

### DIFF
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -2084,7 +2084,156 @@ gb_internal bool check_hash_kind(CheckerContext *c, Ast *call, String const &has
 	return true;
 }
 
+gb_internal bool check_override_foreign_library_directive(CheckerContext *c, Operand *operand, Ast *call) {
+	ast_node(ce, CallExpr, call);
+	ast_node(bd, BasicDirective, ce->proc);
+	String name = bd->name.string;
+	GB_ASSERT(name == "override_foreign_library");
 
+	if (ce->args.count != 2) {
+		error(call, "'#override_foreign_library' expects 2 arguments, got %td", ce->args.count);
+		return false;
+	}
+	if ((c->scope->flags & ScopeFlag_File) == 0) {
+		error(call, "'#override_foreign_library' may only be used at file scope");
+		return false;
+	}
+
+	Ast *target = unparen_expr(ce->args[0]);
+	if (target == nullptr || target->kind != Ast_SelectorExpr) {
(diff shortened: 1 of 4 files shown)


gingerBill:
Why not just modify the library's source code?

hgranthorner:
Two reasons:

1. If the library is already included in your source code, it's one less patch to maintain if you want to update the library.
2. If the library is part of Odin's pre-vendored library collection, it means you don't have to pull the source code into your repository just to change where the external library is.

It's not strictly necessary, but it does allow developers a little more flexibility around what are traditionally linker arguments (i.e typically configurable without third party source changes).

gingerBill:
For both points, I recommend including Odin as part of the project itself (e.g. `git subtree`) and then make the modifications yourself. This is called vendoring and it is the best way in general.

And your idea isn't enough because it doesn't actually solve anything, nor does it allow for a list of the foreign imports. It's too naïve of a solution.

Just modify the library's source code in-place or copy-and-paste and then modify. These are MUCH better solutions in general.

PR #6124, opened 2026-01-11
Add `core:nbio`

laytan:
Package nbio implements a non-blocking I/O and event loop abstraction layer over several platform-specific asynchronous I/O APIs.

Some more examples: https://github.com/odin-lang/examples/pull/161

It is the back bone of the HTTP package that is coming. The original version was implemented years ago at https://github.com/laytan/odin-http/tree/main/nbio. This version is tested more thoroughly, implements missing features (timeouts, cancelling, wake_ups, thread safety, bsds support, sendfile, scatter/gather sending and receiving, blocking until work is to be done, and probably more I am forgetting) and fixes bugs.

All the main targets of Odin are supported: Windows (iocp), Linux (io uring), darwin, openbsd, netbsd, freebsd (kqueue)
I made sure other targets are stubbed out and compile but will error with unimplementedness.

Besides the new package it fixes some stuff I needed fixed across the board.
- core:net support for openbsd and netbsd, and allow compilation for other targets, also cleans up the socket options api so there are no options on one target that are undefined references on other targets
- make sure all documentation examples get through the documentation tester we have running, these apparently don't trip up CI, done in xar, slice, chan
- thread was doing aligned loads on possibly unaligned addresses
- `new_clone` was zeroing memory before putting the cloned value there
- remove checking bsd and arm in linux amd64 ci, they have their own runners now

**I already know I'll get this question: Why callbacks? Callback hell!**

Callbacks are the simplest interface an event loop can reasonably expose. Run this when the operation completes. The loop itself doesn’t need to know how the result is consumed. (You could give control of the entire loop to the user, but that comes with all the problems in that).

Other mechanisms can be built on top of it:
- Coroutines (either via Lua or native coroutine packages): A very rough example can be found here with a port of the minicoro coroutine package https://github.com/laytan/examples/blob/40c2fea3888c6500422ee1c957358eaa1e3a2a5b/nbio/coroutines/main.odin
- rather have a queue you consume at your own time: https://github.com/laytan/examples/blob/nbio/nbio/no-callbacks/main.odin

The choice of one `Operation` type and one callback type for all the operations is partly so it is easy to bind into Lua like or native coroutines.

Callbacks also allow multiple independent users to share the same event loop. A package can register its own operations, and application code can register others, without either seeing or handling the other’s completions.

**future work**

After this I will now continue working on getting the HTTP package into core.

In the future I or somebody else might want to add an `epoll` using Linux target for the cases where IO uring is not wanted/can not be used.

### DIFF
--- a/core/container/pool/pool.odin
+++ b/core/container/pool/pool.odin
@@ -44,7 +44,6 @@ destroy :: proc(p: ^Pool($T)) {
 	elem := sync.atomic_exchange_explicit(&p.free_list, nil, .Acquire)
 
 	sync.atomic_store_explicit(&p.num_ready, 0, .Relaxed)
-	assert(sync.atomic_load_explicit(&p.num_outstanding, .Relaxed) == 0)
 
 	when .Address in ODIN_SANITIZER_FLAGS {
 		for ; elem != nil; elem = _get_next(p, elem) {


core/container/pool/pool.odin
```diff
@@ -0,0 +1,132 @@
...
+
+Example:
+	Elem :: struct {
+		link: ^Elem,
+	}
+
+	p: pool.Pool(Elem)
+	pool.init(&p, "link")
+*/
+Pool :: struct($T: typeid) {
```
gingerBill, line 27: Does this Pool need to be typed or could it be type erased?

e.g. `free_list: rawptr, elem_size: int`

core/container/pool/pool.odin
```diff
@@ -0,0 +1,132 @@
...
+	defer sync.atomic_add_explicit(&p.num_ready, 1, .Relaxed)
+
+	for {
+		head := sync.atomic_load(&p.free_list)
+		_set_next(p, elem, head)
+		if _, ok := sync.atomic_compare_exchange_weak(&p.free_list, head, elem); ok {
+			return
+		}
+	}
+}
```
gingerBill, line 94: If `destroy` is called with `put`/`get`, this is a data race, and you need to do exchange the list with `nil` before traversal.

e.g. `sync.atomic_exchange_explicit(&p.freelist, nil, .Acquire)` to detach and get the free-list head.

In `put`, load with `.Relaxed`  and for the `sync.atomic_compare_exchange_weak_explicit(&p.free_list, ..., .Release, .Relaxed)`

In `get`, use `.Acquire`, and then `sync.atomic_compare_exchange_weak_explicit(&p.free_list, ..., .Acquire, .Relaxed)`

core/nbio/nbio.odin
```diff
@@ -0,0 +1,435 @@
...
+	using _: struct #raw_union {
+		_pool_link: ^Operation,
+		l:          ^Event_Loop,
+	},
+	cb: Callback,
+	using _specifics: Specifics,
+	user_data: [MAX_USER_ARGUMENTS + 1]rawptr,
+	type: Operation_Type,
+	detached: bool,
+}
```
gingerBill, line 56: The layout seems a little messy to me. Why have the "private" stuff first rather than at the end?

e.g.

```odin
Operation :: struct {
	type:      Operation_Type,
	cb:        Callback,
	detached:  bool,
	user_data: [MAX_USER_ARGUMENTS + 1]rawptr,


	using _specifics: Specifics,

	_impl:   _Operation `fmt:"-"`,
	using _: struct #raw_union {
		_pool_link: ^Operation,
		l:          ^Event_Loop,
	},
}
```

core/nbio/nbio.odin
```diff
@@ -0,0 +1,435 @@
...
+	err:         General_Error,
+	refs:        int,
+	now:         time.Time,
+
+	// Queue that is used to queue operations from another thread to be executed on this thread.
+	// TODO: Better data-structure.
+	queue:    queue.Queue(^Operation),
+	queue_mu: sync.Mutex,
+
+	operation_pool: pool.Pool(Operation),
```
gingerBill, line 35: Is this `pool.Pool` ony use for `Operation`?

core/container/pool/pool.odin
laytan: If you want yeah, I like to use generics myself.

Would need to take alignment of the element too, and the offset of the link pointer instead of calculating that using `offset_of` here.

Is there a reason you would want something like this type erased?

core/nbio/nbio.odin
laytan: Yes

core/container/pool/pool.odin
gingerBill: The reason is just to minimize the amount of code generated. But if it's only used once in practice, I don't think it's an issue.

core/nbio/nbio.odin
laytan: `_impl` being the first was so I could do `(^Operation)(event.lpOverlapped)` on Windows, but it does seem nicer to reorder. I have changed it to a series of `container_of` calls: https://github.com/odin-lang/Odin/pull/6124/commits/c637679917397417a4f15e5ee93dc96dc44c9baa.

core/container/pool/pool.odin
laytan: `destroy` will always be a race with `put`/`get` because it frees the memory, if you still have an item it is a use-after-free. The comment at the top also states it is thread-safe between `init` and `destroy`. I did change the proc to exchange but it will never be safe.

Also applied the explicit memory ordering :)

PR #6125, opened 2026-01-12
Add init and fini callback procs to thread pools

Wassimulator:
sometimes you might need the ability to have general thread initializing and deinitializing for workers that need to run once at their creation and shutdown, this especially is useful for any `@thread_local` types, this adds the option to set those at init.

### DIFF
--- a/core/thread/thread_pool.odin
+++ b/core/thread/thread_pool.odin
@@ -20,8 +20,8 @@ Task :: struct {
 	allocator:  mem.Allocator,
 }
 
-Thread_Init_Proc :: #type proc(thread_index: int, user_data: rawptr)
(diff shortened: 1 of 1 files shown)


core/thread/thread_pool.odin
```diff
@@ -20,6 +20,9 @@ Task :: struct {
 	allocator:  mem.Allocator,
 }
 
+Thread_Init_Proc :: #type proc(thread_index: int, user_data: rawptr)
+Thread_Fini_Proc :: #type proc(thread_index: int, user_data: rawptr)
```
gingerBill: I wonder if passing `^Thread` instead of the `thread_index` might be a better idea?

Wassimulator: good point.

PR #6129, opened 2026-01-13
Allow mold linker on UNIX

ghost:
The mold linker isn't just available for Linux, but most UNIX operating systems. It is available on many of the BSDs. Tested on FreeBSD.

### DIFF
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -174,7 +174,7 @@ try_cross_linking:;
 		switch (build_context.linker_choice) {
 		case Linker_Default:  break;
 		case Linker_lld:      section_name = str_lit("lld-link"); break;
-	#if defined(GB_SYSTEM_LINUX)
+	#if defined(GB_SYSTEM_UNIX)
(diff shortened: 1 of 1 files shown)


src/linker.cpp
```diff
@@ -174,7 +174,7 @@ try_cross_linking:;
 		switch (build_context.linker_choice) {
 		case Linker_Default:  break;
 		case Linker_lld:      section_name = str_lit("lld-link"); break;
-	#if defined(GB_SYSTEM_LINUX)
+	#if defined(GB_SYSTEM_UNIX)
```
gingerBill, line 177: Is `mold` available on all "UNIX" platforms?

ghost: I guess not. It is packaged on FreeBSD and NetBSD at least. While it might work on Haiku and OpenBSD, I have not tested that. Not sure if it matters though, isn't using Mold as the linker kinda explicit either way? Seems best to trust the user with the choice.

PR #6147, opened 2026-01-18
Create new `core:sys/darwin/CoreGraphics` package

slowhei:
Apple's CoreGraphics framework provides an API for displays (laptop screens, external monitors, etc.) and the mouse cursor. It also provides many other things, but these two use cases are the motivation behind this PR.

My new `core:sys/darwin/CoreGraphics` package is mostly straightforward; `foreign import` the Core Graphics framework and translate the C types into Odin types.

But there is one decision I made that is worth mentioning; I made `core:sys/darwin/Foundation` import `core:sys/darwin/CoreGraphics`. This means that `CoreGraphics` is not allowed to import `Foundation`, which would be a problem if `CoreGraphics` used, for example, `NSObject`. I searched the C headers of Core Graphics and did not find any types prefixed with `NS`, so we are safe there, but there is a chance of running into cyclic imports when adding more bindings for `CoreGraphics` in the future. If that happens, we can duplicate a couple type definitions or define a third package containing shared type definitions. Not the end of the world.

Other than that, I added a couple bindings for Foundation. And all new bindings, including for Core Graphics, have been run on my machine to verify that they work.

This is my first time proposing a new Odin package. Is there anything I should do relating to package documentation?

### DIFF
--- a/core/sys/darwin/CoreFoundation/CFCGTypes.odin
+++ b/core/sys/darwin/CoreFoundation/CFCGTypes.odin
@@ -1,0 +1,18 @@
+package CoreFoundation
+
+CGFloat :: distinct (f32 when size_of(uint) == 4 else f64)
+
+CGPoint :: struct {
+	x: CGFloat,
+	y: CGFloat,
+}
+
+CGRect :: struct {
+	using origin: CGPoint,
+	using size:   CGSize,
+}
+
+CGSize :: struct {
+	width:  CGFloat,
+	height: CGFloat,
+}
--- a/core/sys/darwin/CoreGraphics/CoreGraphics.odin
+++ b/core/sys/darwin/CoreGraphics/CoreGraphics.odin
@@ -1,6 +1,7 @@
 package CoreGraphics
 
-import "core:c"
+import    "core:c"
+import CF "core:sys/darwin/CoreFoundation"
 
 @(require)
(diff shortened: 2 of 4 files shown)


laytan:
Shouldn't this dependency be the other way around? CoreGraphics depending on Foundation makes way more sense than the other way around.

slowhei:
Core Graphics is a C API that defines the types `CGFloat`, `CGPoint`, `CGRect`, and `CGSize` that Foundation uses. Additionally, in Foundation's Obj-C header files, Core Graphics gets imported (e.g. in `NSGeometry.h`), but the opposite never happens.

(Note: If we want to be really precise, `CGFloat` and the other types I mentioned actually usually get defined in Core Foundation in `CFCGTypes.h`. But Core Graphics contains duplicate definitions of these types in `CGBase.h` that are wrapped in `#ifndef CF_DEFINES_CG_TYPES`, meaning Core Graphics defines these types itself if Core Foundation is never imported.)

laytan:
Aha, I do like CoreFoundation better, it just seems wrong to me how it is now.

So can we put these shared types in CoreFoundation and have both Foundation and CoreGraphics use them instead?

slowhei:
That makes sense to me. I have moved the shared type definitions into CoreFoundation.

PR #6158, opened 2026-01-23
Update SDL_image bindings to SDL 3.4

nowhereware:
These commits update the vendor bindings for SDL_image to version 3.4, including upgrades to the dll and lib files as well as the inclusion of two new libraries (libpng16 and libwebpmux), which I believe are required by the main library given their inclusion in the distributed 3.4 release files.


Kelimion:
Those image dll's aren't required. As stated in `INSTALL.md` in the [distribution](https://github.com/libsdl-org/SDL_image/releases/tag/release-3.4.0) zip:
```
You can include support for additional image formats by including the license and DLL files in the optional directory in your application. They will be automatically loaded by SDL_image as needed.
```

We shouldn't ship them. Our repository is large enough as it is, and if someone wants to use `webp` with SDL3-Image, they can copy `libwebp*.dll`, et al, from that zip's `optional` directory next to their application.

nowhereware:
Gotcha, I'll remove those then. This still updates the bindings to 3.4 so I think the code updates done are still valid for merging, please let me know any other concerns though.

PR #6159, opened 2026-01-23
Fix `in`/`not_in` on constant `bit_set`s

Barinzaya:
This addresses two niche issues with `in`/`not_in` on `bit_set`s evaluated at compile-time:
- With a `bit_set` having a non-zero lower bound, `in` and `not_in` were returning incorrect results when done at compile-time.
- With a `bit_set` of more than 128 bits, `in` always returns false on values that fall within the upper 64 bits.

These issues can be demonstrated with the following snippet:
```odin
#assert(5 in bit_set[1..<128]{5})
#assert(4 not_in bit_set[1..<128]{5})
#assert(100 in bit_set[0..<128]{100})

// also with an enum
E :: enum { A = 1, B, C = 100 }
#assert(.B in bit_set[E]{.B})
#assert(.A not_in bit_set[E]{.B})
#assert(.C in bit_set[E]{.C})
```

On the current `master`, all of these `#assert`s fail. On this branch, they succeed (as expected).

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -4125,15 +4125,19 @@ gb_internal void check_binary_expr(CheckerContext *c, Operand *x, Ast *node, Typ
 				i64 upper = yt->BitSet.upper;
 
 				if (lower <= key && key <= upper) {
-					i64 bit = 1ll<<key;
-					i64 bits = big_int_to_i64(&v.value_integer);
+					BigInt idx = big_int_make_i64(key - lower);
+					BigInt bit = big_int_make_i64(1);
+					big_int_shl_eq(&bit, &idx);
+
+					BigInt mask = {};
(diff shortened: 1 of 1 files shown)


Kelimion:
Good catch, but we may want to `big_int_dealloc` the 3 bigints `idx`, `bit`, and `mask`, or no?

Barinzaya:
> Good catch, but we may want to `big_int_dealloc` the 3 bigints `idx`, `bit`, and `mask`, or no?

Not a single other place in `check_expr.cpp` does so that I can see, and even `big_int_add_eq` leaks a `BigInt` as far as I can tell. The memory management on the `BigInt`s seems to be entirely ignored, so I just did it the same way.

I wouldn't be opposed to adding it if it would be preferred, though.

Kelimion:
No that's fine. Either we do it consistently, or not at all.

PR #6166, opened 2026-01-24
make #+vet tags always work in addition to command line flags

### LINKED ISSUE #6067: `#+vet` tags nullify command line build `-vet` parameters
## Context

Odin:    dev-2025-12
OS:      Debian GNU/Linux forky/sid, Linux 6.17.8-300.fc43.x86_64
RAM:     32004 MiB
Backend: LLVM 19.1.7

The following code snippet illustrates the issue:
```odin
package main

main :: proc() {
	x: int
}
```

- If we run `odin check .` on this, or `odin build`, this builds just fine.
- If we run `odin check . -vet`, it errors,  with `'x' declared but not used` as expected.
- If we add any `#+vet` to the top of the file, the error is ignored:
```odin
#+vet cast
package main

main :: proc() {
	x: int
}
```

`odin check . -vet` and `odin build -vet` show no error and build just fine, even though it should error.

The main case where this is an issue is `#+vet explicit-allocators`, which *must* be a file flag. So it effectively means that `#+vet explicit-allocators` disables every other form of vetting.

laytan:
Fixes #6067

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -6363,13 +6363,11 @@ gb_internal u64 parse_vet_tag(Token token_for_pos, String s) {
 	GB_ASSERT(string_starts_with(s, prefix));
 	s = string_trim_whitespace(substring(s, prefix.len, s.len));
 
+	u64 vet_flags = build_context.vet_flags;
+
 	if (s.len == 0) {
-		return VetFlag_All;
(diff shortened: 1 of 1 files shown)


BigBoyBarney:
Thank you very much! I've been looking forward to this a lot :heart: 

I appreciate your work!

PR #6169, opened 2026-01-24
(Fixed) Update SDL3_image to 3.4.0

nowhereware:
This updates the SDL3 image bindings to version 3.4.0. I apologize for the duplicate PR, but I needed to remove the extraneous image format DLLs. The SDL3_image library files were updated as well which caused an increase in size, but that can't necessarily be helped. Please let me know if there's any further changes that need to be made, although there shouldn't be given that this is a relatively straightforward binding update.


Kelimion:
Turns out we already had the "optional" image format libraries after all, sorry. I refreshed them.

But going forward we'll have to pop up a message on creating a pull request that we're not accepting pull request with object code. As we end up vouching for the authenticity as the project, we're verifying they match upstream DLLs or rebuild things ourselves as a matter of course, so someone outside of the team adding/updating a DLL doesn't actually save us work.

PR #6205, opened 2026-02-01
add intrusive freelist type to the Xar container

Wassimulator:


jakubtomsu:
why add this to the `xar` package, instead of something like `core:container/xar/freelist` or `core:container/intrusive/xar_freelist`?

Wassimulator:
Well this doesn't matter to me much, Which one would be preferred? I can update it.

core/container/xar/freelist.odin
```diff
@@ -0,0 +1,155 @@
...
+		chunk_start_addr := uintptr(chunk)
+		chunk_end_addr := chunk_start_addr + uintptr(chunk_cap * size_of(T))
+		if ptr_addr >= chunk_start_addr && ptr_addr < chunk_end_addr {
+			offset := int(ptr_addr - chunk_start_addr) / size_of(T)
+			return base + offset
+		}
+		base += chunk_cap
+	}
+	return -1
+}
```
jakubtomsu: Are you sure there is no better way to organize the datastructure so you don't have to do this search? All of this can absolutely be O(1). What about always using indexes directly?

gingerBill:
@jakubtomsu `xar` I am thinking `xar` COULD just be the general Exponential Array Related packages. It might not need to be in separate packages.

core/container/xar/freelist.odin
```diff
@@ -0,0 +1,155 @@
+package container_xar
+
+@(require) import "core:mem"
+
+Freelist_Array :: struct($T: typeid, $SHIFT: uint) where
+	0 < SHIFT,
+	SHIFT <= MAX_SHIFT,
+	size_of(T) >= size_of(^T)
+{
```
gingerBill: Reformat it like this to keep `-strict-style` happy:

```odin
Freelist_Array :: struct($T: typeid, $SHIFT: uint) where
	0 < SHIFT,
	SHIFT <= MAX_SHIFT,
	size_of(T) >= size_of(^T) {
```

core/container/xar/freelist.odin
```diff
@@ -0,0 +1,155 @@
...
+	for it.idx < it.freelist_array.array.len {
+		if !freelist_is_freed(it.freelist_array, it.idx) {
+			val = get(&it.freelist_array.array, it.idx)
+			idx = it.idx
+			it.idx += 1
+			return val, idx, true
+		}
+		it.idx += 1
+	}
+	return {}, -1, false
```
gingerBill: I'd probably do `return` with no parameters so that everything is zero. `idx = -1` is probably not a good idea in the use of an iterator.

core/container/xar/freelist.odin
```diff
@@ -0,0 +1,155 @@
...
+	for it.idx < it.freelist_array.array.len {
+		if !freelist_is_freed(it.freelist_array, it.idx) {
+			val = get_ptr(&it.freelist_array.array, it.idx)
+			idx = it.idx
+			it.idx += 1
+			return val, idx, true
+		}
+		it.idx += 1
+	}
+	return nil, -1, false
```
gingerBill: I'd probably do `return` with no parameters so that everything is zero. `idx = -1` is probably not a good idea in the use of an iterator.

PR #6207, opened 2026-02-01
Fix all ThreadSanitizer data races (227 warnings → 0)

JesseRMeyer:
I had Claude address TSAN warnings as they would occur running the compiler on loop running the odin tests and demo on various optimization modes until they went silent on Linux Mint x64 for a duration of 5 minutes.  Claude checked its  work several times along the way.

Replace non-thread-safe patterns with proper atomic operations and synchronization across the concurrent checker and LLVM codegen phases.

Changes by category:

1. Thread pool (Chase-Lev deque) TSan annotations
   - Add TSAN_RELEASE/TSAN_ACQUIRE on ring buffer slots in push/take/steal
   - Annotations establish happens-before for the work-stealing protocol

2. Entity flags: std::atomic<u64> with fetch_or/fetch_and
   - EntityFlag_Used, EntityFlag_Cold, EntityFlag_Disabled, etc.
   - Replaces racy compound assignment (|=, &=~) from concurrent checkers

3. Ident.entity pointer: atomic load/store helpers via std::atomic reinterpret_cast
   - ident_entity_load/ident_entity_store in parser.hpp
   - All read/write sites converted across checker and codegen

4. AST type-and-value (tav): striped locks (64 stripes, cache-line aligned)
   - add_type_and_value always acquires stripe lock keyed by node address
   - Protects memcpy-based tav writes from concurrent tav reads

5. DeclInfo atomic bools: where_clauses_evaluated, defer_use_checked
   - Changed from bool to std::atomic<bool> in checker.hpp

6. viral_state_flags: atomic load/fetch_or helpers
   - ast_viral_flags_load/ast_viral_flags_or in parser.hpp
   - All 29 access sites converted

7. Union type cache fields: atomic load/store via std::atomic reinterpret_cast
   - tag_size and variant_block_size in types.cpp and all LLVM backend readers
   - Idempotent cache writes that race with concurrent readers

8. mutex_try_lock correctness: exchange → compare_exchange_strong
   - Old code could spuriously "succeed" when mutex was already locked

9. parent_proc_decl: atomic load/store via std::atomic reinterpret_cast

All helpers use std::atomic reinterpret_cast (not __atomic_* builtins) because clang's TSan instrumentation does not properly track GCC-style __atomic_load_n/__atomic_store_n with __ATOMIC_RELAXED.

### DIFF
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -2124,7 +2124,7 @@ gb_internal bool check_builtin_procedure_directive(CheckerContext *c, Operand *o
 				if (e == nullptr || (e->flags & EntityFlag_Param) == 0) {
 					error(arg, "'#caller_expression' expected a valid earlier parameter name");
 				}
-				arg->Ident.entity = e;
+				ident_entity_store(arg, e);
 			}
 		}
 
@@ -6947,7 +6947,7 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 			
 			operand->mode = Addressing_Constant;
 			operand->type = t_untyped_integer;
-			operand->value = exact_value_i64(u->Union.variant_block_size);
+			operand->value = exact_value_i64(reinterpret_cast<std::atomic<i64>*>(&u->Union.variant_block_size)->load(std::memory_order_relaxed));
 		}
 		break;
 
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -624,7 +624,7 @@ gb_internal bool check_vet_shadowing_assignment(Checker *c, Entity *shadowed, As
 	if (init->kind == Ast_Ident) {
 		// TODO(bill): Which logic is better? Same name or same entity
 		// bool ignore = init->Ident.token.string == name;
-		bool ignore = init->Ident.entity == shadowed;
+		bool ignore = ident_entity_load(init) == shadowed;
 		if (ignore) {
 			return true;
 		}
@@ -1654,7 +1654,7 @@ retry:;
 	expr = unparen_expr(expr);
 	switch (expr->kind) {
 	case_ast_node(ident, Ident, expr);
-		Entity *e = ident->entity;
+		Entity *e = ident_entity_load(expr);
 		if (e && e->flags & EntityFlag_Overridden) {
 			// GB_PANIC("use of an overriden entity: %.*s", LIT(e->token.string));
 		}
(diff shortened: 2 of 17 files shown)


src/check_builtin.cpp
```diff
@@ -6947,7 +6947,7 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 			
 			operand->mode = Addressing_Constant;
 			operand->type = t_untyped_integer;
-			operand->value = exact_value_i64(u->Union.variant_block_size);
+			operand->value = exact_value_i64(reinterpret_cast<std::atomic<i64>*>(&u->Union.variant_block_size)->load(std::memory_order_relaxed));
```
gingerBill, line 6950: Would it make sense to just make the `variant_block_size` a `std::atomic<i64>` directly?

gingerBill: Et cetera to the rest too?

src/checker.cpp
```diff
@@ -1792,7 +1792,16 @@ gb_internal void add_untyped(CheckerContext *c, Ast *expr, AddressingMode mode,
...
 
-gb_internal void add_type_and_value(CheckerContext *ctx, Ast *expr, AddressingMode mode, Type *type, ExactValue const &value, bool use_mutex) {
+enum { TAV_STRIPE_COUNT = 64 };
+static struct alignas(64) { BlockingMutex mu; } tav_stripes[TAV_STRIPE_COUNT];
+
+static BlockingMutex *tav_mutex_for_node(Ast *node) {
+	uintptr_t h = (uintptr_t)node;
+	h ^= h >> 6; // mix out alignment bits
+	return &tav_stripes[h % TAV_STRIPE_COUNT].mu;
+}
```
gingerBill, line 1802: This uses `static` rather than `gb_internal` and this is a bit of a weird approach. I am not sure I like this in the slightest.

src/checker.cpp
```diff
@@ -1803,14 +1812,8 @@ gb_internal void add_type_and_value(CheckerContext *ctx, Ast *expr, AddressingMo
...
 	}
 
-	BlockingMutex *mutex = &ctx->info->type_and_value_mutex;
-	if (ctx->decl) {
-		mutex = &ctx->decl->type_and_value_mutex;
-	} else if (ctx->pkg) {
-		mutex = &ctx->pkg->type_and_value_mutex;
-	}
-
-	if (use_mutex) mutex_lock(mutex);
```
gingerBill, line 1813: It removing the `use_mutex` is also a bit concerning too but maybe there is a reason for this? I had this be optional because in some cases, we really didn't need it.

gingerBill:
I'm probably going to do the changes manually myself instead of relying on this PR.

gingerBill:
This PR will probably be the one I use: https://github.com/odin-lang/Odin/pull/6215

PR #6208, opened 2026-02-01
Add ThinLTO support via -lto:thin and -lto:thin-files flags

JesseRMeyer:
I asked Claude to implement thin-lto into the compiler, with support at the Package and individual file level.  On my system, this results in a wall clock reduction of 40% in 'odin run' for the tests and demo, largely through parallel compilation gains from LLVM.

- Add -lto:thin and -lto:thin-files CLI flags with validation
- Emit LLVM bitcode (.bc) instead of object files when LTO is enabled
- Pass -flto=thin and -flto-jobs to clang/lld linkers
- Guard linkage corrections to skip declarations without definitions (required for LTO where declarations appear across modules)
- Allow module-per-file with LTO even at higher optimization levels

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -420,6 +420,12 @@ struct BuildCacheData {
 };
 
 
+enum LTOKind : i32 {
+	LTO_None,
+	LTO_Thin,
+	LTO_Thin_Files,
+};
+
 enum LinkerChoice : i32 {
 	Linker_Invalid = -1,
 	Linker_Default = 0,
@@ -560,6 +566,7 @@ struct BuildContext {
 
 	bool   use_single_module;
 	bool   use_separate_modules;
+	LTOKind lto_kind;
 	bool   module_per_file;
 	bool   cached;
 	BuildCacheData build_cache_data;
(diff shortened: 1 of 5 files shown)


src/build_settings.cpp
```diff
@@ -2030,6 +2037,29 @@ gb_internal void init_build_context(TargetMetrics *cross_target, Subtarget subta
 		bc->use_separate_modules = false;
 	}
 
+	if (bc->lto_kind == LTO_Thin || bc->lto_kind == LTO_Thin_Files) {
+#if LLVM_VERSION_MAJOR < 17
+		gb_printf_err("-lto:thin requires LLVM 17 or later\n");
```
laytan, line 2042: All these error messages assume `-lto:thin` when it could also be `-lto:thin-files`.

src/build_settings.cpp
```diff
@@ -2030,6 +2037,29 @@ gb_internal void init_build_context(TargetMetrics *cross_target, Subtarget subta
...
+		gb_exit(1);
+#endif
+		if (bc->build_mode == BuildMode_Assembly || bc->build_mode == BuildMode_LLVM_IR) {
+			gb_printf_err("-lto:thin is incompatible with -build-mode:asm and -build-mode:llvm-ir\n");
```
laytan, line 2046: Could also be -build-mode:llvm, -build-mode:assembly, or -build-mode:assembler

src/llvm_backend.cpp
```diff
@@ -171,15 +171,15 @@ gb_internal void lb_correct_entity_linkage(lbGenerator *gen) {
 		LLVMValueRef other_global = nullptr;
 		if (ec.e->kind == Entity_Variable) {
 			other_global = LLVMGetNamedGlobal(ec.other_module->mod, ec.cname);
-			if (other_global) {
+			if (other_global && (LLVMGetInitializer(other_global) != nullptr || LLVMIsExternallyInitialized(other_global))) {
```
laytan, line 174: This kinda looks like a bodge. As well as the procedure stuff below. I can't put my finger on it though. Why would it matter if the global has an initializer/body?

src/main.cpp
```diff
@@ -643,6 +644,7 @@ gb_internal bool parse_build_flags(Array<String> args) {
 #endif
 
 	add_flag(&build_flags, BuildFlag_Sanitize,                str_lit("sanitize"),                  BuildFlagParam_String,  Command__does_build, true);
+	add_flag(&build_flags, BuildFlag_LTO,                     str_lit("lto"),                       BuildFlagParam_String,  Command__does_build);
```
laytan, line 647: I'm missing an addition to the help output explaining the flag.

PR #6209, opened 2026-02-01
Add TBAA metadata, function attributes, readonly, and re-enable SROA passes

JesseRMeyer:
The TBAA portions may be flat-out rejected due to aliasing concerns, but the "Fix sret parameter" I think is a genuine bug fix across the board, but it was mandatory to finish this implementation.

- TBAA (Type-Based Alias Analysis) metadata on loads/stores, with byte-type exemption for u8/i8 (matching C's char aliasing rule)
- Function attributes: nonnull on non-nil-able pointer returns, noalias on allocator returns, nocapture/readonly on appropriate parameters
- readonly on ABI-indirect value parameters (non-pointer types passed by ref)
- Fix sret parameter attribute offset: use abi_function_type->ret.kind to detect indirect returns instead of pt->Proc.return_by_pointer (which was never set), preventing readonly from landing on the sret slot
- Re-enable SROA passes that were previously disabled

### DIFF
--- a/src/llvm_backend_general.cpp
+++ b/src/llvm_backend_general.cpp
@@ -127,10 +127,6 @@ gb_internal WORKER_TASK_PROC(lb_init_module_worker_proc) {
 		LLVMMetadataRef root_str = LLVMMDStringInContext2(m->ctx, "Odin TBAA", 9);
 		m->tbaa_root = LLVMMDNodeInContext2(m->ctx, &root_str, 1);
 
-		LLVMMetadataRef omni_str = LLVMMDStringInContext2(m->ctx, "omnipotent char", 15);
-		LLVMMetadataRef omni_ops[3] = { omni_str, m->tbaa_root, LLVMValueAsMetadata(LLVMConstInt(LLVMInt64TypeInContext(m->ctx), 0, false)) };
-		m->tbaa_omnipotent = LLVMMDNodeInContext2(m->ctx, omni_ops, 3);
-
 		m->tbaa_kind_id = LLVMGetMDKindIDInContext(m->ctx, "tbaa", 4);
 		map_init(&m->tbaa_access_tags);
 	}
--- a/src/llvm_backend.hpp
+++ b/src/llvm_backend.hpp
@@ -190,7 +190,6 @@ struct lbModule {
 
 	// TBAA metadata
 	LLVMMetadataRef tbaa_root;
-	LLVMMetadataRef tbaa_omnipotent;
 	unsigned        tbaa_kind_id;
 	RecursiveMutex  tbaa_mutex;
 	PtrMap<Type *, LLVMMetadataRef> tbaa_access_tags;
--- a/src/llvm_backend_utility.cpp
+++ b/src/llvm_backend_utility.cpp
@@ -158,7 +158,7 @@ gb_internal lbValue lb_emit_min(lbProcedure *p, Type *t, lbValue x, lbValue y) {
 		LLVMValueRef args[2] = {x.value, y.value};
 		LLVMTypeRef types[1] = {lb_type(p->module, t)};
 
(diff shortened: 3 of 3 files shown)


laytan:
> Re-enable SROA passes that were previously disabled

I haven't read the diff yet but these have been enabled for a while now

JesseRMeyer:
Yes the commit message is too strong (and the comment was wrong too), although, it would appear we were missing instances of it in the pass pipelines (unconfirmed if important).

laytan:
Interesting, what we do with llvm updates is take the passes clang uses and remove coroutine and exception handling related passes. So if we are missing passes they would be missing from clang too.

JesseRMeyer:
I did ask Claude to consider "optimization passes that would most benefit the semantics and typical code patterns of Odin", but that doesn't mean that these new passes are helpful.

src/llvm_backend_general.cpp
```diff
@@ -1089,6 +1101,130 @@ gb_internal void lb_addr_store(lbProcedure *p, lbAddr addr, lbValue value) {
...
+	MUTEX_GUARD(&m->tbaa_mutex);
+
+	LLVMMetadataRef *found = map_get(&m->tbaa_access_tags, type);
+	if (found) {
+		return *found;
+	}
+
+	LLVMMetadataRef type_node = nullptr;
+
+	if (type->kind == Type_Basic) {
```
gingerBill, line 1135: I probably would have made this a switch statement for consistency but that can be changed in the future.

src/llvm_backend_proc.cpp
```diff
@@ -278,6 +290,19 @@ gb_internal lbProcedure *lb_create_procedure(lbModule *m, Entity *entity, bool i
 					lb_add_nocapture_proc_attribute_at_index(p, offset+parameter_index);
 				}
 			}
+			// readonly on non-pointer types passed indirectly by ABI
+			if (!is_type_pointer(e->type) && !is_type_multi_pointer(e->type) && !is_type_rawptr(e->type) &&
```
gingerBill, line 294: `if (!is_type_pointer_like(e->type)) && !is_type_soa_pointer(e->type)`

I am not sure if the soa-pointer thing is even needed too, because it's a ptr+index.

src/llvm_backend_proc.cpp
```diff
@@ -278,6 +290,19 @@ gb_internal lbProcedure *lb_create_procedure(lbModule *m, Entity *entity, bool i
...
+			    !is_type_proc(e->type) && !is_type_soa_pointer(e->type)) {
+				lbFunctionType *ft = p->abi_function_type;
+				if (ft != nullptr && parameter_index < ft->args.count && ft->args[parameter_index].kind == lbArg_Indirect) {
+					unsigned llvm_param_idx = cast(unsigned)(offset + parameter_index - 1);
```
gingerBill, line 298: Is the `- 1` correct in all cases?

src/llvm_backend_passes.cpp
```diff
@@ -17,6 +17,7 @@ inferattrs,
 function<eager-inv>(
 	lower-expect,
 	simplifycfg<bonus-inst-threshold=1;no-forward-switch-cond;no-switch-range-to-icmp;no-switch-to-lookup;keep-loops;no-hoist-common-insts;no-sink-common-insts;speculate-blocks;simplify-cond-branch>,
+	sroa<modify-cfg>,
```
gingerBill, line 22: Is this only re-enabled for version 17 of LLVM or all of them?

laytan: Looks like it yes.

I don't recommend doing it, even though it may be fine, we don't have anyone actively testing this version, and it's not in the CI.

gingerBill:
This needs rewriting a bit now since it has too many conflicts.

JesseRMeyer:
I'll see if I can peel away fixes separate from features.

PR #6210, opened 2026-02-01
Fix debug info sizes, endian flags, and add debug locations for loops / defer

JesseRMeyer:
This I feel the least confident regarding correctness.  But there may be ideas here worth grabbing.

- Fix big-endian float debug types using wrong DI flag (LittleEndian→BigEndian)
- Fix complex32/quaternion64 debug info sizes (were doubled)
- Fix complex64 imag field offset (2*32→1*32)
- Fix quaternion256 debug alignment (32→64)
- Fix string/string16/any debug info to use ptr_bits for field offsets
- Fix SoaPointer debug type to use ptr_bits instead of int_bits
- Add debug locations at loop back-edges and switch cases
- Track source position in deferred procedure calls for accurate debug info
- Fix union constant VLA initialization order
- Fix integer division-by-zero trap fallthrough for constant zero divisor

### DIFF
--- a/src/llvm_backend_debug.cpp
+++ b/src/llvm_backend_debug.cpp
@@ -729,22 +729,22 @@ gb_internal LLVMMetadataRef lb_debug_type_internal(lbModule *m, Type *type) {
 		case Basic_i128be: return lb_debug_type_basic_type(m, str_lit("i128be"), 128, LLVMDWARFTypeEncoding_Signed,   LLVMDIFlagBigEndian);
 		case Basic_u128be: return lb_debug_type_basic_type(m, str_lit("u128be"), 128, LLVMDWARFTypeEncoding_Unsigned, LLVMDIFlagBigEndian);
 
-		case Basic_f16be: return lb_debug_type_basic_type(m,  str_lit("f16be"),   16, LLVMDWARFTypeEncoding_Float,    LLVMDIFlagLittleEndian);
-		case Basic_f32be: return lb_debug_type_basic_type(m,  str_lit("f32be"),   32, LLVMDWARFTypeEncoding_Float,    LLVMDIFlagLittleEndian);
-		case Basic_f64be: return lb_debug_type_basic_type(m,  str_lit("f64be"),   64, LLVMDWARFTypeEncoding_Float,    LLVMDIFlagLittleEndian);
+		case Basic_f16be: return lb_debug_type_basic_type(m,  str_lit("f16be"),   16, LLVMDWARFTypeEncoding_Float,    LLVMDIFlagBigEndian);
+		case Basic_f32be: return lb_debug_type_basic_type(m,  str_lit("f32be"),   32, LLVMDWARFTypeEncoding_Float,    LLVMDIFlagBigEndian);
+		case Basic_f64be: return lb_debug_type_basic_type(m,  str_lit("f64be"),   64, LLVMDWARFTypeEncoding_Float,    LLVMDIFlagBigEndian);
 
 		case Basic_complex32:
 			{
 				LLVMMetadataRef elements[2] = {};
 				elements[0] = lb_debug_struct_field(m, str_lit("real"), t_f16, 0*16);
 				elements[1] = lb_debug_struct_field(m, str_lit("imag"), t_f16, 1*16);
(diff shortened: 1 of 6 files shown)


src/llvm_backend_debug.cpp
```diff
@@ -794,8 +794,8 @@ gb_internal LLVMMetadataRef lb_debug_type_internal(lbModule *m, Type *type) {
 			{
 				LLVMMetadataRef elements[2] = {};
 				elements[0] = lb_debug_struct_field(m, str_lit("data"), t_u8_ptr, 0);
-				elements[1] = lb_debug_struct_field(m, str_lit("len"),  t_int, int_bits);
-				return lb_debug_basic_struct(m, str_lit("string"), 2*int_bits, int_bits, elements, gb_count_of(elements));
+				elements[1] = lb_debug_struct_field(m, str_lit("len"),  t_int, ptr_bits);
+				return lb_debug_basic_struct(m, str_lit("string"), ptr_bits+int_bits, int_bits, elements, gb_count_of(elements));
```
gingerBill, line 798: This change is wrong because the `ptr_size` might be smaller than the `int_size`. The previous logic was correct. `int` has the same alignment as its size, thus there will be padding.

src/llvm_backend_debug.cpp
```diff
@@ -807,8 +807,8 @@ gb_internal LLVMMetadataRef lb_debug_type_internal(lbModule *m, Type *type) {
 			{
 				LLVMMetadataRef elements[2] = {};
 				elements[0] = lb_debug_struct_field(m, str_lit("data"), t_u16_ptr, 0);
-				elements[1] = lb_debug_struct_field(m, str_lit("len"),  t_int, int_bits);
-				return lb_debug_basic_struct(m, str_lit("string16"), 2*int_bits, int_bits, elements, gb_count_of(elements));
+				elements[1] = lb_debug_struct_field(m, str_lit("len"),  t_int, ptr_bits);
+				return lb_debug_basic_struct(m, str_lit("string16"), ptr_bits+int_bits, int_bits, elements, gb_count_of(elements));
```
gingerBill, line 811: This change is wrong because the `ptr_size` might be smaller than the `int_size`. The previous logic was correct. `int` has the same alignment as its size, thus there will be padding.

src/llvm_backend_debug.cpp
```diff
@@ -818,10 +818,12 @@ gb_internal LLVMMetadataRef lb_debug_type_internal(lbModule *m, Type *type) {
...
 		case Basic_any:
 			{
+				unsigned typeid_bits = 8*cast(unsigned)type_size_of(t_typeid);
+				unsigned any_align = cast(unsigned)gb_max(ptr_bits, typeid_bits);
 				LLVMMetadataRef elements[2] = {};
 				elements[0] = lb_debug_struct_field(m, str_lit("data"), t_rawptr, 0);
-				elements[1] = lb_debug_struct_field(m, str_lit("id"),   t_typeid, 64);
-				return lb_debug_basic_struct(m, str_lit("any"), 128, 64, elements, gb_count_of(elements));
+				elements[1] = lb_debug_struct_field(m, str_lit("id"),   t_typeid, ptr_bits);
+				return lb_debug_basic_struct(m, str_lit("any"), ptr_bits+typeid_bits, any_align, elements, gb_count_of(elements));
```
gingerBill, line 826: This change is wrong because the `ptr_size` might be smaller than the `typeid`. `typeid` is defined to be 64 bits on ALL platforms, and requires 8-byte alignment too.

src/llvm_backend_expr.cpp
```diff
@@ -1377,6 +1377,8 @@ gb_internal LLVMValueRef lb_integer_modulo(lbProcedure *p, LLVMValueRef lhs, LLV
 	if (LLVMIsConstant(rhs)) {
 		if (LLVMIsNull(rhs)) {
 			switch (behaviour) {
+			case IntegerDivisionByZero_Trap:
+				break; // fall through to runtime path which emits trap
```
gingerBill, line 1381: A little redundant but whatever. The comment is also a little confusing because it's saying "fall through" on a `break`. I understand the logic but still a little confusing.

gingerBill:
I will probably use this PR instead: https://github.com/odin-lang/Odin/pull/6216

PR #6211, opened 2026-02-01
Fix front-end threading bugs and remove unnecessary serialization

JesseRMeyer:
This is another PR for tsan related fixes but should be completely independent and not collide.  

- Remove broken are_offsets_being_processed early-return in type_set_offsets that let concurrent threads read uninitialized struct offsets
- Fix type_set_offsets fall-through returning false when offsets already set
- Add shared lock on parent scope in scope_insert_with_name to prevent data race when reading parent's elements map
- Remove unnecessary global mutex from hash_exact_value (pure function)
- Convert error collector curr_error_value_set from atomic to plain bool (all access already under mutex)
- Add threading-safety comments for phase-ordered traversals

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -495,15 +495,18 @@ gb_internal Entity *scope_insert_with_name(Scope *s, String const &name, Entity 
 		goto end;
 	}
 	if (s->parent != nullptr && (s->parent->flags & ScopeFlag_Proc) != 0) {
+		rw_mutex_shared_lock(&s->parent->mutex);
 		found = string_map_get(&s->parent->elements, key);
 		if (found) {
 			if ((*found)->flags & EntityFlag_Result) {
 				if (entity != *found) {
 					result = *found;
 				}
+				rw_mutex_shared_unlock(&s->parent->mutex);
 				goto end;
 			}
 		}
+		rw_mutex_shared_unlock(&s->parent->mutex);
 	}
 
 	string_map_set(&s->elements, key, entity);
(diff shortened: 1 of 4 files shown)


src/checker.cpp
```diff
@@ -727,6 +730,8 @@ gb_internal bool check_vet_unused(Checker *c, Entity *e, VettedEntity *ve) {
 	return false;
 }
 
+// NOTE(threading): check_scope_usage runs after all checking is complete (vet phase),
```
gingerBill, line 733: Comment here should not be `NOTE(threading):` but the name of the author so that people can refer back to that individual if necessary. 

I'd do this if you want to point out the threading: 
```odin
// NOTE(SHORT NAME HERE): Threading: ...
```

src/error.cpp
```diff
@@ -22,28 +22,28 @@ struct ErrorCollector {
 
 	Array<ErrorValue> error_values;
 	ErrorValue        curr_error_value;
-	std::atomic<bool> curr_error_value_set;
+	bool curr_error_value_set;
```
gingerBill, line 25: Why was not removed from people `std::atomic<bool>`?

src/exact_value.cpp
```diff
@@ -54,9 +53,6 @@ struct ExactValue {
 gb_global ExactValue const empty_exact_value = {};
 
 gb_internal uintptr hash_exact_value(ExactValue v) {
-	mutex_lock(&hash_exact_value_mutex);
-	defer (mutex_unlock(&hash_exact_value_mutex));
```
gingerBill, line 58: I am not sure if this is necessary or not still. It "might" be?

src/types.cpp
```diff
@@ -4308,26 +4308,23 @@ gb_internal i64 *type_set_offsets_of(Slice<Entity *> const &fields, bool is_pack
 gb_internal bool type_set_offsets(Type *t) {
 	t = base_type(t);
 	if (t->kind == Type_Struct) {
-		if (t->Struct.are_offsets_being_processed.load()) {
-			return true;
-		}
```
gingerBill, line 4313: What does the removal of this fix?

PR #6229, opened 2026-02-04
Add uppercase flag to hex encoding procs

slowhei:
Thoughts about adding an `uppercase` flag to `encode` and `encode_into_writer` in `core:encoding/hex`?

As we can see from the crypto tests I had to modify, this is a breaking change.

### DIFF
--- a/tests/core/encoding/hex/test_core_hex.odin
+++ b/tests/core/encoding/hex/test_core_hex.odin
@@ -23,6 +23,19 @@ hex_encode :: proc(t: ^testing.T) {
 			test[1],
 		)
 	}
+
+	input  := "Hello"
+	output := "48656C6C6F"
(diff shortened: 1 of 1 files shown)


gingerBill:
Is this necessary anywhere?

slowhei:
My use case is displaying binary data in a human-readable hex encoding, and I have a convention of printing hex data in uppercase. I think my use case is simple and not unusual, so not being able to do so with `core:encoding/hex` felt like a shortcoming of the package.

While I think this change is worthwhile, no, this change is not strictly necessary and comes at the cost of breaking a bit of user code.

ske2004:
I would also like this, though wouldn't it be simpler to just add `encode_uppercase` function?

Kelimion:
I'm adding `encode_upper` and `encode_upper_into_writer` in #6275.

PR #6232, opened 2026-02-05
Document atomic synchronization primitives and their differences

lucypero:
## Summary

- Added comprehensive documentation to `Atomic_Mutex`, `Atomic_RW_Mutex`, `Atomic_Cond`, `Atomic_Sema`, and `Atomic_Recursive_Mutex` explaining how they differ from the standard primitives
- Added a "Standard vs Atomic Primitives" section to the package-level documentation in `doc.odin`

The documentation explains:
- **Implementation differences**: Atomic primitives use futexes and atomic operations, while standard primitives may use OS-native implementations (e.g., `SRWLOCK` on Windows)
- **Memory footprint**: Atomic primitives are generally more compact (just a `Futex`/`u32`)
- **Cross-platform consistency**: Atomic primitives guarantee identical behavior across all platforms
- **Compatibility requirements**: e.g., `Atomic_Cond` must be used with `Atomic_Mutex`, not the standard `Mutex`

## Test plan

- [x] Documentation-only change, no code modifications
- [x] Verified documentation renders correctly in Odin doc format

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/sync/doc.odin
+++ b/core/sync/doc.odin
@@ -13,5 +13,32 @@ and blocking the execution of any other threads.
 In lock-free programming the data itself is organized in such a way that threads
 don't intervene much. It can be done via segmenting the data between threads,
 and/or by using atomic operations.
+
+## Standard vs Atomic Primitives
+
+This package provides two families of synchronization primitives:
+
+**Standard primitives** (`Mutex`, `RW_Mutex`, `Cond`, `Sema`, `Recursive_Mutex`):
+These use OS-specific implementations where available. On Windows, they use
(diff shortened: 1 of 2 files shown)


gingerBill:
Can you please not use LLMs to document really important things like this? The problem is that it might _sound correct_ to you but it now means we have to go and verify it ourselves to see if it is even correct rather than it "sounding correct". 

LLMs, including Claude, also have the habit of being way too verbose for things it doesn't need.

I think we need to make a ban on accepting LLM based documentation.

PR #6233, opened 2026-02-05
Bifrost

delaneyj:


gingerBill:
I am going to probably to close this even though it is a draft. 

It appears to be written mostly by an LLM, and not even that well designed in terms of the architecture of the thing.

Will you even maintain it when there are inevitable bugs or flaws with it? I highly doubt it.

And why the heck have you put a http package in `vendor` and not `core` and named it `bifrost` when it is just a http package. 

If you want to develop this elsewhere, please do, but I will not be accepting this into Odin just yet until we have even proven it works.

PR #6235, opened 2026-02-05
Allow test names to be specified as a command line option for tests executable

harold-b:
When running the tests executable generated by `odin build -build-mode:test`, this allows to optionally specify a  set of tests to run at runtime via the command line, instead of being limited only to specifying them as compile-time definitions.

Compile-time definitions still allowed, but the cli list, if present, takes precedence.

### DIFF
--- a/core/testing/runner.odin
+++ b/core/testing/runner.odin
@@ -213,7 +213,7 @@ parse_cli_options :: proc(argv: []string, opts: ^Options, stdout, stderr: io.Wri
 				os.exit(-1)
 			}
 
-			if len(test_names.buf) > 0 {
+			if strings.builder_len(test_names) > 0 {
 				strings.write_byte(&test_names, ',')
 			}
 			strings.write_string(&test_names, tests)


gingerBill:
When we've moved to `os2`, I'll probably need to merge this (also change that import).

harold-b:
No problem. I'll keep an eye out for when `os2` becomes `os` and fix this PR.

core/testing/runner.odin
```diff
@@ -196,6 +196,54 @@ run_test_task :: proc(task: thread.Task) {
...
+			tests := arg[strings.index(arg, ":")+1:]
+			if len(tests) < 1 {
+				fmt.wprintln(stderr, "No test names specified for '-tests:'")
+				os.exit(-1)
+			}
+
+			if len(test_names.buf) > 0 {
+				strings.write_byte(&test_names, ',')
+			}
+			strings.write_bytes(&test_names, transmute([]u8)tests)
```
laytan:
```suggestion
			strings.write_string(&test_names, tests)
```

core/testing/runner.odin
```diff
@@ -196,6 +196,54 @@ run_test_task :: proc(task: thread.Task) {
...
+Options :: struct {
+	// Equivalent to the TEST_NAMES compile-time definition, but used dynamically at runtime.
+	test_names: string,
+}
+
+parse_cli_options :: proc(argv: []string, opts: ^Options, stdout, stderr: io.Writer) {
+	test_names: strings.Builder
+
+	for i := 0; i < len(argv); i+=1 {
+		arg := argv[i]
```
laytan:
```suggestion
	for arg in argv {
```

core/testing/runner.odin
```diff
@@ -196,6 +196,54 @@ run_test_task :: proc(task: thread.Task) {
...
+				exe_name = slashpath.base(path)
+			}
+
+			fmt.wprintfln(stdout, "Usage: %v [OPTIONS]", exe_name)
+			fmt.wprintfln(stdout, "OPTIONS:")
+			fmt.wprintfln(stdout, "        -help:")
+			fmt.wprintfln(stdout, "            Display this help text and exit.")
+			fmt.wprintln(stdout)
+			fmt.wprintfln(stdout, "        -tests:<test_name[,...]>")
+			fmt.wprintfln(stdout, "            Specify a specific set of tests to run by name.\n" +
```
laytan:
```suggestion
			fmt.wprintfln(stdout, "OPTIONS:")
			fmt.wprintln(stdout, "        -help:")
			fmt.wprintln(stdout, "            Display this help text and exit.")
			fmt.wprintln(stdout)
			fmt.wprintln(stdout, "        -tests:<test_name[,...]>")
			fmt.wprintln(stdout, "            Specify a specific set of tests to run by name.\n" +
```

core/testing/runner.odin
```diff
@@ -196,6 +196,54 @@ run_test_task :: proc(task: thread.Task) {
...
+			                      "            The format is: `package.test_name,test_name_only,...`")
+			fmt.wprintln(stdout)
+			os.exit(0)
+		} else {
+			fmt.wprintfln(stderr, "Unknown argument encountered '%v'", arg)
+			os.exit(-1)
+		}
+	}
+
+	opts.test_names = string(test_names.buf[:])
```
laytan:
```suggestion
	opts.test_names = strings.to_string(test_names)
```

core/testing/runner.odin
```diff
@@ -196,6 +196,54 @@ run_test_task :: proc(task: thread.Task) {
...
+	test_names: string,
+}
+
+parse_cli_options :: proc(argv: []string, opts: ^Options, stdout, stderr: io.Writer) {
+	test_names: strings.Builder
+
+	for i := 0; i < len(argv); i+=1 {
+		arg := argv[i]
+		if strings.starts_with(arg, "-tests:") {
+			tests := arg[strings.index(arg, ":")+1:]
```
laytan:
```suggestion
			_, _, tests := strings.partition(arg, ":")
```

core/testing/runner.odin
```diff
@@ -196,6 +196,54 @@ run_test_task :: proc(task: thread.Task) {
...
+			if len(tests) < 1 {
+				fmt.wprintln(stderr, "No test names specified for '-tests:'")
+				os.exit(-1)
+			}
+
+			if len(test_names.buf) > 0 {
```
laytan:
```suggestion
			if strings.builder_len(test_names) > 0 {
```

core/testing/runner.odin
```diff
@@ -196,6 +196,54 @@ run_test_task :: proc(task: thread.Task) {
...
+				strings.write_byte(&test_names, ',')
+			}
+			strings.write_bytes(&test_names, transmute([]u8)tests)
+		} else if arg == "-help" {
+			exe_name := "test"
+			if path, err := os.get_executable_path(context.temp_allocator); err == nil {
+				exe_name = slashpath.base(path)
+			}
+
+			fmt.wprintfln(stdout, "Usage: %v [OPTIONS]", exe_name)
```
laytan, line 226: This should probably use `args[0]` instead of the executable name, `args[0]` would actually reflect what the user typed.

(you would need to adjust line 288 too, which currently slices `args[1:]`)

core/testing/runner.odin
```diff
@@ -205,8 +205,7 @@ Options :: struct {
 parse_cli_options :: proc(argv: []string, opts: ^Options, stdout, stderr: io.Writer) {
 	test_names: strings.Builder
 
-	for i := 0; i < len(argv); i+=1 {
-		arg := argv[i]
+	for arg in argv {
```
harold-b, line 208: I had it in a traditional `for` loop because I was planning on bumping `i` to read the next arg as a value for an option, but I updated it to reflect it to the Odin format (`-opt:value`) and forgot to revert this, thanks.

core/testing/runner.odin
harold-b: And this is this reliable across all platforms?

laytan: It is not reliable to get the executable name, but here you don't want the executable name, you want to get what needs to be typed to execute the program.

For example, if somebody were to symlink this executable from `foo.exe` to `bar.exe` and calls it with `bar.exe -help`, your changes would return `foo`. But typing `foo` wouldn't actually work.

harold-b: Right, I understand what you're suggesting, I just wanted to know if `arg[0]` as the executable path was available in all platforms, as  I don't know the answer to this.

I do think you typically want the executable name, though, as that is what the help is for.

For example, if I have an alias for git named `foogit` and I do `foogit help` I would expect the documentation to reference `git` not `foogit`.

laytan: `arg[0]` as the executable path is not reliable, no.

I was under the impression you would expect "Usage: foogit [OPTIONS]" because that is how you have configured it to be called.

I could be wrong of course, maybe look at what others do?

laytan: Well, I would maybe not expect "foogit" when it's a shell alias, but a symlink.
Or when you do "~/Odin/odin help", I think you'd get "Usage: ~/Odin/odin ..."

harold-b: I see the executable as the identity of the app, and they can often have usage examples referring back to the executable name of app (not any aliases or links). Which is why I would expect the identity name of the exe to show up in the main line.

However, this is in the context of an ephemeral test host exe, where `args[0]` and `get_executable_path` are likely to be exactly the same 99.99% of the time, so I don't have a strong opinion in this case. I'll yield to whatever you prefer.

laytan: I think there's a difference between identity and usage. In this use case it's "Usage: <foo>" and I would expect it to print something I can actually use (copy and paste to run the executable).

For Identity, like having a "foo is a program that does bar", I'd agree you want the actual executable name, not how the user runs it.

I'd say let's do what the compiler does (not at pc to check).

PR #6236, opened 2026-02-05
parser: add support for `switch "partial"` syntax

IDONTUSEGH:
Allow `switch "partial"` syntax to mark partial switches, as an alternative to the existing `#partial switch` directive.
This mimics the `proc "contextless"` syntax for better consistency.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -1062,23 +1062,23 @@ gb_internal Ast *ast_unroll_range_stmt(AstFile *f, Token unroll_token, Slice<Ast
 	return result;
 }
 
-gb_internal Ast *ast_switch_stmt(AstFile *f, Token token, Ast *init, Ast *tag, Ast *body) {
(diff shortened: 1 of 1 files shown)


gingerBill:
No because this is both not necessarily and literally misunderstands that `switch "partial" {` is already VALID syntax.

IDONTUSEGH:
That’s unfortunate; when reading code, the # symbol makes it easy to skip the switch statement entirely, especially if the scopes are folded.

What about copying `#no_bounds_check`?

```
switch foo #partial {
..
}
```

gingerBill:
Why are you suggesting a different syntax for something which already exists and works and is consistent with the rest of the language?

"Copying" `#no_bounds_check` also misunderstands what it is doing. `#partial` is applied to the `switch` itself, not the `{}` of the `switch`.

IDONTUSEGH:
I apologize for the noise; in hindsight, a Discussion would have been the right move here.

PR #6241, opened 2026-02-06
core/fmt: fix incorrect string padding

zen3ger:
Padding for format strings such as "% -6s" incorrectly calculated the length of the argument string - bytes instead of UTF-8 runes - resulting in the padding not being present for input such as "μμμμμ".

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -1495,12 +1495,13 @@ fmt_string :: proc(fi: ^Info, s: string, verb: rune) {
 	switch verb {
 	case 's', 'v':
 		if fi.width_set {
-			if fi.width > len(s) {
+			length := utf8.rune_count_in_string(s)
+			if fi.width > length {
 				if fi.minus {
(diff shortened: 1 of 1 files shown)


OnlyXuul:
That works for multi byte runes, but not multi-rune monospace characters like ö or á. I humbly recommend the following instead:
`_, _, length := utf8.grapheme_count(s)`

grapheme_count I believe would handle all utf8 characters, but not utf16 or utf32. That said, I believe most if not all terminals display utf8 the same. Many differ on utf16 and utf32 because of poor universal font definitions (and multi-emoji's, etc) above utf8. So IMHO, those could be ignored, but I mention it for edification.

And that's just speaking from experience with Latin/Western characters. I have no clue about Chinese/Japanese/Indian, etc fonts or otherwise.

gingerBill:
I'm going to close this because of how difficult it is to actually do "correctly" and I will think about how to do it "correctly" at a later date.

PR #6244, opened 2026-02-08
core:container/handle_map: fix dynamic_add not updating indicies nor …

Mufferaw:
…keeping track of pointers

dynamic_add was using the return from xar.append() (which always returns 1) to get the pointer and set the idx for the handle


gingerBill:
I have just manually fixed it by doing a one line change.

Mufferaw:
I think that needs to be:
i := xar.len(m.items)-1

Issue #6161, opened 2026-01-24
`vendor:box2d` Windows linker issues


blob1807:
## Context
```
Odin:    dev-2026-01:7e39669fd
OS:      Windows 11 Home Basic (version: 24H2), build 26100.7623
CPU:     AMD Ryzen 7 6800HS with Radeon Graphics
RAM:     15608 MiB
Backend: LLVM 20.1.0
```

## Current Behavior
When Box2D was updated to 3.1.1. The binaries were complied with `/DEFAULTLIB:msvcrt`. As opposed to `libcmt`, which it was last time and what Odin uses. Resulting in the following linker errors:
```
LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library
LINK : warning LNK4217: symbol 'fclose' defined in 'libucrt.lib(fclose.obj)' is imported by 'box2d_windows_amd64_sse2.lib(world.c.obj)' in function 'b2World_DumpMemoryStats'
LINK : warning LNK4217: symbol '__stdio_common_vfprintf' defined in 'libucrt.lib(output.obj)' is imported by 'box2d_windows_amd64_sse2.lib(world.c.obj)' in function '_vfprintf_l'
LINK : warning LNK4286: symbol '__stdio_common_vfprintf' defined in 'libucrt.lib(output.obj)' is imported by 'box2d_windows_amd64_sse2.lib(core.c.obj)'
LINK : warning LNK4217: symbol '__stdio_common_vsprintf' defined in 'libucrt.lib(output.obj)' is imported by 'box2d_windows_amd64_sse2.lib(world.c.obj)' in function 'snprintf'
LINK : warning LNK4286: symbol '__stdio_common_vsprintf' defined in 'libucrt.lib(output.obj)' is imported by 'box2d_windows_amd64_sse2.lib(revolute_joint.c.obj)'
LINK : warning LNK4217: symbol '__acrt_iob_func' defined in 'libucrt.lib(_file.obj)' is imported by 'box2d_windows_amd64_sse2.lib(core.c.obj)' in function 'printf'
LINK : warning LNK4217: symbol 'qsort' defined in 'libucrt.lib(qsort.obj)' is imported by 'box2d_windows_amd64_sse2.lib(sensor.c.obj)' in function 'b2SensorTask'
[...]

PR #6249, opened 2026-02-09
fix box2d build and try Git LFS

laytan:
Fixes https://github.com/odin-lang/Odin/issues/6161

Box2D compiled with `-DCMAKE_MSVC_RUNTIME_LIBRARY="MultiThreaded"`

Also giving Git LFS a go for binaries, seeing if we like it for being batteries included but also not blowing up the repository size.


laytan:
Probably want to enable this @gingerBill https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-git-lfs-objects-in-archives-of-your-repository

laytan:
https://github.com/git-lfs/git-lfs/wiki/Limitations

PR #6253, opened 2026-02-10
Simple import graph generator

jakubtomsu:
This implements a very simple import graph output in the graphviz format. All it does is traverse the parser packages and their import declarations.

The output is printed into stderr when `-show-debug-messages` is enabled. My reasoning is that this flag isn't for most users anyway and already prints all kinds of info. I'm not sure a feature like this is important enough to get it's own flag. The output can be copy-pasted pretty easily.

Here how the graph can look like:
image
Each node is a package, and built-in collections get simple colors.

I already have a [Gist](https://gist.github.com/jakubtomsu/730a4343ba0137388d789f507857b9d7) which does something similar using `core:odin/parser`. But there are two main reasons why this is much better:
- resolving collections and packages matches the compiler 100%
- things like file suffixes are considered, so the graph isn't as spammy

In case this isn't wanted directly in the compiler that's fine. But if people need it they can at least refer back to this PR.

@Kelimion mentioned this would have been useful with some recent import bugs, so maybe it can be generally useful to more people? In my case I mostly want to keep my dependencies under control since I'm writing a library.

My PR from yesterday (https://github.com/odin-lang/Odin/pull/6250) is mostly thanks to graphs like this, it makes it so much easier to see what's going on in a project as a whole. By the way, I use a VS Code extension for viewing the actual graphs, because unlike most websites it lets me click on nodes to highlight the dependencies.

One more note: I generally use a command like `odin check my_package -show-debug-messages` to get the graph, but that errors out when `my_package` doesn't have an entry point. For this reason the graph output is placed after all parsing is done but before typechecking.

Just for reference, here's the graph from the image:
<details>
<summary>Graph File</summary>

```dot
digraph odin_import_graph {
        node [shape=box];
        subgraph cluster_0 {
                label = "base";
                node [style=filled, fillcolor=white];
                bgcolor = lightcoral;
                intrinsics;
                builtin;
                "D:/projects/Odin/base/runtime";
                "D:/projects/Odin/base/sanitizer";
        }
        subgraph cluster_1 {
                label = "core";
                node [style=filled, fillcolor=white];
                bgcolor = lightsalmon;
                "D:/projects/Odin/core/sys/windows";
                "D:/projects/Odin/core/hash/xxhash";
                "D:/projects/Odin/core/c";
        }
        subgraph cluster_2 {
                label = "vendor";
                node [style=filled, fillcolor=white];
                bgcolor = lightblue;
                "D:/projects/Odin/vendor/wgpu";
[...]

PR #6254, opened 2026-02-10
More import cleanup

jakubtomsu:
I looked into cleaning up dependencies of some more packages, continuing the work from yesterday (https://github.com/odin-lang/Odin/pull/6250).
I found 3 things I would consider very low-hanging fruit, again a dependency was pulled in for basically a single trivial procedure.

Changes:
- `xxhash` had a dependency on `core:mem` just for a `runtime.new_aligned` wrapper used in a single procedure
- `bytes` only used the `mem.Allocator_Error` in few places and `mem.compare` in one place, so I replaced it with the appropriate equivalents from `runtime` and `intrinsics`.
- `debug/trace` used `fmt` for a single edge case in symbol resolve code path. And the linux path was clearly untested, there was a BUG! So I added a very simple procedure for printing the fallback procedure address. So now all the backends share the same code path. I think in this case this change is very worth it, `debug/trace` seems like a low-level package which should be relatively standalone.

All of this was found thanks to the graph generator from https://github.com/odin-lang/Odin/pull/6253.

### DIFF
--- a/core/debug/trace/trace_cpp.odin
+++ b/core/debug/trace/trace_cpp.odin
@@ -163,7 +163,7 @@ _resolve :: proc(ctx: ^Context, frame: Frame, allocator: runtime.Allocator) -> F
 			} else if info: Dl_info; dladdr(rawptr(address), &info) != 0 && info.dli_sname != "" {
 				frame.procedure = strings.clone_from_cstring(info.dli_sname, btc.allocator)
 			} else {
-				frame.procedure = _format_missing_proc(address, allocator)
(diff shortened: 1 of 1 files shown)


jakubtomsu:
I just found few more things right after this got merged hahah... so I'll submit the rest some other time

Kelimion:
> I just found few more things right after this got merged hahah... so I'll submit the rest some other time

Thanks again, Jakub. :D Sorry, I didn't know more was pending.

PR #6260, opened 2026-02-11
WIP  Remove core:mem dependency from core:strings

Faker-09:
This is a work in progress. Looking for some feedback.
Is this worth doing?

It is strings.compare() that I'm not certain about.  I added some tests to verify the results between using mem.compare() and runtime.memory_compare() and it tests ok.
Are there some other test string pairs that would be useful to test?

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -23,7 +23,7 @@ Returns:
 - res: The cloned string
 - err: An optional allocator error if one occured, `nil` otherwise
 */
-clone :: proc(s: string, allocator := context.allocator, loc := #caller_location) -> (res: string, err: mem.Allocator_Error) #optional_allocator_error {
+clone :: proc(s: string, allocator := context.allocator, loc := #caller_location) -> (res: string, err: runtime.Allocator_Error) #optional_allocator_error {
 	c := make([]byte, len(s), allocator, loc) or_return
 	copy(c, s)
 	return string(c), nil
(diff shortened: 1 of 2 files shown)


Kelimion:
I like the idea, and it makes sense to do this if you can get rid of all references to `core:mem`.

Those tests shouldn't be needed. `strings.compare` calls `mem.compare`, which calls `runtime.memory_compare`.
```odin
compare :: proc "contextless" (lhs, rhs: string) -> (result: int) {
	return mem.compare(transmute([]byte)lhs, transmute([]byte)rhs)
}
```
-> 
```odin
@(require_results)
compare :: proc "contextless" (a, b: []byte) -> int {
	res := compare_byte_ptrs(raw_data(a), raw_data(b), min(len(a), len(b)))
	if res == 0 && len(a) != len(b) {
		return len(a) <= len(b) ? -1 : +1
	} else if len(a) == 0 && len(b) == 0 {
		return 0
	}
	return res
}

@(require_results)
compare_byte_ptrs :: proc "contextless" (a, b: ^byte, n: int) -> int #no_bounds_check {
	return runtime.memory_compare(a, b, n)
}
```

So it looks like your `compare_rt` should be this:
```odin
compare :: proc "contextless" (lhs, rhs: string) -> (res: int) {
	a := transmute([]byte)lhs
	b := transmute([]byte)rhs
	res = runtime.memory_compare(raw_data(a), raw_data(b), min(len(a), len(b)))
	if res == 0 && len(a) != len(b) {
		return len(a) <= len(b) ? -1 : +1
	} else if len(a) == 0 && len(b) == 0 {
		return 0
	}
	return res
}
```

Faker-09:
Don't know how I missed that compare_byte_ptrs was runtime.memory_compare. I was looking at it. Doh!

Your suggestion is pretty much what I had to begin with but I noticed it tested ok without the if {} else if {} stuff and the transmutes. I assume there are some cases that require that? I'm curious what they are.

Thank you for taking a look.

PR #6261, opened 2026-02-11
Remove core:mem dependency from core:strings

Faker-09:


Kelimion:
How does this differ from #6260? If you closed that one by mistake, I could've reopened it. :-)

Faker-09:
Oh, I did close it by mistake.  :/ 
This one is the same as the other and uses your suggested compare.

Kelimion:
`intern.odin` and `conversion.odin` also import `core:mem` to reference `mem.Allocator_Error`.
`builder.odin` likewise, and it also calls `mem.buffer_from_slice`.

What you have here is a good start, but as long as these other files in the package also import `core:mem`, it still remains a dependency for `core:strings`.

Do you want to tackle these files as well, or can I offer some assistance with that (tomorrow, it's late).

Faker-09:
Yeah, I assumed something else would be using core:mem. But sure, I would like to look at those other files.
Good evening and thanks again.

core/strings/strings.odin
```diff
@@ -242,8 +241,16 @@ Inputs:
 Returns:
 - result: `-1` if `lhs` comes first, `1` if `rhs` comes first, or `0` if they are equal
 */
-compare :: proc "contextless" (lhs, rhs: string) -> (result: int) {
-	return mem.compare(transmute([]byte)lhs, transmute([]byte)rhs)
+compare :: proc "contextless" (lhs, rhs: string) -> (res: int) {
+	a := transmute([]byte)lhs
+	b := transmute([]byte)rhs
+	res = runtime.memory_compare(raw_data(a), raw_data(b), min(len(a), len(b)))
```
Barinzaya: Couldn't this just wrap `runtime.string_cmp`?

Kelimion: Well spotted. It could.

Kelimion:
I'm tackling the remaining core packages in #6264.

core/strings/strings.odin
```diff
@@ -242,7 +242,13 @@ Returns:
 - result: `-1` if `lhs` comes first, `1` if `rhs` comes first, or `0` if they are equal
 */
 compare :: proc "contextless" (lhs, rhs: string) -> (res: int) {
-	return runtime.string_cmp(lhs, rhs)
+	res = runtime.string_cmp(lhs, rhs)
+	if res == 0 && len(lhs) != len(rhs) {
+		return len(lhs) <= len(rhs) ? -1 : +1
+	} else if len(lhs) == 0 && len(rhs) == 0 {
+		return 0
+	}
```
Barinzaya: Do these actually have any effect? [`string_cmp` already does the first](https://github.com/odin-lang/Odin/blob/1159110e735ba84d651f4bbc4e9883fd83e9eddc/base/runtime/internal.odin#L541), and as best I can tell `memory_compare` should already return 0 if both lengths are 0.

Faker-09: They don't seem to have any effect in my limited testing. I assumed there was a case where this was needed?

Kelimion: Go ahead and make it an alias for `runtime.string_cmp`. No wrap, just `compare :: runtime.string_cmp`.

PR #6264, opened 2026-02-12
Replace trivial `core:mem` imports with `base:runtime`.

Kelimion:


Kelimion:
`core:strings` is being [worked on](##6261) by @Faker-09.
This PR will evaluate the remainder.

PR #6271, opened 2026-02-13
[core:container/pool]  Fix pool parapoly type error when using multiple Pool types

### LINKED ISSUE #6219: Compiler is confused about parapoly when using two pools
## Context

This is a simplified version of a problem I ran into while working with the `pool` container. For some reason the compiler is getting confused when I added the second pool with a different type.

Adding `-sanitize:address` results in it compiling without problems.

        Odin:    dev-2026-01:f7901cffc
        OS:      Ubuntu 24.04.3 LTS, Linux 6.14.0-37-generic
        CPU:     Intel(R) Xeon(R) CPU E5-1620 v4 @ 3.50GHz
        RAM:     23954 MiB
        Backend: LLVM 18.1.3

```
package main

import "core:container/pool"

@(private)
Bar :: struct {
    link: ^Bar,
}

@(private)
Foo :: struct {
    link: ^Foo,
}

fooPool: pool.Pool(Foo)
barPool: pool.Pool(Bar)

main :: proc() {
    if error := pool.init(&fooPool, "link"); error != nil {
        panic("")
    }

    if error := pool.init(&barPool, "link"); error != nil {
        panic("")
    }

    foo, err1 := pool.get(&fooPool)

    // Uncommenting the next line causes compilation failure:

    /*
/home/matt/odin/core/container/pool/pool.odin(73:19) Error: Cannot assign value 'p' of type '^Pool($T=Foo)' to '^Pool($T=Bar)' in a procedure argument 
    _unpoison_elem(p, elem) 
                  ^ 
 
/home/matt/odin/core/container/pool/pool.odin(73:22) Error: Cannot assign value 'elem' of type '^Foo' to '^Bar' in a procedure argument 
    _unpoison_elem(p, elem) 
                     ^~~^ 
 

    */

    bar, err2 := pool.get(&barPool)
}
```

### Steps to Reproduce
`odin run .` with the attached sample

oxSleep:
## Summary
  - Fix #6219
  - Replace @(disabled=...) attribute on _poison_elem and _unpoison_elem with when blocks inside the procedure body
  - The @(disabled) attribute causes the compiler's type inference to confuse generic parameters across different Pool instantiations

 I'm not sure if this is the intended behavior of @(disabled) on parapoly procedures or a compiler bug, but the when block approach avoids the issue and is consistent with how destroy already handles it in the same file.

### DIFF
--- a/core/container/pool/pool.odin
+++ b/core/container/pool/pool.odin
@@ -113,28 +113,30 @@ _set_next :: proc(p: ^Pool($T), elem: ^T, next: ^T) {
 	(^^T)(uintptr(elem) + p.link_off)^ = next
 }
 
-@(disabled=.Address not_in ODIN_SANITIZER_FLAGS)
 _poison_elem :: proc(p: ^Pool($T), elem: ^T) {
-	if p.link_off > 0 {
-		sanitizer.address_poison_rawptr(elem, int(p.link_off))
-	}
+	when .Address in ODIN_SANITIZER_FLAGS {
+		if p.link_off > 0 {
+			sanitizer.address_poison_rawptr(elem, int(p.link_off))
+		}
 
-	len := size_of(T) - p.link_off - size_of(rawptr)
-	if len > 0 {
-		ptr := rawptr(uintptr(elem) + p.link_off + size_of(rawptr))
-		sanitizer.address_poison_rawptr(ptr, int(len))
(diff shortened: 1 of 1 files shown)


Kelimion:
Thanks for tracking this down. So that was what was confoundig the inference. Perhaps @gingerBill can use this to fix the inference in the presence of `@(disabled)`.

PR #6276, opened 2026-02-14
#lazy imports

xoxorwr:
If you import a package to use a single struct definition, you are forced to (supposedly) inherit its entire dependency tree, its @(init) overhead, and its impact on the linker

I propose `#lazy` import, which tells the compiler to completely skip the parsing, type-checking, and initialization of a package unless a symbol from that package is explicitly referenced in the code

Btw, I am not looking to replace the "file suffixes" for platform logic, i"m looking for a way to keep the dependency graph as lean as posible for when simplicity matters more than the overhead of managing a thousand tiny, fragmented files

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -5136,6 +5136,9 @@ gb_internal void add_import_dependency_node(Checker *c, Ast *decl, PtrMap<AstPac
 		}
 		AstPackage **found = string_map_get(&c->info.packages, path);
 		if (found == nullptr) {
+			if (id->is_lazy) {
+				return;
+			}
 			Token token = ast_token(decl);
 			error(token, "Unable to find package: %.*s", LIT(path));
 			exit_with_errors();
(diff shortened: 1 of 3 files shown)


gingerBill:
This approach would sadly not work because you are not considering if another package imports that package and actually wants that full behaviour.

You cannot just lazily import something like that and hope it works for the rest of the project. 

I understand what you are wanting though and that's mainly just unused imports. Try doing this instead:

```odin
#+vet unused-imports
```

gingerBill:
Was it a specific package in Odin which triggered this PR? If so, which one?

xoxorwr:
> I understand what you are wanting though and that's mainly just unused imports. Try doing this instead:

That's not unused imports, it's lazy imports

```odin
import "core:sys/windows"
import "core:sys/linux"

when ODIN_OS == .Linux {
    sys :: linux
} else when ODIN_OS == .Windows {
    sys :: windows
} else when ODIN_OS == .Freestanding{
    sys :: rt
}
```

When compiling to wasm from Linux:

```
/home/xoxo/dev/odin/core/sys/linux/types.odin:1058:13: Error: Undeclared name: _Arch_Semid_DS 
        Semid_DS :: _Arch_Semid_DS 
                    ^~~~~~~~~~~~~^ 
 
/home/xoxo/dev/odin/core/sys/linux/types.odin:1322:14: Error: Undeclared name: _Arch_User_Regs 
        User_Regs :: _Arch_User_Regs 
                     ^~~~~~~~~~~~~~^ 
 
/home/xoxo/dev/odin/core/sys/linux/types.odin:1327:17: Error: Undeclared name: _Arch_User_FP_Regs 
        User_FP_Regs :: _Arch_User_FP_Regs 
                        ^~~~~~~~~~~~~~~~~^ 
 
/home/xoxo/dev/odin/core/sys/linux/types.odin:1333:18: Error: Undeclared name: _Arch_User_FPX_Regs 
        User_FPX_Regs :: _Arch_User_FPX_Regs 
                         ^~~~~~~~~~~~~~~~~~^ 
 
```

The point is, compiler should be smart enough to not bother processing the unused packages

PR #6287, opened 2026-02-15
Fix build errors in the sys/es package

krnowak:
I haven't got the slightest idea what this "es" platform actually is, but its `sys/es` package does not even compile. The fixes were easy, though.


Kelimion:
It's support for [Essence OS](https://gitlab.com/nakst/essence).

PR #6320, opened 2026-02-22
In handle_map.dynamic_add(), remove redundant assignment

Faker-09:
In handle_map.dynamic_add(), 'item' has been appended to the xar so this additional assignment is unneeded.


gingerBill:
How is this redundant? How else is `item` going to be added otherwise in that branch?

PR #6325, opened 2026-02-23
Update SDL3 to 3.4.2

nowhereware:
These commits update the vendor SDL3 bindings to version 3.4.2. NOTE: I did not update the .lib or .dll files, they'll need to be properly updated based on the released binaries.


Kelimion:
Regrettably can't add the updated libraries, despite "Maintainers are allowed to edit this pull request", because the PR is based on `nowhereware:master`.

Will merge and replace the libraries in a follow-up commit instead.

PR #6331, opened 2026-02-25
[core:text/regex] Follow up to fix #6323 and a test case

### LINKED ISSUE #6323: Invalid regex boundary handling
## Context
Odin:    dev-2026-02-nightly
OS:      Ubuntu 22.04.5 LTS, Linux 6.8.0-94-generic

Taking https://regex101.com/ as a reference regex implementation, with the following regex pattern ``\b[a-z0-9']+\b`` and a string `can, can't, 'can't'` I expect the [iterator](https://pkg.odin-lang.org/core/text/regex/#create_iterator) to match:
```
Capture{pos = [[0, 3]], groups = ["can"]}
Capture{pos = [[5, 10]], groups = ["can't"]}
Capture{pos = [[13, 18]], groups = ["can't"]}
```
but it matches:
```
Capture{pos = [[0, 3]], groups = ["can"]}
Capture{pos = [[5, 10]], groups = ["can't"]}
Capture{pos = [[13, 19]], groups = ["can't'"]}
```

### Reproduce

```odin
package word_count

import "core:strings"
import "core:text/regex"
import "core:fmt"

count_word :: proc(input: string) -> map[string]u32 {
	frequencies := make(map[string]u32)

	lc := strings.to_lower(input)
	defer delete(lc)
	it, _ := regex.create_iterator(lc, `\b[a-z0-9']+\b`)
	defer regex.destroy_iterator(it)

	for capture in regex.match_iterator(&it) {
		word := strings.clone(capture.groups[0])
		if word in frequencies {
			frequencies[word] += 1
			delete(word)
		} else {
			frequencies[word] = 1
		}
	}

	return frequencies
}

main :: proc() {
	fmt.println(count_word("can, can't, 'can't'"))
}
```

GPotoshin:
As was said in the issue discussion I had suspicion that there may be a sibling bug in .Assert_Non_Word_Boundary implementation and I was able to confirm that with `re.findall(r"\B'", "'")` python code. Odin implementation outputed an empty string wherase python gave "'". That is the same bug related to incorrect logic on string ends. This commit makes implementation of those 2 instructions cleaner and correct and adds a test case.

### DIFF
--- a/core/text/regex/virtual_machine/virtual_machine.odin
+++ b/core/text/regex/virtual_machine/virtual_machine.odin
@@ -204,15 +204,8 @@ add_thread :: proc(vm: ^Machine, saved: ^[2 * common.MAX_CAPTURE_GROUPS]int, pc:
 		case .Assert_Word_Boundary:
 			sp := vm.string_pointer+vm.current_rune_size
 
-			left_is_wc := false
-			if sp > 0 {
-				left_is_wc = is_word_class(vm.current_rune)
-			}
-
-			right_is_wc := false
-			if sp < len(vm.memory) {
-				right_is_wc = is_word_class(vm.next_rune)
-			}
+			left_is_wc := sp > 0 && is_word_class(vm.current_rune)
+			right_is_wc := sp < len(vm.memory) && is_word_class(vm.next_rune)
 
 			if left_is_wc != right_is_wc {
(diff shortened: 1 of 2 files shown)


Kelimion:
Thanks for the follow-up, George.

Issue #6065, opened 2025-12-25
`utf8_to_wstring` (win32) does not null-terminate the result


bplu4t2f:
* Odin Version: 2025-12

The function `utf8_to_wstring_buf` from `core/sys/windows` does not null-terminate the result (also affects `utf8_to_utf16` - but in that case it might be intentional).

Also affects `_alloc` variant, but that is only relevant when non-zeroed memory is returned by the used allocator.

## Expected Behavior

The resulting string written to the specified `[]u16` buffer should be null-terminated.

Whether this is expected could be up for debate since I couldn't find something authoritative. However, from what I could read between the lines when I researched this briefly, the `utf8_to_wstring` procedure group appears to be *the* intended way to interoperate Odin strings with the win32 API (`LP[C]WSTR`). It would be strange if this omission was intentional since the vast majority of string related functions in the win32 API expect a null terminated string.

It could be that the function expects the buffer to be zeroed by the caller, but I have found no evidence that that was the intention (nor would I agree that that is a reasonable API contract), so I concluded that this is most likely an oversight.

## Current Behavior

String is not null-terminated after conversion. I.e., no explicit zero is written to the `u16` element at the end of the string in the buffer. This only happens to work if the buffer is zeroed before calling the function.

## Failure Information (for bugs)

This is because of how `MultiByteToWideChar` is used in this procedure group. When used with an explicit length argument, the function does not null-termiante the result (as documented).

In C, this usually doesn't happen because the most frequent use case is to pass in a null-terminated string, let the function infer the length by passing -1, and in that case specifically, the function will ensure that the result is null-terminated.

### Steps to Reproduce

```
buf : [20]u16
for i in 0 ..< len(buf) {
	buf[i] = auto_cast i + 0x20
}
s := "Hello"
converted := win.utf8_to_wstring(buf[:], s)
```

The above code results in a string that is not properly zero-terminated.

I could implement a fix if wanted, as soon as it's clear what the intended behavior is (of `utf8_to_utf16` and `utf8_to_wstring` respectively).

PR #6336, opened 2026-02-25
Fix MultiByteToWideChar usage in utf8_to_wstring_buf

bplu4t2f:
Aims to fix: https://github.com/odin-lang/Odin/issues/6065

---

Current findings/issues:

 * `utf8_to_utf16_alloc` _already did_ null-terminate the string, even though the null character was _not_ part of the result slice.
   * I did my best to consider app compat scenarios. Existing code may rely on that behavior, even though it was never part of the procedure's spec. I decided to implement `utf8_to_utf16_buf` in the same way. This implies that it will now require 1 more element in the provided buffer to succeed, and the added null character is beyond the returned slice.
 * Empty strings now convert to a non-nil value in `utf8_to_wstring`. Rationale: From the perspective of Win32 API functions, `NULL` is different from any non-null pointer pointing to `""`. I believe this was an oversight in the original implementation.
 * The reverse functions (`utf16_to_utf8`, `wstring_to_utf8`) are actually defined quite differently if one looks closely. I have decided to not touch them, and I don't think it is easily possible to make these APIs symmetric without raising compatibility concerns.

### DIFF
--- a/core/sys/windows/util.odin
+++ b/core/sys/windows/util.odin
@@ -103,17 +103,21 @@ utf8_to_utf16_alloc :: proc(s: string, allocator := context.temp_allocator) -> [
 	b := transmute([]byte)s
 	cstr := raw_data(b)
 	n := MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, cstr, c_int(s_length), nil, 0)
-	if n == 0 {
+	if n <= 0 || cast(int)n >= max(int) {
+		// If n is equal to or greater than max(int), then we will not be able
+		// to create a big enough slice with the null terminator.
+		// NOTE: This only affects 32-bit systems and is purely pedantic because
+		//       the system will never be able to allocate that much memory.
 		return nil
 	}
 
-	text := make([]u16, n+1, allocator)
+	text := make([]u16, cast(int)n + 1, allocator)
 	if text == nil {
 		return nil
 	}
 
 	n1 := MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, cstr, c_int(s_length), raw_data(text), n)
-	if n1 == 0 {
+	if n1 <= 0 {
 		delete(text, allocator)
 		return nil
 	}
@@ -128,7 +132,7 @@ utf8_to_utf16_alloc :: proc(s: string, allocator := context.temp_allocator) -> [
 @(require_results)
 utf8_to_utf16_buf :: proc(buf: []u16, s: string) -> []u16 {
 	buf_length := len(buf)
-	if buf_length == 0 {
+	if buf_length < 1 {
 		return nil
 	}
 	s_length := len(s)
@@ -143,7 +147,7 @@ utf8_to_utf16_buf :: proc(buf: []u16, s: string) -> []u16 {
 		buf_length = cast(int)max(c_int)
 	}
 	elements_written := MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, raw_data(s), c_int(s_length), raw_data(buf), cast(c_int)buf_length)
-	if elements_written == 0 {
+	if elements_written <= 0 {
 		// Insufficient buffer size, empty input string, or invalid characters. Contents of the buffer may have been modified.
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


core/sys/windows/util.odin
```diff
@@ -110,48 +110,112 @@ utf8_to_utf16_alloc :: proc(s: string, allocator := context.temp_allocator) -> [
 		return nil
 	}
 
+	// null-terminate the result here, even though the null element is not
+	// part of the slice. This is done to prevent callers which relied on
+	// this behavior, and is also expected by utf8_to_wstring_alloc.
 	text[n] = 0
-	for n >= 1 && text[n-1] == 0 {
```
bplu4t2f, line 114: Rationale: This trims successfully converted nulls from the end, but that does not at all align with the goal of this particular procedure. This procedure takes a fully specified slice of UTF-8 bytes as input, which is decidedly not specified to be null-terminated. Any nulls in the input will appear in the output; this would also be the case for nulls appearing in the middle or at the beginning of the string (which would not have been trimmed by this loop).

Kelimion:
I've followed along so far, and the PR looks very promising. Let us know when you deem it ready, and I'm sure it'll be merged promptly.

bplu4t2f:
> I've followed along so far, and the PR looks very promising. Let us know when you deem it ready, and I'm sure it'll be merged promptly.

Sure, I definitely want to sleep about it though. Maybe I'll have one of those morning shower moments.

core/sys/windows/util.odin
```diff
@@ -91,67 +91,144 @@ LANGIDFROMLCID :: #force_inline proc "contextless" (lcid: LCID) -> LANGID {
 
 @(require_results)
 utf8_to_utf16_alloc :: proc(s: string, allocator := context.temp_allocator) -> []u16 {
-	if len(s) < 1 {
+	s_length := len(s)
+	if s_length < 1 {
+		return nil
+	}
+	if s_length > cast(int)max(c_int) {
```
Kelimion, line 98: This test would equate to `false` on 32-bit Windows if `s_length` is the maximum, and thus not return early.
```odin
a := max(i32) // int is an i32, so the maximum length of a string would be `max(i32)`.
b := max(i32)

fmt.println(a > b, a == b) // false true
```
It would be better to change it to: `if i64(s_length) > i64(max(c_int))` to widen the test, or change it to `>=`. Same in the other proc.

Kelimion: Looks great apart from this one edge case.

bplu4t2f: @Kelimion Maybe I'm just not seeing it, but wouldn't that just be the correct behavior then? Because anything up to int32_max is fine, so it shouldn't return out early.

Kelimion: Yes, up to. But if `len` is 1 byte longer, and the maximum length on a 32-bit platform, it becomes `max(i32)`, then `max(i32)` isn't `> max(i32)`, and so that early out fails.

Kelimion: You're saying "anything up to int32_max". That is, `len(s) <= max(i32) - 1` is fine. And that's true, but what if it's not 1 byte shorter than the max, but the actual max? Then the length is no longer `>` the max, it equals it, and so it falls through to the rest of the body.

bplu4t2f: For the record:

We continued this in DMs - this particular check here is correct on both 64-bit and 32-bit (even though it can never be true on 32-bit). The point here is to protected against overflow when the input string length is equal to `0x8000_0000` or greater on 64-bit systems. Without the check, this would cause a negative value to be passed to the Win32 API function. On 32-bit, this is not a concern because the input string length can be at most `0x7fff_ffff`.

However, in theory, on 32-bit systems only, a specific overflow, which is related to this problem, but it is _not_ the same problem, could happen later when we attempt to allocate a slice with `n+1` elements (`_alloc` variant), OR when we check for sufficient buffer length (`_buf` variant). This would *not* always happen because there is a good chance that the UTF-16 slice will have fewer elements in it than the input string length (depending on the contents of the string). Note that this is completely impractical on 32 bit systems; a conversion of that size has no chance of succeeding, and we're just ensuring that the failure happens gracefully so it cannot be exploited.

PR #6337, opened 2026-02-26
nbio: fix op re-use

andzdroid:
The posix nbio implementation reuses the Operation linked list nodes, but the links are not cleared before re-use. This can cause the re-insertion to loop forever in specific cases.

### DIFF
--- a/tests/core/nbio/nbio.odin
+++ b/tests/core/nbio/nbio.odin
@@ -257,77 +257,49 @@ wake_up :: proc(t: ^testing.T) {
 	}
 }
 
+// Tests that if multiple accepts are queued, and a dial comes in which completes one of them,
+// the rest are queued again properly.
 @(test)
-reused_ops :: proc(t: ^testing.T) {
+still_pending :: proc(t: ^testing.T) {
 	if event_loop_guard(t) {
 		testing.set_fail_timeout(t, time.Minute)
 
-		@static payload := [4]byte{'P', 'I', 'N', 'G'}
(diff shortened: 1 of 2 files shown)


andzdroid:
oops, sorry about the copilot auto review

tests/core/nbio/nbio.odin
```diff
@@ -256,3 +256,78 @@ wake_up :: proc(t: ^testing.T) {
 		}
 	}
 }
+
+@(test)
+reused_ops :: proc(t: ^testing.T) {
+	if event_loop_guard(t) {
+		testing.set_fail_timeout(t, time.Minute)
+
+		@static payload := [4]byte{'P', 'I', 'N', 'G'}
```
gingerBill:
```odin
@(static, rodata) payload := [4]byte{'P', 'I', 'N', 'G'}
```

laytan:
It's a real bug but the description is kinda off.
The issue here is that the accept completes in the kqueue, then we call `handle_completed` on all the accept operations that are in the linked list for this identifier at https://github.com/odin-lang/Odin/blob/c63e3103591acde0f06e8fffc867bf4b433a45b1/core/nbio/impl_posix.odin#L339
this is where I forgot to clear the links of the node, and when an accept in this list can still not complete it add_pending's it again, where it has stale links.

I think I prefer clearing the links at that point instead, so we know why it is needed too.

The test is also doing a lot of extra things not necessary to test this case. Which I will remove.

PR #6339, opened 2026-02-27
[Core:Tokenizer] Tokenizer uses hash table instead of linear lookup for keywords.

Karidus-423:
Saw the TODO while reading the src code.


gingerBill:
By using a hash table, this is not what was meant. It means we needed a static hash map.

Also removing those grouping enum values is bad too because you haven't understood what they were for.

Issue #6189, opened 2026-01-28
Incorrect results by substring and substring_from in core:strings


KittenLord:
## Context

```
Odin:    dev-2025-11:e5153a937
OS:      Arch Linux, Linux 6.18.6-arch1-1
CPU:     Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
RAM:     15852 MiB
Backend: LLVM 21.1.5
```

## Expected Behavior

As per docs (at https://pkg.odin-lang.org/core/strings/#substring):

> Think of it as slicing `s[rune_start:rune_end]` but rune-wise.

Therefore the following code is expected to behave equivalently for `exampleA` and `exampleB`

```odin
package main

import "core:strings"
import "core:fmt"

main :: proc () {
    exampleA : []int  = { 1, 2, 3, 4, 5 }
    exampleB : string = "hello"

    fmt.println(len(exampleA))
    fmt.println(len(exampleB))

    fmt.println(exampleA[0:5])
    fmt.println(strings.substring(exampleB, 0, 5))

    fmt.println(exampleA[0:0])
    fmt.println(strings.substring(exampleB, 0, 0))

    fmt.println(exampleA[5:5])
    fmt.println(strings.substring(exampleB, 5, 5))
}
```

## Current Behavior

```odin
5
5
[1, 2, 3, 4, 5]
hello true
[]
 true
[]
hello false
```

The last example works as intended for slices, but not for `substring` (and neither for `substring_from`)

Doesn't seem too hard to fix. Relevant code: https://github.com/odin-lang/Odin/blob/master/core/strings/strings.odin#L3385

### Steps to Reproduce

Just `odin run` the code snippet above

PR #6340, opened 2026-02-27
fix: `strings.substring` not returning end of range

miningape:
This pull request addresses #6189.

Hello 👋 

This is my first PR to Odin - I tried to look for a `CONTRIBUTING.md` or similar to see if just submitting a PR like this is okay but couldn't find anything (didn't join your discord to ask about it). If I did something wrong, I'm really sorry, and am happy to work with you to fix it 🙂 

I encountered the same bug as the person who created the issue (#6189), and decided to try submitting a PR to fix it. 

As the issue describes, while the slice range operator `slice[start:end]` accepts `start, end = len(slice)` and returns an empty slice - the `strings.substring` function doesn't do the same.

For example:
```Odin
aSlice  : []int  = { 1, 2, 3, 4, 5 }
aString : string = "Hello"

// These are fine
aSlice[0:0] // []int{}
strings.substring(aString, 0, 0) // "", true

aSlice[3:3] // []int{}
strings.substring(aString, 3, 3) // "", true

// Here's the issue
aSlice[5:5] // []int{}
strings.substring(aString, 5, 5) // "Hello", false

// For sanity
aSlice[6:6] // range error
strings.substring(aString, 6, 6) // "Hello", false
```

The idea behind this fix is that after the final iteration of the first `for` loop (without matches) we're now at the end of the string, so taking a slice from that point allows the following `for` loop to return true because we're at the end of the string (`rune_end == rune_i`).

So after applying this PR we get:
```Odin
// Here's the issue
aSlice[5:5] // []int{}
strings.substring(aString, 5, 5) // "", true

// For sanity
aSlice[6:6] // range error
strings.substring(aString, 6, 6) // "", false
```

As you can see this PR does change the `sub` string result for out of range errors, but the `ok` boolean is still false both before and after this PR.

I tried testing a few other scenarios and nothing obvious broke. And I also added a few tests cases for this issue.

Thanks for taking the time to look into this!

p.s. If you'd like me to attempt a larger change to how this function is implemented I'm also open to doing so 😄

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -3420,7 +3420,9 @@ internal_substring :: proc(s: string, rune_start: int, rune_end: int) -> (sub: s
 			}
 			rune_i += 1
 		}
-		if !ok { return }
+		if !ok {
+			sub = sub[rune_i:]
+		}
 	}
 
 	if rune_end >= rune_start {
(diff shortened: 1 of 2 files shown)


Kelimion:
Thanks for this. Merged.

PR #6343, opened 2026-02-28
Kill deprecated file tags and fix assignment of package docs

krnowak:
1. Let's kill the deprecated file tags - I think they were deprecated for long enough for projects to move away from them. Also, there were some issues with it - ignored if they appeared after `#+` file tags, and possibly could end up in package docs (unless some extra preprocessing happened to remove them).

2. So far, the first comment found was marked as package docs, which meant that for the following code, the TODO comment would be assigned to the package declaration instead of the comment directly preceding the
package declaration.

```odin
// TODO: drop after finished with refactoring
#+ feature using-stmt
// Package foo implements this and that.
package foo
```

3. Drop some pointless code, which modifies a local variable after it was stored in an out parameter, so this modification has no effect.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -6725,11 +6725,6 @@ gb_internal bool parse_file(Parser *p, AstFile *f) {
 
 	String filepath = f->tokenizer.fullpath;
 	String base_dir = dir_from_path(filepath);
-	if (f->curr_token.kind == Token_Comment) {
-		consume_comment_groups(f, f->prev_token);
-	}
-
-	CommentGroup *docs = f->lead_comment;
 
 	Array<Token> tags = array_make<Token>(temporary_allocator());
 	bool first_invalid_token_set = false;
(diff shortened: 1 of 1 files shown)


krnowak:
The issue that might pop up is if somewhere the code was like

```odin
// Yadda yadda, some package docs
// More package docs
//+ private
// Continue with package docs
package stuff
…
```

then previously the entire comment counted as package docs, whereas with this PR only the part after the "private" filetag will be assigned to the package declaration. Not sure how serious this issue is, though.

krnowak:
Will update the PR later to drop the deprecated `//+` filetags altogether as discussed on Discord. This should make this PR a bit simpler.

krnowak:
Updated the PR.

src/parser.cpp
```diff
@@ -6754,6 +6746,8 @@ gb_internal bool parse_file(Parser *p, AstFile *f) {
 		}
 	}
 
+	CommentGroup *docs = f->lead_comment;
```
krnowak, line 6749: I'm saving `lead_comment` here, because the `expect_token(f, Token_package)` below changes it (through `advance_token`) and we want to pass this comment into `ast_package_decl` function below. We also pass `line_comment` to `ast_package_decl`, but I'm not sure if I should save `line_comment` from being changed by `expect_token` - I haven't yet understood what `line_comment` exactly is.

gingerBill: `line_comment` is meant for things like this:

```odin
// top comment
declaration: type // line comment
```

PR #6349, opened 2026-03-01
Fix #6348

### LINKED ISSUE #6348: False "Invalid type of a key for a map" error for struct with proc field that circularly references the map's containing type
## Context

        Odin:    dev-2026-02:c63e31035
        OS:      macOS Sonoma 14.4.0 (build 23E214, kernel 23.4.0)
        CPU:     Apple M3
        RAM:     16384 MiB
        Backend: LLVM 20.1.5

## Expected Behavior

This code should compile:

```odin
package odin_mapkey_bug

Foo :: struct { f: proc(^Bar) }
Bar :: struct { m: map[Foo]int }

main :: proc() {}
```

## Current Behavior

Instead, we get this error:
```
Error: Invalid type of a key for a map, got 'Foo' 
        Bar :: struct { m: map[Foo]int }
```

Only triggered when they key type is declared before the map. Swapping the declaration order avoids the error (though programs should not need to care about global declaration order).

Also triggered through indirect chains, e.g. 

```odin
Foo :: struct { f: proc(^Baz) }
Baz :: struct { g: proc(^Bar) }
Bar :: struct { m: map[Foo]int }
```

A proc type alias avoids the error — only inline types in struct fields are affected afaik.

I'll open a PR with a fix for this.

louisnovy:
Here's a first pass that fixes #6348.

This fix feels like kind of a bodge.  Is there a simpler way?


src/check_type.cpp
```diff
@@ -2859,6 +2859,24 @@ gb_internal void add_map_key_type_dependencies(CheckerContext *ctx, Type *key) {
...
+gb_internal void validate_map_key(CheckerContext *ctx, Ast *node, Type *key) {
+	if (!is_type_valid_for_keys(key)) {
+		if (is_type_boolean(key)) {
+			error(node, "A boolean cannot be used as a key for a map, use an array instead for this case");
+		} else {
+			gbString str = type_to_string(key);
+			error(node, "Invalid type of a key for a map, got '%s'", str);
+			gb_string_free(str);
+		}
+	}
```
gingerBill, line 3027: Add a `return;` somewhere to prevent a possible double error printing which is not necessary being printed below. I never did that as it is also an error.

src/check_type.cpp
```diff
@@ -2878,30 +2896,30 @@ gb_internal void check_map_type(CheckerContext *ctx, Type *type, Ast *node) {
...
-		gb_string_free(str);
-	}
-
 	type->Map.key   = key;
 	type->Map.value = value;
 
-	add_map_key_type_dependencies(ctx, key);
+	// If the key type is still being resolved its struct fields may be empty causing false invalid key errors.
+	// Defer to check_deferred_map_key_types which runs after all global entities are resolved.
+	if (key->kind == Type_Named && key->Named.type_name && key->Named.type_name->state == EntityState_InProgress) {
```
gingerBill: I'd probably make an explicit atomic load here:

```cpp
key->Named.type_name->state.load() == EntityState_InProgress
```

src/check_type.cpp
```diff
@@ -2878,30 +2896,30 @@ gb_internal void check_map_type(CheckerContext *ctx, Type *type, Ast *node) {
...
 	init_map_internal_types(type);
 }
 
+gb_internal void check_deferred_map_key_types(Checker *c) {
+	CheckerContext ctx = c->builtin_ctx;
+	ctx.decl = make_decl_info(nullptr, nullptr);
+	for_array(i, c->deferred_map_key_checks) {
+		Ast *node = c->deferred_map_key_checks[i];
+		validate_map_key(&ctx, node, node->tav.type->Map.key);
+	}
```
gingerBill:
```cpp
for (Ast *node : c->deferred_map_key_checks) {
	validate_map_key(&ctx, node, node->tav.type->Map.key);
}
```

src/check_type.cpp
louisnovy: Using else if seemed cleaner and invokes add_map_key_type_dependencies which I believe is the intent.

louisnovy: Oh explicit return makes more sense. I misunderstood the intent.

PR #6356, opened 2026-03-02
Use temporary directory for intermediate .ll files

xoxorwr:
Most systems are configred so that `/tmp` is essentially a ramdisk, so this PR:
- reduces unnecessary disk i/o and wear on SSDs (~30mb on my project everytime i save for hotreload)
- speeds up optimization and linking phases

This is basically a copy paste of `lb_filepath_obj_for_module`'s logic

### DIFF
--- a/src/llvm_backend.cpp
+++ b/src/llvm_backend.cpp
@@ -2655,11 +2655,20 @@ gb_internal void lb_llvm_module_passes_and_verification(lbGenerator *gen, bool d
 }
 
 gb_internal String lb_filepath_ll_for_module(lbModule *m) {
-	String path = concatenate3_strings(permanent_allocator(),
-		build_context.build_paths[BuildPath_Output].basename,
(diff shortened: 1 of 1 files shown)


laytan:
Does `-keep-temp-files` keep the old behavior?

xoxorwr:
> Does `-keep-temp-files` keep the old behavior?

Yes, behavior remain unchanged, it just change the direectory

I should add that this PR should be tested on Windows, I don't have a windows machine with me right now

gingerBill:
> Yes, behavior remain unchanged, it just change the direectory

As silly as this sounds but that is changing the behaviour though. 

I recommend allowing a way to opt-out of the directory behaviour is possible, and not just put it in a random `/tmp` directly. The point of stuff of the command `-keep-temp-files` is to be able to easily inspect where things went wrong, so placing it in a hard to find temp directory isn't a good idea.

PR #6366, opened 2026-03-03
Add an AST dumper

krnowak:
I wrote this AST dumper mostly to compare a tree generated by Odin with a tree generated by my odin tree sitter parser. But I guess it could be useful not only for me, so I'm proposing it here.

This adds two new options to the `build` command - `-dump-ast` and `-dump-ast-opts`, both are taking parameters. `-dump-ast` takes parameters specifying which packages and files should have their ASTs dumped, whereas `-dump-ast-opts` modify the dumper behavior. From the help output:

```
	-dump-ast:<string>
		Dumps AST to files. The passed string can either be '-' or a semicolon-separated list of dump specs.
		Dump spec: <package name>[:<comma-separated list of files>]
		Example dump specs:
			utf8 - dump AST for all the files in the utf8 package
			unicode:doc.odin,fold.odin - dump AST only for doc.odin and fold.odin in unicode package

	-dump-ast-opts:<string>
		Options for -dump-ast. The passed string is a comma-separated list of options.
		Possible options:
			sexp - dump AST in S-expression format
			json - dump AST in JSON format
			xml - dump AST in XML format
			omit-defaults - skip values and attributes with default values (false for booleans, empty strings or empty lists)
			disregard-filetags - ignore build effects of file tags, so this will make the compiler only to parse the files and dump the ast
			no-ids - do not print IDs for dumped AST nodes, useful for less-cluttered diffs
```

I'm not sure which format would be best here, so just wrote three.

An example invocation I was using:

`odin build . -dump-ast:- -dump-ast-opts:sexp,omit-defaults,disregard-filetags`

This dumps all the files (even those that should not be built, because the OS does not match) in all the imported packages using an S-expression format, making the output a bit shorter by skipping fields set to default values.

This turned out to be helpful in finding issues with `#force_inline`, package docs, procedures with diverging results and probably something else too.

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -547,6 +547,10 @@ struct BuildContext {
 	bool   keep_object_files;
 	bool   disallow_do;
 	bool   show_import_graph;
+	bool   dump_ast;
+	bool   dump_ast_omit_defaults;
+	bool   dump_ast_disregard_filetags;
+	StringMap<StringSet> ast_dump_packages;
 
 	bool   webkit_switch_workaround;
 
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -557,6 +557,10 @@ struct BuildContext {
 	bool   keep_object_files;
 	bool   disallow_do;
 	bool   show_import_graph;
+	bool   dump_ast;
+	bool   dump_ast_omit_defaults;
+	bool   dump_ast_disregard_filetags;
+	StringMap<StringSet> ast_dump_packages;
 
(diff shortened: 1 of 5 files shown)


herohiralal:
Is there a reason this hasn't been merged/reviewed?

herohiralal:
bump (?)

krnowak:
If I had to guess then the size of the PR and the rather niche use cases for the AST dumper makes it rather a low priority functionality to spend the time on. Do you need the AST dumper for anything?

PR #6367, opened 2026-03-03
vendor:box2d add GetVersion and SetLogFcn

Faker-09:
This box2d change:

- Adds a binding  for GetVersion
- Adds a binding SetLogFcn  which is left commented out in preparation for the 3.1.2 release.
- Adds a comment to DestroyWorld, "This efficiently destroys all bodies, shapes, and joints in the simulation." Taken from box2d.org
- Aligns formatting in that area of the file.

### DIFF
--- a/vendor/box2d/box2d.odin
+++ b/vendor/box2d/box2d.odin
@@ -47,6 +47,9 @@ FreeFcn :: #type proc "c" (mem: rawptr)
 // Prototype for the user assert callback. Return 0 to skip the debugger break.
 AssertFcn :: #type proc "c" (condition, file_name: cstring, line_number: i32) -> i32
 
+// Prototype for user log callback. Used to log warnings.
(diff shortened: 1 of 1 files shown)


vendor/box2d/box2d.odin
```diff
@@ -47,6 +47,9 @@ FreeFcn :: #type proc "c" (mem: rawptr)
 // Prototype for the user assert callback. Return 0 to skip the debugger break.
 AssertFcn :: #type proc "c" (condition, file_name: cstring, line_number: i32) -> i32
 
+// Prototype for user log callback. Used to log warnings.
+//LogFcn :: #type proc "c" (message: cstring)
```
gingerBill, line 51: Why was this commented out?

Faker-09: LogFcn was added to Box2d recently and is not available until the next release version 3.1.2. We are currently using 3.1.1

Issue #6333, opened 2026-02-25
Missleading error message when using keyword as var name in for loop


xandaron:
If you use a keyword as a variable name in a for loop you get the error message:
```
Syntax Error: Expected a type, got ','
for enum, prefix in config.remove_enum_member_prefix {
        ^
```

This will also cause the compiler to output more errors such as:
```
Syntax Error: Expected ';', got {
if new_enum, exists := config.bit_setify[enum]; exists {
                                                      ^

Syntax Error: You cannot use a for statement in the file scope
for input_base in config.inputs {
^
```

As the compiler no longer seems to be able to parse the file properly.

It would be nice if the compiler could recognise that a keywork was used as a var name and give better error messaging.

Kelimion:
Here's another way to replicate it: `enum, context := 12, 23`, e.g.
```odin
package bug_6333

import "core:fmt"

main :: proc() {
	values := []int{1, 2, 3, 4}
	for enum, prefix in values {
		fmt.println(enum, prefix)
	}

	enum, context := 12, 23
}
```

PR #6368, opened 2026-03-04
chore: better `for` loop syntax error message

miningape:
Aims to address the `for` loop enum issue in https://github.com/odin-lang/Odin/issues/6333 - PR is marked as draft because I'm just looking for some feedback on whether this is the correct direction, and what the next steps to get it across the line look like?

Also wondering if there's some way to clear the previous errors? Right now when executing I am still getting the "bad" error messages below the new one that was added in this PR:

```
/Odin/local/test.odin(6:6) Syntax Error: Expected an expression, got enum type 
        for enum, prefix in x { 
            ^ 
 
/Odin/local/test.odin(6:10) Syntax Error: Expected a type, got ',' 
        for enum, prefix in x { 
                ^ 
 
/Odin/local/test.odin(6:12) Syntax Error: Expected '{', got 'identifier' 
        for enum, prefix in x { 
                  ^ 
 
/Odin/local/test.odin(6:19) Syntax Error: Expected an operand 
        for enum, prefix in x { 
                         ^ 
 
/Odin/local/test.odin(8:2) Syntax Error: Expected ';', got newline 
        } 
        ^ 
 
/Odin/local/test.odin(9:1) Syntax Error: Expected an operand 
        } 
        ^ 
 
/Odin/local/test.odin(9:1) Syntax Error: Expected ';', got newline 
        } 
        ^ 
```

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -4891,13 +4891,15 @@ gb_internal Ast *parse_for_stmt(AstFile *f) {
 		}
 
 		if (f->curr_token.kind != Token_Semicolon) {
+			Token before = f->curr_token;
+
 			cond = parse_simple_stmt(f, StmtAllowFlag_In);
 			if (cond->kind == Ast_AssignStmt && cond->AssignStmt.op.kind == Token_in) {
 				is_range = true;
 			}
 
 			if (cond->kind == Ast_ExprStmt && is_ast_type(cond->ExprStmt.expr)) {
-				syntax_error(cond->ExprStmt.expr->EnumType.token, "Expected an expression, got %.*s", LIT(ast_strings[cond->ExprStmt.expr->kind]));
+				syntax_error(before, "Expected an expression, got %.*s", LIT(ast_strings[cond->ExprStmt.expr->kind]));
 			}
 		}
 


src/parser.cpp
```diff
@@ -4895,6 +4895,10 @@ gb_internal Ast *parse_for_stmt(AstFile *f) {
 			if (cond->kind == Ast_AssignStmt && cond->AssignStmt.op.kind == Token_in) {
 				is_range = true;
 			}
+
+			if (cond->kind == Ast_ExprStmt && is_ast_type(cond->ExprStmt.expr)) {
```
miningape, line 4901: Not sure if there's a better check than `is_ast_type` here? - I tried with `!is_expr_type` but literals and some other types of "expressions" which return data (and are therefore valid conditions in a for loop) aren't included in it

gingerBill:
The problem here is that you used a keyword rather an identifier. The solution to this isn't to use what you did here, nor is it actually a good way of handling it, per se.

I am not sure what the best way would be, even if it might be what it currently is.

miningape:
@gingerBill Thanks for taking the time to look this over 😄 

Yeah agreed this is a hacky rather than structural fix - as you say it'd be best to, for example, make it possible for the parser to parse the `in` into an assign statement and only at that point check that keywords aren't being used in place of identifiers.

PR #6381, opened 2026-03-07
core:testing remove whitespace typo

Faker-09:
This one wasn't me but I leave these spaces around all the time when waking my laptop.


Kelimion:
I use Ctrl for that purpose because it doesn't leave behind any input. :D

PR #6387, opened 2026-03-08
Fix tokenizer init

krnowak:
Without setting column_minus_one to -1 at init, the token line position in the first line of the file was offset by 1.

### DIFF
--- a/src/tokenizer.cpp
+++ b/src/tokenizer.cpp
@@ -379,6 +379,7 @@ gb_internal void advance_to_next_rune(Tokenizer *t) {
 
 gb_internal void init_tokenizer_with_data(Tokenizer *t, String const &fullpath, void const *data, isize size) {
 	t->fullpath = fullpath;
+	t->column_minus_one = -1;
 	t->line_count = 1;
 
 	t->start = cast(u8 *)data;


gingerBill:
Columns are 1-index not 0-index. So how was it wrong before?

krnowak:
AST dump of a simple file like:
```
package foo
```

without the fix:
```
(source-file (version 1) (
	(package-declaration (id 0)
		(token
			(token (string "package") (offset 0) (line 1) (column 2)))
		(name
			(token (string "foo") (offset 8) (line 1) (column 10))))))
```

after the fix:
```
(source-file (version 1) (
	(package-declaration (id 0)
		(token
			(token (string "package") (offset 0) (line 1) (column 1)))
		(name
			(token (string "foo") (offset 8) (line 1) (column 9))))))
```

If I move the `package foo` line to a second line, then the column numbers are correct. I assumed that the `column_minus_one` should be initialized to `-1`, because this is the value it is being set to after getting newline.

Kelimion:
That makes sense. It seems to me then that error reporting on the first line has also been off by 1 column, and that'll be fixed by this patch as well.

And that appears to be the case. This error isn't cleverly pointing out the `r` is out of place so much as reporting the error 1 column over from where it should be.
```
W:/Scratch/main.odin(1:2) Syntax Error: Expected a package declaration at the beginning of the file 
	prackage main 
	 ^ 
```

Kelimion:
And indeed, with this fix. 🎉 
```
W:/Scratch/main.odin(1:1) Syntax Error: Expected a package declaration at the beginning of the file 
	prackage main 
	^ 
```

Kelimion:
Thanks, Krzesimir.

PR #6389, opened 2026-03-08
Partial 1.619.x d3d12 bindings

connnnal:
This new spec mentions [`ID3D12DeviceExtended`](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/D3D12RevisedCreateViews.md) a dozen times, but it seems to have been simplified into `ID3D12Device15`. So the bindings need to catch up before we get most of the new features! Currently we've reached `IDevice8`...

### DIFF
--- a/vendor/directx/d3d12/d3d12.odin
+++ b/vendor/directx/d3d12/d3d12.odin
@@ -174,19 +174,20 @@ PRIMITIVE :: enum i32 {
 }
 
 SRV_DIMENSION :: enum i32 {
-	UNKNOWN          = 0,
-	BUFFER           = 1,
-	TEXTURE1D        = 2,
-	TEXTURE1DARRAY   = 3,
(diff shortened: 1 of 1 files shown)


heavyrain266:
Isn't Agility SDK required for stuff like this?
```odin
@(export, link_name = "D3D12SDKVersion")
D3D12SDKVersion: u32 = 618

@(export, link_name = "D3D12SDKPath")
D3D12SDKPath: cstring = ".\\D3D12\\"
```
Newer versions of DirectX (and extensions like DirectStorage) are distributed independently from Windows SDK, as NuGet packages, so you need the local `D3D12` with DLLs.

connnnal:
Yea for the time being, Win11 is on 616 and Win10 is even older, so you're missing a lot. But these bindings have always targeted latest, unfortunately it's the developer's responsibility to manage and/or ship runtime versions.

PR #6391, opened 2026-03-08
Add formatting flag for space between time.Duration and unit of measurement

mtarik34b:
Just mirror `%#m` for `time.Duration`.

```odin
package pkg

import "core:fmt"
import "core:time"

main :: proc() {
	subsecond := time.Nanosecond
	supersecond := 2 * time.Second

	fmt.printfln("%v", subsecond)
	fmt.printfln("%#v", subsecond)

	fmt.printfln("%v", supersecond)
	fmt.printfln("%#v", supersecond)
}
/*
Before:
1ns
1ns
2s
2s

After:
1ns
1 ns
2s
2 s
*/
```

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -2592,7 +2592,7 @@ fmt_named_buitlin_custom_formatters :: proc(fi: ^Info, v: any, verb: rune, info:
 				prec = 6
 				buf[w] = 'm'
 			}
-			if fi.hash {
+			if fi.space {
 				w -= 1
 				buf[w] = ' '
 			}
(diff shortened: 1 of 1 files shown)


Kelimion:
I would make this switch on `fi.space` instead. `fi.hash` is used to add a prefix, e.g. `fmt.printfln("%#x", 65535) // 0xffff`, whereas `fi.space` is used to replace leading zeroes with a space, e.g. `fmt.printfln("% 08d", 65535) //    65535`.

So I think that `fmt.printfln("% v", subsecond)` better reflects the intent.

mtarik34b:
Did the change, don't know why GH decided to close the PR on commit.

PR #6403, opened 2026-03-11
Fix auto reset event signal deadlock

mifreundorfer:
The status variables need to be updated every iteration, otherwise the thread will spin forever, if another thread waits on the event between the load and the compare exchange, because the old value is never updated.

### DIFF
--- a/core/sync/extended.odin
+++ b/core/sync/extended.odin
@@ -222,9 +222,10 @@ thread.
 */
 auto_reset_event_signal :: proc "contextless" (e: ^Auto_Reset_Event) {
 	old_status := atomic_load_explicit(&e.status, .Relaxed)
-	new_status := old_status + 1 if old_status < 1 else 1
 	for {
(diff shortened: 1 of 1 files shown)


mifreundorfer:
Short test for reproducing this. Without the patch this will likely never print anything because it immediately deadlocks.
```
package test

import "core:sync"
import "core:thread"
import "core:fmt"

event: sync.Auto_Reset_Event

main :: proc() {
	thread := thread.create_and_start(thread_proc)

	iter_count := 0
	for {
		sync.auto_reset_event_signal(&event)
		iter_count += 1

		if iter_count % 10000 == 0 {
			fmt.println("Hello from main")
		}
	}
}

thread_proc :: proc() {
	iter_count := 0
	for {
		sync.auto_reset_event_wait(&event)

		iter_count += 1
		if iter_count % 10000 == 0 {
			fmt.println("Hello from thread")
		}
	}
}
```

Kelimion:
Thanks!

PR #6406, opened 2026-03-12
Fixed Capacity Dynamic Arrays

gingerBill:
`[dynamic; N]T`

* `N` is the capacity
* `T` is the element type.
* Internal structure: `struct {data: [N]T, len: int}`
* Structurally equivalent to `small_array.Small_Array(N, T)`
* `len(array)` accesses the runtime `int` value
* `cap(array)` returns a compile-time known untyped integer equivalent to `N`
* Indexing and slicing are available
* Compound literal support:
  * `[dynamic; 16]u32{1, 4, 9, 16}`
  * `[dynamic; 16]u32{0 = 1, 2 = 9, 1 = 4, 3..<6 = 16}`
* All `@builtin` procedures that supported normal `[dynamic]T` dynamic arrays are now overloaded to support fixed capacity dynamic arrays
  * `append`, `resize`, `unordered_remove`, etc

### DIFF
--- a/core/odin/doc-format/doc_format.odin
+++ b/core/odin/doc-format/doc_format.odin
@@ -12,7 +12,7 @@ String :: distinct Array(byte)
 
 Version_Type_Major :: 0
 Version_Type_Minor :: 3
-Version_Type_Patch :: 1
+Version_Type_Patch :: 2
 
 Version_Type :: struct {
 	major, minor, patch: u8,
@@ -168,32 +168,33 @@ Attribute :: struct {
 }
 
 Type_Kind :: enum u32le {
-	Invalid                = 0,
-	Basic                  = 1,
-	Named                  = 2,
-	Generic                = 3,
-	Pointer                = 4,
-	Array                  = 5,
-	Enumerated_Array       = 6,
-	Slice                  = 7,
-	Dynamic_Array          = 8,
-	Map                    = 9,
-	Struct                 = 10,
-	Union                  = 11,
-	Enum                   = 12,
-	Parameters             = 13,
-	Proc                   = 14,
-	Bit_Set                = 15,
-	Simd_Vector            = 16,
-	SOA_Struct_Fixed       = 17,
-	SOA_Struct_Slice       = 18,
-	SOA_Struct_Dynamic     = 19,
-	Relative_Pointer       = 20,
-	Relative_Multi_Pointer = 21,
-	Multi_Pointer          = 22,
-	Matrix                 = 23,
-	Soa_Pointer            = 24,
-	Bit_Field              = 25,
+	Invalid                      = 0,
+	Basic                        = 1,
+	Named                        = 2,
+	Generic                      = 3,
+	Pointer                      = 4,
+	Array                        = 5,
+	Enumerated_Array             = 6,
+	Slice                        = 7,
+	Dynamic_Array                = 8,
+	Map                          = 9,
+	Struct                       = 10,
+	Union                        = 11,
+	Enum                         = 12,
+	Parameters                   = 13,
+	Proc                         = 14,
+	Bit_Set                      = 15,
+	Simd_Vector                  = 16,
+	SOA_Struct_Fixed             = 17,
+	SOA_Struct_Slice             = 18,
+	SOA_Struct_Dynamic           = 19,
+	Relative_Pointer             = 20,
+	Relative_Multi_Pointer       = 21,
+	Multi_Pointer                = 22,
+	Matrix                       = 23,
+	Soa_Pointer                  = 24,
+	Bit_Field                    = 25,
+	Fixed_Capacity_Dynamic_Array = 26,
 }
 
 Type_Elems_Cap :: 4
@@ -219,13 +220,14 @@ Type :: struct {
 	custom_align: String,
 
 	// Used by:
-	// .Array            - 1   count: 0=len
-	// .Enumerated_Array - 1   count: 0=len
-	// .SOA_Struct_Fixed - 1   count: 0=len
-	// .Bit_Set          - 2   count: 0=lower, 1=upper
-	// .Simd_Vector      - 1   count: 0=len
-	// .Matrix           - 2   count: 0=row_count, 1=column_count
-	// .Struct           - <=2 count: 0=min_field_align, 1=max_field_align
+	// .Array                        - 1   count: 0=len
+	// .Enumerated_Array             - 1   count: 0=len
+	// .SOA_Struct_Fixed             - 1   count: 0=len
+	// .Bit_Set                      - 2   count: 0=lower, 1=upper
+	// .Simd_Vector                  - 1   count: 0=len
+	// .Matrix                       - 2   count: 0=row_count, 1=column_count
+	// .Struct                       - <=2 count: 0=min_field_align, 1=max_field_align
+	// .Fixed_Capacity_Dynamic_Array - 1   count: 0=cap
 	elem_count_len: u32le,
 	elem_counts:    [Type_Elems_Cap]i64le,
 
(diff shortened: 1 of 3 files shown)


core/reflect/types.odin
```diff
@@ -428,6 +434,13 @@ is_simd_vector :: proc(info: ^Type_Info) -> bool {
 	_, ok := type_info_base(info).variant.(Type_Info_Simd_Vector)
 	return ok
 }
+// Returns true when the type is a dynamic-array type ([dynamic]T), false otherwise.
```
Sai-Moen: Copy paste error?

core/encoding/cbor/marshal.odin
```diff
@@ -285,6 +285,32 @@ _marshal_into_encoder :: proc(e: Encoder, v: any, ti: ^runtime.Type_Info) -> (er
...
 
+
+	case runtime.Type_Info_Fixed_Capacity_Dynamic_Array:
+		if info.elem.id == byte {
+			raw := (^[dynamic]byte)(v.data)
+			return err_conv(_encode_bytes(e, raw[:]))
+		}
+
+		array_len := (^int)(uintptr(v.data) + info.len_offset)^
+		array_data := uintptr(v.data)
```
laytan: The bytes encoding looks like a copy paste from dynamic array which wasn't changed

```suggestion
	case runtime.Type_Info_Fixed_Capacity_Dynamic_Array:
		array_data := uintptr(v.data)
		array_len := (^int)(array_data + info.len_offset)^
		if info.elem.id == byte {
					raw := runtime.Raw_Slice{v.data, array_len}
					return err_conv(_encode_bytes(e, transmute([]byte)raw))
		}

```

core/odin/ast/ast.odin
```diff
@@ -786,6 +786,16 @@ Dynamic_Array_Type :: struct {
 	elem:        ^Expr,
 }
 
+Fixed_Capacity_Dynamic_Array_Type :: struct {
+	using node: Expr,
```
laytan:
```suggestion
	using node:  Expr,
```

core/reflect/reflect.odin
```diff
@@ -1919,6 +1940,23 @@ equal :: proc(a, b: any, including_indirect_array_recursion := false, recursion_
...
 		return true
+	case Type_Info_Fixed_Capacity_Dynamic_Array:
+		a_count := (^int)(uintptr(a.data) + v.len_offset)^
+		b_count := (^int)(uintptr(b.data) + v.len_offset)^
+		if a_count != b_count {
+			return false
+		}
+
+		for i in 0..<a_count {
+			// size the data is stored at the start, this is possible
```
laytan:
```suggestion
			// since the data is stored at the start, this is possible
```

src/checker.cpp
```diff
@@ -2276,6 +2276,11 @@ gb_internal void add_type_info_type_internal(CheckerContext *c, Type *t) {
 		add_type_info_type_internal(c, t_int);
 		break;
 
+	case Type_FixedCapacityDynamicArray:
+		add_type_info_type_internal(c, bt->FixedCapacityDynamicArray.elem);
+		add_type_info_type_internal(c, t_allocator);
```
laytan: The fixed capacity dynamic array won't use the allocator type right?

And I see on the other container types that the pointer of the elem type is also added. ie `add_type_info_type_internal(c, alloc_type_pointer(bt->FixedCapacityDynamicArray.elem));`

Also think `t_int` is supposed to be added here.

src/checker.cpp
```diff
@@ -2514,6 +2519,10 @@ gb_internal void add_min_dep_type_info(Checker *c, Type *t) {
 		add_min_dep_type_info(c, t_int);
 		break;
 
+	case Type_FixedCapacityDynamicArray:
+		add_min_dep_type_info(c, bt->FixedCapacityDynamicArray.elem);
```
laytan, line 2523: I see on the other container types that the pointer of the elem type is also added. ie `add_min_dep_type_info(c, alloc_type_pointer(bt->FixedCapacityDynamicArray.elem));`

src/types.cpp
```diff
@@ -2515,6 +2557,11 @@ gb_internal bool type_has_nil(Type *t) {
 	case Type_DynamicArray:
 	case Type_Map:
 		return true;
+
+	case Type_FixedCapacityDynamicArray:
+		// TODO(bill): should it have `nil`?
```
laytan: I'd say no, like normal arrays

src/types.cpp
```diff
@@ -2642,6 +2689,9 @@ gb_internal bool is_type_comparable(Type *t) {
 	case Type_Matrix:
 		return is_type_comparable(t->Matrix.elem);
 
+	case Type_FixedCapacityDynamicArray:
+		return false;
```
laytan, line 2693: Why wouldn't it be comparable if the element is?

tests/internal/test_fixed_capacity_dynamic_array.odin
laytan, line 1: I'd put this under tests/internal, not in tests/core at least, since it is not a core thing.

src/checker.cpp
gingerBill: This is a copy and paste error.

src/checker.cpp
gingerBill: I think I need to add the array of the type too not just the pointer.

src/types.cpp
gingerBill: My view is no. Since it is not a consistent comparable about due to the `len` parameter.

karl-zylinski:
I think this is a good idea, since its something I do so often in the language, so it's nice to have better support than "Small_Array".

leecommamichael:
Might be a nice replacement for this pattern which is all over my project:
```odin
  _elem_storage: [N]T,
  elems:         [dynamic]T,
...
elems = slice.into_dynamic(_elem_storage[:])
```

Though I have an alternative proc which opts for the `nil_allocator` rather than `into_dynamic`'s panic allocator. What would appending to a full array do given these changes?

Kelimion:
It appends the minimum of `(cap(array) - len(array)` - i.e. available space - and the count of new items to append.
If the capacity is 15, and 15 items have already been appended, that's zero, and that's what's returned. No OOM.
```odin
@builtin
append_fixed_capacity_elems :: proc "contextless" (array: ^$T/[dynamic; $N]$E, #no_broadcast args: ..E) -> (n: int) {
	Raw :: Raw_Fixed_Capacity_Dynamic_Array(N, E)
	raw := (^Raw)(array)

	n = min(N - len(array), len(args))

	when size_of(E) != 0 {
		for i in 0..<n {
			#no_bounds_check raw.data[raw.len + i] = args[i]
		}
	}

	raw.len += n
	return n
}
```

base/runtime/core_builtin.odin
```diff
@@ -686,13 +855,70 @@ append_string :: proc(array: ^$T/[dynamic]$E/u8, args: ..string, loc := #caller_
...
+append_fixed_capacity_elems :: proc "contextless" (array: ^$T/[dynamic; $N]$E, #no_broadcast args: ..E) -> (n: int) {
+	Raw :: Raw_Fixed_Capacity_Dynamic_Array(N, E)
+	raw := (^Raw)(array)
+
+	n = min(N - len(array), len(args))
+
+	when size_of(E) != 0 {
+		for i in 0..<n {
+			#no_bounds_check raw.data[i] = args[i]
+		}
```
leecommamichael: Lower bound is zero, wouldn't that make this more like an assignment of the front of the array? i.e. successive calls to append would overwrite the same region?

~~Unrelated: Am I reading right that this proc considers it not-a-failure so long as one of the varargs was appended?~~ Jeroen answered that

gingerBill: Yes. This is just a bug.

Kelimion: Now that you say it, I missing an offset of `raw.data` to account for the current length. It does read like it would "insert at front", @gingerBill.

If it has a capacity of 15, and len(arr) is 14, then it would only insert the first item of the given arguments, and return 1. In that sense it resembles the existing `append`, which also returns the number of elements it was able to insert, and `copy`, which does the same.

blob1807:
The doc format & command needs updating with the new type. If you're not already working on it.

PR #6411, opened 2026-03-13
fix various misspellings of "if" as "iff" across the codebase

NicePotato-MS:


blob1807:
`iff` means `if and only if`. it comes from the maths world.
https://en.wikipedia.org/wiki/If_and_only_if

NicePotato-MS:
> `iff` means `if and only if`. it comes from the maths world. https://en.wikipedia.org/wiki/If_and_only_if

oh huh i had no idea

blob1807:
no worries mate.

PR #6424, opened 2026-03-15
encoding/cbor: fix order-dependent partial unmarshals

andzdroid:
This changes CBOR struct unmarshalling to iterate over all map entries instead of stopping after len(destination_struct_fields) entries.

This fixes partial unmarshals that can break depending on the field ordering. Since the old behaviour only reads n fields, if the desired field is encoded at a later index, the loop would end before reaching the field.

The old code was iterating through all fields anyways just to discard them. I can add a field counter for similar behavour to discard remaining entries once all desired struct fields have been decoded, let me know if you want this.

### DIFF
--- a/core/encoding/cbor/unmarshal.odin
+++ b/core/encoding/cbor/unmarshal.odin
@@ -661,8 +661,7 @@ _unmarshal_map :: proc(d: Decoder, v: any, ti: ^reflect.Type_Info, hdr: Header, 
 		unknown := length == -1
 		fields := reflect.struct_fields_zipped(ti.id)
 
-		idx := 0
-		for ; idx < len(fields) && (unknown || idx < length); idx += 1 {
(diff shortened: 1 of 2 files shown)


laytan:
The patch looks like a fine fix. But:

I am confused by

> I can add a field counter for similar behavour to discard remaining entries once all desired struct fields have been decoded, let me know if you want this.

Could you elaborate? Looking at the code after this patch, it seems to handle skipping the other map entries that do not have a struct field already.

andzdroid:
Once the struct is filled, it can skip the loop on lines 680-695 where it tries to find the matching field and go straight to discarding the map entry. Just an optimization, not needed for correctness. As you said, it already discards the remaining map entries.

laytan:
Ah I see what you mean, I don't think that is necessary.

PR #6442, opened 2026-03-17
src: Support clang-style `+`/`-` prefixes in target features

Yawning:
This is primarily to un-fuck LLVM's auto-vectorization thinking it is a good idea to slow down the xorshiro256** PRNG by using vector instructions, but is probably useful for other things in the future.

Known issue:
- Functions that get inlined automatically by the LLVM inliner, apparently treats a `target-features` attribute that
  is full of disable directives as a sub-set of something that has all the features enabled, and will **SILENTLY** ignore it.

So this is only really useful for functions that can be `#force_inline`-ed, or cases where the annotations are appropriate all the way up.

### DIFF
--- a/core/math/rand/rand_xoshiro256.odin
+++ b/core/math/rand/rand_xoshiro256.odin
@@ -3,6 +3,18 @@ package rand
 import "base:intrinsics"
 import "base:runtime"
 
+when ODIN_ARCH == .amd64 || ODIN_ARCH == .i386 {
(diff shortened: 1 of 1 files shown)


alsakandari:
didn't I do it already in #5794

Kelimion:
> didn't I do it already in #5794

No, that's a project-wide command line argument. This could disable a cpu feature locally.

PR #6436, opened 2026-03-17
Update build_odin.sh to support LLVM 22

nathanchere:
I've only tried building this on Linux where it works fine building with LLVM 22.
I haven't verified that it 'just works' for Win or OSX builds.

PR #6444, opened 2026-03-18
Allow building with LLVM 22

JackJ30:
Sorry for the duplicate pr (#6436)

My distro updated to llvm 22 so I made this branch. I'm not totally sure if this is everything needed for the update, I saw in the release notes that they added the "AllocToken" pass and I'm not sure how to support that.

Feel free to close if you want to go with the other PR or do it yourself. I'm going to keep looking into the update though, to learn.


laytan:
For optimizations we use:
1. `-o:none` (no optimizations)
2. `-o:minimal` (just inlining iirc)
3. `-o:speed` the output of `opt -O2 -print-pipeline-passes <some-llvm-file>` (llvm file being a .ll which odin outputs with -build-mode:llvm)
4. `-o:aggressive` the same as 3, but with `-O3` instead

From 3 and 4 we remove passes regarding coro or openmp

I see that LLVM 22 has updated these a bit, so that will need to be changed too.

These are all defined in `llvm_backend_passes.cpp`

JackJ30:
I'm not very experienced with LLVM, but from what I can tell the same optimizations should work for 22. I'll probably close this PR though, since it seems like he's making progress in the other one.

PR #6460, opened 2026-03-21
Add `vendor:windows/wasapi`

jakubtomsu:
This PR adds bindings to [WASAPI](https://learn.microsoft.com/en-us/windows/win32/coreaudio/wasapi), a low-latency Windows audio API.

I have been using it in my audio engine for a while and it works great. I made the bindings by hand and with a bit of search-and-replace automation. I directly used the windows header C COM definitions as a base. Everything I tested was absolutely without issues, though of course I cannot 100% guarantee there aren't any typos.

It's not a part of `core:sys/windows` because it's a completely separate COM API in the windows headers. That said it does have a dependency on quite a few windows definitions.

There are a few duplicated structures like `WAVEFORMMAT*` which are in both windows and WASAPI. The windows ones could possibly get removed if `waveOut` got a separate package. Or they could just be updated to the newer variant.

Also IIRC @karl-zylinski mentioned he might want to add this as a backend to his Karl2D lib, so I hope it's useful to other odin users as well.

The basic API usage goes like this:
- Create `IMMDeviceEnumerator` COM object
- Get default audio endpoint `IMMDevice`
- Get the audio client and configure it with a mix format
- Get a render client
- in a loop, wait for the buffer event and fill audio samples


jakubtomsu:
I'll move this to `vendor:windows/wasapi`, I didn't realize there was a directory like this

PR #6466, opened 2026-03-23
Bit field intrinsics and fixes

jakubtomsu:
Changes:
- fixes `offset_of(some_bit_field)`, now it's an error instead of an internal compiler assertion
- implements `intrinsics.type_field_bit_offset(typeid, string)` to get a comptime bit field field offset in bits.
- implements `intrinsics.type_field_bit_size(typeid, string)` to get a comptime bit field field size in bits (the value after `|`, not the backing type size)

Example:
```odin
import "base:intrinsics"
import "core:fmt"

Bit_Field :: bit_field u16 {
    a: i32 | 5,
    b: u16 | 11,
}

#assert(intrinsics.type_field_type(Bit_Field, "b") == u16)
#assert(intrinsics.type_field_bit_offset(Bit_Field, "b") == 5)
#assert(intrinsics.type_field_bit_size(Bit_Field, "b") == 11)
```

Now it's also possible to do this for example. By default, bit fields use extremely slow `runtime::__read/__write_bits` procedures which manipulate individual bits. This works with raw 64-bit ints and can extract any field with a single shift and mask. This can serve as a temporary fix before the compiler does something more reasonable.
```odin
bit_field_extract :: proc(val: $T, $Field: string) -> intrinsics.type_field_type(T, Field)
where
    intrinsics.type_is_bit_field(T),
    size_of(T) <= size_of(u64)
{
    mask: u64 = (1 << intrinsics.type_field_bit_size(T, Field)) - 1
    extracted := (cast(u64)val >> intrinsics.type_field_bit_offset(T, Field)) & mask
    return cast(intrinsics.type_field_type(T, Field))extracted
}
```

This isn't the only use-case however. For example, one could write a procedure which manipulates bits of multiple bit field fields at once. These intrinsics just help us avoid hardcoding offsets and sizes.

### DIFF
--- a/base/intrinsics/intrinsics.odin
+++ b/base/intrinsics/intrinsics.odin
@@ -208,8 +208,6 @@ type_field_type :: proc($T: typeid, $name: string) -> typeid ---
 type_field_bit_size :: proc($T: typeid, $name: string) -> int where type_is_bit_field(T) ---
 type_field_bit_offset :: proc($T: typeid, $name: string) -> int where type_is_bit_field(T) ---
 
-type_bit_field_size :: proc($T: typeid, $name: string) -> typeid ---
-
 type_proc_parameter_count :: proc($T: typeid) -> int where type_is_proc(T) ---
 type_proc_return_count    :: proc($T: typeid) -> int where type_is_proc(T) ---
 


base/intrinsics/intrinsics.odin
```diff
@@ -205,6 +205,11 @@ type_bit_set_underlying_type :: proc($T: typeid) -> typeid where type_is_bit_set
 type_has_field  :: proc($T: typeid, $name: string) -> bool ---
 type_field_type :: proc($T: typeid, $name: string) -> typeid ---
 
+type_field_bit_size :: proc($T: typeid, $name: string) -> int where type_is_bit_field(T) ---
+type_field_bit_offset :: proc($T: typeid, $name: string) -> int where type_is_bit_field(T) ---
+
+type_bit_field_size :: proc($T: typeid, $name: string) -> typeid ---
```
gingerBill: Seems a little redundant if `type_field_bit_size` exists, but I understand the rationale.

jakubtomsu: oh whoops, `type_bit_field_size` is actually a typo, looks like I didn't remove that definition. I initially named the intrinsics `type_bit_field_*` but then changed them to `type_field_bit_*`

PR #6469, opened 2026-03-24
Revert "added get_page_size() to `core:mem/virtual`"

Kelimion:
Reverts odin-lang/Odin#6467

Turns out we already have `DEFAULT_PAGE_SIZE`.

### DIFF
--- a/core/mem/virtual/virtual.odin
+++ b/core/mem/virtual/virtual.odin
@@ -45,17 +45,6 @@ release :: proc "contextless" (data: rawptr, size: uint) {
 	_release(data, size)
 }
 
-get_page_size :: proc() -> int {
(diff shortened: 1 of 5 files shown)


Xaryen:
Probably worth mentioning there seems to be a second DEFAULT_PAGE_SIZE in mem.virtual which isn't an alias or even a constant but `DEFAULT_PAGE_SIZE := uint(4096)`

Carlyle-Foster:
> Probably worth mentioning there seems to be a second DEFAULT_PAGE_SIZE in mem.virtual which isn't an alias or even a constant but `DEFAULT_PAGE_SIZE := uint(4096)`

that's just the declaration of the same variable, not a second one. `DEFAULT_PAGE_SIZE` isn't a constant, but rather initialized be `_platform_memory_init()`

PR #6471, opened 2026-03-24
rename DEFAULT_PAGE_SIZE to PAGE_SIZE

Carlyle-Foster:
to reflect that it's often not just a default

### DIFF
--- a/core/crypto/argon2id/argon2id.odin
+++ b/core/crypto/argon2id/argon2id.odin
@@ -143,7 +143,7 @@ derive :: proc(
 	m_ := 4 * u64(p) * (m / u64(4 * p))
 	b := mem.alloc_bytes_non_zeroed(
 		int(m_) * BLOCK_SIZE_BYTES,
-		alignment = mem.DEFAULT_PAGE_SIZE,
+		alignment = mem.PAGE_SIZE,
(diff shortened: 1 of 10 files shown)


Kelimion:
And we have another hardcoded one in `core:mem/alloc.odin`.
```odin
DEFAULT_PAGE_SIZE ::
	64 * 1024 when ODIN_ARCH == .wasm32 || ODIN_ARCH == .wasm64p32 else
	16 * 1024 when ODIN_OS == .Darwin && ODIN_ARCH == .arm64 else
	4 * 1024
```

@gingerBill may want to consider whether and how to consolidate these.

Carlyle-Foster:
> And we have another hardcoded one in `core:mem/alloc.odin`.
> 
> ```odin
> DEFAULT_PAGE_SIZE ::
> 	64 * 1024 when ODIN_ARCH == .wasm32 || ODIN_ARCH == .wasm64p32 else
> 	16 * 1024 when ODIN_OS == .Darwin && ODIN_ARCH == .arm64 else
> 	4 * 1024
> ```
> 
> @gingerBill may want to consider whether and how to consolidate these.

my latest commits are a draft of consolidation

PR #6473, opened 2026-03-25
Add fast_obj bindings to vendor

blob1807:
Adds [`fast_obj`](https://github.com/thisistherk/fast_obj) bindings to vendor. This PR don't include any binaries, so the CI runner should fail. I ran `odin check` with all the same flags locally with that in mind.

It's opened as a draft because I've only tested on Windows haven't tested on Linux or WASM yet, and can't test on Darwin. 
For WASM, the only thing missing from the `vendor:libc-shim` is `memmove`. For now I've assumed Odin already defines it.

The build scripts & foreign import stuff were taken from `vendor:cgltf`, so they should work fine.

Why not something like [`tinyobjloader-c`](https://github.com/syoyo/tinyobjloader-c)? Admittedly, mostly personal preference. I prefer `fast_obj`'s simpler api, and it also meshes better with Odin IMO.

Edit:
This is for the [last commit](https://github.com/thisistherk/fast_obj/commit/d620667f10a548dee94dbc8c144bb22f79162176) as of making, and not the latest [release](https://github.com/thisistherk/fast_obj/releases/tag/v1.3). The only api difference being support for face lines (`l` command).


gingerBill:
A direct port would probably be a much better idea, especially considering how small the library is.

blob1807:
I had thought about doing that, but this was the easier path.

It was also going to be one questions I had, when I opened for review. 

I'll leave it open until a source port of either is made, or you can feel free to close it instead.

gingerBill:
> I had thought about doing that, but this was the easier path.

That's not necessarily a good argument when this is going to be used by other people, and we want high quality stuff.

blob1807:
Understandable. Would you be okay if I submitted a source port PR instead?
I'm already going to be doing now to remove an external dep.

PR #6481, opened 2026-03-26
Fix cg_emit_transmute inner switch for int sources with raw mismatch

cuiweixie:
In `cg_emit_transmute`, when the value is TB_INT and the raw encoding differs from the destination, the inner `switch` repeated `value.node->dt.type` (still `TB_INT`), so only the inner `TB_INT` arm could run. The `TB_FLOAT` and `TB_PTR` arms were dead. Switch on `dt.type` so int→float and int→pointer transmutes emit the intended TB ops.


gingerBill:
I would not waste any time on the tilde stuff. I am going to delete it all soon.

PR #6517, opened 2026-04-04
core/os: Add dir procedure

juaumjuaumjuaum:
Adds a `dir` procedure which complements the `base` procedure already existing.

Let me know if there's anything I should have done properly, happy to learn.

Cheers,


juaumjuaumjuaum:
@Kelimion 

I am now wondering, should an empty path just return an empty string as well?

Need to update core:path/filepath.dir aswell

Related:  #6528

PR #6518, opened 2026-04-04
Replace table driven bits.len with direct calculation

DuchGhast:


gingerBill:
We wanted these separate procedures and not to made generic like that.

DuchGhast:
Would you be ok with using this technique instead of the table approach, if I would separate them into separate procedures and not used the generic approach?

PR #6525, opened 2026-04-07
[win32 api] Add Reg Association API

blob1807:
Adds the registry association API.


core/sys/windows/types.odin
```diff
@@ -4277,17 +4305,19 @@ SHACF_AUTOAPPEND_FORCE_OFF  :: 0x80000000  // Ignore the registry default and fo
 
 LWSTDAPI :: HRESULT
 
-CLSID_FileOpenDialog := &GUID{0xDC1C5A9C, 0xE88A, 0x4DDE, {0xA5, 0xA1, 0x60, 0xF8, 0x2A, 0x20, 0xAE, 0xF7}}
-CLSID_FileSaveDialog := &GUID{0xC0B4E2F3, 0xBA21, 0x4773, {0x8D, 0xBA, 0x33, 0x5E, 0xC9, 0x46, 0xEB, 0x8B}}
-CLSID_TaskbarList    := &GUID{0x56FDF344, 0xFD6D, 0x11d0, {0x95, 0x8A, 0x00, 0x60, 0x97, 0xC9, 0xA0, 0x90}}
+CLSID_FileOpenDialog    := &GUID{0xDC1C5A9C, 0xE88A, 0x4DDE, {0xA5, 0xA1, 0x60, 0xF8, 0x2A, 0x20, 0xAE, 0xF7}}
+CLSID_FileSaveDialog    := &GUID{0xC0B4E2F3, 0xBA21, 0x4773, {0x8D, 0xBA, 0x33, 0x5E, 0xC9, 0x46, 0xEB, 0x8B}}
+CLSID_TaskbarList       := &GUID{0x56FDF344, 0xFD6D, 0x11d0, {0x95, 0x8A, 0x00, 0x60, 0x97, 0xC9, 0xA0, 0x90}}
+CLSID_QueryAssociations :=  GUID{0xa07034fd, 0x6caa, 0x4954, {0xac, 0x3f, 0x97, 0xa2, 0x72, 0x16, 0xf9, 0x8a}}
```
krnowak, line 4311: Why not a pointer? (Missing `&` before `GUID`?)

blob1807: Because it's only used with `AssocCreate`, which just takes a plain `GUID`.
The pointer is just a qol, as that's how you'd normally use them with something like `CoCreateInstance`.

core/sys/windows/types.odin
```diff
@@ -5254,3 +5369,47 @@ Com_Error_Bits :: enum DWORD {
 }
 Com_Error :: distinct bit_set[Com_Error_Bits; DWORD]
 
+FILETYPEATTRIBUTEFLAGS_Bits :: enum DWORD {
```
gingerBill:
```odin
FILETYPEATTRIBUTEFLAG :: enum DWORD {
```

blob1807: Change made.

blob1807: For the future, is there a general rule for when to use `_Bits` vs non-plural?

PR #6550, opened 2026-04-10
address hint to linux virtual mem reserve & made `make_aligned` a builtin

sudokit:
- added a `address_hint` parameter to be passed down to mmap on linux as a hint to around where the user would like .to reserve the memory.  defaults to 0, which it was hardcoded to before.
- made `make_aligned` a builtin


gingerBill:
Please make these separate PRs since they are literally separate problems.

Issue #5821, opened 2025-10-17
core:virtual/arena: reserved amount is wrong for rare cases because of alignment, causing runtime assert.


RilleP:
## Context

odin version dev-2025-10:4b8751beb

## Current Behavior

Assertion is hit core/mem/virtual/virtual.odin(178:2) runtime assertion: block.committed <= block.reserved

### Steps to Reproduce

```
import "core:mem"
import "core:mem/virtual"
import "core:fmt"
main :: proc() {
	arena: virtual.Arena;
	error := virtual.arena_init_growing(&arena, mem.Megabyte*16);	
	assert(error == nil);

	for ii in 0..<32 {
		bytes, alloc_err := virtual.arena_alloc(&arena, mem.Megabyte*1, 4);
		fmt.printf("Allocated %d bytes, %d\n", len(bytes), arena.curr_block.used);
	}
}
```

I have done some investigation and i believe that this line, in memory_block_alloc, should not add alignment.
`total_size     := reserved + alignment + size_of(Platform_Memory_Block)`

And also the call to memory_block_alloc in arena_alloc_unguarded, should not pass the alignment. 
`new_block := memory_block_alloc(needed, block_size, alignment, {}) or_return`

Though this fix breaks other things, so maybe not.

PR #6552, opened 2026-04-10
Fix arena overcommit bug (revisited)

alektron:
### Version
dev-2026-04:98d6b3bc8

### Context

In #6398 and #5821 a bug with the virtual arena allocator was encountered and attempted to fix.
This PR reverts the change in #6398 as it does not correctly fix the bug and can cause other issues down the line, and fixes the actual root cause.

### Bug
As the other PR/issue correctly identified, passing an aligment other than 0 to `memory_block_alloc` can cause an assert to later be triggered in `do_commit_if_necessary`.
The assert seems to trigger due to overcommitting the reserved memory when in reality the size of the reserved memory is just not calculated correctly and the commit was actually safe.

### Fix
Not passing in the alignment (as implemented by #6398, even with the changes to `block_size`) can cause other issues as the `base_offset`, and by extension the base pointer, will not be calculated correctly, should the alignment requirement be bigger than that of `Platform_Memory_Block`.

The real issue comes from the fact that the `Platform_Memory_Block` and `Memory_Block` disagree on the size of actual reserved memory. `Platform_Memory_Block` correctly assumes `total_size` to be reserved, which includes the alignment and the header size.
`Memory_Block` however only assumes the initially requested `reserved`. This does not include the header size, which is intended, BUT it does also not include the alignment which would be correct, if `alignment` would actually contribute to `base_offset`. But it does not do that if the alignment requirement is already met by `align_of(Platform_Memory_Block)`.

Removing `alignment` from the calculation of `total_size` (as suggested in #5821) is however the wrong approach since it would break the cases where `alignment` actually contributes to `base_offset`.

The correct change is therefore to adjust the calculation of the `Memory_Block`'s `reserved` size to use `total_size - base_offset` instead.

### DIFF
--- a/core/mem/virtual/arena.odin
+++ b/core/mem/virtual/arena.odin
@@ -141,9 +141,9 @@ arena_alloc_unguarded :: proc(arena: ^Arena, size: uint, alignment: uint, loc :=
 
 			needed := mem.align_forward_uint(size, alignment)
 			needed = max(needed, arena.default_commit_size)
-			block_size := max(needed, arena.minimum_block_size) + alignment
+			block_size := max(needed, arena.minimum_block_size)
 
-			new_block := memory_block_alloc(needed, block_size) or_return
+			new_block := memory_block_alloc(needed, block_size, alignment) or_return
 			new_block.prev = arena.curr_block
 			arena.curr_block = new_block
 			arena.total_reserved += new_block.reserved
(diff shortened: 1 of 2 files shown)


Faker-09:
Thanks a lot for tracking that down and providing the correct fix.

PR #6558, opened 2026-04-12
`address_hint` onto virtual memory reserve proc

sudokit:


core/mem/virtual/virtual_darwin.odin
```diff
@@ -2,8 +2,8 @@ package mem_virtual
 
 import "core:sys/posix"
 
-_reserve :: proc "contextless" (size: uint) -> (data: []byte, err: Allocator_Error) {
-	result := posix.mmap(nil, size, {}, {.ANONYMOUS, .PRIVATE})
+_reserve :: proc "contextless" (size: uint, address_hint := uintptr(0)) -> (data: []byte, err: Allocator_Error) {
```
gingerBill: It's a better idea if all of the `_reserve` signatures were `(size: int, address_hint: uintptr)` instead of using default values.

`reserve` should be fine with default values.

sudokit: Yeah fair enough. Fixed!

gingerBill (approved):

PR #6575, opened 2026-04-15
core:path/filepath Create wrappers for allocating procedures

juaumjuaumjuaum:
- removes a duplicate procedure in core:os and core:path/filepath
- core:path/filepath wraps core:os procedures with a context.allocator as  the default
- fixes a bug where an empty path call to os.dir() returns "." instead of ""

### DIFF
--- a/core/path/filepath/path.odin
+++ b/core/path/filepath/path.odin
@@ -2,6 +2,7 @@
 // To process paths such as URLs that depend on forward slashes regardless of the OS, use the slashpath package.
 package filepath
 
+import "base:runtime"
 import "core:os"
 import "core:strings"
 
@@ -32,7 +33,15 @@ Join all `elems` with the system's path separator and normalize the result.
 
 For example, `join_path({"/home", "foo", "bar.txt"})` will result in `"/home/foo/bar.txt"`.
 */
-join :: os.join_path
+join :: proc(
+	elems: []string,
+	allocator := context.allocator,
+) -> (
+	joined: string,
+	err: runtime.Allocator_Error,
+) {
+	return os.join_path(elems, allocator)
+}
 
 /*
 	Returns leading volume name.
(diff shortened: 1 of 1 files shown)


core/os/path.odin
```diff
@@ -390,11 +390,11 @@ e.g.
 	'/home/foo/bar.tar.gz' -> '/home/foo'
 	'path/to/name.tar.gz'  -> 'path/to'
 
-Returns "." if the path is an empty string.
+Returns "" if the path is an empty string.
 */
 dir :: proc(path: string) -> string {
 	if path == "" {
-		return "."
+		return ""
```
Kelimion: `os.base` also returns `"."` if the path is empty, so we should probably change that one as well for consistency.

core/os/path.odin
Kelimion, line 379: `return ""` +  the comment, to mirror the change to `dir`.

juaumjuaumjuaum:
clean_path also returns "." if the path is empty, starting to wonder if there's a reason why that is?

gingerBill:
Returning `"."` when the path is empty is not a bug and actually required behaviour as it represents the current directory.

Kelimion:
Hmm, true.

juaumjuaumjuaum:
I have reverted the changes on the output for empty paths, sorry for the trouble.

core/path/filepath/path.odin
```diff
@@ -136,15 +145,32 @@ long_ext :: os.long_ext
...
+}
 
 /*
 Returns the result of replacing each path separator character in the path
 with the specific character `new_sep`.
 
 *Allocates Using Provided Allocator*
 */
-replace_path_separators := os.replace_path_separators
+replace_separators :: proc(
```
Kelimion, line 164: This used to be `replace_path_separators`, now it's `replace_separators`. Why?

juaumjuaumjuaum: Ah sorry, I forgot to ask.

In every other procedure the path is taken out of the name, to this was the only one that still has it

Issue #6543, opened 2026-04-10
sdl3_mixer CreateMixerDevice's spec parameter should be nullable


cndofx:
The official sdl3 docs and their examples both show that null is a valid value for this parameter, but odin's bindings don't allow this:

`CreateMixerDevice          :: proc(devid: SDL.AudioDeviceID, #by_ptr spec: SDL.AudioSpec) -> ^Mixer ---`

https://wiki.libsdl.org/SDL3_mixer/MIX_CreateMixerDevice
https://examples.libsdl.org/SDL3_mixer/basics/01-load-and-play/

PR #6576, opened 2026-04-16
[SDL3_mixer] [bug] Removed #by_ptr from args that are valid as nil

victor-Lopez25:
The docs say nil is not valid for the rest of the functions that still have `#by_ptr`

Should I just remove `#by_ptr` entirely to keep it consistent or keep it?

### DIFF
--- a/vendor/sdl3/mixer/sdl3_mixer.odin
+++ b/vendor/sdl3/mixer/sdl3_mixer.odin
@@ -76,8 +76,8 @@ foreign lib {
 	Quit                       :: proc() ---
 	GetNumAudioDecoders        :: proc() -> c.int ---
 	GetAudioDecoder            :: proc(index: c.int) -> cstring ---
-	CreateMixerDevice          :: proc(devid: SDL.AudioDeviceID, spec: ^SDL.AudioSpec) -> ^Mixer ---
-	CreateMixer                :: proc(spec: ^SDL.AudioSpec) -> ^Mixer ---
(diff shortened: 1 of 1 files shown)


victor-Lopez25:
This is not a problem because you can pass `(^Type)(nil)^` to those functions

victor-Lopez25:
Reopening since `(^Type)(nil)^` will try to read nil which causes a crash

victor-Lopez25:
fixes this issue: #6543

cndofx:
It doesn't look like the `MIX_CreateMixer` spec in particular should be nullable though. null makes sense in `MIX_CreateMixerDevice` because a default format can be obtained from the given device, but not otherwise.

https://wiki.libsdl.org/SDL3_mixer/MIX_CreateMixer
> An audio format must be specified. This is the format it will output in. This cannot be NULL.

victor-Lopez25:
@cndofx you're right but it's very strange to have only one of the two use a pointer, I left it like this because I thought this would be much less confusing, do you think I should leave `CreateMixer` with `#by_ptr`?

victor-Lopez25:
It does get checked anyway in SDL_mixer: https://github.com/libsdl-org/SDL_mixer/blob/main/src/SDL_mixer.c#L821

cndofx:
I agree it looks a little strange, but I personally think the `CreateMixer` binding should stay as it was. If anything, I think the `Maybe(^SDL.AudioSpec)` is the more confusing one as it doesn't clearly communicate that the parameter is not to be modified, but I don't think much can be done about that without a distinct const pointer type in odin. I admit I'm really not an expert on idiomatic odin though so I can't really say too much about the best way to handle something like this

PR #6580, opened 2026-04-16
[vendor/sdl3] move satellite libraries to relative import

A1029384756:
this should make the sdl3 bindings easier to vendor by making their imports relative. before, vendored bindings would still reference the copy of `sdl3` in `vendor`, resulting in duplicate symbol errors


gingerBill:
If you're vendoring SDL yourself, it was clearly only 4 lines of code change, and thus not that much harder to maintain. So I don't see the big need for it.

PR #6581, opened 2026-04-16
Omit value when formatting map[T]struct{}

jakubtomsu:
Consider the following code example where I use a map as a set and print it:
```odin
m: map[int]struct{}
// ...
fmt.println(m)
```
The output looks something like this:
```
map[1={}, 3={}, 0={}, 2={}]
```
The `={}` is clearly redundant, the map is a *set*.

This change just skips the value output in the case map value has size 0, so the result would look more like this:
```
map[1, 3, 0, 2]
```


gingerBill:
This is honestly not a bad decision!

PR #6582, opened 2026-04-16
Fix array counts for floats that are exactly representable as ints

Znarf64:


gingerBill:
Can you split the changes please.

Znarf64:
I hate git, give me a second

PR #6589, opened 2026-04-17
[vendor/sdl3] added bindings for tray click properties added in 3.2.0

A1029384756:
Allows for tray click actions introduced in SDL 3.2.0


A1029384756:
misread the git blame, this won't be live until 3.6.0

PR #6592, opened 2026-04-19
Fix `bit_set`s inside of `bit_field`s in vulkan.

Znarf64:
This fixes eg. `vk.AccelerationStructureInstanceKHR`:
```C
typedef struct VkAccelerationStructureInstanceKHR {
    VkTransformMatrixKHR          transform;
    uint32_t                      instanceCustomIndex:24;
    uint32_t                      mask:8;
    uint32_t                      instanceShaderBindingTableRecordOffset:24;
    VkGeometryInstanceFlagsKHR    flags:8; // This was translated to GeometryInstanceFlagKHR (singular) previously
    uint64_t                      accelerationStructureReference;
} VkAccelerationStructureInstanceKHR;
```

### DIFF
--- a/src/check_type.cpp
+++ b/src/check_type.cpp
@@ -1134,6 +1134,15 @@ gb_internal void check_bit_field_type(CheckerContext *ctx, Type *bit_field_type,
 				error(f->bit_size, "A bit_field's specified bit size cannot exceed its type, got %lld, expect <=%lld", cast(long long)bit_size_i64, cast(long long)sz);
 				bit_size_i64 = sz;
 			}
+			if (is_type_bit_set(type)) {
+				Type *bit_set = base_type(type);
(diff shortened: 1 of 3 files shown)


Kelimion:
Was this intended to be closed a minute after opening?

Znarf64:
uh it does not compile :sweat:

Znarf64:
because `bit_fields`s can not include `bit_set`s

gingerBill:
So naively supporting `bit_set`s to `bit_field` fields is a really bad idea. They were not forgotten about, but rather they need very special care to actually implement them correctly. 

I am going to close this PR specifically for that reason.

Znarf64:
Just out of curiousity, what is so special about them? Anyway the vulkan bindings should probably be changed

PR #4961, opened 2025-03-23
[doc-core:math] Documentation for `core:math` package. (partials)

kelreeeeey:
Documented: (all of this will be gradually updated)

- [x] `math.Float_Class` enum
- [x] `math.sqrt` proc group
- [x] `math.sin` proc group
- [x] `math.cos` proc group
- [x] `math.tan` proc group

---

- [x] `math.to_radians` proc group
- [x] `math.to_degrees` proc group

---

- [x] `DEG_PER_RAD` constant
- [x] `RAD_PER_DEG` constant

---

- [x] `math.asin` proc group
- [x] `math.acos` proc group
- [x] `math.atan` proc group

---

More detail underlying explanation/issue in https://github.com/odin-lang/Odin/issues/2358#issuecomment-2746308757

PR #6598, opened 2026-04-22
rework documentation for core:math package (Issue #2358)

daninaga:
Created my own PR since the original (#4961) seems stale and not being worked on. Examples have been updated with verifiable output and more succinct description applied.
The idea is to finally complete the documentation of the math module. 

AfterNote: only the main `math.odin` will be taken care in this PR, the procs for the standalone files are to be handled on a later contribution.

Related to #2358 

The following are tracked so far:
Constants
COMPLETION: 74/74 (100%)
- [x] DONE
--------------------------------------
Proc Group
COMPLETION: 44/44 (100%) [8 SKIPPED]
- [x] ~~std-copy_sign~~
- [x] ~~std-erf~~
- [x] ~~std-erfc~~
- [x] ~~std-gamma~~
- [x] ~~std-lgamma~~
- [x] ~~std-log1p~~
- [x] ~~std-sincos~~
- [x] ~~std-ln~~
- [x] DONE
----------------------------------
Procedures
COMPLETION: 462/462 (100%) [424 SKIPPED] --> need a way to account for the proc groups
- [x] DONE


daninaga:
for whoever is reviewing, I did my best; please let me know if something is not up-to-par

PR #6599, opened 2026-04-22
Fixes #6594

### LINKED ISSUE #6594: Declaring variable and assigning it to itself triggers assert
## Context

The compiler was built from (unmodified) source fetched from github.

```
        Odin:    dev-2026-04:a896fb2b4
        OS:      Windows 10 Professional (version: 22H2), build 19045.6466
        CPU:     Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
        RAM:     16344 MiB
        Backend: LLVM 20.1.0
```

## Expected Behavior
Compiler should give an error instead of triggering assert in the llvm backend.

## Current Behavior
When declaring a variable and assigning it to itself (e.g. `a := a`), an assert is triggered.
```
C:\Users\Brink\src\third_party\Odin-github\src\llvm_backend_general.cpp(1890): Assertion Failure: `type != t_invalid`
```
(Obviously you would never do this deliberately, but I happened to stumble upon this).

### Steps to Reproduce
Compiling the following program is enough for me to trigger the error:
```odin
package main

a := a

main :: proc() {
    a + 1 // needs to be used to be triggered
}

```

Using this command in the directory with the file above:
```
Odin build .
```

### Failure Logs

3rd-Party-Guy:
Fixes #6594 
Checker now reports "Illegal declaration cycle" before codegen, preventing a compiler crash.
Added a unit test inside tests/issues.

### DIFF
--- a/src/check_decl.cpp
+++ b/src/check_decl.cpp
@@ -1771,6 +1771,12 @@ gb_internal void check_global_variable_decl(CheckerContext *ctx, Entity *e, Ast 
 
 	Operand o = {};
 	check_expr_with_type_hint(ctx, &o, init_expr, e->type);
+	if (check_vet_shadowing_assignment(ctx->checker, e, init_expr)) {
+		error(e->token, "Illegal declaration cycle of `%.*s`", LIT(e->token.string));
+		o.mode = Addressing_Invalid;
+		o.type = t_invalid;
+		e->type = t_invalid;
+	}
 	check_init_variable(ctx, e, &o, str_lit("variable declaration"));
 	if (e->Variable.is_rodata && o.mode != Addressing_Constant) {
 		ERROR_BLOCK();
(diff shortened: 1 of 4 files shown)


src/check_decl.cpp
```diff
@@ -1980,6 +1986,18 @@ gb_internal void check_entity_decl(CheckerContext *ctx, Entity *e, DeclInfo *d,
...
+		case Entity_Variable:
+		case Entity_Constant:
+		case Entity_TypeName:
+			track_cycle_path = true;
+			break;
+		}
+		if (track_cycle_path) {
+			check_type_path_push(&c, e);
+			defer (check_type_path_pop(&c));
+		}
```
gingerBill, line 1999: `defer` is scope bound, not function bound.

Write this:

```odin
if (track_cycle_path) {
	check_type_path_push(&c, e);
}
defer (if (track_cycle_path) {
	check_type_path_pop(&c)
});
```

gingerBill (changes requested):

src/check_decl.cpp
3rd-Party-Guy: Done in latest push

PR #6605, opened 2026-04-23
Native Array Casting Semantics

gingerBill:
```odin
a := [4]i32{1, 4, 9, 16}
b := ([4]f32)(a)
```

### DIFF
--- a/core/math/linalg/general.odin
+++ b/core/math/linalg/general.odin
@@ -368,11 +368,8 @@ cubic :: proc "contextless" (v1, v2, v3, v4: $T/[$N]$E, s: E) -> T #no_bounds_ch
 
 
 @(require_results)
-array_cast :: proc "contextless" (v: $A/[$N]$T, $Elem_Type: typeid) -> (w: [N]Elem_Type) #no_bounds_check {
-	for i in 0..<N {
-		w[i] = Elem_Type(v[i])
-	}
-	return
+array_cast :: proc "contextless" (v: $A/[$N]$T, $Elem_Type: typeid) -> [N]Elem_Type #no_bounds_check {
+	return ([N]Elem_Type)(v)
 }
 
 @(require_results)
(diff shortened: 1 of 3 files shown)


core/math/linalg/general.odin
```diff
@@ -385,28 +382,28 @@ matrix_cast :: proc "contextless" (v: $A/matrix[$M, $N]$T, $Elem_Type: typeid) -
...
-@(require_results) to_u64  :: #force_inline proc(v: $A/[$N]$T) -> [N]u64  { return array_cast(v, u64)  }
-@(require_results) to_uint :: #force_inline proc(v: $A/[$N]$T) -> [N]uint { return array_cast(v, uint) }
-
-@(require_results) to_complex32     :: #force_inline proc(v: $A/[$N]$T) -> [N]complex32     { return array_cast(v, complex32)     }
-@(require_results) to_complex64     :: #force_inline proc(v: $A/[$N]$T) -> [N]complex64     { return array_cast(v, complex64)     }
-@(require_results) to_complex128    :: #force_inline proc(v: $A/[$N]$T) -> [N]complex128    { return array_cast(v, complex128)    }
-@(require_results) to_quaternion64  :: #force_inline proc(v: $A/[$N]$T) -> [N]quaternion64  { return array_cast(v, quaternion64)  }
-@(require_results) to_quaternion128 :: #force_inline proc(v: $A/[$N]$T) -> [N]quaternion128 { return array_cast(v, quaternion128) }
-@(require_results) to_quaternion256 :: #force_inline proc(v: $A/[$N]$T) -> [N]quaternion256 { return array_cast(v, quaternion256) }
+@(require_results) to_f16  :: #force_inline proc(v: $A/[$N]$T) -> [N]f16  { return #force_inline array_cast(v, f16)  }
```
Kelimion, line 385: I'd change this to `@(require_results) to_f16  :: #force_inline proc(v: $A/[$N]$T) -> [N]f16  { return ([N]f16)v }`, and so forth for the rest.

gingerBill: I'll change it after the merge.

blob1807:
Is there a reason not to have an upper limit to the element count, or not generating a loop like broadcasting does? 
I wouldn't think generating a 1000 inline casts would be a good idea.

gingerBill:
> Is there a reason not to have an upper limit to the element count, or not generating a loop like broadcasting does? I wouldn't think generating a 1000 inline casts would be a good idea.

I forgot to add this, but that is a very good point. After a certain size, that should be in a `for` loop.

PR #6610, opened 2026-04-24
Add `#optional_allocator_error` to `filepath.join` and `filepath.clean`

mr0x13f:
Add `#optional_allocator_error` to `filepath.join` and `filepath.clean`.

This tripped me up porting older code using dev-2026-02 where the directive was present:
https://github.com/odin-lang/Odin/blob/dev-2026-02/core/path/filepath/path_windows.odin?plain=1#L90
https://github.com/odin-lang/Odin/blob/dev-2026-02/core/path/filepath/path.odin?plain=1#L279


gingerBill:
No.

PR #6612, opened 2026-04-24
core/sys/posix/dlfcn: add RTLD_NOLOAD and RTLD_NODELETE definitions

rnpnr:
RTLD_NOLOAD is used to get a handle to a library which is already loaded. This is useful for example to get a handle to the Renderdoc in application API when Renderdoc is attached to the application.

RTLD_NODELETE is used to dispose of the handle without destroying the handle's static storage which will remain unchanged if the handle is reopened later.

### DIFF
--- a/core/sys/posix/dlfcn.odin
+++ b/core/sys/posix/dlfcn.odin
@@ -74,9 +74,11 @@ foreign lib {
 }
 
 RTLD_Flag_Bits :: enum c.int {
-	LAZY   = log2(RTLD_LAZY),
-	NOW    = log2(RTLD_NOW),
-	GLOBAL = log2(RTLD_GLOBAL),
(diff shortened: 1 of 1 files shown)


laytan:
These flags are not defined in the POSIX standard which this package targets. https://pubs.opengroup.org/onlinepubs/9799919799/functions/dlopen.html

rnpnr:
> These flags are not defined in the POSIX standard which this package targets. https://pubs.opengroup.org/onlinepubs/9799919799/functions/dlopen.html

They are defined in every implementation of dlopen Odin is targeting though. This is the only way to obtain an equivalent of windows GetModuleHandle() function.

May I ask how an application is expected to use dlopen? Because it seems kind of silly to have an odin wrapper for it when we just have to define the interface ourselves anyways.

PR #6617, opened 2026-04-25
Fixing bug introduced with ef275c5 (allowing the use of typeid slice literals).

LeandroLibanio28H:
Adding the check of Addressing_Type that is present in `check_expr_with_type_hint` to `check_multi_expr_with_type_hint` so types used as typeid values are not treated as invalid.

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -958,7 +958,7 @@ gb_internal i64 check_distance_between_types(CheckerContext *c, Operand *operand
 			}
 		}
 	}
-	
+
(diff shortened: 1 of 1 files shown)


gingerBill:
I won't accept this purely because of all of the whitespace changes.

LeandroLibanio28H:
> I won't accept this purely because of all of the whitespace changes.

Sorry, didn't see the whitespace changes, I switched my desktop to linux and I really don't know why this happened, the editor (nvim) and plugins are all the same.

PR #6631, opened 2026-04-28
removed Haiku from supported targets

heavyrain266:
Changes were split into logical commits, the `core:posix` package comes with most significant set of them.


laytan:
Don't forget to remove here: https://odin-lang.org/docs/install/

heavyrain266:
Thanks, will do. Tests should pass now (tested on Ubuntu 24.04)

gingerBill:
I'm not sure if we should remove Haiku, even if it has little active use. Essence was a good idea to remove though.

heavyrain266:
The idea popped up during an audit because I was unable to get it running on any of the hardware at hand (x86 and ARM). Are there any people actively testing Odin on Haiku? We don't have a CI runners either.

gingerBill:
It's a good question if we have any active Haiku users. The problem is we cannot know unless people tell us. We don't have telemetry (thank God) in the compiler.

heavyrain266:
I'm going to close the PRs for now.

gingerBill:
I'm going to leave this open because we might remove it.

kenaryn:
Perhaps adding a pinned poll on Discord or a kind of top banner on the website's homepage (more work required but better visibility, and could be reused as a template for further endeavours to probe the community) for a short period like 3-4 weeks could bypass the lack of telemetry and provide a actionable feed-back.

heavyrain266:
In any case, I tried to keep it easy to walk through the history and pull out commits to make it work again. Only the changes with sole build tags were merged with larger ones.

System APIs are in C++, not C, most packages don't even work there, POSIX oddities were hard to untangle, rewritten `os` package didn't received a port either, and nobody complained thus far.

heavyrain266:
I did some research and tried to build the compiler on an HP thin client t520 and in VMs. The LLVM port I tried was 16. Without networking support, I tried to install `llvm16-*.hpkg` manually, but it is not possible due to broken packages, as seen below:

<br>

<br>

waddlesplash:
(Haiku developer here.) Hello, just came across this. Personally I don't use Odin, so I have no stake in this, but happy to help others in case there's interest; figured I'd answer some of the questions in here:

> System APIs are in C++, not C

Pretty much all the APIs you should need for porting a language are in C. We are pretty compliant with POSIX-2024 (any discrepancies besides the very few explicitly mentioned in our documentation, please file a bug), and have various non-POSIX C APIs for things POSIX doesn't cover (e.g. RAM usage) or implement a BSD extension.

> POSIX oddities were hard to untangle

Not sure what you mean by this.

> Without networking support

Did networking not even work in VMs? On bare metal, if nothing else USB tethering should work. But PCIe ethernet hardware usually works fine too (we use FreeBSD's drivers.)

> I tried to install llvm16-*.hpkg manually, but it is not possible due to broken packages

This looks like you've tried to install some mismatched combination of packages, or tried to install a set of packages in the wrong order. Did you try using `pkgman` instead of the GUI? It might work better for this, or at least give a clearer idea of what dependencies you missed (just use `pkgman install file.hpkg file2.hpkg ...`).

PR #6635, opened 2026-04-29
core/crypto/mlkem: Support ML-KEM (FIPS 203)

Yawning:
~~If it looks like people want ML-KEM-512 it would be trivial to add, but consensus seems to be settling around ML-KEM-768.~~

This supports all 3 parameter sets, since it indeed was trivial to add.


Yawning:
CI is a burning trashfire of random failures that can't be my fault (eg: the windows documentation test).

PR #6636, opened 2026-04-29
Update the boolean values constants.odin in vendor/glfw

celestine-of-the-night:
changed
- `TRUE :: true` to `TRUE :: 1`  
- `FALSE :: false` to `FALSE :: 0`. 
in vendor/glfw

GFLW defines TRUE and FALSE as 1 and 0. you have to convert to i32 or c.int with behavior as is
[see in the Macros section of the docs](https://www.glfw.org/docs/latest/group__init.html#Macros)

### DIFF
--- a/vendor/glfw/constants.odin
+++ b/vendor/glfw/constants.odin
@@ -10,8 +10,8 @@ VERSION_MINOR    :: 4
 VERSION_REVISION :: 0
 
 /* Booleans */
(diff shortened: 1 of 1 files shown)


gingerBill:
`vendor/glfw` is already a wrapper around the actual direct bindings of `vendor/glfw/bindings`, so trying to match it does not make sense in this context.

PR #6637, opened 2026-04-30
[vet] moved premature returns to a warning instead of an error

A1029384756:
the following code is currently a compilation error:
```odin
main :: proc() {
	return
	foo: [3]int = [?]int{1, 2, 3}
}
```
although this is useful at times, it adds substantial friction when debugging and prototyping. this keeps the check but places it behind `-vet` or `-vet-premature-return` to cut down on friction but still catch mistakes. "premature return" was used here to avoid overloading "early return"

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -310,10 +310,11 @@ enum VetFlags : u64 {
 	VetFlag_Tabs            = 1u<<9,
 	VetFlag_UnusedProcedures = 1u<<10,
 	VetFlag_ExplicitAllocators = 1u<<11,
+	VetFlag_PrematureReturn    = 1u<<12,
 
 	VetFlag_Unused = VetFlag_UnusedVariables|VetFlag_UnusedImports,
 
-	VetFlag_All = VetFlag_Unused|VetFlag_Shadowing|VetFlag_UsingStmt|VetFlag_Deprecated|VetFlag_Cast,
+	VetFlag_All = VetFlag_Unused|VetFlag_Shadowing|VetFlag_UsingStmt|VetFlag_Deprecated|VetFlag_Cast|VetFlag_PrematureReturn,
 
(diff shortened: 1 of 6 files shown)


A1029384756:
went ahead and changed the vet flat to `-unreachable-code` as it better matches what c does and allows flexibility in the future around adding `-unreachable-code:break,continue,fallthrough` options as well without breaking things for users (though i don't see these being nearly as important)

gingerBill:
I'm hesitant on this being a `-vet` thing. I understand people coming from languages that allow you to put `return` anywhere, but escape hatch was already relatively simple: `if true do return`.

So why make it EVEN easier when it is probably a _bug_ for most people?

A1029384756:
i have a couple things i should probably mention:

first, `if true do return/break/continue` *is* a simple escape hatch but i would argue is generally unintuitive behavior to new (and even experienced) users of the language, given that it's just "hiding" the statement from the checker under an unconditionally taken branch

second, at least anecdotally, i haven't particularly seen this bug pop up as a *bug* particularly commonly. statements such as these are generally added very intentionally, and if that is taken to be true, then the existence of the statement is not an issue and the code after it is more of a code smell (rather than impacting behavior), hence the classification under `-vet`

if classification as a `-vet` is not desired, another option that i have considered (and that may address your concerns) would be to make this a *warning*, not an error. i understand warnings are generally rare in odin but in cases like this where users are generally *intentionally* inserting these early exit statements, it allows their code to compile while *also* providing a persistent reminder that their code may not be correct

BigBoyBarney:
I'd like to add a minor correction. The escape hatch currently is even more basic, no `if` needed at all.

```odin
main :: proc() {
	{ return }
	fmt.println("Very sneaky!")
}
```

A1029384756:
closing this as i take it that this will not be merged

PR #6639, opened 2026-04-30
sys/linux, sys/unix: fully emulate dup2 on arm64 and riscv64

### LINKED ISSUE #6620: dup2 fails with EINVAL if newfd equals oldfd on ARM64
## Context

Odin:    dev-2026-04:a896fb2b4
OS:      Fedora Linux Asahi Remix 43 (KDE Plasma Desktop Edition), Linux 6.18.15-400.asahi.fc43.aarch64+16k
CPU:     ARM64
RAM:     7504 MiB
Backend: LLVM 20.1.8

## Expected Behavior

dup2 should do nothing and return newfd.

## Current Behavior

Errors with EINVAL.

### Steps to Reproduce

1. Be on an ARM64 machine.
2. Call dup2 with newfd equal to oldfd.
3. profit???

### Failure Logs

Here is a minimal example that demonstrates the issue.

```odin
package main
import "core:fmt"
import "core:sys/linux"
main :: proc() {
    fd, errno := linux.dup2(0, 0)
    fmt.printfln("fd = %v, errno = %v", fd, errno)
}
```

BunterSchatten:
#6632 does not fully fix #6620: we need to actually return the old file descriptor on EINVAL, but only if it is valid.
Additionally, if oldfd == newfd `dup3` does not check if it got a valid file descriptor, it immediatly returns EINVAL. I do that check manually with `fcntl(F_GETFD)`.

I also implemented the same fix for sys/unix.

relevant man page snippets:
```
dup2() makes newfd be the copy of oldfd, closing newfd first if necessary, but note the following:
* If oldfd is not a valid file descriptor, then the call fails, and newfd is not closed.
* If oldfd is a valid file descriptor, and newfd has the same value as oldfd,
  then dup2() does nothing, and returns newfd.
```
```
dup3() is the same as dup2(), except that:
* The caller can force the close-on-exec flag to be set for the new file descriptor
  by specifying O_CLOEXEC in flags. [...]
* If oldfd equals newfd, then dup3() fails with the error EINVAL.
```

### DIFF
--- a/core/sys/linux/sys.odin
+++ b/core/sys/linux/sys.odin
@@ -418,16 +418,21 @@ dup :: proc "contextless" (fd: Fd) -> (Fd, Errno) {
 */
 dup2 :: proc "contextless" (old: Fd, new: Fd) -> (Fd, Errno) {
 	when ODIN_ARCH == .arm64 || ODIN_ARCH == .riscv64 {
-		ret := syscall(SYS_dup3, old, new, 0)
-
 		// Differences between dup2 and dup3 are:
 		//   - dup3 takes a flags argument
 		//   - dup2 does not return EINVAL
-		fd, errno := errno_unwrap(ret, Fd)
-		if errno == .EINVAL {
-			errno = .NONE
+		//   - if old == new dup3 returns EINVAL, dup2 returns new if it is a valid file descriptor
+		if old == new {
+			_, err := fcntl(new, F_GETFD)
+			if err == nil {
+				return new, .NONE
+			} else {
+				default_fd: Fd
+				return default_fd, .EBADF
+			}
 		}
-		return fd, errno
+		ret := syscall(SYS_dup3, old, new, 0)
+		return errno_unwrap(ret, Fd)
 	} else {
 		ret := syscall(SYS_dup2, old, new)
 		return errno_unwrap(ret, Fd)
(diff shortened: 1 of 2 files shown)


jasonKercher:
Ah yeah, I forgot that unwrap is not truly returning multiple values.  Would this not be the more direct fix:

```diff
diff --git a/core/sys/linux/sys.odin b/core/sys/linux/sys.odin
index 87e41fa29..e908f29e6 100644
--- a/core/sys/linux/sys.odin
+++ b/core/sys/linux/sys.odin
@@ -426,6 +426,9 @@ dup2 :: proc "contextless" (old: Fd, new: Fd) -> (Fd, Errno) {
                fd, errno := errno_unwrap(ret, Fd)
                if errno == .EINVAL {
                        errno = .NONE
+                       if old == new {
+                               fd = new
+                       }
                }
                return fd, errno
        } else {
```

E: You know, it's kind of funny that our original test case was:
```odin
fd, errno := linux.dup2(0, 0)
```
... which works flawlessly.

DenverLacey:
Thanks for the quick fix. Sorry about the test case. I tried it out in C to see if dup2 didn't work there either where i tested mapping 0 -> 0, 1 -> 1 and 2 -> 2 and they all worked and i didn't check when I translated it to Odin. My bad.

BunterSchatten:
That's what I tried at first, it works in almost all cases, excpet when `old == new` and old is not a vaid fd: dup3 will return EINVAL so we return new, but dup2 would check the fd and return EBADF.

jasonKercher:
I see. It changes the error. Not the result. 

👍

PR #6648, opened 2026-05-03
Clean up unused things in parser

krnowak:
- init in unroll range statement is never set to anything
- count in map type and is_using in enum type are not used anywhere
- removed duplicated dead code
- removed unused parameter


gingerBill:
It is placeholder code, not "unused" code.

PR #6649, opened 2026-05-03
Better error reporting for unary decrement/increment operators

krnowak:
There was a missing routing for those operators in `parse_stmt` function.

For code like:

```odin
package p

main :: proc() {
	i : int
	++i
	assert(i == 1)
}
```

Old error message: `…/main.odin(5:2) Syntax Error: Expected a statement, got '++'`
New error message: `…/main.odin(5:2) Syntax Error: Unary '++' operator is not supported`

### DIFF
--- a/core/odin/parser/parser.odin
+++ b/core/odin/parser/parser.odin
@@ -1390,7 +1390,7 @@ parse_stmt :: proc(p: ^Parser) -> ^ast.Stmt {
 	     .Pointer,
 	     .Asm, // Inline assembly
 	     // Unary Expressions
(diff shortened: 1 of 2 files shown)


gingerBill:
Any changes that are made to parser.cpp must also be made to `core:odin/parser`.

krnowak:
Updated `core:odin/parser` too. While doing it, I noticed that handling of C-like pointer dereference (`*ptr`) is absent in `core:odin/parser`, but adding it is probably a bit more involved, since handling of this case in C++ code is split between the parser and a checker. For now I left it as is, so the PR covers only the decrement/increment operators.

PR #6650, opened 2026-05-03
Make typed literal values on left-hand side more consistent

krnowak:
There are two forms of literal values:

- naked (`{ … }`)
- typed (`type { … }` or `pkg.type { … }`)

Left-hand side expressions beginning with a naked literal value are not accepted by the parser and this is controlled by parse_operand with its lhs parameter.

Left-hand side expressions beginning with a typed literal value are only accepted by the parser if they come in the form of `pkg.type { … }`. This is because `pkg` is an lhs operand, `.type` is an lhs atom expr and `{` becomes an rhs atom expr. Braces are accepted in rhs atom expressions.

I'd argue that typed literal values of both forms should be accepted in such expressions even if their usefulness is rather limited. This makes language a bit more consistent and the parse_atom_expr a bit simpler.

And in contexts where typed literal values should not be accepted, setting `expr_level` to -1 should be used (like in parse_type_or_ident).

Demo code:

In "pkg" directory:

```odin
package pkg

Some_Type :: struct {
	ptr : ^int
}
```

In parent directory:

```odin
package main
import p "./pkg"

T :: p.Some_Type

main :: proc() {
	i : int
	p.Some_Type{ ptr = &i }.ptr^ = 42
	assert(i == 42)
	// Does not compile.
	// Error is "Syntax Error: Expected ';', got {".
	// This PR fixes it.
	T{ ptr = &i }.ptr^ = 13
	assert(i == 13)
}
```

### DIFF
--- a/core/odin/parser/parser.odin
+++ b/core/odin/parser/parser.odin
@@ -1658,7 +1658,7 @@ parse_type_or_ident :: proc(p: ^Parser) -> ^ast.Expr {
 	p.expr_level = -1
 
 	lhs := true
-	return parse_atom_expr(p, parse_operand(p, lhs), lhs)
+	return parse_atom_expr(p, parse_operand(p, lhs))
 }
 parse_type :: proc(p: ^Parser) -> ^ast.Expr {
 	type := parse_type_or_ident(p)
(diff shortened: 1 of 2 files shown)


gingerBill:
Any changes that are made to `parser.cpp` must also be made to `core:odin/parser`. 

There is an inconsistency, but it's also kind of one of things which is probably not a good idea in the first place, and something I'll have to consider how to make more consistent.

krnowak:
> Any changes that are made to `parser.cpp` must also be made to `core:odin/parser`.

Updated.

> There is an inconsistency, but it's also kind of one of things which is probably not a good idea in the first place, and something I'll have to consider how to make more consistent.

Probably a way to consistently ban typed literal values on left hand side would be to avoid setting `lhs` (or `is_lhs` in `core:odin/parser`) to `false` after first iteration of the loop in `parse_atom_expr`.

PR #6654, opened 2026-05-04
`-bedrock`

gingerBill:
Name subject to change.

This is a mode developed for things like kernel development, creating a subset of Odin with restricted features.

Disabled features:

* `map`
* 128-bit integers
* RTTI
* Non-Constant Global Variables
* `@(init)` `@(fini)`
* Anything Objective-C related
* The default paths to the library collections `core` and `vendor`

New flags:

* `-disable-non-constant-globals`
* `-disable-init-fini`

Flags enabled by default when `-bedrock` is enabled:

* `-disable-non-constant-globals`
* `-disable-init-fini`
* `-no-rtti`

-------

Minor new feature: conditional inclusion in procedure groups:

```odin
group :: proc{
	a,
	b where COND,
	c,
	d,
}
```

-------

~~Considered features: default to `"contextless"` calling convention meaning `context` has to be explicitly opt-in. Not sure if it should be or not, yet.~~


base/runtime/entry_unix.odin
```diff
@@ -42,9 +44,9 @@ when ODIN_BUILD_MODE == .Dynamic {
 		_start_odin :: proc "c" (argc: i32, argv: [^]cstring) -> ! {
 			args__ = argv[:argc]
 			context = default_context()
```
Barinzaya, line 46: This will still pull in the default allocator/RNG implementations even if it'll never be used (e.g. a `contextless` `main`), won't it?

src/check_decl.cpp
```diff
@@ -1523,10 +1527,26 @@ gb_internal void check_proc_decl(CheckerContext *ctx, Entity *e, DeclInfo *d) {
...
-			if (pt->calling_convention != default_calling_convention()) {
-				error(e->token, "Procedure 'main' cannot have a custom calling convention");
+			if (build_context.bedrock) {
+				switch (pt->calling_convention) {
+				case ProcCC_Odin:
+				case ProcCC_Contextless:
+					// Okay
+					break;
+				default:
+					error(e->token, "Procedure 'main' cannot have a custom calling convention beyond \"odin\" and \"contextless\" with '-bedrock'");
```
Barinzaya, line 1544: Thoughts on allowing a `contextless` `main` outside of bedrock mode too?

base/runtime/entry_unix.odin
gingerBill: Yes, I haven't figured out how to make this "optional" yet.

src/check_decl.cpp
gingerBill: I cannot think of an __actual__ need to allow this.

base/runtime/entry_unix.odin
gingerBill: It doesn't matter since it's kind of necessary on the `context`.

Issue #6656, opened 2026-05-05
fmt.print with flush=false no longer works


GigaGrunch:
## Context

Output of `odin report`:
```
Odin:    dev-2026-05-nightly:ea5175d
OS:      Windows 11 Enterprise (version: 25H2), build 26200.8246
CPU:     Intel(R) Core(TM) Ultra 9 275HX
RAM:     64980 MiB
Backend: LLVM 20.1.0
```

## Expected Behavior

`fmt.print` can also be used with `flush=false` (if followed by a flush).

## Current Output

`fmt.print` does not do anything (observable) with `flush=false`.

### Steps to Reproduce

* Run this code snippet:
```
fmt.print("no flush\n", flush=false)
fmt.print("flush\n", flush=true)
```
* Only `flush` is printed, but `no flush` is not.

## What I already figured out

The bug was introduced by this PR: https://github.com/odin-lang/Odin/pull/6567. Or maybe it was just made visible by that PR because before it was just always flushing which is not desired either. The author of the PR also acknowledges the problem specifically:
> But the buffer is on the stack here so I don't know what the caller expects to happen when flush=false in this situation.

JackJ30:
So it seems like buffered IO isn't implemented for files, and the buffer on the stack never goes anywhere. With the current implementation, it makes sense to remove the flush parameter from the file print function (and others that rely on it).

We should decide if we actually want buffered IO for files by default, and implement it soon.
Libc has buffered IO in its file abstraction. Is Odin's is meant to be lower level than that, since os.File "represents an OS-specific file-handle"
How do we feel about a buffered writer in the File struct? Or would it go in a stream? Or maybe another abstraction on top?
I found a commented `open_buffered` proc in file.odin, so it seems like there were plans at some point.

LunarEclipse1185:
I stumbled upon this problem when doing some TUI toys. This is quite a problem.

PR #6657, opened 2026-05-06
Update fmt_os printing to always flush

Faker-09:
Change fmt_os printing procs to ignore the flush parameter and always flush.
Add a comment to the existing comments stating flush is ignored.
This addresses #6656, the case where `flush=false` is passed to these procs.


gingerBill:
This is a bodge. Not a solution in the slightest.

Faker-09:
Ok, it was meant to be an interim bodge to maintain behaviour. 
Any thoughts on how you want to handle the case where flush=false is used with these printing procs?

Issue #6655, opened 2026-05-04
os.remove_all() returns ENOTEMPTY on Linux when 3 or more files are present in a directory in an NTFS file-system.


Ribiveer:
## Context

I'm building a project I've previously been making on a Windows-machine, and I'm building it for Linux. Right now, I'm working on the build script, which needs to call `os.remove_all` on some directories filled with files.

However, it kept only deleting a few files within the directory, and then returning ENOTEMPTY, which seemed out-of-character for `os.remove_all`. I've whittled this behaviour down to a minimum example, and found out it was only an issue on NTFS file-systems.

Here is my `odin report` output:

```
Odin:    dev-2026-04:a896fb2b4
OS:      EndeavourOS, Linux 6.19.11-arch1-1
CPU:     AMD Ryzen 5 5600X 6-Core Processor             
RAM:     32030 MiB
Backend: LLVM 21.1.8
```

## Expected Behavior

The directory gets deleted, along with every file it contains.

## Current Behavior

Only a few files inside of the directory get deleted, and the procedure returns ENOTEMPTY.

### Steps to Reproduce

1. On an NTFS drive, create a directory structure like so:

```
repro
L dir-og (directory)
    L  1.txt (empty file)
    L  2.txt (empty file)
    L  3.txt (empty file)
    L  4.txt (empty file)
    L  5.txt (empty file)
    L  6.txt (empty file)
    L  7.txt (empty file)
    L  8.txt (empty file)
    L  9.txt (empty file)
    L 10.txt (empty file)
    L 11.txt (empty file)
    L 12.txt (empty file)
L main.odin
```

2. In `main.odin`, paste the following code:

```odin
package repro

import "core:fmt"
import "core:os"

main :: proc() {
	assert_err :: proc(err : os.Error, loc := #caller_location) {
		fmt.assertf(err == nil, "Error: {}", err, loc = loc)
	}

	if !os.exists("dir") {
		err := os.copy_directory_all("dir", "dir-og")
		assert_err(err)
	} 

	cwd, cwd_err := os.get_working_directory(context.temp_allocator)
	assert_err(cwd_err)

	err := os.remove_all("dir")
	assert_err(err) // Fails with ENOTEMPTY, but like... Yeah I sure hope it does?
	fmt.println("All good!")
}
```

3. Call `odin run .`

On the first run, text files 1, 2, 3, 6, 7, and 10 get deleted.
On the second run, text files 8, 9, 11, and 12 get deleted.
[...]

PR #6662, opened 2026-05-07
rework os.remove_all (Linux)

jasonKercher:
This fixes issue #6655 with `os.remove_all` on Linux and an NTFS partition.  The original implementation would remove files while iterating the directory entries.  On NTFS, this breaks the directory entries.  This new implementation scans the directory before removing anything.

E: The force pushes are squashing commits that remove an allocation and the other makes -vet happy.
E2: The 3rd force push is a reset to the original because we actually wanted that allocation =]


jasonKercher:
This could be done entirely with the temp_allocator... back to the drawing board.

PR #6683, opened 2026-05-12
core:os (Linux): Convert (f)stat calls to statx

jasonKercher:
This provides access to `os.Stat.creation_time` for Linux.

NOTE: Sneaky edit to string_interner.cpp: Added a cast for 32-bit builds of the compiler.

I converted all uses of `linux.Stat` to `Statx` which wasn't strictly necessary.  Statx is supposed to be a bit faster since you request what you want to be filled out instead of just getting everything.  Also, regular `(f)stat` was giving me file size issues in a 386 container that I used for testing (both Odin and C).

The downside is it cuts off people with kernels older than 4.11 (released April 2017).  This takes out anything Ubuntu 16.04 and below.  I could add a fallback wrapper if we want to support older kernels.

### DIFF
--- a/core/os/dir_linux.odin
+++ b/core/os/dir_linux.odin
@@ -98,7 +98,7 @@ _read_directory_iterator_init :: proc(it: ^Read_Directory_Iterator, f: ^File) {
 	}
 
 	stat: linux.Statx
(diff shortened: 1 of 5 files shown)


gingerBill:
I am thinking we should not probably care for Linux Kernels older than that any way, and deal with it when it is actually a problem.

PR #6608, opened 2026-04-23
[vendor] added initial raylib 6.0 bindings

RobinsAviary:
Hello! I went through the itemized "Changelog" file and implemented all of the changes in the new version. I did not include binaries as I assume you want to add them yourself, but I did test it on Windows and had no issues compiling a simple test program. I also updated the LICENSE and README files from the repo. rlgl was also still labeled as "5.0", so I updated it to 6.0 fully, and checked for anything that was missed.

PR #6718, opened 2026-05-23
raylib: Update to 6.0

lxmcf:
* Updated raylib bindings to match raylib 6.0
* Updated rlgl bindings to match rlgl 6.0
* Updated binaries to 6.0

Unfortunately only have access to test on Linux as no MacOS nor windows devices handy so could not test these.


lxmcf:
Didn't realise this was a duplicate of #6608 

Closing off

PR #6729, opened 2026-05-24
Update handle map procedures to accept pointers

cephei8:
Static_Handle_Map has several procedures that accept handle map by copy. 
Other read-only procedures like `get` already accept it by pointer.

Update Static_Handle_Map and Dynamic_Handle_Map procedures to accept the handle map argument by pointer.
Dynamic_Handle_Map was updated for consistency.


gingerBill:
Odin's semantics does not work that way. Anything over 16 bytes is implicitly passed by reference, and thus it has no performance issues.

PR #6737, opened 2026-05-27
Fix unbalanced parentheses

obiwan87:
Fix syntax typo in intrinsics.odin


obiwan87:
weird that the check fails °_°

PR #6738, opened 2026-05-27
Added cubic root (cbrt) to core:math

heavyrain266:
The `sign(x)` and `abs(x)` are kind off colour science specific as it guards against silent NaN from off-gamut and malformed inputs.


gingerBill:
This is not necessarily a good implementation if you want to optimize specifically for the cube root.

I'll just add it manually.

heavyrain266:
Thanks.

PR #6739, opened 2026-05-27
Explicit quick sort

jakubtomsu:
The original quick sort implementation was replaced by smooth sort because of generated code size. Instantiating the polymorphic procedure implementation hundreds of times across a large codebase is a problem.

## Info

This PR brings back the old implementation with a number of performance improvements, and adds an ***explicit public API*** for people who actually need the performance. So no code bloat.
This actually makes a big difference in practice, I personally ran into issues with slow sorting in my rendering engine in multiple places (depth sorting and batch reordering).

The default is still smoothsort. Only exception is sorting basic integer/float types with `slice.sort` (not sure about this, maybe this behavior is too "magical"? My reason for this is that I often use something like `slice.sort(transmute([]u64)my_slice_of_u64_sized_structs)`)

## API

There are now `quick_sort`, `quick_sort_by` and `quick_sort_by_cmp` procedures with exact same signatures as the default `sort_*` equivalents.

## Performance
This results in a quicksort that's generally ***5x*** faster than the smooth sort implementation, and usually ***1.2x - 2.0x*** faster than the old quick sort implementation depending on the data.

Tweaks to the old quicksort:
- Much larger insertion sort threshold, this alone is a big speedup
- Slightly better insertion sort implementation
- Adapted shell sort
- Force inline in a few places where it matters

All of these parameters were independently benchmarked on various datasets to find the optimum.

Here is my [benchmarking code](https://gist.github.com/jakubtomsu/0d7c80769fdead33f2b2bbec356ca75c).

Just for reference, here is the commit that removed the quicksort: f40fc2792f01b1fa265f62fb2aa4ef39e7529903

### DIFF
--- a/core/slice/slice.odin
+++ b/core/slice/slice.odin
@@ -95,7 +95,7 @@ reinterpret :: proc "contextless" ($T: typeid/[]$U, s: []$V) -> []U {
 }
 
 
-swap :: proc(array: $T/[]$E, a, b: int) {
+swap :: #force_inline proc(array: $T/[]$E, a, b: int) {
 	when size_of(E) > 8 {
 		ptr_swap_non_overlapping(&array[a], &array[b], size_of(E))
 	} else {
--- a/core/slice/sort.odin
+++ b/core/slice/sort.odin
@@ -37,7 +37,12 @@ cmp_proc :: proc($E: typeid) -> (proc(E, E) -> Ordering) where ORD(E) {
 // sort sorts a slice
 // This sort is not guaranteed to be stable
 sort :: proc(data: $T/[]$E) where ORD(E) {
-	when size_of(E) != 0 {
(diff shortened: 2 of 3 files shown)


TheRadischen:
if you really care about speed and you have primitiv types, (int, float) try using radix sort

and quicksort can be improved by about 2x with branchless optimazation
https://arxiv.org/abs/1604.06697

works quite well, but not quite done, doesnt handle generic data and  no heapsort for worst case, but thats easy
only uses some extra stack space
https://github.com/TheRadischen/odin-utils/blob/main/block_quicksort.odin

jakubtomsu:
I know about radix sort and have considered implementing it. But this PR is just to bring back the old quicksort for cases where that’s a good choice. It’s nice and flexible, even for non basic types.

PR #6746, opened 2026-05-28
vendor/OpenCL: add constants and bindings to the C library

devNegative-asm:
opencl + egl + gl extensions. Should be up to at least 3.0 but I'm not sure of the exact version.

I could make odin-style wrappers around these functions if necessary. Might do that later.

### DIFF
--- a/vendor/OpenCL/opencl.odin
+++ b/vendor/OpenCL/opencl.odin
@@ -1,0 +1,987 @@
+// bindings for [[ OpenCL ; https://registry.khronos.org/OpenCL/ ]].
+#+build windows, linux, darwin
+package vendor_opencl
+
+import "core:c"
+
+cl_platform_id :: distinct rawptr
+cl_device_id :: distinct rawptr
+cl_context :: distinct rawptr
+cl_command_queue :: distinct rawptr
+cl_mem :: distinct rawptr
+cl_program :: distinct rawptr
+cl_kernel :: distinct rawptr
+cl_event :: distinct rawptr
+cl_sampler :: distinct rawptr
+svm_pointer :: distinct uintptr
+
+cl_char :: i8
+cl_uchar :: u8
+cl_short :: i16
+cl_ushort :: u16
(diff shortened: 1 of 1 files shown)


devNegative-asm:
I did some limited testing on linux. Most functions aren't tested, but I was able to validate at least these:
clGetPlatformIDs
clGetPlatformInfo
clGetDeviceIDs
clGetDeviceInfo
clCreateContextFromType
clCreateProgramWithSource
clBuildProgram
clCreateBuffer
clCreateKernel
clCloneKernel
clSetKernelArg
clCreateCommandQueueWithProperties
clWaitForEvents
clEnqueueNDRangeKernel
clEnqueueReadBuffer
clEnqueueWriteBuffer
clEnqueueTask
clEnqueueFillBuffer

devNegative-asm:
I pushed an update with some intel extensions. new functions I tested:
clEnqueueMemFillINTEL
clEnqueueMemcpyINTEL
clSetKernelArgMemPointerINTEL
clSharedMemAllocINTEL
clSetKernelArgMemPointerINTEL

jwaxy:
You can use "link_prefix" to remove the cl prefix:
```odin

@(default_calling_convention="c", link_prefix="cl")
foreign opencl { ... }
```
so you can call procs like this: `cl.GetPlatformIDs()`,
instead of `cl.clGetPlatformIDs()`

gingerBill:
I won't accept this for a few reasons:

* It doesn't strip the `cl` prefixes, which is unlike any of the other `vendor` packages.
* `cl_platform_id` etc, have been made `distinct rawptr` rather than in keeping with the original API which would be `platform_id :: struct {}`
* No need to declare `cl_char` when `c.char` is equivalent in all cases.
* `cl_device_type_bits` didn't need to define anything else, and probably should have just been a normal `u32` instead BECAUSE it would have been easier to allow for the custom bits than the `_p#` approach

And many more reasons.

Issue #6721, opened 2026-05-23
Syntax error on math expressions in struct initialization


thetarnav:
code
```odin
Type :: struct {a: [2]f32}
t := Type{{1, 2} - 3}
```

checker error
```
Syntax Error: Expected '}', got '-' 
        t := Type{{1, 2} - 3} 
                         ^ 
 
Syntax Error: Expected ';', got integer 
        t := Type{{1, 2} - 3} 
                          ^ 
```

report
```
Odin:    dev-2026-05:ea5175d86
OS:      Arch Linux, Linux 7.0.3-arch1-2
CPU:     Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
RAM:     15684 MiB
Backend: LLVM 22.1.3
```

thetarnav:
Looks like it's fine now

PR #6760, opened 2026-06-02
Fix support for array expression in struct literals

jqcorreia:
In `src/parser.cpp` the `parse_value()` function has a conditional `parse_literal_value()` call that removes the possibility of having arbitrary expression with array literals in struct literals. 

Removing this condition fixes the behaviour reported at #6721 
This logic surely has a point to exist but I fail to understand it, so I thought I would open this PR to discuss. The `examples/demo.odin` program runs without apparent problem and a minimal example works as intended. 
This was the code used to test this behaviour:

```odin
package main

import "core:fmt"

S :: struct {
	vec: [2]f32,
	x: i32,
	y: i32
}

main :: proc() {
	v: S = {{1, 2} * 10, 100 + 20, 200}
	w: S = {vec = {1, 2} * 5, x = 100, y = 314 }

	fmt.println(v.vec, v.x, v.y) // {10,20}, 120, 200
	fmt.println(w.vec, w.x, w.y) // {5,10}, 100, 314
}
``` 
This yields the expected output.

I'm happy to do any changes or just scrap it if it doesn't make any sense and/or inserts a regression.
Cheers.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -2039,9 +2039,6 @@ gb_internal Ast *parse_literal_value(AstFile *f, Ast *type) {
 }
 
 gb_internal Ast *parse_value(AstFile *f) {
-	if (f->curr_token.kind == Token_OpenBrace) {
-		return parse_literal_value(f, nullptr);
-	}
 	Ast *value;
 	bool prev_allow_range = f->allow_range;
 	f->allow_range = true;


jqcorreia:
@gingerBill Not trying to be combative here, just to learn. Did I do something egregiously wrong here? Should I do something different in the future, or is just a plain bad idea do to this?

gingerBill:
@jqcorreia you did nothing wrong. I just had to make sure it was correct, but I also had to apply it to `core:odin/parser/parser.odin` too to be consistent, which is quicker to do manually than waiting for it.

jqcorreia:
@gingerBill  Cool, glad to hear that. Contributing to anything these days can be a something like walking a minefield :-) To resolve my curiosity if you can spare the time: I got the feeling that this logic is vestigial and made sense somewhere along the compiler developer but not anymore. Am I correct?

gingerBill:
The logic was vestigial.

jqcorreia:
Glad I could help, even if partially. Thanks for the response

PR #6765, opened 2026-06-03
fix: Get meminfo from `/proc/meminfo` over `sysinfo()`

### LINKED ISSUE #6764: `core:sys/info` `ram_stats()` returning `MemFree` through `sysinfo()` over `MemAvailable` on Linux
## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except under certain conditions.

        Odin:    dev-2026-05:ea5175d86
        OS:      CachyOS, Linux 7.0.10-2-cachyos
        RAM:     32008 MiB
        Backend: LLVM 21.1.8

## Expected Behavior

All other programs on my machine report 5.87 (used) / 31.26 (total) GiB at the time of writing. However, in my Odin program, the memory usage reported is entirely different.

Given RAM in order: Used/Free/Total.
`RAM:       29.6 GiB/1.6 GiB/31.3 GiB`

The expected output given from `ram_stats()` should be 25.4 GiB free and 31.3 GiB total, allowing used to be calculated at 5.9 GiB, which is what KDE System Monitor and Fastfetch both report.

## Current Behavior

What is the current behavior?

The current behavior being done is that Odin is relying on the syscall `sysinfo()` to get the RAM information. This is apparently known for being rather inaccurate, so most programs instead rely on reading `/proc/meminfo` to get the most detailed memory information and using what they need from there. Odin should probably reflect this since everything else on Linux does as well. I'm unsure, however, if this is intentional or not for Linux. I will probably work on implementing a fix internally for my program to read from `/proc/meminfo` for now.

MightyChubz:
This is a change to `core:sys/info` to fix inaccurate memory reads in Linux. I haven't really touched direct Odin core library code before, so this is my first attempt. If there are _any_ mistakes or improvements to be made, let me know. I'm more than willing to correct any dumb code I may have written.

Also, the reason the size of the buffer here is 4096 is because the actual size of `/proc/meminfo` on my machine is around 1500 bytes, and it felt a little too close to 2048 for me to be comfortable with it. I can adjust this if needed; I just assumed 4096 would work fine here.

Closes #6764

### DIFF
--- a/core/sys/info/platform_linux.odin
+++ b/core/sys/info/platform_linux.odin
@@ -81,6 +81,8 @@ _os_version :: proc (allocator: runtime.Allocator, loc := #caller_location) -> (
 
 @(private)
 _ram_stats :: proc() -> (total_ram, free_ram, total_swap, free_swap: i64, ok: bool) {
+	// The approach is to read /proc/meminfo for the memory information _over_ sysinfo(),
+	// since sysinfo() just returns MemFree over the value we actually want, MemAvailable.
 	fd, errno := linux.open("/proc/meminfo", {})
 	if errno != .NONE {
 		// This should never happen since something would be wrong with the system
--- a/core/sys/info/platform_linux.odin
+++ b/core/sys/info/platform_linux.odin
@@ -2,6 +2,7 @@ package sysinfo
 
 import "base:intrinsics"
 import "base:runtime"
+import "core:strconv"
 import "core:strings"
 import "core:sys/linux"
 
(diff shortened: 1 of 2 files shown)


Yawning:
FWIW, I approve the approach (and commented on the relevant issue wrt. the discrepancy).  Adding a comment in the file as to why this approach is taken would be a good idea.

The fallback path will over-estimate the amount of available memory relative to `MemAvailable` as the latter caps how much the kernel will free, but that code should never be called in the first place (If people are running on Odin on Linux < 3.14, they have a mountain of other problems).

MightyChubz:
> FWIW, I approve the approach (and commented on the relevant issue wrt. the discrepancy). Adding a comment in the file as to why this approach is taken would be a good idea.
> 
> The fallback path will over-estimate the amount of available memory relative to `MemAvailable` as the latter caps how much the kernel will free, but that code should never be called in the first place (If people are running on Odin on Linux < 3.14, they have a mountain of other problems).

I can do a comment to explain the approach for sure inside the proc. The fallback isn't really intended to ever be reached; it's more of a worst-case scenario in the event `MemAvailable` doesn't exist or is broken to some capacity (or what you said with people running Odin on Linux < 3.14). Accuracy at that point kind of takes a minor hit.

Optionally, I could opt for just making the fallback `MemFree` and working off that? Could be a little more accurate overall since the reasoning there would be: if `MemAvailable` isn't there, then just return `MemFree` without the extra stuff.

Yawning:
> I can do a comment to explain the approach for sure inside the proc. The fallback isn't really intended to ever be reached; it's more of a worst-case scenario in the event `MemAvailable` doesn't exist or is broken to some capacity (or what you said with people running Odin on Linux < 3.14). Accuracy at that point kind of takes a minor hit.
> 
> Optionally, I could opt for just making the fallback `MemFree` and working off that? Could be a little more accurate overall since the reasoning there would be: if `MemAvailable` isn't there, then just return `MemFree` without the extra stuff.

I personally would just opt for returning `MemFree` (with a comment noting that something is badly wrong/ancient if this happens) because it's simpler.

MightyChubz:
> > I can do a comment to explain the approach for sure inside the proc. The fallback isn't really intended to ever be reached; it's more of a worst-case scenario in the event `MemAvailable` doesn't exist or is broken to some capacity (or what you said with people running Odin on Linux < 3.14). Accuracy at that point kind of takes a minor hit.
> > Optionally, I could opt for just making the fallback `MemFree` and working off that? Could be a little more accurate overall since the reasoning there would be: if `MemAvailable` isn't there, then just return `MemFree` without the extra stuff.
> 
> I personally would just opt for returning `MemFree` (with a comment noting that something is badly wrong/ancient if this happens) because it's simpler.

I will do this then since it will be better and a little more predictable in behavior imo.

core/sys/info/platform_linux.odin
```diff
@@ -79,17 +80,101 @@ _os_version :: proc (allocator: runtime.Allocator, loc := #caller_location) -> (
...
-	sys_info: linux.Sys_Info
-	errno := linux.sysinfo(&sys_info)
-	assert_contextless(errno == .NONE, "Good luck to whoever's debugging this, something's seriously cucked up!")
-
-	total_ram  = i64(sys_info.totalram)  * i64(sys_info.mem_unit)
-	free_ram   = i64(sys_info.freeram)   * i64(sys_info.mem_unit)
-	total_swap = i64(sys_info.totalswap) * i64(sys_info.mem_unit)
-	free_swap  = i64(sys_info.freeswap)  * i64(sys_info.mem_unit)
-	ok         = true
+_ram_stats :: proc() -> (total_ram, free_ram, total_swap, free_swap: i64, ok: bool) {
```
gingerBill: Why was `"contextless"` removed here?

core/sys/info/platform_linux.odin
```diff
@@ -79,17 +80,101 @@ _os_version :: proc (allocator: runtime.Allocator, loc := #caller_location) -> (
...
+		ok = true
+
+		return
+	}
+	meminfo := string(meminfo_buf[:n])
+
+	// Fallback in the event MemAvailable is not found or is invalid in its value
+	mem_free: i64
+
+	for line in strings.split_lines_iterator(&meminfo) {
```
gingerBill, line 136: If you need a temporary context here, just use these procedures, use: 
```odin
context = runtime.default_context()
```

gingerBill (changes requested):

core/sys/info/platform_linux.odin
```diff
@@ -79,17 +80,101 @@ _os_version :: proc (allocator: runtime.Allocator, loc := #caller_location) -> (
...
+	if free_ram == 0 || free_ram > total_ram {
+		// We opt to return MemFree here if MemAvailable is not found or is broken to some degree.
+		// This will act as a predictable fallback, but shouldn't ever really occur unless the user
+		// is on Linux < 3.14
+		free_ram = mem_free
+	}
+
+	mem_unit :: 1024
+	total_ram *= mem_unit
+	free_ram *= mem_unit
```
Kelimion: I'd move up the `mem_unit` constant above the loop, and move `* mem_unit` into the switch.

It won't change the `if free_ram == 0 || free_ram > total_ram` calculation. They'll just be done in bytes instead of kilobytes, but that doesn't affect the result.

core/sys/info/platform_linux.odin
MightyChubz: I kind of assumed this was going to be an issue, but I was unsure how to approach it. Several `strings` procedures require the context, though they don't do any allocation (at least, explicitly showing they do). So my initial approach was just removing `"contextless"`, but sort of leaving the door open for that to be changed if needed. I considered using `context = runtime.default_context()`, but didn't know if that would be considered a kind of misdirection since the procedure shows itself as `"contextless"`, but still uses the context. I'll make the changes now and get it all sorted.

core/sys/info/platform_linux.odin
MightyChubz: I'll make these changes and get the procedure a little more cleaned up.

core/sys/info/platform_linux.odin
MightyChubz: I added this to the very start of the function so it's easier to spot (not really needed for most, but I like having it there).

gingerBill (approved):

PR #6768, opened 2026-06-03
changed slice._stable_sort_general from insertion sort to O(nlog²n) inplace merge sort

TheRadischen:
uses O(logn) stack memory because of recursion

tested with stable_sort, stable_sort_by, stable_sort_by_cmp

for small inputs <10k about as fast as slice.sort, becomes slower because of worse growth

used in c++ std lib:
https://github.com/llvm-mirror/libcxx/blob/a12cb9d211019d99b5875b6d8034617cbc24c2cc/include/algorithm#L4452
https://thomas.baudel.name/Visualisation/VisuTri/inplacestablesort.html


gingerBill:
Could you provide some benchmarks to show this is actually better that the current approach? Both in terms of performance, and show the (stack) memory cost too, please.

TheRadischen:
made a few tests. they are a bit of a mess, 

the readme has the reasons for changing the algo

https://github.com/TheRadischen/odin-utils/tree/main/sort

TheRadischen:
i was wondering if i should change the function signiture from  (arr: $T/[]$E) to (arr: $T) that way ppl could sort soa as well, witch is more interesting for stable sorting anyway.
but looking at the rest of core:slice all of it uses the former version

TheRadischen:
hmm... forgot this was on master.
it should be on a feature branch last i checked, but if you need it back, let me know

PR #6777, opened 2026-06-04
[checker] disallow `*` and `/` for bit sets

A1029384756:
before, the following code would compile:
```odin
Foo :: bit_set[enum {A,B,C}]

a := Foo{.A}
b := Foo{.B}
c := a * b
d := b / a
```


gingerBill:
That's an amazing find!

PR #6781, opened 2026-06-05
WASM: Conditionally compile os-dependent procs in vendor/OpenGL

evaporei:
Example below fails `odin check gl_wasm_example.odin -target:js_wasm32`:

```odin
package main

@(require) import gl "vendor:OpenGL"

main :: proc() {  }
```

Error:
```
gl_wasm_example.odin(4:1) Error: Compile time panic: core:os is unsupported on js/wasm 
        #panic("core:os is unsupported on js/wasm") 
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^ 
 ```

### DIFF
--- a/vendor/OpenGL/helpers.odin
+++ b/vendor/OpenGL/helpers.odin
@@ -3,10 +3,8 @@ package vendor_gl
 
 // Helper for loading shaders into a program
 
-import    "core:os"
 import    "core:fmt"
 import    "core:strings"
-@(require) import "core:time"
 import    "base:runtime"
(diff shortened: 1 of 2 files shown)


gingerBill:
Because WASM doesn't support `vendor:OpenGL`, it supports `vendor:wasm/WebGL`

evaporei:
@gingerBill OpenGl does work correctly on browser if you use emscripten (it ports to WebGL calls for you). We use it in our game engine.

Issue #5639, opened 2025-08-30
False positives from `-vet-unused-procedures` with nested procedures


FrancisTheCat:
```
#+vet unused-procedures
package test

main :: proc() {
	foo :: proc() {
		
	}

	bar :: proc() {
		foo()
	}

	bar()
}
```

```
main.odin(5:2) Error: 'foo' declared but not used
        foo :: proc() {
```

This only happens when the two functions are declared in the same scope or the when called function is in a more outer scope.

```
odin version dev-2025-08:2f0b018db
```

bennyyip:
I can reproduce it in `odin version dev-2026-05-nightly:ea5175d`.

PR #6785, opened 2026-06-06
fix: checker flagging polymorphic procedure as unused

MrEliasen:
Hey gingerBill/maintainer(s)!

---
## Disclaimer

Immediate up front LLM disclaimer: AST is beyond my knowledge and I would not call myself a very competent cpp dev, so I used an LLM to write the code.

I made sure all the test came back OK, I can see what it is doing at a very high level but honestly I am far from experienced enough to judge this proerply.

Please feel free to close, change or otherwise make any changes you feel is necessary regarding this PR and its content.

---

## The issue

Right to the issue: `odin check` flags polymorphic procedures as unused, when one of the procedures within was indeed in use. This was causing me a lot of grief in my own projects and I couldn't find any existing issue or PRs mentioning this issue - unless I also suck at github search, which wouldn't surprise me.

If it is helpful:
OS: MacOS 15.7.5 (24G624)
Odin Version: dev-2026-05

To reproduce the issue, you can try this:

```odin
// main.odin

package main

import "thing" // 'thing' declared but not used flagged

do_thing :: proc($T: typeid, x: int) -> int {
    return thing.plus_one(x)
}
```

```odin
// thing/thing.odin

package thing 

plus_one :: proc(x: int) -> int {
    return x + 1
}
```

Screenshot 2026-06-06 at 15 39 31

The fix is to walk the polymorphic procedures and mark any imports they use  as 'used'. This seemed like a good idea to me when i kicked this off, but please do scrutinise this to death.

I hope this PR is at least in some way helpful or can in some way help resolve my original issue in some way (even if just to surface it).

Cheers!

EDIT: Likely related: https://github.com/odin-lang/Odin/issues/5639

### DIFF
--- a/src/check_decl.cpp
+++ b/src/check_decl.cpp
@@ -1256,6 +1256,319 @@ gb_internal void check_foreign_procedure(CheckerContext *ctx, Entity *e, DeclInf
 	mutex_unlock(&ctx->info->foreign_mutex);
 }
 
+gb_internal void check_mark_referenced_imports_used(CheckerContext *ctx, Ast *node) {
+	if (node == nullptr) {
+		return;
+	}
+
+	#define WALK(field_)       check_mark_referenced_imports_used(ctx, node->field_)
+	#define WALK_ARRAY(field_) for (Ast *child_ : node->field_) check_mark_referenced_imports_used(ctx, child_)
+
+	switch (node->kind) {
+	case Ast_Ident: {
+		AstFile *file = node->file();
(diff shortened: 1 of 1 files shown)


gingerBill:
This is more of a bodge since it is relying that the semantics don't change because of the syntax.

MrEliasen:
Makes sense, appreciate the look! Hopefully this is something which can be solve properly down the road - wish I could help, but I am too dumb for this.

Cheers!

PR #6787, opened 2026-06-06
Improved fmt_os documentation

BigBoyBarney:


BigBoyBarney:
Woops, should be fine now :)

Issue #6788, opened 2026-06-06
Linux target default option changes.


Yawning:
## Context

This came up in the context of an exotic target (termux) and is reminiscent of #4448, but there is no good reason not to do the following by default across all Linux targets.

- Compiler (always): `-reloc-mode:pic`
- Link: `-extra-linker-flags:"-pie -Wl,-z,now,-Wl,-z,relro"` (full relro) or just `-extra-linker-flags:"-pie -Wl,-z,relro"` (partial relro)

This behavior is supported on everything that is not massively out of date, and is the expected/default behavior when building Linux executables/libraries (Eg: https://wiki.archlinux.org/title/Arch_package_guidelines/Security).

PR #6790, opened 2026-06-07
[linux] change default settings to enable pie and full relro

A1029384756:
#6788 changes the default linux options to enable pie and full relro

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -1906,7 +1906,7 @@ gb_internal void init_build_context(TargetMetrics *cross_target, Subtarget subta
 		switch (metrics->arch) {
 		case TargetArch_arm64:
 		case TargetArch_amd64:
-			bc->no_plt = true;
+			bc->no_plt = LLVM_VERSION_MAJOR >= 19;
 			break;
 		}
 	} else if (metrics->os == TargetOs_openbsd) {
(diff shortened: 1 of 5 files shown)


Yawning:
At a glance LGTM.  Should confirm that libraries also get this treatment (I don't see why it shouldn't).

For the record, I forgot to note when I filed the issue, that Full RELRO will cause a performance regression in startup time and on dlopen because the dynamic linker is forced to load all symbols immediately (to generate the GOT and then make it read-only), versus lazily filling out the GOT on-use.  While Partial RELRO does not have this downside, "overwrite a GOT entry" is "easy money" in terms of writing exploits.  The Full RELRO case has NO other runtime overhead.

Additionally, with Full RELRO, we can enable `-fno-plt` behavior on supported (ELF/Mach-O) targets (Linux Intel, Aarch64 with LLVM >= 19), since there shouldn't be a PLT section in the first place, for better code-gen (nb: people in the LLVM bug tracker seem to be arguing about it (cf: LLVM issue 126176, along with the link at the way bottom that shows what Huawei does with their RHEL fork), so this is a can of worms that can be delayed till later).

A1029384756:
went ahead and added the `no-plt` behavior on the suggested targets (linux amd64 and arm64, llvm 19+). it should be noted that there still is a `.plt` section with `__cxa_finalize` but when tested against clang with the same flags, this section also existed with the same contents

PR #6800, opened 2026-06-08
[net] Unix sockets

Tetralux:
Adds `dial_unix()`, `listen_unix()`, `accept_unix()`, `recv_unix()`, `send_unix()`, and a test.

Also, since I was touching the relevant code and noticed this while doing so, this also alters the default listen() backlog for each operating system from a magic 1000, to a more suitable number, according to the maximum number that must be supported by the implementation. e.g. on POSIX, this is `posix.SOMAXCONN`.

NOTE: As of writing this, I have not tested the FreeBSD/NetBSD implementations myself, though the CI can help with that.

NOTE: Unix sockets behave a lot like TCP sockets from what I gather, and as such I have transferred all the errors from that across to Unix socket operations. I have not personally tested all of those are correct, or that there aren't missing ones.

NOTE: The backlog values currently don't really have much validation to check if the selected number is valid.
The Windows impl merely checks it's < max(i32), and the others don't really check at all.
This may be fine however, since if it's not valid the syscall will probably either return an error(?) or simply pick a more suitable number. It is, after all, merely a hint on at least some of the supported systems.

### DIFF
--- a/core/nbio/errors.odin
+++ b/core/nbio/errors.odin
@@ -61,17 +61,19 @@ error_string :: proc(err: Error) -> string {
 
 error_string_recv :: proc(recv_err: Recv_Error) -> string {
 	switch err in recv_err {
-	case TCP_Recv_Error: return error_string(err)
-	case UDP_Recv_Error: return error_string(err)
-	case:                return "Unknown"
+	case TCP_Recv_Error:  return error_string(err)
+	case UDP_Recv_Error:  return error_string(err)
+	case Unix_Recv_Error: return error_string(err)
+	case:                 return "Unknown"
 	}
 }
 
 error_string_send :: proc(send_err: Send_Error) -> string {
 	switch err in send_err {
-	case TCP_Send_Error: return error_string(err)
-	case UDP_Send_Error: return error_string(err)
-	case:                return "Unknown"
+	case TCP_Send_Error:  return error_string(err)
+	case UDP_Send_Error:  return error_string(err)
+	case Unix_Send_Error: return error_string(err)
+	case:                 return "Unknown"
 	}
 }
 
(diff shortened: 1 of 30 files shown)


Yawning:
Some minor comments/feature requests:
- `AF_LOCAL` is a name that people have been trying to push, but `AF_UNIX` is probably fine.
- `sendmsg`/`recvmsg` should exist (non-Windows implementations support passing file descriptors between processes, and this is also needed for `SCM_CREDENTIALS`)
- (Linux, may "just work" already) There is an "abstract" namespace feature (first byte of the socket path is `\0`) which results in the socket not having a corresponding file.  This is convenient for various reasons (primarily around cleanup).

Tetralux:
@Yawning 
> Some minor comments/feature requests:
> 
> * `AF_LOCAL` is a name that people have been trying to push, but `AF_UNIX` is probably fine.
> * `sendmsg`/`recvmsg` should exist (non-Windows implementations support passing file descriptors between processes, and this is also needed for `SCM_CREDENTIALS`)
> * (Linux, may "just work" already) There is an "abstract" namespace feature (first byte of the socket path is `\0`) which results in the socket not having a corresponding file.  This is convenient for various reasons (primarily around cleanup).

I've not looked into `send_msg`/`recv_msg`.

With the socket path starting with a zero, is that a binary feature? Meaning that you are just selecting whether the UNIX socket has a file or not, and nothing else?

Separately, does "cleanup" mean "it can be nice because it avoids having to remove the socket file later", or anything else?

Yawning:
> @Yawning
>
> I've not looked into `send_msg`/`recv_msg`.
> 
> With the socket path starting with a zero, is that a binary feature? Meaning that you are just selecting whether the UNIX socket has a file or not, and nothing else?

"\0arbitraryidentifierlabel" sticks it in the "abstract" namespace, meaning no file, no permissions (because no file), and it goes away automatically when all references to the socket are gone.

The lack of file system permission checking does not prevent enforcing authentication (`SO_PEERCRED`), but would be left to the developer.

> Separately, does "cleanup" mean "it can be nice because it avoids having to remove the socket file later", or anything else?

Mostly "no need to deal with removing the file (which may be stale from a previous instance of the app)".

PR #6801, opened 2026-06-08
fix(formatting): Fixed incorrect formatting in dxgidebug.odin

MauriceElliott:
Literally just notice some out of line whitespace, feel free to ignore and fix yourselves.


gingerBill:
This is a double error, and you've not made it better.

PR #6806, opened 2026-06-10
vendor/vulkan: improve Vulkan ergonomics slightly

dayllenger:
This change removes `distinct` from `vk.DeviceSize`.

`vk.WHOLE_SIZE` constant is `u64`. You have to cast it, and of course you need to cast arbitrary offsets and sizes:
```c
vk.CmdFillBuffer(cmdbuf, buf, 0, vk.DeviceSize(vk.WHOLE_SIZE), 0x0)
vk.CmdFillBuffer(cmdbuf, buf, vk.DeviceSize(offset), vk.DeviceSize(size), 0x0)
write_ubo(&{buf, 0, vk.DeviceSize(vk.WHOLE_SIZE)})
```

Some functions take `[^]DeviceSize`, so you may need to make an array for this distinct type.

With Odin's strict type system you cannot confuse `u64` with `int` or `u32`. So, one can store API-agnostic `u64` for buffer sizes and offsets in the engine, and have slightly less Vulkan-specific code.

### DIFF
--- a/vendor/vulkan/_gen/create_vulkan_odin_wrapper.py
+++ b/vendor/vulkan/_gen/create_vulkan_odin_wrapper.py
@@ -943,7 +943,7 @@ API_VERSION_VARIANT :: proc "contextless" (version: u32) -> u32 {
 // Base types
 Flags         :: distinct u32
 Flags64       :: distinct u64
-DeviceSize    :: distinct u64
+DeviceSize    :: u64
 DeviceAddress :: distinct u64
 SampleMask    :: distinct u32
 
--- a/vendor/vulkan/core.odin
+++ b/vendor/vulkan/core.odin
@@ -42,7 +42,7 @@ API_VERSION_VARIANT :: proc "contextless" (version: u32) -> u32 {
 // Base types
 Flags         :: distinct u32
 Flags64       :: distinct u64
-DeviceSize    :: distinct u64
+DeviceSize    :: u64
 DeviceAddress :: distinct u64
 SampleMask    :: distinct u32
 


gingerBill:
I am not sure what ergonomics this is improving exactly.

dayllenger:
Before-after:
```c
region := vk.BufferCopy{0, 0, vk.DeviceSize(vk.WHOLE_SIZE)}
```
```c
region := vk.BufferCopy{0, 0, vk.WHOLE_SIZE}
```

Before-after:
```c
write_ubo(&{renderer.globals_ubo[frame_index].buffer, 0, vk.DeviceSize(vk.WHOLE_SIZE)})
```
```c
write_ubo(&{renderer.globals_ubo[frame_index].buffer, 0, vk.WHOLE_SIZE})
```

Before-after:
```c
vk.CmdFillBuffer(cmdbuf, curr_frame_buf.buffer, 0, vk.DeviceSize(vk.WHOLE_SIZE), 0x0)
```
```c
vk.CmdFillBuffer(cmdbuf, curr_frame_buf.buffer, 0, vk.WHOLE_SIZE, 0x0)
```

Before-after:
```c
size := vk.DeviceSize(len(data) * size_of(E))
res := vma.CopyMemoryToAllocation(vmaa, raw_data(data), buf.allocation, vk.DeviceSize(buf_offset), size)
```
```c
size := u64(len(data) * size_of(E))
res := vma.CopyMemoryToAllocation(vmaa, raw_data(data), buf.allocation, buf_offset, size)
```

We could change just `vk.WHOLE_SIZE` to be `vk.DeviceSize` but it is a breaking change.

D3D12 uses `u64` or `SIZE_T :: uint`. wgpu uses `u64`. Metal is distinct though. I prefer to have `u64` for memory offsets/sizes since you rarely see 64-bit integers in a renderer for other purposes.

diniamo:
I agree with this, the casting is kinda annoying. Especially because Vulkan isn't even consistent with using DeviceSize across the API (I've seen u64 and int too for sizes and offsets).

PR #6824, opened 2026-06-12
Update Vulkan

diniamo:


diniamo:
Of course they would add a double pointer...

diniamo:
Finally.

I couldn't come up with an easy solution that doesn't involve exceptions here: https://github.com/odin-lang/Odin/pull/6824/changes#diff-bdf9420678b007337a02c791dd77fe1ed6008da90c1ba874739ffc6918d7bec8R387-R388
Is this fine or is there a better way I can do it?

gingerBill:
@diniamo I have merged this PR, but for those "exceptions", I'd just hard code them and be done if necessary.

spencer-lunarg:
> Of course they would add a double pointer...

@diniamo randomly came across this... we (the Vulkan Working Group) are painfully aware of our sins we made with `VK_EXT_host_image_copy`, sorry, glad you found a simple enough solution

diniamo:
Well that's not a comment I was expecting lol. When I said that I was just frustrated, because it was my first time contributing to the generator, and I originally thought it handled all cases properly, so I first thought I also had to add a full solution. Turns out it's just a hacky mess (which doesn't matter as long as it outputs what we want), so an edge case was fine.

Otherwise, I have since tried the host image copy API, and not having to deal with staging buffers is a pretty big deal, so I generally like it. It also somehow outperformed staging buffers on the latest Mesa amdgpu driver with a desktop GPU in the case I was testing by like 40-60ms. While less surprising, it also won on a Windows Nvidia laptop with combined RAM.

spencer-lunarg:
sorry, `VK_EXT_host_image_copy` itself... awesome!

the double pointer to query things... we could have done better :parrot:

PR #6829, opened 2026-06-12
Make the type hash 63-bit to not crash OMG JIT in WebKit under WASM.

### LINKED ISSUE #6810: core:fmt (reflection/any path) crashes the `WebKit/Safari` tab in `js_wasm32` builds after ~3300 calls, no allocation, no imports (impacts `rl.TextFormat`)
## Context

* Operating System & Odin Version: macOS 26.5.1, Odin versions tested: 2026-05 and 2026-06

        OS:      macOS Tahoe 26.5.1 (build 25F80, kernel 25.5.0)
        CPU:     Apple M5
        RAM:     16384 MiB
        Backend: LLVM 22.1.7

## Expected Behavior

Calling `core:fmt` formatting (e.g. `fmt.bprintf` into a fixed buffer) once per `requestAnimationFrame` in a `js_wasm32` build should not crash the browser in `WebKit/Safari`.

Also impacts vendored raylib's `rl.TextFormat` by extension. Not an issue in chromium browsers.

## Current Behavior

Calling `fmt.bprintf()` ~3300 times in a `js_wasm32` build crashes the window forcing a reload. 

Did some investigation and:

| function called | result after 3300 calls |
| -- | -- |
| fmt.bprintf(buf, "frame %d", n)         | crash |
| fmt.bprint(buf, "constant") (no % verb) | crash |
| strconv.write_int(buf, n, 10)           | survives |
| strings.Builder + write_string          | survives |
| hand-rolled int→string                  | survives |

So it is narrowed down to `core:fmt`'s `any/reflection` value-formatting path: not the `%` verb (crashes without one), not `strconv`, not `strings.Builder`/`io.Writer` layer fmt writes through, with the only thing the crashing `fmt.bprint` does that the surviving Builder path doesn't is pass its argument as an any and reflect on it.

I have found that when using the vendored raylib TextFormat my game was crashing only in Safari.

To solve this I have resorted to rolling a rudimentary text format helper that skips the `fmt.bprintf` replacing `rl.TextFormat`:

```odin
@(private = "file")
fmt_bufs: [8][64]u8
@(private = "file")
fmt_idx: int

@(private = "file")
take_buf :: proc() -> []u8 {
	b := fmt_bufs[fmt_idx][:]
	fmt_idx = (fmt_idx + 1) % len(fmt_bufs)
	return b
}

@(private = "file")
put_str :: proc(buf: []u8, pos: int, s: string) -> int {
	p := pos
	for i in 0 ..< len(s) {
		buf[p] = s[i]
		p += 1
	}
	return p
}

@(private = "file")
put_uint :: proc(buf: []u8, pos: int, value: i32, pad: int) -> int {
	v := value
	if v < 0 {
		v = 0
	}
	tmp: [16]u8
	n := 0
	for {
		tmp[n] = u8('0' + (v % 10))
		v /= 10
		n += 1
		if v == 0 {
			break
		}
	}
	p := pos
	for _ in 0 ..< (pad - n) {
		buf[p] = '0'
		p += 1
	}
	for i := n - 1; i >= 0; i -= 1 {
		buf[p] = tmp[i]
		p += 1
	}
	return p
}

fmt_int :: proc(prefix: string, value: i32) -> cstring {
	buf := take_buf()
	p := put_str(buf, 0, prefix)
	p = put_uint(buf, p, value, 1)
	buf[p] = 0
	return cstring(raw_data(buf))
}

fmt_time :: proc(prefix: string, minutes, seconds: i32) -> cstring {
	buf := take_buf()
	p := put_str(buf, 0, prefix)
	p = put_uint(buf, p, minutes, 2)
	buf[p] = ':'
	p += 1
	p = put_uint(buf, p, seconds, 2)
	buf[p] = 0
	return cstring(raw_data(buf))
}
```

which behaves correctly both native and in both Safari and Chromium in WASM builds.

Calling `fmt.bprintf()` ~3300 times causes a webkit/Safari window to crash from a browser-process leak, WASM memory stays flat.

In the minimum reproduction example I linked further nothing crosses into JS and wasm memory is flat, so I would expect the root is almost certainly in WebKit's WASM engine. 

Likely worth filing a bug in WebKit maybe?

### Steps to Reproduce

Minimum reproducible repo:

[https://github.com/krisfur/odin-corefmt-wasm-issue](https://github.com/krisfur/odin-corefmt-wasm-issue)

Or if you want to do this yourself I suppose:

main.odin (swap the MODE value for testing different calls)

```odin
package main

import "base:runtime"
import "core:fmt"
import "core:strconv"
import "core:strings"

MODE :: 1
CALLS_PER_FRAME :: 5

n: int
buf: [64]u8

main :: proc() {}

@(export)
step :: proc "contextless" () -> i32 {
	context = runtime.default_context()
	for _ in 0 ..< CALLS_PER_FRAME {
		n += 1
		switch MODE {
		case 1:
			_ = fmt.bprintf(buf[:], "frame %d", n)
		case 2:
			manual(n)
		case 3:
			_ = fmt.bprint(buf[:], "constant")
		case 4:
			_ = strconv.write_int(buf[:], i64(n), 10)
		case 5:
			sb := strings.builder_from_bytes(buf[:])
			strings.write_string(&sb, "constant")
		}
	}
	return i32(buf[0]) // observed so the formatting isn't optimised away
}

@(export)
calls :: proc "contextless" () -> i32 {
	return i32(n)
}

manual :: proc(value: int) {
	v := value
	tmp: [20]u8
	m := 0
	for {
		tmp[m] = u8('0' + (v % 10))
		v /= 10
		m += 1
		if v == 0 do break
	}
	p := 0
	for i := m - 1; i >= 0; i -= 1 {
		buf[p] = tmp[i]
		p += 1
	}
	buf[p] = 0
}
```

build with:

```bash
odin build . -target:js_wasm32 -out:main.wasm
cp "$(odin root)/core/sys/wasm/js/odin.js" odin.js
```

serve in your favourite way of serving HTML, with index.html of the form:

```HTML
<!doctype html>
<html lang="en-us">
<head>
	<meta charset="utf-8">
	<title>fmt-wasm-repro</title>
	<style>body { font: 14px monospace; background: #111; color: #eee; padding: 10px } #last { color: #f66 }</style>
</head>
<body>
	<div id="last"></div>
	<div id="live">starting…</div>
	<script src="odin.js"></script>
	<script>
		const wmi = new odin.WasmMemoryInterface();
		WebAssembly.instantiateStreaming(fetch("main.wasm"), odin.setupDefaultImports(wmi)).then((o) => {
			const e = o.instance.exports;
			wmi.setExports(e);
			wmi.setMemory(e.memory);
			e._start();

			let frames = 0;
			const t0 = performance.now();
			// Previous run's last numbers, persisted each second so they survive the crash-reload.
			try { const p = localStorage.getItem("repro"); if (p) last.textContent = "LAST RUN BEFORE RELOAD -> " + p; } catch {}
			setInterval(() => {
				const line = "t=" + ((performance.now() - t0) / 1000).toFixed(1) + "s" +
					" frames=" + frames + " calls=" + e.calls() +
					" heap=" + (e.memory.buffer.byteLength / 1048576).toFixed(1) + "MB";
				live.textContent = line;
				try { localStorage.setItem("repro", line); } catch {}
			}, 1000);

			(function loop() { frames++; e.step(); requestAnimationFrame(loop); })();
		});
	</script>
</body>
</html>
```

> Note: Safari caches `main.wasm` hard, so if you rebuild you should serve with `no-store` or `hard-reload`.

If you want to verify that nothing crosses into JS, you can do it with Node (no browser): the wasm only imports `odin_env.write` and `rand_bytes`; instrumenting both with counters and calling step() 20,000 times shows 0 calls to either and memory.buffer.byteLength flat at flat at 17 pages (1.06 MB).

```js
const b=require("fs").readFileSync("main.wasm"), m=new WebAssembly.Module(b), c={};
const imp={odin_env:{}};
for(const i of WebAssembly.Module.imports(m)){c[i.name]=0; imp.odin_env[i.name]=(...a)=>(c[i.name]++, i.name==="write"?a[2]:0);}
const e=new WebAssembly.Instance(m,imp).exports; e._start();
for(let i=0;i<20000;i++) e.step();
console.log(c, e.memory.buffer.byteLength/65536, "pages");
```

### Failure Logs

No console logs or http server logs are produced, the browser reload gives no useful information, that would be too helpful. 

The minimum reproducible example repo produces a debug log in the form:
`LAST RUN BEFORE RELOAD -> t=11.0s frames=657 calls=3285 heap=1.1MB`

It took me too many hours to narrow this down...

Happy to help contribute to a fix if it is possible to resolve Odin side rather than WebKit side only.

krisfur:
Closes #6810 

When compiling Odin to WASM:

`fmt`'s type-switch over `any` lowers to a chain of equality compares against `typeid` values. `WebKit`'s `OMG` (optimizing wasm JIT) reconstructs a switch from that chain and computes its case-value span as a signed `i64` (max - min) to decide whether to build a jump table.

Because `typeids` are full-range 64-bit hashes, the cases span both signs, so `max - min` exceeds `2^63` and overflows: WebKit then allocates a pathologically-sized jump table and OOM-crashes the tab (deterministically, once the function tiers up to OMG; in practice after ~3300 `fmt` calls).

This constrains the `typeid` to 63 bits by clearing the top bit, so any switch over `typeids` has a span < `2^63` and the overflow can't occur. This doubles the type collision possibility, but it is still negligible. 

To lower the impact for other users it is behind the `-webkit-switch-workaround` build flag.

If you would prefer not to constrain `typeids` globally, the alternative is lowering any/union type-switches to switch on the dense type-info index instead of the hash, but the index isn't stable between compilations so not a great fix.

Also reported as a bug in WebKit: https://bugs.webkit.org/show_bug.cgi?id=317022

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -539,6 +539,8 @@ struct BuildContext {
 	bool   disallow_do;
 	bool   show_import_graph;
 
+	bool   webkit_switch_workaround;
+
 	IntegerDivisionByZeroKind integer_division_by_zero_behaviour;
 
 	LinkerChoice linker_choice;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -312,6 +312,7 @@ enum BuildFlagKind {
 	BuildFlag_Debug,
 	BuildFlag_DisableAssert,
 	BuildFlag_NoBoundsCheck,
+	BuildFlag_WebkitSwitchWorkaround,
 	BuildFlag_NoTypeAssert,
 	BuildFlag_NoDynamicLiterals,
 	BuildFlag_DynamicLiterals,
@@ -545,6 +546,7 @@ gb_internal bool parse_build_flags(Array<String> args) {
 	add_flag(&build_flags, BuildFlag_Debug,                   str_lit("debug"),                     BuildFlagParam_None,    Command__does_check);
 	add_flag(&build_flags, BuildFlag_DisableAssert,           str_lit("disable-assert"),            BuildFlagParam_None,    Command__does_check);
 	add_flag(&build_flags, BuildFlag_NoBoundsCheck,           str_lit("no-bounds-check"),           BuildFlagParam_None,    Command__does_check);
+	add_flag(&build_flags, BuildFlag_WebkitSwitchWorkaround,  str_lit("webkit-switch-workaround"),  BuildFlagParam_None,    Command__does_check);
 	add_flag(&build_flags, BuildFlag_NoTypeAssert,            str_lit("no-type-assert"),            BuildFlagParam_None,    Command__does_check);
 	add_flag(&build_flags, BuildFlag_NoThreadLocal,           str_lit("no-thread-local"),           BuildFlagParam_None,    Command__does_check);
 	add_flag(&build_flags, BuildFlag_NoDynamicLiterals,       str_lit("no-dynamic-literals"),       BuildFlagParam_None,    Command__does_check);
@@ -1234,6 +1236,9 @@ gb_internal bool parse_build_flags(Array<String> args) {
 							break;
 						case BuildFlag_NoBoundsCheck:
 							build_context.no_bounds_check = true;
+							break;
+						case BuildFlag_WebkitSwitchWorkaround:
+							build_context.webkit_switch_workaround = true;
 							break;
 						case BuildFlag_NoTypeAssert:
 							build_context.no_type_assert = true;
@@ -2889,6 +2894,11 @@ gb_internal int print_show_help(String const arg0, String command, String option
 	if (run_or_build) {
 		if (print_flag("-no-bounds-check")) {
 			print_usage_line(2, "Disables bounds checking program wide.");
+		}
+
+		if (print_flag("-webkit-switch-workaround")) {
+			print_usage_line(2, "Constrains 'typeid' values to 63 bits to avoid an OMG JIT crash in WebKit when running WASM builds.");
+			print_usage_line(2, "Only needed for 'js_wasm32'/'js_wasm64p32' targets run in Safari/WebKit. See: https://github.com/odin-lang/Odin/issues/6810");
 		}
 
 		if (print_flag("-no-crt")) {
--- a/src/name_canonicalization.cpp
+++ b/src/name_canonicalization.cpp
@@ -520,8 +520,17 @@ gb_internal u64 type_hash_canonical_type(Type *type) {
 	type_writer_make_hasher(&w, &w.hash_ctx);
 	write_type_to_canonical_string(&w, type);
 	u64 hash = typeid_hash_context_fini(&w.hash_ctx);
-	hash &= 0x7fffffffffffffffull;
-	hash = hash ? hash : 1;
+	if (build_context.webkit_switch_workaround) {
+		// Clear the top bit so every `typeid` is in [1, 2^63). A `switch` over a
+		// typeid (e.g. a type switch over `any` in core:fmt) then has a case-value
+		// span < 2^63. WebKit's B3/OMG wasm JIT computes a switch's value range as
+		// a signed i64 (max - min); a span >= 2^63 overflows and makes it build a
+		// pathologically-sized jump table, OOM-crashing the tab.
+		// WebKit bug: https://bugs.webkit.org/show_bug.cgi?id=317022
+		// Odin issue/PR: https://github.com/odin-lang/Odin/issues/6810
+		hash &= 0x7fffffffffffffffull;
+		hash = hash ? hash : 1;
+	}
 
 	type->canonical_hash.store(hash, std::memory_order_relaxed);
 
--- a/src/name_canonicalization.cpp
+++ b/src/name_canonicalization.cpp
@@ -520,6 +520,8 @@ gb_internal u64 type_hash_canonical_type(Type *type) {
 	type_writer_make_hasher(&w, &w.hash_ctx);
 	write_type_to_canonical_string(&w, type);
 	u64 hash = typeid_hash_context_fini(&w.hash_ctx);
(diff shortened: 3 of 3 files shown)


gingerBill:
We've been looking for this bug for a very long time and it has eluded us. 

The problem is that this "fix" is honestly something we should not be fixing since it is clearly a bug in WebKit. This fix at best is a bodge just for WebKit on WASM and nothing else.

Please keep your Issue #6810 and this PR open please!

Kelimion:
@krisfur, it's likely going to take a long, long time before Apple fixes this bug.

We could merge this if it was put behind a build option, e.g. `-webkit-sucks` so those affected can work around it until WebKit's been patched and their fix has percolated through the install base, without narrowing the hash for everyone else.

```diff
gb_internal u64 type_hash_canonical_type(Type *type) {
	if (type == nullptr) {
		return 0;
	}
	u64 prev_hash = type->canonical_hash.load(std::memory_order_relaxed);
	if (prev_hash != 0) {
		return prev_hash;
	}

	TypeWriter w = {};
	type_writer_make_hasher(&w, &w.hash_ctx);
	write_type_to_canonical_string(&w, type);
	u64 hash = typeid_hash_context_fini(&w.hash_ctx);
+	if (build_context.webkit_switch_workaround) {
+		hash &= 0x7fffffffffffffffull;
+		hash = hash ? hash : 1;
+	}

	type->canonical_hash.store(hash, std::memory_order_relaxed);

	return hash;
}
```

Kelimion:
It should also have a comment above that branch that links to the bug and this PR.

krisfur:
Great shout @Kelimion, done and commented - also need to say this codebase is so tidy it's quite nice adding stuff haha

Kelimion:
When merged, we should reopen #6810 and leave it open until WebKit fixes things on their end so we have a reminder.

PR #6830, opened 2026-06-13
Stack canaries

A1029384756:
adds support for/enables by default stack canaries on all desktop platforms, barring arm. arm seemingly has some odd behavior where glibc is compiled without `-fPIC`, resulting in some linking errors. i will test this on other platforms to see if this is just github CI but disabled this to be safe - edit: tested on both fresh ubuntu and fedora arm machines, the issue persists. i can make it work in clang though so this is my issue

additionally, i reordered the linux pic/plt logic since as it was (my mistake), it would break the android subtarget

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -1952,18 +1952,9 @@ gb_internal void init_build_context(TargetMetrics *cross_target, Subtarget subta
 		}
 	}
 
-	// note: there are some stack protection toolchain issues on arm
-	// platforms that will clash with PIC/PIE:
-	// `unresolvable R_AARCH64_ADR_PREL_PG_HI21 relocation against symbol`
-	if (metrics->os == TargetOs_linux &&
-			metrics->arch != TargetArch_arm32 &&
-			metrics->arch != TargetArch_arm64
-	) {
-		if (bc->stack_protector == StackProtector_Default) {
-			bc->stack_protector = StackProtector_Ssp;
-		}
-	} else if (metrics->os == TargetOs_windows ||
+	if (metrics->os == TargetOs_windows ||
 			metrics->os == TargetOs_darwin ||
+			metrics->os == TargetOs_linux ||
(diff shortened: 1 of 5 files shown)


A1029384756:
linux arm builds are now fixed. the issue turned out to be that we weren't setting the reloc mode for the module. this was causing some codegen that was improperly assuming that PIC *wasn't* enabled (now the reason as to why that's only an issue on linux + arm eludes me)

Yawning:
Maybe default to `sspstrong`, the heuristics are a bit more stringent.

ssp:
```
* Character arrays larger than ssp-buffer-size (default 8).
* Aggregates containing character arrays larger than ssp-buffer-size.
* Calls to alloca() with variable sizes or constant sizes greater than ssp-buffer-size.
```

sspstrong:
```
 * Arrays of any size and type
 * Aggregates containing an array of any size and type.
 * Calls to alloca().
 * Local variables that have had their address taken.
```

Packaging standards are:
>All Fedora packages are compiled with -fstack-protector since Fedora Core 5, and -fstack-protector-strong since Fedora 20.[19][20] Most packages in Ubuntu are compiled with -fstack-protector since 6.10.[21] Every Arch Linux package is compiled with -fstack-protector since 2011.[22] All Arch Linux packages built since 4 May 2014 use -fstack-protector-strong.[23] Stack protection is only used for some packages in Debian,[24] and only for the FreeBSD base system since 8.0.[25] Stack protection is standard in certain operating systems, including OpenBSD,[26] Hardened Gentoo[27] and DragonFly BSD.[citation needed]

We do have bounds-checking enabled by default so this may seem redundant at a glance, but have ways to opt out of  bounds-checking, call external code via FFI, and multi-pointers.

A1029384756:
sounds good, went ahead and updated the default

gingerBill:
I am going to merge this PR but I am going to change the default stack protection level to `None` because even though it's useful, it should still be opt-in.

PR #6834, opened 2026-06-13
Add newline when using `odin root` command

juaumjuaumjuaum:


juaumjuaumjuaum:
Otherwise this happens 
image

laytan:
It's made to be used in scripting, `$(odin root)/core` for example. We don't want a newline there.

juaumjuaumjuaum:
Ah sorry, should have asked first!

PR #6835, opened 2026-06-13
Bug/stringprecision

cleanshaven:
This is to fix a difference in fmt between how C handles precision and how Odin handles Precision.

I noticed that when using the following with wprintf(stream, "%5.4s", "12345") I would get the result of "12345" instead of what I expected was " 1234" which is the C result.

I altered fmt.fmt_string and noticed there was already code that did this with structure tags.  The code to do this from the wprintf was not present. I added the code ensuring that the precision is taken into effect only if there was no structure tag.

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -1487,7 +1487,10 @@ fmt_string :: proc(fi: ^Info, s: string, verb: rune) {
 	s, verb := s, verb
 	if ol, ok := fi.optional_len.?; ok {
 		s = s[:clamp(ol, 0, len(s))]
-	}
+	} else if fi.prec_set {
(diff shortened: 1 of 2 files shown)


gingerBill:
Your formatting does not adhere to the code base. We use tabs for indentation, not spaces.

gingerBill:
Also, we are not trying to be identical to C, so this behaviour may or may not be the "correct" idea.

cleanshaven:
> Also, we are not trying to be identical to C, so this behaviour may or may not be the "correct" idea.

However, You already support this behaviour in the fmt tags for struct.

gingerBill:
> However, You already support this behaviour in the fmt tags for struct.

I have no idea what you mean by this statement.

PR #6839, opened 2026-06-14
`core:rexcode`

gingerBill:
# rexcode

High-performance multi-architecture instruction encoder/decoder/printer library written in Odin. Developed by @dotbmp, original source: https://github.com/dotbmp/rexcode

## Architectures

| Package    | ISA                                                              | Coverage |
|------------|------------------------------------------------------------------|----------|
| `x86`      | x86-64 + i386 (legacy/SSE/AVX/AVX-512/BMI/FMA/AES-NI)             | LLVM-verified |
| `arm32`    | ARMv8 AArch32 (A32 + T32 + Thumb-1 + VFP + NEON + crypto/CRC)     | LLVM-verified, 100% sweep |
| `arm64`    | ARMv8 AArch64 (base integer + FP scalar; SVE/SME WIP)             | LLVM-verified |
| `mips`     | MIPS I/II/III/IV + R6 + COP1 FPU + COP0 + GTE + PS2 EE MMI + DSP ASE | LLVM-verified |
| `riscv`    | RV32GC / RV64GC                                                  | LLVM-verified |
| `ppc`      | Power ISA 3.1 + AltiVec/VSX/MMA/HTM/DFP/BookE/SPE/SPE2 + Paired Singles + VMX128 | 3327 entries, LLVM-verified |
| `ppc_vle`  | Freescale e200 VLE (sibling to `ppc`)                            | 222 entries, binutils-verified |
| `mos6502`  | NMOS 6502 + undocumented + 65C02 + HuC6280                       | da65-verified |
| `mos65816` | W65C816S (SNES, Apple IIgs)                                      | ca65-verified |
| `rsp`      | N64 RSP (MIPS-derived scalar + vector unit)                      | armips-verified |

Every package follows the same API contract (see `docs/cross_arch_design.md`).

## Design

- **Encoder**: assembles instructions to machine code with label resolution
  and per-arch relocation support.
- **Decoder**: disassembles machine code back to structured instructions.
- **Printer**: emits assembly text output with optional syntax-highlighting
  tokens.
- **Table-driven**: O(1) opcode lookup via precomputed encoding/decoding
  tables.
- **Zero allocations** on the hot path: caller provides all buffers.

The `isa/` package owns the parts that are the same on every ISA — labels,
result/error types, the print framework, token types, and shared
formatting helpers. Each architecture package owns its registers, memory
model, operand types, mnemonics, encoding tables, and the actual
`encode_one`/`decode_one` bytes.

## Performance (x86)

To be re-announced when optimizations have settled.

## Usage

```odin
import "x86"

instructions := []x86.Instruction{
	x86.inst_r_r(.MOV, x86.RAX, x86.RDI),
	x86.inst_r_r(.ADD, x86.RAX, x86.RSI),
	x86.inst_none(.RET),
}

code: [4096]u8
relocs: [dynamic]x86.Relocation
errors: [dynamic]x86.Error
result := x86.encode(instructions[:], nil, code[:], &relocs, &errors)

decoded_insts:  [dynamic]x86.Instruction
decoded_info:   [dynamic]x86.Instruction_Info
decoded_labels: [dynamic]x86.Label_Definition
decode_errors:  [dynamic]x86.Error
x86.decode(code[:result.byte_count], nil, &decoded_insts, &decoded_info, &decoded_labels, &decode_errors)

x86.print(decoded_insts[:], decoded_info[:], decoded_labels[:])
disasm := x86.tprint(decoded_insts[:], decoded_info[:], decoded_labels[:])
```

The same shape works for every other arch — change the import.

## Instruction Builders

```odin
x86.inst_none(.RET)
x86.inst_r(.PUSH, x86.RAX)
x86.inst_r_r(.MOV, x86.RAX, x86.RBX)
x86.inst_r_i(.MOV, x86.EAX, 42, 4)
x86.inst_r_r_r(.VADDPS, x86.XMM0, x86.XMM1, x86.XMM2)
x86.inst_m_r(.MOV, x86.mem_base_only(x86.RSP), 8, x86.RAX)
x86.inst_r_m(.MOV, x86.RAX, x86.mem_base_disp(x86.RBP, -8), 8)
x86.inst_rel(.JMP, label_id, 1)
```

## Memory Operands

```odin
x86.mem_base_only(x86.RAX)                        // [RAX]
x86.mem_base_disp(x86.RBP, -16)                   // [RBP - 16]
x86.mem_base_index(x86.RAX, x86.RCX, 4)           // [RAX + RCX*4]
x86.mem_base_index_disp(x86.RAX, x86.RCX, 8, 32)  // [RAX + RCX*8 + 32]
x86.mem_rip_disp(0)                               // [RIP + disp]
```

## Labels

```odin
labels: [dynamic]x86.Label_Definition
instructions: [dynamic]x86.Instruction

loop := x86.label(&labels, &instructions)
x86.emit_r(&instructions, .DEC, x86.RDI)
x86.emit_rel(&instructions, .JNZ, loop)

done := x86.label_forward(&labels)
x86.emit_rel(&instructions, .JMP, done)
labels[done] = x86.Label_Definition(len(instructions))

result := x86.encode(instructions[:], labels[:], code[:], &relocs, &errors)
```

For named labels:

```odin
lm: x86.Label_Map
x86.label_map_init(&lm)
defer x86.label_map_destroy(&lm)

loop := x86.label_named(&lm, "loop", &instructions)
done := x86.label_reserve(&lm, "done")
x86.label_set(&lm, "done", &instructions)

result := x86.encode(instructions[:], lm.labels[:], code[:], &relocs, &errors)

// Printer wants id→name; Label_Map stores name→id, so invert once.
id_to_name := make(map[u32]string, len(lm.names), context.temp_allocator)
for name, id in lm.names { id_to_name[id] = name }
x86.print(decoded_insts[:], decoded_info[:], lm.labels[:], label_names = &id_to_name)
```

## Running Tests

Each package has its own test suite:

```sh
odin run x86/tests
odin run arm32/tests
odin run arm64/tests
odin run mips/tests
odin run mos6502/tests
odin run mos65816/tests
odin run ppc/tests
odin run ppc_vle/tests
odin run riscv/tests
odin run rsp/tests
```

## Verification harnesses

Each arch has a verification harness under `<arch>/tools/`:
- `dump_verify_input.odin` — emits the per-entry hex/asm manifest.
- `verify_against_<tool>.*` — runs the canonical external assembler/
  disassembler and compares. LLVM-mc for the seven modern archs, plus
  `da65`/`ca65`/`armips`/`powerpc-eabivle-as` for retro/embedded ISAs.

## Project Structure

```
rexcode/
	isa/                # shared core: labels, status, print framework, label-inference
	docs/               # cross-arch design + per-arch design docs
	x86/                # x86-64 / i386
	arm32/              # AArch32
	arm64/              # AArch64
	mips/               # MIPS (R1..R6 + ASEs + coprocessors)
	mos6502/            # NMOS 6502 family
	mos65816/           # W65C816S
	ppc/                # PowerPC (Power ISA 3.1)
	ppc_vle/            # Freescale VLE (sibling of ppc)
	riscv/              # RISC-V
	rsp/                # N64 RSP
```

Per-package layout (canonical, enforced by the cross-arch contract):

```
<arch>/
	encoder.odin         # encode() — two-pass, label/reloc-aware
	decoder.odin         # decode()
	printer.odin         # sb/sbln/print/println/aprint/aprintln/tprint/tprintln/bprint/bprintln/fprint/fprintln/wprint/wprintln
	registers.odin       # Register, REG_* classes, typed enums
	operands.odin        # Operand, Memory, Operand_Kind, op_* constructors
	instructions.odin    # Instruction, inst_* builders
	encoding_types.odin  # Encoding, Encoding_Flags, isa re-exports
	encoding_table.odin  # ENCODING_TABLE: [Mnemonic][]Encoding
	decoding_tables.odin # generated dispatch tables
	mnemonics.odin       # Mnemonic enum (u16, INVALID=0)
	reloc.odin           # Relocation_Type + Relocation
	tests/               # smoke, pipeline_smoke, sweep
	tools/               # gen_decode_tables, dump_verify_input, verify_against_*
```

## Cross-architecture API design

See [docs/cross_arch_design.md](docs/cross_arch_design.md) for the
naming contract every arch package follows.


blob1807:
Does the original author have a good understanding of assembler & experience dealing with it?  I assume they do, otherwise this PR wouldn't exist, but it would still be nice to hear it directly.  Since I'm probably not alone in having reservations about LLM generated code, especially in large amounts.

gingerBill:
@blob1807 He does have a good understanding, that's why I am accepting it. I have been discussing with him directly on this.

blob1807:
@gingerBill Okay, good to know the was some pre-discussion. Thank you for the reply.

dotbmp:
> Does the original author have a good understanding of assembler & experience dealing with it? I assume they do, otherwise this PR wouldn't exist, but it would still be nice to hear it directly. Since I'm probably not alone in having reservations about LLM generated code, especially in large amounts.

I can assure you I don't use Claude like the typical vibecoder. I designed the API very explicitly myself for performance and ergonomic usage, as well as the entire architecture and how everything would be done. I read the generated code and went back and forth extensively to get quality results, and added automated encoding and decoding validation against LLVM and other toolchains (for ISAs that LLVM doesn't support) to ensure that everything is as validated as possible. There are some undocumented and/or reverse-engineered extensions on some ISAs that are somewhat less validated. I think in particular some PSP stuff, but I'd like to add validation using PPSSPP tooling (though, obviously, PSP codegen is a very minor target).

The main part that was left entirely up to the AI was the generation of the encoding tables themselves, entirely from LLM-memory. This was because all of the authoritative data sources for these ISAs are a mess (XML), not in any common format between ISAs, some ISAs do not have them, and for some ISAs many of the desired extensions would be missing. However I did this knowing there would be errors, and the validation layer exists to catch those errors and fix them. As of Bill merging the library, all validations were passing, and validation coverage is very high.

The design is also structured intentionally to be readable and maintainable by humans, and I hope adding this to the Odin core libraries means we'll get a lot of high-quality usage, feedback and fixes/improvements. Odin is also, by design, very resistant to AI slop because the language doesn't support features that result in disgusting code. I'd also add that in my experience, Claude and other AIs produce far better quality code in Odin than any other language I've tested them with, with less steering, as a result of this (and probably a high-quality training set).

Igonato:
I know it's not an *architecture* per se, but can WASM instructions be a part of this? Or is there something blocking that?

gingerBill:
> I know it's not an _architecture_ per se, but can WASM instructions be a part of this? Or is there something blocking that?

@Igonato One of the differences between WASM and the other ISAs is that WASM is a stack-machine rather than a register machine, so the approach to machine code is a little different entirely, and might not work as intended. 

I'd have to refer to @dotbmp on whether it is even possible or a good idea to support something like wASM.

Igonato:
> WASM is a stack-machine rather than a register machine

Should it matter for decoding/encoding instructions though? `.wasm` file structure is wildly different to a regular executable, but the instructions still come one after another, and [the spec looks pretty neat](https://webassembly.github.io/spec/core/binary/instructions.html), not a mess of XML (mess of TeX instead 😅)

core/rexcode/isa/x86/printer.odin
```diff
@@ -0,0 +1,749 @@
...
+sbprintln :: proc(
+	sb:           ^strings.Builder,
+	instructions: []Instruction,
+	inst_info:    []Instruction_Info,
+	label_defs:   []Label_Definition,
+	tokens:       ^[dynamic]Token = nil,
+	options:      ^Print_Options = nil,
+	label_names:  ^map[u32]string = nil,
+) {
+	#force_inline sbprint(sb, instructions, inst_info, label_defs, tokens, options, label_names)
```
Kelimion, line 563: I'm not sure that `#force_inline` is appropriate here. The printers aren't exactly in the hot path, and it's a lot of code to duplicate for each of the print variants.

core/rexcode/isa/x86/printer.odin
```diff
@@ -0,0 +1,749 @@
...
+// print: Print instructions to stdout
+print :: proc(
+	instructions: []Instruction,
+	inst_info:    []Instruction_Info,
+	label_defs:   []Label_Definition,
+	tokens:       ^[dynamic]Token = nil,
+	options:      ^Print_Options = nil,
+	label_names:  ^map[u32]string = nil,
+) {
+	sb := strings.builder_make(context.temp_allocator)
```
Kelimion, line 576: Leaks on the temp allocator. While it could be followed up with `free_all(context.temp_allocator)`, it's more user friendly if `print` (and `println`) use `runtime.DEFAULT_TEMP_ALLOCATOR_TEMP_GUARD()` to roll back the tempory allocation on return.

core/rexcode/isa/x86/printer.odin
```diff
@@ -0,0 +1,749 @@
...
+	w:            io.Writer,
+	instructions: []Instruction,
+	inst_info:    []Instruction_Info,
+	label_defs:   []Label_Definition,
+	tokens:       ^[dynamic]Token = nil,
+	options:      ^Print_Options = nil,
+	label_names:  ^map[u32]string = nil,
+) {
+	sb := strings.builder_make(context.temp_allocator)
+	sbprint(&sb, instructions, inst_info, label_defs, tokens, options, label_names)
```
Kelimion, line 736: For some reason these don't `#force_inline`. Good, but it makes me wonder why the other ones do.

Kelimion: Could also use the `runtime.DEFAULT_TEMP_ALLOCATOR_TEMP_GUARD()`.

Kelimion (commented):
It may also be beneficial to move the various tests to their usual location, using `core:testing` so they run in parallel.

dotbmp:
> I know it's not an _architecture_ per se, but can WASM instructions be a part of this? Or is there something blocking that?

I already plan on adding WASM. As well as other IR-level things, like SPIR-V, DXIL and AIR for compute/shader code. Like Bill said, they're not shaped the same way, so the API for the ISAs won't port cleanly to these - they may be in a separate package, and it's up to Bill how or whether to ship them in the core library.

PR #6844, opened 2026-06-14
Fix #6811: collect @(test) procs from #+private file by iterating c->…

### LINKED ISSUE #6811: [Bug] `#+private file` file tag silently ignores all tests in the file
## Context
Odin:    dev-2026-06:daa56de55
OS:      Debian GNU/Linux forky/sid, Linux 6.19.14-101.fc44.x86_64
RAM:     32004 MiB
Backend: LLVM 21.1.8

## Issue
Setting the file tag `#+private file` silently ignores all tests in that file. This makes the procs in that file effectively untestable, unless one sets the `private` attribute on all the individual procs instead of the file.

Minimal repro:
Running `odin test` on the following finishes with `No tests to run.`.
```odin
#+private file
package main

import "core:testing"

@(test)
test :: proc (t: ^testing.T) {}
```

Uncommenting the `#+private file` line makes the test run as expected.

astenmark:
collect @(test) procs from #+private file by iterating c->info.entities instead of pkg->scope->elements.

Test procs in #+private file files were silently dropped by the test runner because collect_testing_procedures_of_package iterated pkg->scope->elements, which only contains entities exported to the package scope. File-private entities are placed in the file scope instead and never reach pkg->scope. Fix by iterating c->info.entities filtered by package, matching how @(init) collection works in generate_minimum_dependency_set_internal. Test procs are already exempted from laziness, so they are guaranteed to appear in c->info.entities by the time collection runs.

Fixes #6811

### DIFF
--- a/src/checker.cpp
+++ b/src/checker.cpp
@@ -2875,9 +2875,11 @@ gb_internal void collect_testing_procedures_of_package(Checker *c, AstPackage *p
 	InternedString interned = string_interner_insert(str_lit("Test_Signature"), 0, &hash);
 	Entity *test_signature = scope_lookup_current(testing_scope, interned, hash);
 
-	Scope *s = pkg->scope;
-	for (auto const &entry : s->elements) {
-		Entity *e = entry.value;
+	for_array(i, c->info.entities) {
+		Entity *e = c->info.entities[i];
+		if (e->pkg != pkg) {
+			continue;
+		}
 		if (e->kind != Entity_Procedure) {
 			continue;
 		}


astenmark:
The Windows CI failure in tests/core/crypto/wycheproof/pqc.odin appears to be pre-existing and unrelated to this PR.

BigBoyBarney:
Thank you so much @astenmark! I can confirm this solves the issue! I've been looking forward to this a lot.

Thanks again!

astenmark:
@BigBoyBarney The pleasure is all mine. Enjoy your testing!

PR #6845, opened 2026-06-14
Fix #6814: reject transitive #subtype-only implicit conversion in checker

### LINKED ISSUE #6814: Passing a doubly-nested #subtype causes a compiler panic.
	Odin:    dev-2026-06-nightly:7ab61e4
	OS:      Arch Linux, Linux 7.0.5-arch1-1
	CPU:     Intel(R) Core(TM) i5-4570S CPU @ 2.90GHz
	RAM:     7841 MiB
	Backend: LLVM 20.1.8

## Expected Behavior

I'm going to assume that `#subtype` shouldn't nest (unlike using) and that this should be an error reported to the user.

In my example below, it should probably tell me that I can't pass `Baz` to `test`.

## Current Behavior

The compiler panics and the compilation fails.

### Reproducable Example
```odin
package main

import "core:fmt"

Foo :: struct {
	num: int,
}

Bar :: struct {
	#subtype foo: Foo,
}

Baz :: struct {
	#subtype bar: Bar,
}

Bap :: struct {
	#subtype foo: Foo,
	#subtype bar: Bar,
}

Beep :: struct {
	#subtype bar: Bar,
	#subtype foo: Foo,
}

test :: proc(foo: Foo) {
	fmt.println(foo.num)
}

main :: proc() {
	// This works fine
	bar := Bar{foo={num=10}}
	test(bar)

	// CRASH
	// baz := Baz{bar={foo={num=20}}}
	// test(baz)

	// Note that this works.
	bap := Bap{
		foo = {num=30}, // test takes in this foo here
		bar = {},
	}
	test(bap)

	// And this is just checking if the order of foo matters, which it doesn't seem to
	beep := Beep {
		bar = {},
		foo = {num=40},
	}
	test(beep)
}
```

### Failure Logs

```
main::main
lb_emit_conv: src -> dst
Not Identical Baz != Foo
Not Identical struct {bar: Bar} != struct {num: int}
Not Identical 7fb19d1486c0 != 7fb19d1478a0
Not Identical 7fb19d148768 != 7fb19d147948
src/llvm_backend_expr.cpp(3291): Panic: Invalid type conversion: 'Baz' to 'Foo' for procedure 'main::main'
```

astenmark:
Fixes #6814

When a struct Baz has #subtype bar: Bar and Bar has #subtype foo: Foo, the checker was accepting a Baz value passed to a procedure expecting Foo. This caused a GB_PANIC("Invalid type conversion") in llvm_backend_expr.cpp because the backend's lookup_subtype_polymorphic_selection only follows transitive subtype chains through fields that also have using (EntityFlag_Using), while the checker's check_is_assignable_to_using_subtype was recursing into any field with EntityFlags_IsSubtype (which includes plain #subtype fields that only set EntityFlag_Subtype).

The fix aligns the checker's recursion guard with the backend: transitive descent now only happens when the intermediate field also has EntityFlag_Using (i.e. declared with using or using #subtype). Single-hop #subtype and transitive using #subtype chains are unaffected.

This previously panicked, now gives a proper checker error:

```
package main

Foo :: struct { x: int }
Bar :: struct { #subtype foo: Foo }
Baz :: struct { #subtype bar: Bar }

takes_foo :: proc(f: Foo) {}

main :: proc() {
    baz: Baz
    takes_foo(baz) // Error: Cannot assign value 'baz' of type 'Baz' to 'Foo' in a procedure argument
}
```

### DIFF
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -5030,9 +5030,15 @@ gb_internal isize check_is_assignable_to_using_subtype(Type *src, Type *dst, isi
 				return level+1;
 			}
 		}
-		isize nested_level = check_is_assignable_to_using_subtype(f->type, dst, level+1, src_is_ptr, allow_polymorphic);
-		if (nested_level > 0) {
-			return nested_level;
+		// Only follow the chain transitively when the field also has `using`, which is
+		// what the backend's lookup_subtype_polymorphic_selection requires (it gates
+		// recursion on EntityFlag_Using). A plain `#subtype`-only field enables a
+		// single-hop conversion but not a two-or-more hop transitive one.
+		if (f->flags & EntityFlag_Using) {
+			isize nested_level = check_is_assignable_to_using_subtype(f->type, dst, level+1, src_is_ptr, allow_polymorphic);
+			if (nested_level > 0) {
+				return nested_level;
+			}
 		}
 	}
 


astenmark:
The Windows CI failure in tests/core/crypto/wycheproof/pqc.odin appears to be pre-existing and unrelated to this PR. I don't know why the Ubuntu build is timing out. I don't see how that would be related to this PR.

gingerBill:
I should not have merged this since it has fundamentally broken the functionality of `#subtype`

gingerBill:
I have fixed it by forced `EntityFlags_IsSubtype` everywhere rather than just `EntityFlag_Using`

PR #6855, opened 2026-06-18
Fix map reserve calculation

### LINKED ISSUE #6854: [Bug?] `reserve` and `make` don't account for map resize rules
Currently, when a user `reserve`s or `make_map_cap`s a map, the map's capacity does not account for the fact that the map will resize itself when it's 75% full.

When I reserve `X` in a map, I expect to be able to insert `X` elements without the map resizing itself.

The following snippet illustrates the issue:
```odin
x := make_map(map[int]struct{})
reserve(&x, 16) // Map capacity is 16

// Map resizes itself when the 13th element is inserted
// due to the 75% allocation rule.
for i in 0 ..< 13 {
	x[i] = {}
}

fmt.printfln("Map capacity: %d", cap(x)) // => 32
```

This could be solved with relative ease by changing the logic in `map_reserve_dynamic()`, by calculating with an extra 33% allocation. I can't think of anything this would break.

Thoughts?

BigBoyBarney:
Fixed #6854 

Copying parts of that issue here for convenience:
Currently, when a user `reserve`s or `make_map_caps` a map, the map's capacity does not account for the fact that the map will resize itself when it's 75% full, which is not the case for other types `reserve` and `make` act upon.

When I reserve X in a map, I expect to be able to insert X elements without the map resizing itself. That is the entire point of `reserve`.

This PR calculates the new capacity ceiling by taking the load factor into account.

-----

Aparte: There is no new "surprise allocation size" effect introduced by this change, since `reserve` and `make` already allocated more than what the user explicitly asked for, always rounding up to the next power of 2.

I.e.: `reserve(&map, 20)` already allocated 32.

### DIFF
--- a/base/runtime/dynamic_map_internal.odin
+++ b/base/runtime/dynamic_map_internal.odin
@@ -608,15 +608,17 @@ map_reserve_dynamic :: #force_no_inline proc "odin" (#no_alias m: ^Raw_Map, #no_
 		m.allocator = context.allocator
 	}
 
-	new_capacity := new_capacity
+	// Take into account the fact that the map will resize itself at
+	// MAP_LOAD_FACTOR capacity.
+	new_capacity_with_resize := new_capacity * 100 / MAP_LOAD_FACTOR
 	old_capacity := uintptr(map_cap(m^))
 
-	if old_capacity >= new_capacity {
+	if old_capacity >= new_capacity_with_resize {
 		return nil
 	}
 
 	// ceiling nearest power of two
-	log2_new_capacity := ceil_log2(new_capacity)
+	log2_new_capacity := ceil_log2(new_capacity_with_resize)
 
 	log2_min_cap := max(MAP_MIN_LOG2_CAPACITY, log2_new_capacity)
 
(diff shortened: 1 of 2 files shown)


base/runtime/dynamic_map_internal.odin
```diff
@@ -608,15 +608,17 @@ map_reserve_dynamic :: #force_no_inline proc "odin" (#no_alias m: ^Raw_Map, #no_
 		m.allocator = context.allocator
 	}
 
-	new_capacity := new_capacity
+	// Take into account the fact that the map will resize itself at
+	// MAP_LOAD_FACTOR capacity.
+	new_capacity_with_resize := new_capacity * 100 / MAP_LOAD_FACTOR
```
gingerBill, line 623: If we are to merge this (still not decided), this calculation is not a good one because it could overflow too early for large (but valid) `map`s on 32-bit systems. So to minimize this risk, I'd recommend doing something like a conservative overflow check.

BigBoyBarney: Thank you for  considering this.

I completely forgot about 32-bit systems haha. Added a check now.

gingerBill: That's not a good check still. And again, why are you doing `100 / 75` when `4 / 3`  is clearly the same thing and will have a smaller change of overflowing?

gingerBill: In fact, the check is a little tricky to get right too because it's not a simple calculation.

PR #6857, opened 2026-06-19
core/crypto: RSA support

Yawning:
- [x] Arbitrary precision big integers in Z[0,n) (n should be odd)
- [x] RSA
  - [x] Public key s11n
  - [x] Private key s11n
  - [x] Key generation
  - [x] Private key verification
- [x] RSA PKCS 1.5 (Ewwwwwww)
  - [x] Signatures
    - [x] Verification
    - [x] Signing
  - [x] Decryption (TLS <= 1.2 only, no IRF) 
- [x] RSA PSS
  - [x] Verification
  - [x] Signing
- [x] RSA OAEP
  - [x] Decryption
  - [x] Encryption
- [x] Tests
  - [x] Standalone
  - [x] Wycheproof
- [x] Benchmarks


core/crypto/_bigint/i31.odin
```diff
@@ -0,0 +1,900 @@
...
+	return
+}
+
+// Subtract `b` from `a` and return the `carry` (`0` or `1`).
+// If `ctl` is `0`, then `a` is unmodified, but the carry is still computed
+// and returned.
+//
+// The slices `a` and `b` MUST have the same announced bit length (in subscript `0`)
+//
+// `a` and `b` MAY be the same array, but partial overlap is not allowed.
```
Kelimion, line 187: Feels like these two comments also apply to `i31_add`. It's practically the same code apart from the operation.

Kelimion (approved):

core/crypto/_bigint/i31.odin
Yawning: Fixed.

PR #6871, opened 2026-06-22
fix `odin root` command missing a newline

msaher:


Kelimion:
It's missing on purpose so it can be used in scripts.

PR #6873, opened 2026-06-23
core:text/regex: improve compiler performance

bennyyip:
Improving regex compiler performance by reducing the usage of `inject_at`.

### DIFF
--- a/core/text/regex/compiler/compiler.odin
+++ b/core/text/regex/compiler/compiler.odin
@@ -449,15 +449,14 @@ compile :: proc(tree: Node, flags: common.Flags) -> (code: Program, class_data: 
 
 		// `.*?`
 		split := Split{.Split, i16(SPLIT_SIZE + size_of(byte) + JUMP_SIZE), i16(SPLIT_SIZE)}
-		inject_raw(&code, pc_open, split)
-		pc_open += size_of(Split)
-
-		inject_at(&code, pc_open, Opcode.Wildcard)
-		pc_open += size_of(byte)
-
 		jump := Jump{.Jump, i16(-size_of(byte) - SPLIT_SIZE)}
-		inject_raw(&code, pc_open, jump)
-		pc_open += size_of(Jump)
+		pack := struct {
+			a: Split,
+			b: Opcode,
+			c: Jump,
+		} { split, Opcode.Wildcard, jump }
+		inject_raw(&code, pc_open, pack)
+		pc_open += size_of(Split) + size_of(byte) + size_of(Jump)
 	}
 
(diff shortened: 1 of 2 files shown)


core/text/regex/compiler/compiler.odin
```diff
@@ -141,15 +148,13 @@ map_all_classes :: proc(tree: Node, collection: ^[dynamic]Rune_Class_Data) {
...
+	append(code, ..data[:])
 }
 inject_raw :: #force_inline proc(code: ^Program, start: int, data: $T) {
 	// NOTE: This is system-dependent endian.
-	for b, i in transmute([size_of(T)]byte)data {
-		inject_at(code, start + i, cast(Opcode)b)
-	}
+	data := transmute([size_of(T)]Opcode)data
+	inject_at(code, start, ..data[:])
 }
```
gingerBill, line 158: I wonder if these should take a pointer instead to minimize the extra stack copy.

```odin
append_raw :: #force_inline proc(code: ^Program, data: ^$T) {
	// NOTE: This is system-dependent endian.
	ptr := (^[size_of(T)]Opcode)(data)
	append(code, ..ptr[:])
}

inject_raw :: #force_inline proc(code: ^Program, start: int, data: ^$T) {
	// NOTE: This is system-dependent endian.
	ptr := (^[size_of(T)]Opcode)(data)
	inject_at(code, start, ..ptr[:])
}
```

And then any occurence like `save := Save{...}` should be `save := &Save{...}`

core/text/regex/virtual_machine/virtual_machine.odin
```diff
@@ -49,6 +49,35 @@ Opcode :: enum u8 {
...
+	operand: Opcode,
+}
+Wait_For_Rune_Class_Negated :: struct #packed {
+	opcode: Opcode,
+	operand: Opcode,
+}
+Save :: struct #packed {
+	opcode: Opcode,
+	operand: Opcode,
+}
```
gingerBill, line 80: Try to align the fields with themselves.

core/text/regex/compiler/compiler.odin
bennyyip: They are almost the same. Without `-o:speed`, they are exactly the same. With `-o:speed`, original version is 1 movw + 3 movb, ptr version is 2 movw + 1 movb. 
I think without ptr is more clear.

https://odin.godbolt.org/z/EMMa6hTx1

gingerBill: Good testing! I was just wondering if they were or not.

PR #6879, opened 2026-06-24
Fix literal integer `%%`

### LINKED ISSUE #6877: `%%` is broken on literals
## Context

        Odin:    dev-2026-06-nightly:7ab61e4
        OS:      Windows 11 Education (version: 25H2), build 26200.8037
        CPU:     AMD Ryzen 7 3700X 8-Core Processor
        RAM:     32689 MiB
        Backend: LLVM 20.1.0

```odin
package main
import "core:fmt"

main :: proc() {
	D: int = 1
	d: int = -3
	fmt.printfln("%v mod %v = %v", D, d, D %% d)
	fmt.println(int(1) %% int(-3))
}
```

## Expected Behavior

1 mod -3 = -2
-2

## Current Behavior

1 mod -3 = -2
1

bennyyip:
According to overview, `%%` is "remainder (floored)". So it `%%` for literals should not use `big_int_euclidean_mod`.
I implemented `big_int_floor_mod` and use it instead.

These two printfln should be the same.
```odin
package main
import "core:fmt"

main :: proc() {
	x, y: int = 1, -3
	fmt.printfln("%v mod %v = %v", x, y, x %% y)
	fmt.printfln("%v mod %v = %v", 1, -3, 1 %% -3)
	assert(x %% y == 1 %% -3)
}
```

Where should I add test for this PR?

close #6877

### DIFF
--- a/src/big_int.cpp
+++ b/src/big_int.cpp
@@ -448,6 +448,17 @@ gb_internal void big_int_euclidean_mod(BigInt *z, BigInt const *x, BigInt const 
 	}
 }
 
+gb_internal void big_int_floor_mod(BigInt *z, BigInt const *x, BigInt const *y) {
+	BigInt y0 = {};
+	big_int_init(&y0, y);
+
+	BigInt q = {};
+	big_int_quo_rem(x, y, &q, z);
+	if (z->sign != y0.sign && !mp_iszero(z)) {
+		big_int_add(z, z, &y0);
+	}
+	big_int_dealloc(&q);
+}
 
 
 gb_internal void big_int_and(BigInt *dst, BigInt const *x, BigInt const *y) {
--- a/src/exact_value.cpp
+++ b/src/exact_value.cpp
@@ -780,7 +780,7 @@ gb_internal ExactValue exact_binary_operator_value(TokenKind op, ExactValue x, E
 		case Token_Quo:    return exact_value_float(fmod(big_int_to_f64(a), big_int_to_f64(b)));
 		case Token_QuoEq:  big_int_quo(&c, a, b); break; // NOTE(bill): Integer division
 		case Token_Mod:    big_int_rem(&c, a, b); break;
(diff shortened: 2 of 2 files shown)


src/big_int.cpp
```diff
@@ -448,6 +448,17 @@ gb_internal void big_int_euclidean_mod(BigInt *z, BigInt const *x, BigInt const
 	}
 }
 
+gb_internal void big_int_floor_mod(BigInt *z, BigInt const *x, BigInt const *y) {
+	BigInt y0 = {};
+	big_int_init(&y0, y);
```
krnowak, line 453: This is being leaked, isn't it?

gingerBill: It's not a problem it is being leaked.

src/big_int.cpp
```diff
@@ -448,6 +448,17 @@ gb_internal void big_int_euclidean_mod(BigInt *z, BigInt const *x, BigInt const
...
+
+	BigInt q = {};
+	big_int_quo_rem(x, y, &q, z);
+	if (z->sign != y0.sign && !mp_iszero(z)) {
+		big_int_add(z, z, &y0);
+	}
```
gingerBill, line 459: This is the wrong calculation.

bennyyip: It's not. I follow the implementation in this [paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/divmodnote-letter.pdf). [stb](https://github.com/nothings/stb/blob/master/stb_divide.h#L250-L261) does the same.


If you want `%%` to be `big_int_euclidean_mod`,  then we should change this to be euclidean:
https://github.com/odin-lang/Odin/blob/f58d7edb399fd679f923fea15c0892acb5163588/src/llvm_backend_expr.cpp#L438-L445

There is no way this assert fails: 
```odin
x, y: int = 1, -3
assert(x %% y == 1 %% -3)
```

Issue #6798, opened 2026-06-07
Compiler panic when iterating over SoA slice of type with 'using'


IvanB101:
## Context

Odin:    dev-2026-06-nightly:23f57a4
OS:      Ubuntu 24.04.4 LTS, Linux 6.17.0-35-generic
CPU:     AMD Ryzen 5 5500U with Radeon Graphics
Backend: LLVM 20.1.8

## Expected Behavior

Program should correctly compile

## Current Behavior

Building produces the following error
```
src/llvm_backend_utility.cpp(1603): Panic: un-gep-able type [^]Inner
This is a compiler error. Please report this.
[1]    23463 illegal hardware instruction (core dumped)  ~/Downloads/odin-linux-amd64-nightly+2026-06-06/odin build .
```

## Failure Information (for bugs)

The code that produces the error
```
package main

import "core:fmt"

Inner :: struct {
	value: f32,
}

Outer :: struct {
	using inner: Inner,
}

Data :: struct {
	arr: #soa[10]Outer,
}

main :: proc() {
	data := Data{}

    for &outer in data.arr[:] {
        fmt.println(outer.value)
        // NOTE: this version works fine
        // fmt.println(outer.inner.value)
    }
}
```

### Steps to Reproduce

Build the program provided in the previous point.

### Failure Logs

```
src/llvm_backend_utility.cpp(1603): Panic: un-gep-able type [^]Inner
This is a compiler error. Please report this.
[1]    23463 illegal hardware instruction (core dumped)  ~/Downloads/odin-linux-amd64-nightly+2026-06-06/odin build .
```

Issue #6725, opened 2026-05-24
Compiler error when using nested SOA structs


michael-weems:
## Context

```bash
        Odin:    dev-2026-04:a896fb2b4
        OS:      Windows 11 Home Basic (version: 25H2), build 26200.8457
        CPU:     AMD Ryzen 7 6800H with Radeon Graphics
        RAM:     15630 MiB
        Backend: LLVM 20.1.0
```

I am new to odin, so I may be using features incorrectly. 

## Expected Behavior

Either the compiler tells me this is incorrect usage of a language feature or the compiler properly compiles it. 

## Current Behavior

```bash
Odin\src\llvm_backend_utility.cpp(1596): Panic: un-gep-able type [^]Mid
This is a compiler error. Please report this.
```

## Failure Information (for bugs)

 Nested `using` usage alongside `#soa` usage on the user side seems to cause this compiler behavior.

### Steps to Reproduce

```odin
package repro

Inner :: struct {
	x: int,
	y: int,
}

Mid :: struct {
	using inner: Inner,
	z:           int,
}

Outer :: struct {
	using mid: Mid,
	w:         int,
}

main :: proc() {
	arr: #soa[]Outer = make(#soa[]Outer, size_of(Outer) * 4)
	_ = arr[0].x
}
```

### Failure Logs

```bash
Odin\src\llvm_backend_utility.cpp(1596): Panic: un-gep-able type [^]Mid
This is a compiler error. Please report this.
```

Issue #4943, opened 2025-03-16
Accessing nested struct field in an SOA array indirectly via `using` while iterating causes LLVM panic


jrachele:
## Context

This is very similar to bug #4172, but with a different, perhaps more common way of reproducing it.

```
        Odin:    dev-2025-02:584fdc0d4
        OS:      macOS Sequoia 15.3.2 (build 24D81, kernel 24.3.0)
        CPU:     Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
        RAM:     16384 MiB
        Backend: LLVM 19.1.7
```

## Expected Behavior

You should be able to access the fields of a nested struct that is fused with `using` without it throwing a panic.

## Current Behavior

A panic occurs when accessing the fields of a nested struct indirectly, via `using`

## Failure Information (for bugs)

`src/llvm_backend_utility.cpp(1456): Panic: un-gep-able type [^]My_inner_struct`

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.
```odin
package repro

My_inner_struct :: struct {
	x: u32,
}

My_struct :: struct {
	using inner: My_inner_struct,
}

main :: proc() {
	entities: #soa[dynamic]My_struct

	append_soa(&entities, My_struct{})
	for &entity in entities {
		entity.inner.x = 1 // This is okay
		entity.x = 1 // This is not
	}
}
```

Note: This happens specifically in the range loop, probably due to internally fetching the slice within the dynamic array, therefore invoking the same problem as #4172

neraidware:
I'm facing this exact same scenario currently. Able to reproduce in the latest revision

PR #6881, opened 2026-06-24
Fix for SoA arrays of structs with nested using structs crashing the compiler

sa-pro:
there are a few open issues about this, examples:

https://github.com/odin-lang/Odin/issues/6725
https://github.com/odin-lang/Odin/issues/6798
https://github.com/odin-lang/Odin/issues/4943

just searching for "using soa" in the open issues finds a lot more.

the fix seems to be very simple: just making sure the element is a ^T instead of [^]T before getting the element pointer fixes this. tried it out and seems to fix it properly even with nested of nested structs (working example after the fix):

```
First :: struct {
    i: i32,
}

Second :: struct {
    using first: First,
    j:           i32,
}

Third :: struct {
    using second: Second,
    k:            i32,
}


main :: proc() {

    data: #soa[dynamic]Third
    append(&data, Third{i = 1, j = 2, k = 3})

    fmt.printfln("data: {}", data[0].i)
    fmt.printfln("data: {}", data[0].j)
    fmt.printfln("data: {}", data[0].k)
}
```

edit: sorry, apparently vscode removed a few white spaces that didnt show in the commit changes, thats why it has a few extra white space removal lines.. but the fix is literally just changing 1 LoC from one place to before a function call.

### DIFF
--- a/src/llvm_backend_expr.cpp
+++ b/src/llvm_backend_expr.cpp
@@ -53,11 +53,11 @@ gb_internal lbValue lb_emit_logical_binary_expr(lbProcedure *p, TokenKind op, As
 	incoming_blocks[done->preds.count] = p->curr_block->block;
 
 	lb_emit_jump(p, done);
-	lb_start_block(p, done);	
-	
+	lb_start_block(p, done);
+
 	LLVMTypeRef dst_type = lb_type(m, t_llvm_bool);
 	LLVMValueRef phi = nullptr;
-	
+
 	GB_ASSERT(incoming_values.count == incoming_blocks.count);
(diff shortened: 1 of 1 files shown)


gingerBill:
I am closing this because it was meant to be a one line change and you introduced way too many whitespace changes.

sa-pro:
you mean you copied my fix and gave me no credit after some bullshit about whitespaces that could be reverted in a simple commit.
i lost almost 2 hours finding the bug, it might be easy for you because you know the code base, even though you clearly didnt care about it because that bug has been reported since at least 2023. yes its a simple fix, but took time to get it.

yeah whatever.

PR #6890, opened 2026-06-26
[core:slice] Myers' Difference Algorithm

brodeuralexis:
Implement Myers' Difference Algorithm for slices of simply comparable values. This implies computing a slice of edits to apply to the original slice to obtain the expected slice.

I've been writing a lot of Odin recently, and while I like that testing is builtin, but `testing.expect_value` is making me angry, since I am trying to diff tokens and AST representations of [Valve 220 MAP files](https://book.leveldesignbook.com/appendix/resources/formats/map). (even a simple map file can be around ~200 tokens). 

If this is accepted, I believe that further work can be done to improve `core:testing`'s assertion helpers to also display differences between their `value`, and `expected` argument.  With `os.is_tty`, logic can be applied to use `core:terminal/ansi` instead of prefixes to denote inserted or deleted values.

While the algorithm works and computes differences as expected, I know the number of memory allocations can be drastically reduced (hopefully linear memory usage with fixed number of calls to `make`). If there is interest in this PR, I'll take time to improve the algorithm.

### DIFF
--- a/core/slice/diff.odin
+++ b/core/slice/diff.odin
@@ -8,26 +8,20 @@ import "base:intrinsics"
 
 import "core:mem"
 
-// An edit script command to insert a value.
-Diff_Insert :: struct($T: typeid) {
-	value: T,
-}
-
-// An edit script command to delete a value.
-Diff_Delete :: struct($T: typeid) {
-	value: T,
-}
-
-// An edit script command to keep a value.
-Diff_Keep :: struct($T: typeid) {
-	value: T,
+// A kind of difference in an edit script.
+Diff_Kind :: enum {
+	// A value should be inserted in the original slice.
+	Insert,
+	// A value should be deleted from the original slice.
+	Delete,
+	// A value should be kept in the original slice.
+	Keep,
 }
 
 // A difference in an edit script.
-Diff :: union($T: typeid) #no_nil {
-	Diff_Insert(T),
-	Diff_Delete(T),
-	Diff_Keep(T),
+Diff :: struct($T: typeid) {
+	kind:  Diff_Kind,
+	value: T,
 }
 
 /*
@@ -70,7 +64,7 @@ diff :: proc(
 
 	stack := make([dynamic]Subproblem, allocator)
 	defer delete(stack)
-	append(&stack, Subproblem{ax = 0, ay = 0, bx = len(a), by = len(b)})
+	append(&stack, Subproblem{ax = 0, ay = 0, bx = len(a), by = len(b)}) or_return
 
 	for len(stack) != 0 {
 		p := pop(&stack)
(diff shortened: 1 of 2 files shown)


core/slice/diff.odin
```diff
@@ -0,0 +1,276 @@
...
+Diff_Keep :: struct($T: typeid) {
+	value: T,
+}
+
+// A difference in an edit script.
+Diff :: union($T: typeid) #no_nil {
+	Diff_Insert(T),
+	Diff_Delete(T),
+	Diff_Keep(T),
+}
```
gingerBill:
```odin
Diff_Kind :: enum {
	Keep,
	Insert,
	Delete,
}

Diff :: struct($S: typeid/[]$T) {
	kind: Diff_Kind,
	value: []S,
}
```

core/slice/diff.odin
```diff
@@ -0,0 +1,276 @@
...
+- allocator: The allocator used to create the edit script (default is context.allocator).
+
+Returns:
+- result: The edit script to transform the original list in the expected list.
+- err: An `Allocator_Error`, if allocation failed.
+*/
+diff :: proc(
+	value, expected: []$T,
+	allocator := context.allocator,
+) -> (
```
gingerBill, line 44:
```odin
diff :: proc(value, expected: $S/[]$T, allocator := context.allocator) -> (result: []Diff(S), err: mem.Allocator_Error) {
```

core/slice/diff.odin
```diff
@@ -0,0 +1,276 @@
...
+				append(&script, Diff(T)(Diff_Delete(T){value = deletes[i]})) or_return
+			} else if i in keeps {
+				append(&script, Diff(T)(Diff_Keep(T){value = keeps[i]})) or_return
+			}
+		}
+	}
+
+	for insert in end_inserts {
+		append(&script, Diff(T)(Diff_Insert(T){value = insert})) or_return
+	}
```
gingerBill, line 165:
```odin
append(&script, Diff(S){kind = .Insert, end_inserts[:])
```

gingerBill: You are storing individual elements rather ranges which is what you should be doing because most things will be giant ranges.

core/slice/diff.odin
```diff
@@ -0,0 +1,276 @@
...
+	) -> (
+		result: Maybe(Middle_Snake),
+		err: mem.Allocator_Error,
+	) {
+		n := p.bx - p.ax
+		m := p.by - p.ay
+		delta := n - m
+		odd := (delta % 2 != 0)
+
+		// TODO: Convert to fixed-length slices.
```
gingerBill: Should this be done already?

brodeuralexis: Yep, I'll work on improving memory usage for now.

core/slice/diff.odin
brodeuralexis: I just checked, and parametric polymorphic specializations are only valid for procedures:
```
/home/chaton/workspace/brodeuralexis/Odin/core/slice/diff.odin(22:29) Error: '$T' is not a type 
        Diff :: struct($S: typeid/[]$T) { 
                                    ^^ 
 
/home/chaton/workspace/brodeuralexis/Odin/core/slice/diff.odin(22:30) Error: Invalid use of a polymorphic parameter '$T' 
        Diff :: struct($S: typeid/[]$T) { 
```

PR #6899, opened 2026-06-27
Fix #6786

### LINKED ISSUE #6786: Failure to assign to Maybe in Using struct
## Context
I am trying to construct a struct with the {name= value} syntax. The struct has a using that uses another struct with a Maybe and it fails to assign the value.

        Odin:    dev-2026-05-nightly:ea5175d
        OS:      Windows 11 Home Basic (version: 25H2), build 26200.8457
        CPU:     13th Gen Intel(R) Core(TM) i9-13900HX
        RAM:     32507 MiB
        Backend: LLVM 20.1.0

## Expected Behavior
It should assign the value to the name in the struct

## Current Behavior
The assertion fails in the following example.

### Steps to Reproduce
The following fails the assertion
```
B :: struct {
	something : Maybe(bool)
}

A :: struct {
	color : Maybe([4]f32),
	using _ : B,
}

my_proc :: proc (a : A) -> bool {
	fmt.printf("a : %v", a);
	return a.something.(bool);
}

assert(my_proc({color = [4]f32{1,0,0,1}, something = true}));
```

Removing the Maybe or moving the Maybe to A and the behavior changes to the expected:
```
B :: struct {
	something : bool,
}

A :: struct {
	color : [4]f32,
	using _ : B,
}

my_proc :: proc (a : A) -> bool {
	fmt.printf("a : %v", a);
	return a.something;
}

assert(my_proc({color = [4]f32{1,0,0,1}, something = true}));
```

korvahkh:
We weren't coercing the constant to the type of the field we're trying to assign to.

This resulted in union tags not being set when assigned through `using`.

I'm fairly certain there is no case in which not coercing the type is desirable.

### DIFF
--- a/src/llvm_backend_const.cpp
+++ b/src/llvm_backend_const.cpp
@@ -2103,7 +2103,7 @@ gb_internal lbValue lb_const_value(lbModule *m, Type *type, ExactValue value, Ty
 									}
 								}
 								if (is_constant) {
-									LLVMValueRef elem_value = lb_const_value(m, tav.type, tav.value, tav.type, cc).value;
+									LLVMValueRef elem_value = lb_const_value(m, cv_type, tav.value, tav.type, cc).value;
 									if (LLVMIsConstant(elem_value) && LLVMIsConstant(values[index])) {
 										values[index] = llvm_const_insert_value(m, values[index], elem_value, idx_list, idx_list_len);
 									} else if (is_local) {


korvahkh:
Upon further inspection I'm still seeing some cases that have issues, so this PR probably needs further work to actually be an adequate fix.

For example when I run this program:

```odin
package main

import "core:fmt"

T :: union{int, bool}

B :: struct {
	something: T,
}

A :: struct {
	color: [16]u8,
	using x: B,
}

x :: proc(a: A) -> A {
	return a
}

raw_bytes :: proc(a: $T) -> [size_of(T)]byte {
	a := a
	return transmute([size_of(T)]byte)a
}

main :: proc() {
	a := A{color = 1, something = true}
	b := x(A{color = 1, something = true})
	fmt.printfln("a: %02x", raw_bytes(a))
	fmt.printfln("b: %02x", raw_bytes(b))
	fmt.printfln("a: %v", a)
	fmt.printfln("b: %v", b)
}
```

The output is:
```
a: [01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, f7, d9, 6d, e5, 55, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00]
b: [01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 01, 00, 00, 00, 00, 00, 00, 00, 02, 00, 00, 00, 00, 00, 00, 00]
a: A{color = [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1], x = B{something = nil}}
b: A{color = [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1], x = B{something = true}}
```

(the values of the `f7, d9, 6d, e5, 55, 00, 00` bytes in the first `a` line are non-deterministic).

korvahkh:
It looks like the problem I'm seeing is caused by LLVM's handling of converting between the struct generated in the branch starting at [llvm_backend_const.cpp:968](https://github.com/odin-lang/Odin/blob/a4dcf8f5d65f940fa4261c9712e535461fec1506/src/llvm_backend_const.cpp#L968) and the underlying struct of the union type.

The number of fields in generated struct ends up different for variants smaller than the largest due to the padding logic in the non-LLVM 14 branch here:
https://github.com/odin-lang/Odin/blob/a4dcf8f5d65f940fa4261c9712e535461fec1506/src/llvm_backend_const.cpp#L1004-L1013

My current theory is that if we had an implementation of `llvm_const_pad_to_size` for versions after 14, and used it instead of a separate padding field, that might solve the issue.
Unfortunately, I'm not familiar enough with LLVM's API to create an implementation to actually test that theory.

gingerBill:
@korvahkh Thank you for looking into this. I'll give a good look now to see if I can improve it.

gingerBill:
Thank you for figuring out the EXACT source of the problem. I have now fixed it thanks to you!

gingerBill:
Fixed it this way: https://github.com/odin-lang/Odin/commit/e227c736e66859d5ebcdf3a787232832cdc14721

PR #6900, opened 2026-06-27
Playdate Support

MauriceElliott:
After rigorous testing I've narrowed support down to this change.
There are also the following target-features:"~~-fp64,-fpregs64,-vfp2,-vfp3d16,-vfp4d16,+vfp2sp,+vfp3d16sp,+vfp4d16sp,~~+no-movt" turns out only no-movt is actually required, the rest is handled by the triple and calling convention.
Most of the testing has been done with this repo: https://tangled.org/codedawa.dev/gutwound/tree/testing_odin_playdate_target

There is still lots to do on the playdate end but at least it runs!


MauriceElliott:
I retested this on the same commit on my own repo and got a green build: https://github.com/MauriceElliott/Odin/actions/runs/28335769073/job/83941604708
Any idea what might be causing this here or reckon we could just requeue the failed job?

MauriceElliott:
Have run into an issue with floats, going to convert this to a draft for now and continue work from there.

MauriceElliott:
I have chosen to specifically target the cortex-m7 for this change to minimise the impact, I don't own any of the other hardware this may affect so can't currently test. Its ready for review though and has been tested thoroughly on the playdate.

PR #6913, opened 2026-06-30
Add more constant type cases

### LINKED ISSUE #6831: #simd type casts are wrong and/or crash compiler
## Context

        Odin:    dev-2026-06:a99ac952d
        OS:      Arch Linux, Linux 7.0.11-arch1-1
        CPU:     11th Gen Intel(R) Core(TM) i5-11600K @ 3.90GHz
        RAM:     15852 MiB
        Backend: LLVM 22.1.6

## Expected Behavior

Type casts to #simd types should work correctly.

## Current Behavior

Some casts do not work and/or crash the compiler.

### Steps to Reproduce

```odin
package main

import "core:fmt"

main :: proc() {
    x := cast(#simd[16]u8) 3.14
    fmt.println(x) // prints <0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>

    y := cast(#simd[16]u8) 256 // crashes compiler
}
```

### Failure Logs

```
[DEBUG] 256
u8 -> 1
src/llvm_backend_const.cpp(524): Assertion Failure: `sz >= max_count` max_count: 2, sz: 1, written: 0, type u8
Illegal instruction        (core dumped) odin build .
```

tf2spi:
SIMD vector types are not checked if they're constant, so add this case in order to improve accuracy of other checks enough to resolve #6831

### DIFF
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -1471,6 +1471,10 @@ gb_internal bool is_type_constant_type(Type *t) {
 		return is_type_constant_type(t->Array.elem);
 	case Type_EnumeratedArray:
 		return is_type_constant_type(t->EnumeratedArray.elem);
+	case Type_SimdVector:
+		return is_type_constant_type(t->SimdVector.elem);
+	case Type_Matrix:
+		return is_type_constant_type(t->Matrix.elem);
 	}
 	return false;
 }


tf2spi:
Are there other types that should be handled in ``is_type_constant_type``? I didn't want to be too eager and make types constant that shouldn't have been constant.

PR #6918, opened 2026-06-30
Remove LLVM 14 support

Kelimion:
LLVM 17 is now the minimum supported version. That also means the old pass manager has been removed, as it was only necessary on LLVM versions older than 17.

The build script now also suggests using `apt.llvm.org` (or the redhat equivalent) if it can't find a supported `llvm-config` version.


Kelimion:
~TODO: Patch to update the install instruction on the website to point to e.g. apt.llvm.org if your ancient OS install doesn't have at least LLVM 17 available.~

Edit: Done.

PR #6924, opened 2026-07-01
Fix misuse of lb_emit_(min|max) in simd_(min|max)

### LINKED ISSUE #6815: Compiler segfault: `simd.min`/`simd.max` on float vectors with `-target:freestanding_wasm32 -target-features:+simd128`
# Context

* Operating System & Odin Version: macOS Sequoia 15.6.1 (arm64, Apple M4 Max), Odin dev-2026-06 (Homebrew)

```
	OS:      macOS Sequoia 15.6.1 (build 24G90, kernel 24.6.0)
	CPU:     Apple M4 Max
	RAM:     49152 MiB
	Backend: LLVM 22.1.7
```

## Expected Behavior

`simd.min` / `simd.max` on float vectors (`#simd[4]f32`) compiles for
`-target:freestanding_wasm32` with `-target-features:+simd128`, ideally
lowering to the wasm `f32x4.min` / `f32x4.pmin`-style operations.

## Current Behavior

The compiler itself crashes with SIGSEGV (exit code 139) while emitting code.
No diagnostic is printed. Crashes identically for `simd.min` and `simd.max`,
in `-build-mode:obj` and `-build-mode:llvm-ir`, and at every optimization
level (`-o:none` through `-o:speed`), but only in non-debug builds; adding
`-debug` makes the same source compile cleanly.

The same source compiles fine when any one of the trigger conditions is
removed:

| variant | result |
|---|---|
| `-target:freestanding_wasm32 -target-features:+simd128` (float vectors) | **compiler SIGSEGV** |
| `-target:freestanding_wasm32` without `+simd128` | OK |
| same flags + `-debug` | OK |
| `-target:js_wasm32 -target-features:+simd128` | OK |
| native arm64 darwin | OK |
| same flags, but integer vectors (`#simd[4]u32`) | OK |

Other float-vector simd ops (`lanes_lt/le/gt/ge`, `select`, `floor`, `abs`,
`reduce_*`, arithmetic) compile fine under the same flags, so it appears
isolated to the float min/max lowering path.

Workaround in user code: `simd.select(simd.lanes_lt(a, b), a, b)` in place of
`simd.min(a, b)` compiles and runs correctly under the same flags (modulo
NaN/±0 semantics).

Under lldb the crash is a null `llvm::Value*` passed as the condition into
LLVM's constant folder while the backend builds a select.

### Steps to Reproduce

1. Save the following as `repro.odin` in an empty directory:

```odin
package main

import "core:simd"

@(export, link_name="t")
t :: proc "c" (a, b: f32) -> f32 {
	av := simd.f32x4(a)
	bv := simd.f32x4(b)
	return simd.extract(simd.min(av, bv), 0)
}
```

2. Build it for wasm32 with simd128 enabled:

```sh
odin build . -target:freestanding_wasm32 -build-mode:obj -out:repro.o -o:none -target-features:+simd128
```

3. The compiler exits with `Segmentation fault: 11` (exit code 139).

### Failure Logs

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: libLLVM.dylib`llvm::ConstantFolder::FoldSelect(llvm::Value*, llvm::Value*, llvm::Value*) const
libLLVM.dylib`llvm::ConstantFolder::FoldSelect:
->  ldrb   w8, [x1]        ; first argument (the select condition) is null
    ldrb   w9, [x2]
    ldrb   w10, [x3]
    cmp    w8, #0x15
```

(Deeper frames unavailable: the release compiler is built with
`-fomit-frame-pointer`, so lldb cannot unwind past frame 0.)

tf2spi:
lb_emit_(min|max) is a utility used for emitting the (min|max) of selecting the scalar/array-like itself after comparison, but it doesn't implement the selection logic for the individual elements needed for simd_(min|max), so replace their usages in simd_(min|max) by inlining the logic for selection (similar to the int case) and LLVM intrinsics.

Resolves #6815

### DIFF
--- a/src/llvm_backend_expr.cpp
+++ b/src/llvm_backend_expr.cpp
@@ -3878,6 +3878,7 @@ gb_internal lbValue lb_emit_comp(lbProcedure *p, TokenKind op_kind, lbValue left
 			break;
 		}
 
+		GB_ASSERT(res.value != nullptr);
 		return res;
 
 	} else if (is_type_soa_pointer(a)) {
--- a/src/llvm_backend_proc.cpp
+++ b/src/llvm_backend_proc.cpp
@@ -1693,15 +1693,8 @@ gb_internal lbValue lb_build_builtin_simd_proc(lbProcedure *p, Ast *expr, TypeAn
 		return res;
 	case BuiltinProc_simd_min:
 		if (is_float) {
-			bool use_llvm_intrinsic = !is_arch_wasm();
-			if (use_llvm_intrinsic) {
-				LLVMValueRef args[2] = {arg0.value, arg1.value};
-				LLVMTypeRef types[1] = {lb_type(p->module, res.type)};
-				res.value = lb_call_intrinsic(p, "llvm.minnum", args, gb_count_of(args), types, gb_count_of(types));
-			} else {
-				LLVMValueRef cond = LLVMBuildFCmp(p->builder, LLVMRealOLT, arg0.value, arg1.value, "");
-				res.value = LLVMBuildSelect(p->builder, cond, arg0.value, arg1.value, "");
-			}
+			LLVMValueRef cond = LLVMBuildFCmp(p->builder, LLVMRealOLT, arg0.value, arg1.value, "");
+			res.value = LLVMBuildSelect(p->builder, cond, arg0.value, arg1.value, "");
 		} else {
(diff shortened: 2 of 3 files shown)


gingerBill:
I am not sure if using `minnum`/`maxnum` is a good idea since it is not the same behaviour as `min`/`max`. 

`min` and `max` are defined very dumb on purpose, and `minnum`/`maxnum` variants are a little different.

PR #6964, opened 2026-07-07
Faster win32 directory iteration.

rmn64k:
I noticed that using the `core:os` directory walker was very slow for a folder hierarchy I have on my machine. It's somewhat large, consisting of 289 directories and 558522 files.

This PR brings the directory walking time down from 22-30 s to about 200-300 ms on my machine, so it's a significant improvement in that regard. There are drawbacks however. This will no longer fill in inode information, and it will no longer distinguish between `Named_Pipe`, `Character_Device`, `Regular` or `Undetermined` file types. I choose to have it report only `Regular` as that would mostly be the case I think, but `Undetermined` would be more truthful.

I guess the crux of this is if the speed up is worth the drawbacks for the standard library.

The main changes in this PR are:
1. Use `FindFirstFileExW` to fetch directory data in larger chunks and skip populating the short name. Falls back to the previous `FindFirstFileW` on error.
2. Skip doing `CreateFileW`, `GetFileType`, `GetFileInformationByHandleEx` and `CloseHandle` per entry. This is the main thing that slows the original code down. However we can't find out as much info per entry as per the drawbacks above.
3. Perform less work when building the `fullpath` and `name` fields by putting the basepath, directory separator and utf8 conversion directly into the allocated buffer.

### DIFF
--- a/core/os/dir_windows.odin
+++ b/core/os/dir_windows.odin
@@ -2,8 +2,35 @@
 package os
 
 import "base:runtime"
+import "core:slice"
 import "core:time"
 import win32 "core:sys/windows"
+
+@(private="file")
+find_data_fullpath :: proc(base_path: string, d: ^win32.WIN32_FIND_DATAW, allocator: runtime.Allocator) -> (path, name: string, err: Error) {
+	wname_len := i32(slice.linear_search(d.cFileName[:], 0) or_else len(d.cFileName))
+	wname := cstring16(raw_data(d.cFileName[:]))
+
+	name_len := win32.WideCharToMultiByte(win32.CP_UTF8, win32.WC_ERR_INVALID_CHARS, wname, wname_len, nil, 0, nil, nil)
+	if name_len == 0 {
+		return "", "", _get_platform_error()
+	}
+
+	path_buf := make([]byte, len(base_path)+1+int(name_len), allocator) or_return
+	copy(path_buf, base_path)
+	path_buf[len(base_path)] = '\\'
+
+	name_buf := path_buf[len(base_path)+1:]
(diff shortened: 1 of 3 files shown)


core/os/dir_windows.odin
```diff
@@ -15,28 +42,18 @@ find_data_to_file_info :: proc(base_path: string, d: ^win32.WIN32_FIND_DATAW, al
...
 	fi.creation_time     = time.unix(0, win32.FILETIME_as_unix_nanoseconds(d.ftCreationTime))
 	fi.modification_time = time.unix(0, win32.FILETIME_as_unix_nanoseconds(d.ftLastWriteTime))
 	fi.access_time       = time.unix(0, win32.FILETIME_as_unix_nanoseconds(d.ftLastAccessTime))
 
-	if file_id_info: win32.FILE_ID_INFO; handle != nil && win32.GetFileInformationByHandleEx(handle, .FileIdInfo, &file_id_info, size_of(file_id_info)) {
-		#assert(size_of(fi.inode) == size_of(file_id_info.FileId))
-		#assert(size_of(fi.inode) == 16)
-		runtime.mem_copy_non_overlapping(&fi.inode, &file_id_info.FileId, 16)
-	}
-
```
gingerBill, line 39: Why was this removed?

rmn64k: The handle variable was removed earlier in this function since I removed opening the current entry.

Kelimion:
I think such a speedup is great, but we'll have to have a look at the other platforms and see what info they return from the corresponding call. We may want to adopt this as a complementary fast directory iterator if e.g. Linux returns all the info we'd be dropping but does so in a timeframe comparable with this patch, if that makes sense.

laytan:
Another option is doing a similar thing as `os.process_info` where you give it a bit set of what properties you actually want from the files. Then you can specifically say you want less info to make it faster because it doesn't need to call the functions.

PR #6969, opened 2026-07-07
fix: delete FilePathList.capacity field from the raylib bindings

yuraiz:
Now layout FilePathList should match the c one. I don't know how did that even work for so long

Screenshot 2026-07-07 at 12 58 23


Kelimion:
These APIs were added in raylib 6.0, which we merged only recently. You're quite probably the first to have run into this. Thanks for the fix.

PR #6970, opened 2026-07-07
Fix some wrong bindings for new Raylib 6 model animations

JonathanTron:
Raylib 6 changed how animations are handled and thus `Mesh`, `Model`, `ModelAnimation` structs changed and a new `ModelSkeleton` type was introduced.


Kelimion:
I'm already reworking the complete raylib 6.0 bindings right now, going through them line by line.

Thanks, though.

JonathanTron:
Thanks @Kelimion

PR #6973, opened 2026-07-07
🐛 Prevent git output with signature information

drepper:
When users have something like

[log]
        showSignature = true

in their .gitconfig files, invocations of the log or show git sub-command emit additional information about signatures.  This additional output disturbs the generation of the GIT_SHA and GIT_DATE values for the compiler command line.  The additional text is copied verbatim into the file in unexpected places.  The result is a parse error when invocing the compiler.

To fix it always suppress the output of the signature information.  This has no effects when the setting is disabled anyway.

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -21,7 +21,7 @@ if "%VSCMD_ARG_TGT_ARCH%" neq "x64" (
 
 where /Q git.exe || goto skip_git_hash
 if not exist .git\ goto skip_git_hash
-for /f "tokens=1,2" %%i IN ('git show "--pretty=%%cd %%h" "--date=format:%%Y-%%m-%%d" --no-patch --no-notes HEAD') do (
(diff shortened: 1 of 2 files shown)


Kelimion:
For Windows, the git command needs to be this to avoid the "git: 'false' is not a git command. See 'git --help'." error.
```bat
for /f "tokens=1,2" %%i IN ('git -c "log.showSignature=false" show "--pretty=%%cd %%h" "--date=format:%%Y-%%m-%%d" --no-patch --no-notes HEAD') do (
	set CURR_DATE_TIME=%%i
	set GIT_SHA=%%j
)
```

(The other platforms are spurious CI hardware failures.)

Kelimion:
Thanks for the head's up on this, Ulrich. I've merged this and added the quotes to `build.bat`.

PR #6977, opened 2026-07-08
Support JSON marshalling endian-specific types

rats159:
The unmarshaller already handles endian-specific types for integers and floats, and the marshalling code uses the `cast_any_int_to_u128` proc for integer marshalling, which already includes handling for endian specific types (and uintptr). All this PR does is add additional cases for logic that was already mostly implemented

### DIFF
--- a/core/encoding/json/marshal.odin
+++ b/core/encoding/json/marshal.odin
@@ -187,10 +187,10 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err: 
 		// allow uints to be printed as hex
 		if opt.write_uint_as_hex && (opt.spec == .JSON5 || opt.spec == .MJSON) {
 			switch i in a {
-			case u8, u16,   u32,   u64,   u128, 
-			         u16be, u32be, u64be, u128be, 
-					 u16le, u32le, u64le, u128le,
-					 uintptr:
+			case u8, u16, u32, u64, u128, 
+				 u16be, u32be, u64be, u128be, 
+				 u16le, u32le, u64le, u128le,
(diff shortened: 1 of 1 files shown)


core/encoding/json/marshal.odin
```diff
@@ -187,7 +187,10 @@ marshal_to_writer :: proc(w: io.Writer, v: any, opt: ^Marshal_Options) -> (err:
 		// allow uints to be printed as hex
 		if opt.write_uint_as_hex && (opt.spec == .JSON5 || opt.spec == .MJSON) {
 			switch i in a {
-			case u8, u16, u32, u64, u128:
+			case u8, u16,   u32,   u64,   u128, 
+			         u16be, u32be, u64be, u128be, 
+					 u16le, u32le, u64le, u128le,
+					 uintptr:
```
gingerBill: This formatting is just wildly off here. Make sure you align with spaces but indent with tabs.

rats159: What would be the preferred way to do it? Is the issue just the consecutive spaces instead of tabs, or should the extra indent to align the integer widths be removed altogether? Sorry I missed it, `-vet-tabs -strict-style` didn't complain about it.

gingerBill:
```odin
			case u8, u16, u32, u64, u128, 
			     u16be, u32be, u64be, u128be, 
			     u16le, u32le, u64le, u128le,
			     uintptr:
```

rats159: Done, thanks

gingerBill: I'll merge it and fix the formatting problems still 😄

PR #6978, opened 2026-07-08
vendor:box3d: fix ShapeDef layout by adding missing name field

SwiftBike3317:
# vendor:box3d: fix ShapeDef layout by adding missing name field

## Summary

`vendor:box3d`'s `ShapeDef` is missing the first field from Box3D's C
definition:

```c
const char* name;
```

The generated Odin binding represents this field as:

```odin
name: cstring,
```

Because `ShapeDef` is passed to C by pointer, the missing 8-byte pointer
changes the layout of every following field. Box3D then reads incorrect
values from the structure.

This PR adds the missing `name: cstring` field to `ShapeDef` so that the
Odin structure layout matches the C definition.

## Reproduction

```odin
package main

import b3 "vendor:box3d"

main :: proc() {
    worldDef := b3.DefaultWorldDef()
    worldId := b3.CreateWorld(worldDef)

    bodyDef := b3.DefaultBodyDef()
    bodyId := b3.CreateBody(worldId, bodyDef)

    hull := b3.MakeBoxHull(50.0, 10.0, 50.0)

    shapeDef := b3.DefaultShapeDef()

    b3.CreateHullShape(bodyId, shapeDef, &hull.base)
}
```

## Actual result

```
BOX3D ASSERTION: def->internalValue == B3_SECRET_COOKIE,
/Users/defton/box3d/src/shape.c, line 263

zsh: trace trap  odin run .
```

## Investigation

The current `vendor:box3d` binding contains:

```odin
ShapeDef :: struct {
    userData: rawptr,
    materials: [^]SurfaceMaterial `fmt:"v,materialCount"`,
    materialCount: c.int,
    baseMaterial: SurfaceMaterial,
    density: f32,
    explosionScale: f32,
    filter: Filter,
    enableCustomFiltering: bool,
    isSensor: bool,
    enableSensorEvents: bool,
    enableContactEvents: bool,
    enableHitEvents: bool,
    enablePreSolveEvents: bool,
    invokeContactCreation: bool,
    updateBodyMass: bool,
    internalValue: c.int,
}
```

Regenerating bindings directly from Box3D's headers using
`odin-c-bindgen` produces:

```odin
ShapeDef :: struct {
    name: cstring,
    userData: rawptr,
    materials: ^SurfaceMaterial,
    materialCount: i32,
    baseMaterial: SurfaceMaterial,
    density: f32,
    explosionScale: f32,
    filter: Filter,
    enableCustomFiltering: bool,
    isSensor: bool,
    enableSensorEvents: bool,
    enableContactEvents: bool,
    enableHitEvents: bool,
    enablePreSolveEvents: bool,
    invokeContactCreation: bool,
    updateBodyMass: bool,
    internalValue: i32,
}
```

The regenerated bindings work correctly with the same `box3d.a` binary.
The library binary was verified to be identical by MD5, so this is not a
library version mismatch.

## Fix

Add the missing field as the first member of `ShapeDef`:

```odin
ShapeDef :: struct {
    /// Optional shape name for debugging
    name: cstring,

    userData: rawptr,
    ...
}
```

This restores the correct C ABI layout.

### DIFF
--- a/vendor/box3d/box3d_types.odin
+++ b/vendor/box3d/box3d_types.odin
@@ -507,6 +507,9 @@ ShapeType :: enum c.int {
 // Used to create a shape
 // @ingroup shape
 ShapeDef :: struct {
+	/// Optional shape name for debugging
+	name:                  cstring,
+
 	// Use this to store application specific shape data.
 	userData: rawptr,
 


gingerBill:
Why are you adding a field which does not even exist in the C headers that are used to build the library?

Where is this field that you are talking about?

https://github.com/odin-lang/Odin/blob/master/vendor/box3d/src/include/box3d/types.h#L456-L513

```c
typedef struct b3ShapeDef
{
	/// Use this to store application specific shape data.
	void* userData;

	/// Surface material used on mesh shapes per triangle. Ignored for convex shapes. Ignored for compound shapes.
	b3SurfaceMaterial* materials;

	/// Surface material count.
	int materialCount;

	/// The base surface material. Ignored for compound shapes.
	b3SurfaceMaterial baseMaterial;

	/// The density, usually in kg/m^3.
	float density;

	/// Explosion scale for b3World_Explode. non-dimensional
	float explosionScale;

	/// Contact filtering data.
	b3Filter filter;

	/// Enable custom filtering. Only one of the two shapes needs to enable custom filtering. See b3WorldDef.
	bool enableCustomFiltering;

	/// A sensor shape generates overlap events but never generates a collision response.
	/// Sensors do not have continuous collision. Instead, use a ray or shape cast for those scenarios.
	/// Sensors still contribute to the body mass if they have non-zero density.
	/// @note Sensor events are disabled by default.
	/// @see enableSensorEvents
	bool isSensor;

	/// Enable sensor events for this shape. This applies to sensors and non-sensors. False by default, even for sensors.
	bool enableSensorEvents;

	/// Enable contact events for this shape. Only applies to kinematic and dynamic bodies. Ignored for sensors. False by default.
	bool enableContactEvents;

	/// Enable hit events for this shape. Only applies to kinematic and dynamic bodies. Ignored for sensors. False by default.
	bool enableHitEvents;

	/// Enable pre-solve contact events for this shape. Only applies to dynamic bodies. These are expensive
	///	and must be carefully handled due to multithreading. Ignored for sensors.
	bool enablePreSolveEvents;

	/// When shapes are created they will scan the environment for collision the next time step. This can significantly slow down
	/// static body creation when there are many static shapes.
	/// This is flag is ignored for dynamic and kinematic shapes which always invoke contact creation.
	bool invokeContactCreation;

	/// Should the body update the mass properties when this shape is created. Default is true.
	bool updateBodyMass;

	/// Used internally to detect a valid definition. DO NOT SET.
	int internalValue;

} b3ShapeDef;
```

gingerBill:
I am closing this PR because the source code needs to be updated for us too, not just fixed for you end.

SwiftBike3317:
@gingerBill  The link you posted points to the vendored copy of Box3D inside the Odin repository.

My comparison was against the upstream Box3D headers:

https://github.com/erincatto/box3d/blob/ef8ef0187a6fb7d93fc847872a538096b4a5833d/include/box3d/types.h#L456

There, `b3ShapeDef` begins with:

```c
/// Optional shape name for debugging
const char* name;
```

I regenerated the bindings directly from those upstream headers using `odin-c-bindgen`, which also produced the `name: cstring` field.

Could you clarify which upstream Box3D revision the vendored copy was generated from? I'm trying to understand why the vendored headers differ from the current upstream definition.

PR #6988, opened 2026-07-09
core/rbtree: match find_or_insert semantics with AVL

antisaling:
I know the docs explained it but this was a poor choice and very confusing.


Kelimion:
You're right.

I'll merge this and then add `upsert` to preserve the current semantics of inserting, or updating if already present.

PR #6999, opened 2026-07-10
Allow custom `-collection`s to be used for `foreign import`s.

callframe:
I noticed that foreign imports with absolute paths weren't working on Linux. After looking through the implementation, I found that absolute foreign imports were only permitted on Windows.

This PR changes that behavior so that foreign import now accepts absolute paths on both Windows and POSIX platforms. It also allows Windows-style absolute paths to be recognized from POSIX.

While investigating how absolute paths are detected, I came across what appears to be a bug in gb.h, which this PR also fixes.

I'm open to discussion if there's a reason the previous behavior was intentional.

### DIFF
--- a/src/gb/gb.h
+++ b/src/gb/gb.h
@@ -5613,7 +5613,7 @@ gb_inline b32 gb_path_is_absolute(char const *path) {
 	b32 result = false;
 	GB_ASSERT(path != NULL);
 #if defined(GB_SYSTEM_WINDOWS)
-	result == (gb_strlen(path) > 2) &&
+	result = (gb_strlen(path) > 2) &&
 	          gb_char_is_alpha(path[0]) &&
 	          (path[1] == ':' && path[2] == GB_PATH_SEPARATOR);
 #else
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -648,6 +648,9 @@ try_cross_linking:;
 							//                local to the executable (unless the system collection is used, in which case we search
 							//                the system library paths for the library file).
 							if (string_ends_with(lib, str_lit(".a")) || string_ends_with(lib, str_lit(".o")) || string_ends_with(lib, str_lit(".so")) || string_contains_string(lib, str_lit(".so."))) {
+								if (lib.len > 0 && lib[0] == '/') {
+									lib = substring(lib, 1, lib.len);
+								}
 								lib_str = gb_string_append_fmt(lib_str, " -l:\"%.*s\" ", LIT(lib));
 							} else {
 								// dynamic or static system lib, just link regularly searching system library paths
--- a/src/linker.cpp
+++ b/src/linker.cpp
@@ -648,9 +648,6 @@ try_cross_linking:;
 							//                local to the executable (unless the system collection is used, in which case we search
 							//                the system library paths for the library file).
 							if (string_ends_with(lib, str_lit(".a")) || string_ends_with(lib, str_lit(".o")) || string_ends_with(lib, str_lit(".so")) || string_contains_string(lib, str_lit(".so."))) {
-								if (lib.len > 0 && lib[0] == '/') {
-									lib = substring(lib, 1, lib.len);
-								}
 								lib_str = gb_string_append_fmt(lib_str, " -l:\"%.*s\" ", LIT(lib));
 							} else {
 								// dynamic or static system lib, just link regularly searching system library paths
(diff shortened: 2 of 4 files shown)


laytan:
I don't think we want absolute paths. We want builds to not be tied to absolute paths on somebody's system.

Kelimion:
Agreed. If `foreign import` allows absolute paths on Windows, that's a bug.

backtr4ce:
Windows allows absolute paths but it is not allowed on POSIX also I need it for my archives which are not located relative to the binary. I have a usecase for this feature and it seems to be supported on Windows

laytan:
Well, I'm sorry to say it, but you are probably relying on a bug.

backtr4ce:
Which one?

laytan:
Absolute paths being allowed in foreign imports on windows.

Kelimion:
That Windows accepts absolute paths in this place is a bug. We shouldn't be replicating that bug on other platforms.
Hardcoding paths in source files isn't a good idea.

I am open to extending the `-collection:mylib=path_to_mylib` system to foreign imports, so instead of e.g.
```odin
foreign import lib {
	"windows/raylibdll.lib" when RAYLIB_SHARED else "windows/raylib.lib" ,
	"system:Winmm.lib",
	"system:Gdi32.lib",
	"system:User32.lib",
	"system:Shell32.lib",
}
```
Where `system:` is the special prefix for system libraries, then instead of `windows/raylibdll.lib` relative to the package, this could then be done as:
```odin
foreign import lib {
	"mylib:windows/raylibdll.lib" when MY_LIB_SHARED else "mylib:windows/raylib.lib",
}
```

That's similar to the way `import "core:fmt"` works, and still allows someone to place their libraries somewhere not immediately relative to the package source.

@gingerBill, what's your take?

backtr4ce:
Whether the Windows behavior was intentional or not, I think the capability itself is valuable. Relative paths remain the default, but allowing absolute paths enables hermetic build systems like Bazel, where build artifacts are referenced by their full paths.

backtr4ce:
@Kelimion I think your proposal would fix my issue too

Kelimion:
I believe it will. You could `-collection:foo=/usr/lib/foo` and be on your merry way.

backtr4ce:
I could remove the absolute path feature from both Windows and Posix and try to add the flag if I should.

`-collection:foo=/usr/lib/foo`

is "foo" a directory here?

Kelimion:
> I could remove the absolute path feature from both Windows and Posix and try to add the flag if I should.
> 
> `-collection:foo=/usr/lib/foo`
> 
> is "foo" a directory here?

Correct. It's normally used for e.g. `import "core:fmt"`, where `fmt` is a directory (package) inside of the `core` collection (ODIN_ROOT/core). But you can also supply your own collections for package imports this way.

You can find it listed under `odin help build`.

Reusing that existing custom collection machinery for `foreign import lib { "collection:file.lib" }` seems sensible enough.

If you want to fix that Windows absolute path bug in the process, that'd be grand.

backtr4ce:
@Kelimion Looks like Odin did, by accident, already support this. I added according documentation

backtr4ce:
I'm working on a fix

backtr4ce:
Both test's now get checked successfully locally

Kelimion:
Merged (as a single commit, so as to keep `git bisect` happier).

Good luck with your project.

backtr4ce:
@Kelimion I have a slightly off-topic question regarding command dispatch. From what I saw, Odin uses CreateProcessW on Windows, while the POSIX implementation uses system(). Is there a reason the POSIX implementation doesn't use posix_spawnp() (or a similar direct process API) instead? I was wondering whether there was a design reason for keeping the implementations different.

A1029384756:
@callframe is this for the linker or `run`? `run` was ported away from `system` and uses `_spawnv` on windows and `posix_spawn` on everything else but i didn't touch the linker stuff

Kelimion:
I don't know the why of that. Perhaps `git blame` can identify commit messages or pull requests that illuminate the reasoning behind it.

backtr4ce:
@A1029384756 There is one call to `system()` in `src/main.cpp:155` which is inside of the `system_exec_command_line_app_internal` method which is invoked by `system_exec_command_line_app` which in turn is invoked multiple times across the linker in the main file and bundlers

A1029384756:
yeah, afaik that code dates back quite a long time. it was probably just the first implementation that got made. `run_subprocess` (added by the recent `odin run` changes) would require some modifications to the surrounding command builders but would work just fine as well

PR #7001, opened 2026-07-10
`core:flags` Add support for "short" subtag in Unix style flag parsing

prnvbn:
adds support for "short" subtag in core:flags for `.Unix`

for the following struct with `.Unix` style
```odin
Options :: struct {
	version: bool `args:"short=v" usage:"show version number"`,
}
```
`--version` and `-v` are both valid

All changes to the `.Unix` style:

- Long flags use --flag strictly; -flag is not supported.
- Short flags can be declared with args:"short=f" and used as -f.
- Boolean short flags can be bundled, so -abc is the same as -a -b -c.


prnvbn:
@FourteenBrush pls review if you can

FourteenBrush:
> @FourteenBrush pls review if you can

Tested this with some configurations and seems to be working fine.

So just to get a good understanding of what you are trying to achieve, this seems to address the fact that the .Unix parsing style parses flags in the form of --flag or -flag, but that there is no possibility to use a short hand syntax -f as of right now.

So this pr seems to give us three possible ways of parsing .Unix style flags now: `--flag`, `-flag` and `-f` (if short=f wouldve been set). I am not quite sure if we actually want to keep the second case (-flag) when overriding the short option, this currently does not show up in the --help command, and I personally havent seen much unix programs provide three possible ways to supply one flag. This may or may not be desired, I'd probably want someone elses opinion on this as well.

prnvbn:
> So just to get a good understanding of what you are trying to achieve, this seems to address the fact that the .Unix parsing style parses flags in the form of --flag or -flag, but that there is no possibility to use a short hand syntax -f as of right now.

yep.

> So this pr seems to give us three possible ways of parsing .Unix style flags now: --flag, -flag and -f (if short=f wouldve been set). I am not quite sure if we actually want to keep the second case (-flag) when overriding the short option, this currently does not show up in the --help command, and I personally havent seen much unix programs provide three possible ways to supply one flag. This may or may not be desired, I'd probably want someone elses opinion on this as well.

I dislike the `-flag` in .Unix flags as well and am in favor of removing it! The only reason I didn't is because of backward compatibility. Can remove it if the maintainers agree.

rm'ing  support for `-flag`  will also allow the .Unix style to support flag clustering (you can do `ls -a -l` and `ls -al`)

prnvbn:
also, is there an `odinfmt.json` file? the default `ols` formatter keeps formatting unrelated parts of the code as they use some different formatting

FourteenBrush:
@prnvbn There is not, just ensure your code is formatted correctly already, and there's no need to rely on odinfmt at all :)

prnvbn:
also @FourteenBrush you in the Odin discord? the link in the readme does not work

FourteenBrush:
@prnvbn No I am not, also what link in the readme, the discord link in the Odin readme?

prnvbn:
@FourteenBrush yes. This one - https://github.com/odin-lang/Odin#odin-discord

prnvbn:
@gingerBill do you have any thoughts on this?

prnvbn:
seems like this PR will take a while, created this in the meantime - https://github.com/prnvbn/pflags


mainly for use in my tiny cli - https://github.com/prnvbn/idgen

MightyChubz:
Seconding this. It matches more closely the GNU style of argument parsing for CLI applications. I've used Odin arguments and Unix interchangeably depending on the project for a while now, but still largely prefer Unix due to how flexible it is in general for CLI applications (what I work on primarily right now). Odin flags lacking shorthand arguments and bundling of flags has been something I've wanted changed for a while now. It may take a while for Ginger Bill to review this, mainly just because there are higher-priority items on the docket right now, I imagine, but I would like to see this merged in the future.

As for functionality mentioned in the earlier comments, I would opt for removing the `-flag` for `.Unix` and stick with it. It keeps the `.Odin` argument style and `.Unix` style more separated and distinct. Makes it clear to me what type of argument parsing you're getting, rather than having `.Unix` mix with Odin's argument style.

Like what I see so far! :+1:

prnvbn:
>  I would opt for removing the -flag for .Unix and stick with it. It keeps the .Odin argument style and .Unix style more separated and distinct.

agreed! and has been done in the PR :)

PR #7003, opened 2026-07-10
[core:unicode/utf8]: Fix truncated text for grapheme clusters

### LINKED ISSUE #6912: core:unicode/utf8: decode_grapheme_iterate returns truncated/invalid text (slices by display width, not bytes)
## Context

The bug is in pure Odin code in core:unicode/utf8 (grapheme.odin) and is platform-independent. It reproduces on the version below and on current master (the offending line is identical there). It affects any non-ASCII grapheme cluster; ASCII-only input happens to work because 1 byte == 1 display cell.

- Operating System & Odin Version: macOS Tahoe 26.5.1 (arm64, Apple M3 Pro) — Odin dev-2026-06:285f6d87b (also confirmed on current master)
- Odin report output:
	OS:      macOS Tahoe 26.5.1 (build 25F80, kernel 25.5.0)
	CPU:     Apple M3 Pro
	RAM:     18432 MiB
	Backend: LLVM 22.1.7

## Expected Behavior

The first return value (text) of utf8.decode_grapheme_iterate should be the full byte span of the current grapheme cluster. For the US flag 🇺🇸 (U+1F1FA U+1F1F8, 8 bytes, 2 runes, 1 cluster), it should be:

text="🇺🇸"  len=8 bytes  width=1

## Current Behavior

Text is sliced using grapheme.width (a monospace display-cell count) as if it were a byte length, so it returns a truncated and frequently invalid UTF-8 fragment. For 🇺🇸:

text="\xf0"  len=1 bytes  width=1

The companion fields byte_index and rune_index are correct; only text is wrong.

The defect is in decode_grapheme_iterate (core/unicode/utf8/grapheme.odin):
text = it.str[byte_index:][:grapheme.width]

Two compounding problems:

1. Wrong unit: grapheme.width comes from normalized_east_asian_width (display cells), not bytes. Cells != bytes (a flag is 8 bytes / 1–2 cells; a CJK ideograph is 3 bytes / 2 cells; e+combining-acute is 3 bytes / 1 cell).
2. Premature emission: the cluster is emitted at its first rune (it.grapheme_count > it.last_grapheme_count), before trailing Extend/ZWJ/RI runes are consumed, so the cluster's full byte length isn't known at that point.

This goes unnoticed because the sibling APIs grapheme_count and decode_grapheme_clusters discard text, so there is no internal test coverage; only external users of the iterator hit it.

Suggested fix direction: slice by byte range, not width — emit each cluster spanning [start_byte, next_boundary_byte). In practice, defer emission by one cluster (remember the pending start offset, emit it.str[start:byte_index] when the next boundary is found, and flush the final cluster at end-of-string). byte_index already provides the needed offsets.

Related: possibly shares a root area with #5939 (incorrect width for complex emoji). Fixing width alone will not fix text (different units), but they touch the same code.

## Steps to Reproduce

1. Save the following as grapheme_bug.odin:
```odin
package main
import "core:fmt"
import "core:unicode/utf8"

main :: proc() {
    s := "\U0001F1FA\U0001F1F8" // 🇺🇸 — 8 bytes, 2 runes, 1 grapheme cluster
    it := utf8.decode_grapheme_iterator_make(s)
    for text, g in utf8.decode_grapheme_iterate(&it) {
        fmt.printf("text=%q  len=%d bytes  width=%d\n", text, len(text), g.width)
    }
}
```
2. Run odin run grapheme_bug.odin -file
3. Observe that text is a single, incomplete byte ("\xf0") instead of the full "🇺🇸"

## Failure Logs
  Actual:
    text="\xf0"  len=1 bytes  width=1
  Expected:
    text="🇺🇸"  len=8 bytes  width=1

## Additional cases affected by the same root cause (each returns a truncated text):
- Combining mark: "e\u0301" (3 bytes, 1 cluster) → returns "e".
- CJK ideograph (3 bytes, width 2) → returns 2 bytes (invalid UTF-8).
- Emoji ZWJ sequence 👨‍👩‍👧‍👦 and skin-tone modifier 👍🏽 → truncated.

jpgleeson:
When iterating grapheme clusters, the iterator sliced each cluster's text using the display width as the byte count. We emitted that as the first rune before we knew the actual byte length of the cluster. Multi-byte/multi-rune clusters returned truncated values. Example from the bug report is the flag emoji returning just the first byte of the 8 byte sequence. CRLF returned an empty string.

Now we only emit clusters when we encounter a new boundary, and flush the in-flight cluster when we have consumed all input. `byte_index`, `rune_index`, and `width` are unaffected.

Test cases are taken from the original bug report, plus plain ASCII, CRLF, and an adjacent flag pair to verify the boundary between the regional indicator pairs gives correct byte spans rather than slicing to the end of input.

Fixes #6912

### DIFF
--- a/core/unicode/utf8/grapheme.odin
+++ b/core/unicode/utf8/grapheme.odin
@@ -54,6 +54,9 @@ Grapheme_Iterator :: struct {
 
 	current_sequence:           Grapheme_Cluster_Sequence,
 	continue_sequence:          bool,
+
+	current_grapheme:           Grapheme,
+	continue_grapheme:          bool,
 }
 
 
@@ -147,13 +150,13 @@ decode_grapheme_iterate :: proc(it: ^Grapheme_Iterator) -> (text: string, graphe
 
 			if it.grapheme_count > it.last_grapheme_count {
 				it.width += normalized_east_asian_width(this_rune)
-				grapheme = Grapheme{
-					byte_index,
-					it.rune_count,
-					it.width - it.last_width,
+				if it.continue_grapheme {
+					grapheme = it.current_grapheme
+					text = it.str[it.current_grapheme.byte_index:byte_index]
+					ok = true
 				}
-				text = it.str[byte_index:][:grapheme.width]
-				ok = true
+				it.current_grapheme = Grapheme{byte_index, it.rune_count, it.width - it.last_width}
+				it.continue_grapheme = true
 
 
 				it.last_grapheme_count = it.grapheme_count
@@ -383,6 +386,15 @@ decode_grapheme_iterate :: proc(it: ^Grapheme_Iterator) -> (text: string, graphe
 		//
 		// GB999: Any ÷ Any
 		it.grapheme_count += 1
+	}
+
+	// Flush the remaining grapheme - the loop only flushes when
+	// a new grapheme is encountered.
+	if !ok && it.continue_grapheme {
+		grapheme = it.current_grapheme
+		text = it.str[it.current_grapheme.byte_index:]
+		ok = true
+		it.continue_grapheme = false
 	}
 
 	return
--- a/tests/core/unicode/test_core_unicode.odin
+++ b/tests/core/unicode/test_core_unicode.odin
@@ -9,6 +9,11 @@ Test_Case :: struct {
 	expected_clusters: int,
 }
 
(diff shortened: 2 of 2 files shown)


jpgleeson:
Missed a comma in the struct definition. 🤦

Kelimion:
Thanks!

PR #7015, opened 2026-07-12
Fix arm64 ABI: '#simd' vectors and vector aggregates not passed in registers (fixes #7010)

### LINKED ISSUE #7010: macOS: #simd parameters in proc "c" are passed in general-purpose registers instead of SIMD registers
## Context

Odin:    dev-2026-07:301c287de
	CPU:     Apple M1 Max
	RAM:     32768 MiB
	Backend: LLVM 22.1.8

## Expected Behavior

Able to pass SIMD parameters to C or Objective-C code following the platform's vector calling convention, or a compiler error if not supported.

## Current Behavior

incorrect values which may cause a crash

While experimenting with an Odin port of Metal by Tutorials demos, I came across an issue when calling `[MDLMesh initSphereWithExtent:segments:inwardNormals:geometryType:allocator:]` which has SIMD parameters.

### Steps to Reproduce

I want to be upfront. I've been having Fable 5 run ahead of me to implement Metal in Odin to uncover what challenges I'll face. I won't be offended if you close this issue as LLM reported or whatever, but based on the crashes I've seen and what I've read, it does seem like it could be a legitimate issue that I shouldn't just keep to myself.

The following test case was written by Fable -- tested on an M1 (not tested on an Intel Mac):

### #simd parameters in proc "c" don't follow the AArch64 C calling convention (passed in general-purpose registers instead of SIMD registers)

Passing a #simd[N]T value to a C function through a foreign procedure (or intrinsics.objc_send) mis-passes the argument on darwin/arm64. AAPCS64 classifies short vectors (8/16 bytes) as SIMD/FP-register arguments (v0–v7), and clang passes/expects simd_float4, vector_float3, simd_uint2, etc. accordingly — but Odin passes #simd parameters in general-purpose registers, so the callee reads garbage lanes and all subsequent arguments shift by one or more registers.

```c
#include <simd/simd.h>
float probe_sum(simd_float4 v, simd_uint2 u) {
    return v.x + v.y * 10 + v.z * 100 + (float)u.x * 1000 + (float)u.y * 10000;
}
```

```odin
package cabi_probe

import "core:fmt"

foreign import ctest "ctest.o"
foreign ctest {
	probe_sum :: proc "c" (v: #simd[4]f32, u: #simd[2]u32) -> f32 ---
}

main :: proc() {
	got := probe_sum({1, 2, 3, 0}, {4, 5})
	fmt.println("got:", got, "expected: 54321")
}
```

```sh
clang -c ctest.c -o ctest.o
odin run .
```

Expected: got: 54321
Actual: got: 1.1802771e+13

This makes every Apple API with simd types in its signature uncallable from Odin — for example all of Model I/O's procedural initializers (`-[MDLMesh initSphereWithExtent:segments:inwardNormals:geometryType:allocator:]`), parts of GameplayKit, SceneKit, and MetalPerformanceShaders. Called via intrinsics.objc_send, the argument shift is observable as a crash inside objc_retain where the pointer being retained is the bit pattern of the preceding vector argument (e.g. 0x1e0000001e for a #simd[2]u32{30, 30}). Given the Odin 2027 goal of bringing "the full core Objective-C library natively," a correct vector ABI (or a loud diagnostic) seems load-bearing.

Note: only parameters were tested; vector returns through proc "c" are untested and may have the same issue.

### Failure Logs

Minimal C repro (odin run . after clang -c ctest.c -o ctest.o):

```
got: 1.1802771e+13 expected: 54321
```

Real-world manifestation — calling `-[MDLMesh initSphereWithExtent:segments:inwardNormals:geometryType:allocator:]` via intrinsics.objc_send with extent: #simd[4]f32{0.75, 0.75, 0.75, 0} and segments: #simd[2]u32{30, 30}. The process crashes inside the Objective-C runtime, which attempts to retain an "object pointer" whose value is exactly the bit pattern of the segments vector (0x0000001e_0000001e = {30, 30}), showing that the vector landed in a general-purpose register and shifted the following arguments:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x1e0000001e)
    frame #0: 0x0000000183d5f914 libobjc.A.dylib`objc_retain + 8
libobjc.A.dylib`objc_retain:
->  0x183d5f914 <+8>:  ldr    x16, [x0]
    0x183d5f918 <+12>: and    x17, x16, #0x7ffffffffff8
    0x183d5f91c <+16>: ldr    x17, [x17, #0x20]
```

BlackGoku36:
Fixes #7010.

Fixed passing and returning #simd vector through function with "c" calling convention on arm64 devices.

Tested on Apple device with macOS.

Note: I haven't added test file for this in `run.bat` as I don't have windows device to test build.

Let me know if you have any need any changes.

### DIFF
--- a/src/llvm_abi.cpp
+++ b/src/llvm_abi.cpp
@@ -1147,6 +1147,10 @@ namespace lbAbiArm64 {
 		case LLVMDoubleTypeKind:
 		case LLVMPointerTypeKind:
 			return true;
+		case LLVMVectorTypeKind:{
+				i32 sz = lb_sizeof(type);
+				return sz == 8 || sz == 16;
+			}
 		}
 		return false;
 	}
@@ -1228,6 +1232,15 @@ namespace lbAbiArm64 {
 			if (base_type_) *base_type_ = type;
 			if (member_count_) *member_count_ = 1;
 			return true;
+		case LLVMVectorTypeKind:{
+			i64 sz = lb_sizeof(type);
+			if (sz == 8 || sz == 16){
+				if (base_type_) *base_type_ = type;
+				if (member_count_) *member_count_ = 1;
+				return true;
+			}
+			return false;
+		}
 		case LLVMArrayTypeKind:
 			return is_homogenous_array(c, type, base_type_, member_count_);
 		case LLVMStructTypeKind:
--- a/src/llvm_abi.cpp
+++ b/src/llvm_abi.cpp
@@ -1148,7 +1148,7 @@ namespace lbAbiArm64 {
 		case LLVMPointerTypeKind:
 			return true;
 		case LLVMVectorTypeKind:{
-				i32 sz = lb_sizeof(type);
+				i64 sz = lb_sizeof(type);
 				return sz == 8 || sz == 16;
 			}
 		}
--- a/tests/issues/run.sh
+++ b/tests/issues/run.sh
@@ -81,6 +81,10 @@ else
 fi
 $ODIN check ../test_issue_6979.odin -no-entry-point $COMMON
 
+
+clang -c ../test_issue_7010.c -o test_issue_7010_c.o
+$ODIN test ../test_issue_7010.odin $COMMON
+
 set +x
 
 popd
--- a/tests/issues/test_issue_7010.c
+++ b/tests/issues/test_issue_7010.c
@@ -1,0 +1,51 @@
+// Tests issue: https://github.com/odin-lang/Odin/issues/7010
+
+// #include <simd/types.h>
+// #ifdef __APPLE__
+// #include <simd/simd.h>
+// #else
+typedef __attribute__((__ext_vector_type__(2))) float simd_float2;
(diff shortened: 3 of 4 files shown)


nathany:
@BlackGoku36 Thanks for looking into this issue.

I built Odin with your fix on my M1 Mac and Fable confirmed that both the small test case from the issue and usage in MDLMesh sphereWithExtent worked.

> fix PR #7015 by BlackGoku36 arrived within a day (params + returns + vector
aggregates). Nathan built the patched compiler in his Odin checkout; on 2026-07-12 both
repros verified fixed (C control exact, MDLMesh sphere via direct objc_send works —
992 verts, extent intact)

```
=== C control probe:
got: 54321 expected: 54321

=== MDLMesh objc_send probe (previously SIGBUS):
device: Apple M1 Max
vertexCount: 992
max |position coord|: 0.75 
```

PR #6838, opened 2026-06-14
core:encoding/asn1 (strict DER) + core:crypto/x509 (X.509 v3)

kalsprite:
core:encoding/asn1: strict DER for the PKIX subset.
  - Reader: a Cursor; read_* return views (only oid_* allocate). Minimal lengths/integers, definite form enforced.
  - Writer: a declarative Value tree -> encode (no alloc, into a caller buffer) or marshal (single alloc).

  core:crypto/x509: v3 parse, signature verification, chain validation; zero-copy where possible.
  - parse (raw views preserved for verify/pinning) + verify_signature / verify_chain (RFC 5280 §6: validity, name chaining,
  BC/keyCertSign/pathLen, EKU nesting; fails closed on unhandled critical exts + name constraints; DoS-bounded) + verify_hostname (RFC 6125, SAN only).
  - RSA and P-521 -> .Unsupported_Algorithm.

  Testing: unit + coverage-guided / mutational / bit-flip fuzzers (view-bounds invariants under ASan) + OOM sweeps; x509 adversarial path-validation fixtures (cross-signed cycle, pathlen, EKU, expired, not-a-CA, NC fail-closed), developed against x509-limbo.

Notes:

  - RSA, P-521 Error Out
  - Name Constraints fail closed (refused / not decoded). Implementing Name Constraints is intended as a future pr.

  @Yawning

PR #7022, opened 2026-07-13
win32: Add bindings to query the Windows certificate store

Kelimion:
Preparation for TLS support: win32 bindings needed to query the OS certificate store.


Kelimion:
The cross platform cert store will rely on #6838, and uses these bindings for Windows. Bindings for macOS forthcoming in another PR, e.g. `SecTrustSettingsCopyCertificates`, `SecCertificateCopyData`, etc.

Linux doesn't need additional bindings.

PR #7032, opened 2026-07-14
Improve error message for never executed statements after a divergin proc call

Creativty:
Statements that are never executed after a diverging procedure calls produce the error at the proc call site, which sometimes misleads the user to thinking the error being in the call itself

Example:
```odin
main :: proc() {
  os.exit(0)
  return
}
```
The mistake is obviously the useless `return` but the current error message would have you thinking otherwise.

Current message: `Statements after a diverging procedure call are never executed` pointing at the **call_expr**
Proposed message: `Statements after divergin procedure call "<call_expr>" are never executed` pointing at the **stmt**

### DIFF
--- a/src/check_stmt.cpp
+++ b/src/check_stmt.cpp
@@ -121,7 +121,9 @@ gb_internal void check_stmt_list(CheckerContext *ctx, Slice<Ast *> const &stmts,
 
 			case Ast_ExprStmt:
 				if (is_diverging_stmt(n)) {
-					error(n, "Statements after a diverging procedure call are never executed");
+					Ast *n_after = stmts[i + 1];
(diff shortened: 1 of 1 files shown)


gingerBill:
I am not sure if this is an improvement or not.

Creativty:
> I am not sure if this is an improvement or not.

I am also hesitant to consider this a fully fledged improvement, but the problem is there. as this has stumped a beginner user on discord (and to no fault of their own in my opinion).
A proper error report in my view would be keeping the same location for the error, while printing the remainder of the block to demonstrate both the diverging call, and the statement**s** that are not going to be executed. This way the user gets to choose which to interpret as the offender.
Since the current scheme assumes that the issue lies in the call, and this PR assumes that the issues is the statements that follow it.

PR #7034, opened 2026-07-14
replace makefiles in vendor with shell scripts; update ci.

quarkstuff:
Of the three Makefiles in `vendor` (`cgltf`, `miniaudio`, and `stb`), 2 of them (`cgltf` and `stb`) rely on GNU extensions.  This means users on BSD systems, or Linux users who prefer to use a different make implementation have to endure a bit of friction to build those dependencies.  It would also improve portability to not rely on a specific variant of Make.

The new version uses shell scripts since I didn't want to remove the condition.  IMO, using Make here (basically as a command runner) was overkill anyway.

The new scripts have been tested using a posix shell locally.


quarkstuff:
:skull:

Kelimion:
Great idea, to be honest. Will merge it after you direct the other platforms using `make` to also use the new scripts.

Kelimion:
New failures make sense. After `cd`, the `./vendor/stb/src/` path prefix for the script is no longer needed, and doesn't exist relative to itself.

Kelimion:
Thanks for this! 🎉 

Going to squash merge to keep breakage to a minimum for `git bisect` purposes.

quarkstuff:
NEVER trust me with CI again omg :sob:

Kelimion:
Perseverance won the day!

PR #7042, opened 2026-07-14
compiler: preserve recursive struct aliases

antisaling:


antisaling:
Drafted until issue fix can be confirmed by @godalming123

PR #7043, opened 2026-07-14
[core:crypto] Add TurboSHAKE

zhibog:
Note:
I omitted the f from the keccakf procedures to denote that this is no longer just used for keccak-f.

I'm unsure about the separate file for the TurboSHAKE init, I can also easily move it into one of the other files.
Just felt like it didn't fit there.

Setting the round start after the init call is on purpose, to defend against someone explicitly setting that to `0` within the internal init, sometime in the future.


Yawning:
Looks good to me.  Separate file is fine, esp if someone decides to do K12 at some point.

PR #7047, opened 2026-07-15
Fixes a couple of bugs in core:math/fixed + compilation for unsigned backing types

corleypc:
The floor() proc in the core:math/fixed package works incorrectly for negative numbers with no fractional part (e.g. -2.0 floors to -3, etc). This was introduced in https://github.com/odin-lang/Odin/commit/7f6f97128492092b82af6298e1c3d863dea07518.

floor() also doesn't compile for the Fixed7_1 type, due to the shifting shenanigans in the negative integer part case (shifting by 1 << (Fraction_Width - 2) errors when Fraction_Width is 1).

Additionally, to_f64() doesn't work properly for the minimum value of the Backing type for signed Backings, it flips the sign of the result. This is also masked by the existing test incorrectly expecting 8.0 instead of -8.0 for a raw value of 128 and Fixed(i8, 4).

Additionally, init_from_f64() does not compile for unsigned Backing types.

This PR reverts floor() to its correct (and simpler) original form, simplifies to_f64() drastically, fixes the bug in the existing test, fixes init_from_f64() compilation for unsigned Backings and adds a bunch of additional tests for floor() and for round tripping through init_from_f64() and to_f64().


corleypc:
The last commit https://github.com/odin-lang/Odin/pull/7047/commits/a4a51eb6731c453104050c60a64ac32b124ce199 also fixes crashes in write() and adds a test to reproduce. write() would crash with either a destination buffer being too short, or when its internal buffer is overflown when using large backings with big Fraction_Width.

It also limits Fraction_Width to widths that actually compile (and work). In particular, full bit width Fraction_Width breaks almost everything for signed backings (e.g. Fixed(i8, 8)). Fraction_Width of zero (e.g. Fixed(i8, 0)) breaks compilation of round(). Fraction_Width = 0 is mostly academic, so I think excluding it is not an issue. Fix for the first would complicate code since it needs type branching in many procedures. Can be done if needed? Note that this is technically a breaking change, if someone happened to only use procedures that did compile (very unlikely).

Lastly, init_from_f64() was doing redundant work, the committed version is equivalent for target range valid values, although the two versions can break differently for invalids. If anything, the committed version has _less_ pieces that can break.

PR #7048, opened 2026-07-15
Fix function names in sdl3_render.odin

ghost:
Fix incorrect function names that caused linking errors when using them.

The reference for these functions are here: [SDL_GetDefaultTextureScaleMode](https://wiki.libsdl.org/SDL3/SDL_GetDefaultTextureScaleMode) and [SDL_SetDefaultTextureScaleMode](https://wiki.libsdl.org/SDL3/SDL_SetDefaultTextureScaleMode)

### DIFF
--- a/vendor/sdl3/sdl3_render.odin
+++ b/vendor/sdl3/sdl3_render.odin
@@ -253,8 +253,8 @@ foreign lib {
 	GetRenderVSync                   :: proc(renderer: ^Renderer, vsync: ^c.int) -> bool ---
 	RenderDebugText                  :: proc(renderer: ^Renderer, x, y: f32, str: cstring) -> bool ---
 	RenderDebugTextFormat            :: proc(renderer: ^Renderer, x, y: f32, fmt: cstring, #c_vararg args: ..any) -> bool ---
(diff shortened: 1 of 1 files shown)


gingerBill:
Huh? Those are real procedures: 
* https://wiki.libsdl.org/SDL3/SDL_SetRenderTextureAddressMode
* https://wiki.libsdl.org/SDL3/SDL_GetRenderTextureAddressMode

They just have the wrong signature. And it also means that the others are missing, not incorrectly named.

ghost:
> Huh? Those are real procedures:

Yes, and they are defined in the sdl3_render.odin file. However,

>     * https://wiki.libsdl.org/SDL3/SDL_SetRenderTextureAddressMode
> 
>     * https://wiki.libsdl.org/SDL3/SDL_GetRenderTextureAddressMode

These are not the names of the functions I changed. They are about a RenderTexture, not the DefaultTexture. These functions are unchanged and do still exist in the commit.

> They just have the wrong signature. And it also means that the others are missing, not incorrectly named.

They are not missing because `SDL_[Get/Set]DefaultTextureAddressMode` doesn't exist in the official docs. What you probably meant was to write `SDL_[Get/Set]DefaultTextureScaleMode` (the parameters corroborate this). 

If I came across as rude in this message, it's not my intent and I'm sorry.

PR #7049, opened 2026-07-15
Expose memory block resizing in core:mem/virtual

godalming123:
See my use case at https://discord.com/channels/568138951836172421/1526888907747102812
I'm quite new to such low-level programming, and it's entirely possible that this change in the standard library is completely unnecessary for my use case.


Kelimion:
After some back and forth internally, we've decided not to adopt this change.

We recommend taking `core:mem/virtual` and copying it into your project tree and adopting those changes locally. As you're writing a custom allocator based on this package, that way is also more robust against future changes to this package.

PR #7068, opened 2026-07-16
Fixes fast_math attribute plumbing to llvm

corleypc:
The (@fast_math) supplied flags get incorrectly passed to LLVM in lb_run_fast_float_math_pass(), because fast_math_flags is tested against ordinals instead of the corresponding bit positions.

Fixing it unmasks another issue. Some of the flags can't be applied to some of the LLVM instructions in the switch [here](https://github.com/odin-lang/Odin/blob/master/src/llvm_backend_opt.cpp?#L81), because they alias other flags. Here is a minimal example (for LLVMUIToFP) of reassoc aliasing nneg and turning unsigned conversions to signed conversions (build with -o:speed AFTER fixing the flags in lb_run_fast_float_math_pass(), otherwise it is essentially passing a random combination of flags):

```odin
package fast_math_test

import "core:fmt"
import "core:os"

@(fast_math={.Allow_Reassoc})
to_f64_fast :: proc(x: u64) -> f64 {
	return f64(x) // emits signed conversion because reassoc aliases nneg
}

main :: proc() {
	x := u64(0x8000_0000_0000_0000) + u64(len(os.args)) // runtime dependent so that conversion isn't optimized out
	fast := #force_no_inline to_f64_fast(x)

	fmt.printfln("reference: %f", f64(x))
	fmt.printfln("fast math: %f", fast)
}
```

To fix this, the commit replaces the explicit switch with a call to LLVMCanValueUseFastMathFlags() which should exclude conversions (and likely include a few other benefiting cases not in the switch) and should be more future proof.

All this is guarded by an LLVM version check in the commit, cause [LLVM 17](https://github.com/llvm/llvm-project/blob/release/17.x/llvm/include/llvm-c/Core.h) doesn't seem to have support for fast math flags.

### DIFF
--- a/src/llvm_backend_opt.cpp
+++ b/src/llvm_backend_opt.cpp
@@ -51,6 +51,8 @@
 **************************************************************************/
 
 gb_internal void lb_run_fast_float_math_pass(lbProcedure *p) {
+#if LLVM_VERSION_MAJOR >= 18
+
 	Entity *e = p->entity;
 	if (e == nullptr) {
 		return;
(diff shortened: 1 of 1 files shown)


corleypc:
The additional instruction types getting the fast_math treatment from the LLVMCanValueUseFastMathFlags() check in the PR appear to be call, select and phi. So with the PR e.g. min(), clamp(), intrinsics.sqrt(), etc benefit from the fast_math flags too.

Here is min() (lowered to a call to the llvm.minnum intrinsic) on x64, o:speed, def target, LLVM22:

no fast_math flags:
```asm
        movaps  %xmm1, %xmm2
        minss   %xmm0, %xmm2
        cmpunordss      %xmm0, %xmm0
        blendvps        %xmm0, %xmm1, %xmm2
        movaps  %xmm2, %xmm0
        retq
```

with @(fast_math={.No_NaNs, .No_Signed_Zeros}):
```asm
        minss   %xmm1, %xmm0
        retq
```


Also, PR now verified working correctly with LLVM 23.

LLVM 23 changes flags handling, flags no longer alias, and LLVMCanValueUseFastMathFlags() correctly re-enables fast_math for conversions on this LLVM version, so the future proof comment seems to be right.

PR #7070, opened 2026-07-17
Fix ENet Buffer data/length ordering issue

raymcveety:
Our team discovered this issue when we were trying to send a UDP broadcast packet using enet_socket_send().

On Linux, the broadcast works as expected, but on Windows the function returns -1.
Calling WSAGetLastError() returns a code of [WSAEFAULT](https://learn.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2) indicating an error with the pointer parameters.

This lead us to inspect the [parameters accepted by enet.socket_send](https://github.com/lsalzman/enet/blob/5a9c537fd464b3c6d3c55e1d3bd47588faf71b42/win32.c#L323) and how they were subsequently transformed/translated into the parameters for WSASendTo where we noticed the ENetBuffer passed in "buffers" as [LPWSABUF](https://github.com/lsalzman/enet/blob/5a9c537fd464b3c6d3c55e1d3bd47588faf71b42/win32.c#L341)

Here's essentially what we were passing to socket_send for the buffer:
```
b := enet.Buffer {
    data = raw_data(bytes),
    dataLength = len(bytes)
}

bytes_sent := enet.socket_send(network.broadcast_sender, &addr, &b, 1)
// below we report error 
``` 

Currently [enet.Buffer is defined in win32.odin](https://github.com/odin-lang/Odin/blob/7c2b21925901d1b125ddc9fb290b6a7e7ed9dd0d/vendor/ENet/win32.odin#L62) with the order (data first):
```
Buffer :: struct {
	data:       rawptr,
	dataLength: uint,
}
```

But [LPWSABUF](https://learn.microsoft.com/en-us/windows/win32/api/ws2def/ns-ws2def-wsabuf) (which is a long pointer to WSABUF) expects the opposite ordering of the data and length (length first):
```
typedef struct _WSABUF {
  ULONG len;
  CHAR  *buf;
} WSABUF, *LPWSABUF;
```

So the crash was due to the length being interpreted as the data/data being interpreted as length.

The fix is simple (and we verified locally it eliminates the WSAEFAULT) which is to reverse the order of the members of the Buffer struct in win32.odin
```
Buffer :: struct {
	dataLength: uint,
	data:       rawptr,
}
```

We verified the forked compiler builds with these changes, our project builds on the forked compiler, and the WSAEFAULT no longer occurs when we call enet.socket_send using the enet.Buffer struct.

### DIFF
--- a/vendor/ENet/win32.odin
+++ b/vendor/ENet/win32.odin
@@ -60,8 +60,8 @@ Socket :: distinct SOCKET
 SOCKET_NULL :: Socket(~uintptr(0))
 
 Buffer :: struct {
-	data:       rawptr,
 	dataLength: uint,
+	data:       rawptr,
 }
 
 SocketSet :: distinct fd_set


vendor/ENet/win32.odin
```diff
@@ -60,8 +60,8 @@ Socket :: distinct SOCKET
 SOCKET_NULL :: Socket(~uintptr(0))
 
 Buffer :: struct {
-	data:       rawptr,
 	dataLength: uint,
+	data:       rawptr,
```
raymcveety, line 64: This is ultimately passed in as [LPWSABUF ](https://github.com/lsalzman/enet/blob/5a9c537fd464b3c6d3c55e1d3bd47588faf71b42/win32.c#L341) which [is expected to be length first](https://learn.microsoft.com/en-us/windows/win32/api/ws2def/ns-ws2def-wsabuf)

raymcveety:
Can I bump this for review? 

I see @FourteenBrush  gave a 👍  but nothing since

PR #7076, opened 2026-07-18
Fix errors in encoding/csv example docs

tarsir:
Hello Odin maintainers! Firstly, thanks for the awesome language - I couldn't have gotten so far in my current game project without Odin :)

I was using the CSV module in a project and had to fix the `reader_init` call from the example docs to what I have in this PR. I also checked if the rest of the docs satisfied my LSP and found the second fix here.

Please let me know if anything needs changed or if these fixes are incorrect!

### DIFF
--- a/core/encoding/csv/doc.odin
+++ b/core/encoding/csv/doc.odin
@@ -49,7 +49,7 @@ Example:
 			fmt.eprintfln("Unable to open file: %v. Error: %v", filename, err)
 			return
 		}
-		csv.reader_init(&r, handle.stream)
+		csv.reader_init(&r, os.to_stream(handle))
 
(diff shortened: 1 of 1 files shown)


core/encoding/csv/doc.odin
```diff
@@ -49,7 +49,7 @@ Example:
 			fmt.eprintfln("Unable to open file: %v. Error: %v", filename, err)
 			return
 		}
-		csv.reader_init(&r, handle.stream)
+		csv.reader_init(&r, os.to_stream(handle))
```
Kelimion, line 52: Right, looks like we forgot to update this when we moved to the new version of `core:os`. Well spotted.

PR #7078, opened 2026-07-18
tokenizer: fix keyword LUT init race

antisaling:


antisaling:
seems unrelated to user problems, I found this out when using ols so might still be relevant for some other report. keeping it drafted until I can find a repro to justify this patch.

PR #7084, opened 2026-07-19
make build scripts portable and pwd independent + fix instructions

antisaling:
Also addressed some linter warnings on the script. it seems that the wasm target was using a Makefile-only `shell` thing.

### DIFF
--- a/vendor/stb/src/build_stb.sh
+++ b/vendor/stb/src/build_stb.sh
@@ -1,17 +1,19 @@
 #!/usr/bin/env sh
+set -e
 
 cc=${CC:-cc}
 ar=${AR:-ar}
+ODIN_ROOT=${ODIN_ROOT:-$(cd "$(dirname "$0")/../../.." && pwd)}
 
 build_wasm() {
 	mkdir -p ../lib
-	$cc -c -Os --target=wasm32 --sysroot=$(shell odin root)/vendor/libc-shim stb_image.c        -o ../lib/stb_image_wasm.o        -DSTBI_NO_STDIO
-	$cc -c -Os --target=wasm32 --sysroot=$(shell odin root)/vendor/libc-shim stb_image_write.c  -o ../lib/stb_image_write_wasm.o  -DSTBI_WRITE_NO_STDIO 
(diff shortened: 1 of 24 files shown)


vendor/stb/src/build_stb.sh
```diff
@@ -57,20 +58,24 @@ build_darwin() {
 	$cc -arch x86_64 -c -O2 -Os -fPIC stb_sprintf.c -o stb_sprintf-x86_64.o -mmacosx-version-min=10.12
 	$cc -arch arm64  -c -O2 -Os -fPIC stb_sprintf.c -o stb_sprintf-arm64.o -mmacosx-version-min=10.12
 	lipo -create stb_sprintf-x86_64.o stb_sprintf-arm64.o -output ../lib/darwin/stb_sprintf.a
-	rm *.o
+	rm ./*.o
```
antisaling, line 62: btw this avoids files with dashes to be parsed as options, more of a nit thing than something that would actually happen IMO

antisaling:
I just realized that we might want to get user-set `ODIN_ROOT` on the script. will fix

antisaling:
ahh also need to fix the CI and the other vendor build scripts. damn

antisaling:
@Kelimion I see that the netbsd script is using an old path that's unavailable, so it's genuinely broken now. I can take the opportunity to fix it on this PR so we can validate my changes properly, wdyt?

Only question that remains is whether we want to keep using 10.0 or upgrade to 11.0: https://cdn.netbsd.org/pub/pkgsrc/packages/NetBSD/x86_64/

Kelimion:
@antisaling, I think it makes sense to fix NetBSD 11 in a separate PR, merge that first, and then rebase this PR on top of that.

antisaling:
should I extend the script cleanup for e.g. box3d? I didn't really notice it on a first pass

Kelimion:
> should I extend the script cleanup for e.g. box3d? I didn't really notice it on a first pass

We just wrote the scripts for box3d, but if you want to equalize them with the other scripts to be more portable, go right ahead if you want to, @antisaling.

Kelimion:
Thanks, @antisaling.

PR #7085, opened 2026-07-19
ci: update netbsd to 10_2026Q1

antisaling:


Kelimion:
I'm hosting a dinner party, so I'll have to look at this later tonight or tomorrow.

antisaling:
have fun, I'll try to fix this weird new failure.

antisaling:
so apparently the Q2 has a bad index, I'm going to try using the base 11.0, is that a problem?

antisaling:
the new failure seems to be from very slow download speeds on their side, I'll keep the fallback `cat` either way...

antisaling:
ah, netbsd 11 is on RC6, so that's the problem, I will just try to update to 10_2026Q1 so we have a non-broken CI. testing it on QEMU so I can spare the CI

Kelimion:
Sounds like a fair approach to me. I'll have a look in the morning.

Kelimion:
Thanks for that. :-)

PR #7086, opened 2026-07-19
[core:test] Windows: Prevent exception handler from continuing.

### LINKED ISSUE #7035: Multiple tests using testing.expect_assert crash or hang the test runner on Windows
## Context

I have encountered an issue where some tests in one of my test suites crashes, but only on Windows.
I have been able to make a small reproduction of the tests that have been crashing in a reliable way. 
I have not been able to pin-point the cause of this crash, other than having a feeling that it might
be related to how signal handling on Windows dealing with asserts, since this does not seem to happen on Linux, both locally for me and in CI for my test suite. Doesn't trigger for macOS in CI either. 

### System Information
    Odin:    dev-2026-07:301c287de<br>
    OS:      Windows 10 Home Basic (version: 22H2), build 19045.7058<br>
    CPU:     AMD Ryzen 9 3900X 12-Core Processor<br> 
    RAM:     32693 MiB<br>
    Backend: LLVM 20.1.0<br>

## Expected Behavior

Tests should pass successfully without any crash or hang, both when run in parallel and sequentially using a single thread.

## Current Behavior

When running both tests in parallel, one of them will crash and the test runner stops execution.
Running both tests sequentially, i.e. with a single thread, the test runner seemingly hangs indefinitely.
Running each of the tests separately succeeds.

### Steps to Reproduce

Matrix showing the different commands and expected results in a simple way.

| Run | Command | Result |
|---|---|---|
| Both tests, default threads | `odin test .` | Crash / silent death|
| Both tests, 1 thread | `odin test . -define:ODIN_TEST_THREADS=1` | Hangs indefinitely |
| Test one alone | `odin test . -define:ODIN_TEST_NAMES=test_assert_one` | Passes |
| Test two alone | `odin test . -define:ODIN_TEST_NAMES=test_assert_two` | Passes |

Simple reproduction that triggers the issue:

```odin
package main

import "core:testing"

boom_a :: proc() {
	assert(false, "boom a")
}

boom_b :: proc() {
	assert(false, "boom b")
}

@(test)
test_assert_one :: proc(t: ^testing.T) {
	testing.expect_assert(t, "boom a")
	boom_a()
}

@(test)
test_assert_two :: proc(t: ^testing.T) {
	testing.expect_assert(t, "boom b")
	boom_b()
}

```

## Failure Logs

Running binary of both tests with default threads using RAD Debugger:<br>
Triggers on `ud2` (Illegal instruction)<br>
0x00007ff78982c985  ud2

blob1807:
Mirror POSIX and idle in our Exception Handler. Instead of continuing to next handler.  This prevent Windows from close the process before the runner is done with its reporting.

I thought it was already doing this, but must of been getting lucky with Window's handler being or something.

Fixes #7035

### DIFF
--- a/core/sys/windows/kernel32.odin
+++ b/core/sys/windows/kernel32.odin
@@ -185,6 +185,7 @@ foreign kernel32 {
 	GetExitCodeThread    :: proc(thread: HANDLE, exit_code: ^DWORD) -> BOOL ---
 	TerminateThread      :: proc(thread: HANDLE, exit_code: DWORD) -> BOOL ---
 	SuspendThread        :: proc(hThread: HANDLE) -> DWORD ---
+	ExitThread           :: proc(dwExitCode: DWORD) ---
 
 	GetProcessAffinityMask :: proc(
 		hProcess: HANDLE,
--- a/core/testing/signal_handler_windows.odin
+++ b/core/testing/signal_handler_windows.odin
@@ -146,10 +146,19 @@ This is a dire bug and should be reported to the Odin developers.
 		intrinsics.atomic_store(&stop_test_passed, passed)
 		intrinsics.atomic_store(&stop_test_alert, 1)
 
+		for {
+			// Idle until this thread is terminated by the runner,
+			// otherwise we will crash after the base handler is run.
+			win32.SwitchToThread()
+		}
+
+		// As a fallback, we exit the thread.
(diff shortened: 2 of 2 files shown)


Skarsh:
Can confirm this passes all the cases in the #7035 issue I posted. 😄

PR #7088, opened 2026-07-19
ci: fetch LFS for vendored platforms

antisaling:


Kelimion:
@antisaling, if this patch is only needed for the box3d tests, then perhaps given you've previously rewritten those build scripts, the CI can compile box3d before running the tests.

That way we don't incur unnecessary LFS-related costs.

gingerBill:
LFS does cost us quite a bit per month, so minimizing that cost, for things that get ran all the time helps a ton.

gingerBill:
I'll close this PR because of the LFS cost.

laytan:
Another possibility is doing the git checkout (with lfs) once per pipeline and saving it as an artifact. All the jobs can then pull that artifact instead of doing a full checkout again.

This would also be a nice upgrade to the release pipeline, so that will also start costing less.

antisaling:
@laytan I'm open to try that fix, but how can we actually confirm that we will save money from this? I also don't want to blast the CI testing it.

Issue #5830, opened 2025-10-19
[Bug] Deferring a labeled block results in a compile assertion


BigBoyBarney:
Hi!

## Context

* Operating System & Odin Version: Linux, dev-2025-10
* Please paste `odin report` output:
```
	Odin:    dev-2025-10
	OS:      Debian GNU/Linux forky/sid, Linux 6.16.8-200.fc42.x86_64
	CPU:     AMD Ryzen 7 5800X3D 8-Core Processor
	RAM:     32004 MiB
	Backend: LLVM 19.1.7
```

The following code results in a compile assertion:
```odin
main :: proc() {
	one: {
		defer one
	}
}
```
```sh
src/llvm_backend_general.cpp(1839): Assertion Failure: `type != t_invalid`
fish: Job 1, 'odin build Prototyping/' terminated by signal SIGILL (Illegal instruction)
```

-----

Aparte: I don't expect this to compile 😄

Kelimion:
That should definitely be caught in the frontend and raise a syntax error. 😁

Issue #7108, opened 2026-07-22
compiler assertion on named blocks


juaumjuaumjuaum:
## Context

```odin
package main

main :: proc() {
	named: {
		named
	}
}
```

### Failure Logs

```
src/llvm_backend_general.cpp(1993): Assertion Failure: `type != t_invalid`
This is a compiler error. Please report this.
```

juaumjuaumjuaum:
Related to this #5830.

`defer named` and `do named` also raises the same assertion

PR #7130, opened 2026-07-25
fix: reject the use of labels as expressions

aelobdog:
This should fix bugs #5830 and #7108


aelobdog:
The Ubuntu CI seems to have exited due to an issue in the "Download LLVM" step.

PR #7132, opened 2026-07-25
Fix error hint for vet tag

Chaosus:
My first contribution to this awesome language 😃 Just a fix for outdated error hint in cases like that:

```odin
#+vet foo
package main
//...
```

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -6586,7 +6586,8 @@ gb_internal u64 parse_vet_tag(Token token_for_pos, String s, u64 base_vet_flags)
 			error_line("\tusing-stmt\n");
 			error_line("\tusing-param\n");
 			error_line("\tstyle\n");
-			error_line("\textra\n");
+			error_line("\tsemicolon\n");
+			error_line("\tdeprecated\n");
 			error_line("\tcast\n");
 			error_line("\ttabs\n");
 			error_line("\texplicit-allocators\n");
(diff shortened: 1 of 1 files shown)


src/parser.cpp
```diff
@@ -6586,7 +6586,8 @@ gb_internal u64 parse_vet_tag(Token token_for_pos, String s, u64 base_vet_flags)
 			error_line("\tusing-stmt\n");
 			error_line("\tusing-param\n");
 			error_line("\tstyle\n");
-			error_line("\textra\n");
```
gingerBill, line 6589: Why did you remove `extra`?

Chaosus: It seems dosn't working at all and I looked into a code and its not even mentioned here - https://github.com/odin-lang/Odin/blob/e1691f9face81eb0ff0e81f04f60d9f603eab998/src/build_settings.cpp#L306-L357

tf2spi: Ah, good catch! That being said, I think it'd be better to introduce a new VetFlag ``VetFlag_Extra`` that is another combination of the ``VetFlag`` params, since that seems like what it is intended for.

I don't know what combination of ``VetFlag``s would be best. Maybe we can make it ``VetFlag_NONE`` for now until a combination of flags is agreed upon later, even if later in this PR.

Chaosus:
Okay, I misundestooding that "extra" tag is intended to be existed - I bringed it back

PR #7140, opened 2026-07-26
fix segfault on polymorphic arguments when no arguments are expected

edizeqiri:
Hi, I've been learning Odin and trying to build an ECS with it and stumbled upon a segfault while compiling my code.

Bug
---
```odin
package main

ArchetypeRow :: struct($Components: typeid) {
}

Archetype :: struct {
  components: ArchetypeRow
}

Player :: struct {
}

spawn :: proc(entity: Entity) {
  archetype := Archetype(entity) 
}

ecs :: proc() {
  player := Player {}
  spawn(player)
}

```
output:
```bash
odin check -file bug.odin                                                                                                                                                                                                     [13:47:17]
[1]    31783 segmentation fault  odin check -file bug.odin
```
This happens with the latest release and main.

        Odin:    dev-2026-07:819fdc7a8
        CPU:     Apple M4 Pro
        RAM:     49152 MiB
        Backend: LLVM 22.1.8

Fix
---
Since I would like to contribute in the future, I tried to find and fix the error in the compiler.

From my understanding, the problem lies in the struct 'Archetype' which should have a generic input like 'ArcheTypeRow' and which the compiler expects, but the naive user (me) forgot it in the struct definition but not in the initialization of it.

This is the following error message the user now gets:
```bash
.../bug.odin(14:16) Error: No params are expected!
        archetype := Archetype(entity)
                     ^~~~~~~~~~~~~~~~^
```

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -8217,7 +8217,7 @@ gb_internal CallArgumentError check_polymorphic_record_type(CheckerContext *c, O
 
 	TypeTuple *tuple = get_record_polymorphic_params(original_type);
 	if (tuple == nullptr) {
-		error(call, "No params are expected!");
+		error(call, "Type '%s' does not expect any polymorphic parameters",type_to_string(operand->type));
 		err = CallArgumentError_TooManyArguments;
 		return err;
 	}
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -8216,6 +8216,11 @@ gb_internal CallArgumentError check_polymorphic_record_type(CheckerContext *c, O
 	}
 
 	TypeTuple *tuple = get_record_polymorphic_params(original_type);
+	if (tuple == nullptr) {
+		error(call, "No params are expected!");
+		err = CallArgumentError_TooManyArguments;
+		return err;
+	}
 	isize param_count = tuple->variables.count;
 	isize minimum_param_count = param_count;
 	for (; minimum_param_count > 0; minimum_param_count--) {


antisaling:
can you confirm your patch also keeps the intended error message for a type assignment? e.g.
```
Error: Cannot assign a type 'int' to variable 'x'
        x := int
             ^~^
```

so your output should look like

```
.../bug.odin(14:16) Error: No params are expected!
        archetype := Archetype(entity)
                     ^~~~~~~~~~~~~~~~^
.../bug.odin(14:16) Error: Cannot assign a type 'Archetype(entity)' to variable 'archetype'
        archetype := Archetype(entity)
                     ^~~~~~~~~~~~~~~~^
```

src/check_expr.cpp
```diff
@@ -8216,6 +8216,11 @@ gb_internal CallArgumentError check_polymorphic_record_type(CheckerContext *c, O
 	}
 
 	TypeTuple *tuple = get_record_polymorphic_params(original_type);
+	if (tuple == nullptr) {
+		error(call, "No params are expected!");
```
antisaling: I also think we don't really want an exclamation mark on an error. Something more neutral like `Type 'X' does not expect any polymorphic parameters` would be more in line with the current errors

edizeqiri:
> can you confirm your patch also keeps the intended error message for a type assignment? e.g.
> 
> ```
> Error: Cannot assign a type 'int' to variable 'x'
>         x := int
>              ^~^
> ```
> 
> so your output should look like
> 
> ```
> .../bug.odin(14:16) Error: No params are expected!
>         archetype := Archetype(entity)
>                      ^~~~~~~~~~~~~~~~^
> .../bug.odin(14:16) Error: Cannot assign a type 'Archetype(entity)' to variable 'archetype'
>         archetype := Archetype(entity)
>                      ^~~~~~~~~~~~~~~~^
> ```

The current output does not emit that error:
```bash
./odin run -file debug/test.odin                                                                                                                                                                                                                                                                                                                                                            
/Users/edi/Documents/GitHub/Odin/debug/test.odin(1:1) Error: Undefined entry point procedure 'main'
	package main
	^

/Users/edi/Documents/GitHub/Odin/debug/test.odin(7:3) Error: Invalid parameter type
	components: ArchetypeRow
	^~~~~~~~~~~~~~~~~~~~~~~^

/Users/edi/Documents/GitHub/Odin/debug/test.odin(7:15) Error: Invalid use of a non-specialized polymorphic type 'ArchetypeRow'
	components: ArchetypeRow
	            ^~~~~~~~~~~^

/Users/edi/Documents/GitHub/Odin/debug/test.odin(13:23) Error: Undeclared name: Entity
	spawn :: proc(entity: Entity) {
	                      ^~~~~^

/Users/edi/Documents/GitHub/Odin/debug/test.odin(14:16) Error: Type 'Archetype' does not expect any polymorphic parameters
	archetype := Archetype(entity)
	             ^~~~~~~~~~~~~~~~^

/Users/edi/Documents/GitHub/Odin/debug/test.odin(19:9) Error: Cannot assign value 'player' of type 'Player' to 'invalid type' in a procedure argument
	spawn(player)
	      ^~~~~^
```

While debugging I found out that the error you mentioned would be in check_decl.cpp:60. But when debugging it never went past the first if condition.

[`check_decl.cpp:47-68`](https://github.com/odin-lang/Odin/blob/94d188c75fa6092ed2df8a53e115f19f2d5a7e89/src/check_decl.cpp)

```cpp {47-68}
	if (operand->mode == Addressing_Type) {
		if (e->type != nullptr && is_type_typeid(e->type) && !is_type_polymorphic(operand->type)) {
			add_type_info_type(ctx, operand->type);
			add_type_and_value(ctx, operand->expr, Addressing_Value, e->type, exact_value_typeid(operand->type));
			return e->type;
		} else {
			ERROR_BLOCK();

			gbString t = type_to_string(operand->type);
			defer (gb_string_free(t));
			if (is_type_polymorphic(operand->type)) {
				error(operand->expr, "Cannot assign a non-specialized polymorphic type '%s' to variable '%.*s'", t, LIT(e->token.string));
			} else {
				error(operand->expr, "Cannot assign a type '%s' to variable '%.*s'", t, LIT(e->token.string));
			}
			if (e->type == nullptr) {
				error_line("\tThe type of the variable '%.*s' cannot be inferred as a type and does not have a default type\n", LIT(e->token.string));
			}
			e->type = operand->type;
			return nullptr;
		}
	}
```
I do not think that my patch is causing this to not show the error, but I am not certain. What do you think?

src/check_expr.cpp
edizeqiri: That's a better idea, thx. I changed it to your suggestion with the type:
```bash
../test.odin(14:16) Error: Type 'Archetype' does not expect any polymorphic parameters
	archetype := Archetype(entity)
 ```

antisaling (commented):
So your patch [causes `check_call_expr_as_type_cast` to poison the operand](https://github.com/odin-lang/Odin/blob/94d188c75fa6092ed2df8a53e115f19f2d5a7e89/src/check_expr.cpp#L8662). I'm not completely sure on what's the clear path here to preserve the second error because `check_call_expr_as_type_cast` would need to be smarter about "recoverable" errors.

LGTM?

edizeqiri:
After this gets merged, I could open an issue for `check_call_expr_as_type_cast` and try to come up with a solution to preserve the other error.

Issue #7104, opened 2026-07-22
tables.odin has Excess ideographic spaces, potentially breaking binary_search


LemonatedCat:
I'm not 100% Sure this is intentional, but 
in 
https://github.com/odin-lang/Odin/blob/master/core/unicode/tables.odin#L477
and
https://github.com/odin-lang/Odin/blob/master/core/unicode/tables.odin#L495

There are extra ranges from 0x3000 to 0x3000 where they shouldn't be, breaking the ordering of the table. Said tables are also used in binary search, which is why i doubt it's intentional.

Seemingly, a comment was replaced with them.

PR #7142, opened 2026-07-26
Fix duplicated entries within unicode_spaces and space_ranges

LemonatedCat:
Somewhere on the creation of these tables, the `0x3000-0x3000` range seemingly slipped in an additional time.

Since binary search is done on both `unicode_spaces` and `space_ranges`, i doubt this is intentional behavior, as this unsorts the tables.


LemonatedCat:
Also closes #7104 from a few days ago

PR #7147, opened 2026-07-28
[debug/trace] sync tracking allocator with core:mem

A1029384756:
adds functionality that was added to the tracking allocator in `core:mem` after the base of the current `core:debug/trace` was created. this has been tested on a few programs and should allow it to basically be dropped in in place of the current tracking allocator in `core:mem`

the reason this was duplicated instead of being brought into `core:mem` was because this package imports many other `core` packages which is not a good fit for `core:mem`

### DIFF
--- a/core/debug/trace/allocator.odin
+++ b/core/debug/trace/allocator.odin
@@ -39,12 +39,18 @@ Example:
 	}
 */
 Tracking_Allocator :: struct {
-	backing:              mem.Allocator,
-	internals_allocator:  mem.Allocator,
-	allocation_map:       map[rawptr]Tracking_Allocator_Entry,
-	bad_free_array:       [dynamic]Tracking_Allocator_Bad_Free_Entry,
-	mutex:                sync.Mutex,
-	clear_on_free_all:    bool,
+	backing:                  mem.Allocator,
+	allocation_map:           map[rawptr]Tracking_Allocator_Entry,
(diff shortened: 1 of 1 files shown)


core/debug/trace/allocator.odin
```diff
@@ -63,38 +69,94 @@ Tracking_Allocator_Bad_Free_Entry :: struct {
...
+@(no_sanitize_address)
+tracking_allocator_bad_free_callback_panic :: proc(
+	t: ^Tracking_Allocator,
+	memory: rawptr,
+	backtrace: Capture_Const,
+	location: runtime.Source_Code_Location,
+) {
+	buf: [256]byte
+	offset := 0
+	_ = runtime.write_string(&offset, buf[:], "Tracking allocator error: Bad free of pointer ")
```
laytan, line 133: Should we make this print the back trace?

A1029384756: i tested this and it already does if they set the assertion failure proc (which is part of the recommended setup). if i *do* print the trace here, it would result in a double-print

core/debug/trace/allocator.odin
laytan: Let's remove the todo now that we looked into it

core/debug/trace/allocator.odin
laytan: Ah, right!

PR #7148, opened 2026-07-28
add odin build/run -cached flags

i-api:
so we have the option to reuse cached binaries if it doesn't require a rebuild. makes runs after the first much less painful now

@gingerBill lmk how you feel about this, cool / not cool?


antisaling:
Even if we want something like this I think you completely overlooked that we still need checking `mtime` for statically-linked foreign libs.

i-api:
you are right

i-api:
pushed my static libs fix

i-api:
@gingerBill @Kelimion please review, thanks

i-api:
Rebased onto current master (e7b554e60) — no conflicts. Verified locally: build/run -cached hits skip codegen, flag/compiler/static-lib changes correctly invalidate, committed tests/issues cached checks all pass, and run shares the single build cache path (test rides along via the same flag registration, no test-specific logic).

i-api:
@gingerBill @Kelimion  this is rebased on current master and I verified it locally. Cache hits skip codegen, changing flags or the compiler or a static lib forces a rebuild, and the cached tests in tests/issues all pass. run -cached is just build -cached plus exec on the same path, and test comes along for free with no test specific logic. Ready for review when you have time. Happy to change anything, especially the rules for what counts as a cache hit.

i-api:
@Kelimion @gingerBill can you take over this PR for me? Unfortunately I won't have time to shepherd this change in the coming months

PR #7160, opened 2026-07-29
Respect TERM=dumb in test runner to prevent ANSI sequences, color

### LINKED ISSUE #7137: TERM=dumb ignored by test runner (regression?) (good first issue?)
## Context

I am a dinosaur, and as such my workflow requires that I use the emacs compilation buffer. The test runner doesn't seem to respect the TERM environment variable at the moment, causing the fancy progress and ANSI escapes to clutter the buffer. Likely a regression?

Is this a good first issue? Seems like it should be a simple fix and I'm happy to tackle it. Some other terminal programs simply check the TERM environment variable and unconditionally turn off all ANSI if its value is 'dumb'.

Related:
- #5189 
- #4100

Operating System & Odin Version:

	Odin:    dev-2026-07:c5e58ad05
	OS:      NixOS 26.05 (Yarara), Linux 6.18.33.2-microsoft-standard-WSL2
	RAM:     7899 MiB
	Backend: LLVM 21.1.8

## Expected Behavior

When TERM=dumb, a nice clean text-only display for us simple-minded dinosaurs that like things in black and white.

## Current Behavior

Not that. (See log below.)

### Steps to Reproduce

Run `odin test .` on a simple test project.

### Failure Logs

```
-*- mode: compilation; default-directory: "~/dev/odin/term-dumb/" -*-
Compilation started at Sun Jul 26 12:06:57

echo $TERM && ../Odin/odin test .
dumb
[?25l[?7lmain  [[90m|                       [0m]    0/   1 :: [90mtest_one[0m
[?7h1 thread                            0/   1 :: total
[2F[J]2;Odin test runner (1/1)\[0m[INFO ] --- [0m[2026-07-26 00:06:59] Starting test runner with 1 thread. Set with -define:ODIN_TEST_THREADS=n.
[0m[INFO ] --- [0m[2026-07-26 00:06:59] The random seed sent to every test is: 4661628076924. Set with -define:ODIN_TEST_RANDOM_SEED=n.
[0m[INFO ] --- [0m[2026-07-26 00:06:59] Memory tracking is enabled. Tests will log their memory usage if there's an issue.
[0m[INFO ] --- [0m[2026-07-26 00:06:59] < Final Mem/ Total Mem> <  Peak Mem> (#Free/Alloc) :: [package.test_name]
[?7lmain  [[32m|                       [0m]         1 :: [package done]
[?7h
[?25hFinished 1 test in 232.994µs. The test was [32msuccessful.[0m


Compilation finished at Sun Jul 26 12:07:00, duration 2.60 s

```

mocompute:
Fixes: #7137

### DIFF
--- a/core/terminal/internal_os.odin
+++ b/core/terminal/internal_os.odin
@@ -72,9 +72,9 @@ get_environment_color :: proc() -> Color_Depth {
 
 @(init)
 init_terminal :: proc "contextless" () {
-	_init_terminal()
-
 	context = runtime.default_context()
+
+	_init_terminal()
 
 	// We respect `NO_COLOR` specifically as a color-disabler but not as a
 	// blanket ban on any terminal manipulation codes, hence why this comes
--- a/core/terminal/internal_os.odin
+++ b/core/terminal/internal_os.odin
@@ -21,6 +21,15 @@ get_no_color :: proc() -> bool {
 	return false
 }
 
+get_is_dumb :: proc() -> bool {
+	buf: [128]u8
+	if term, err := os.lookup_env(buf[:], "TERM"); err == nil {
+		// "dumb" terminal overrides all other color and ansi capabilities logic
+		return term == "dumb"
+	}
+	return false
+}
+
 get_environment_color :: proc() -> Color_Depth {
 	buf: [128]u8
 	// `COLORTERM` is non-standard but widespread and unambiguous.
--- a/core/terminal/terminal.odin
+++ b/core/terminal/terminal.odin
@@ -24,6 +24,11 @@ is_terminal :: proc(f: $T) -> bool {
 }
 
 /*
+This is true if the terminal is dumb, i.e. cannot process ANSI control sequences.
+*/
+is_dumb: bool
+
+/*
 This is true if the terminal is accepting any form of colored text output.
 */
 color_enabled: bool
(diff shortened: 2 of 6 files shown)


Kelimion:
I'll have a look at this tomorrow, because I used to have ANSI spam in Sublime Text at first, and that was fixed. This seems to be related to that, so I want to have a look whether this is the right place for this check, or perform it in the [terminal package](https://github.com/odin-lang/Odin/blob/master/core/terminal/internal_os.odin). That way apps other than the test runner would also benefit from the "dumb" detection.

mocompute:
Thank you, I wasn't sure if `terminal` was the right place, so I went with the targeted fix in the test runner for the time being. 

terminal correctly determines colour depth, and it already queries the TERM environment variable to help do that. But colour depth and ansi escape capability are orthogonal, and the package doesn't yet directly address ansi capability. I agree that would be a more general place.

Kelimion:
I'll have a think about it tomorrow. We might want to expose e.g. `terminal.is_dumb` (or conversely, `terminal.ansi_enabled`) so apps can disable control sequences irrespective of color capabilities. You're definitely on the right track. The main question I need to ponder is whether to move the detection to the terminal package for wider use.

Kelimion:
Unrelated CI failure. Merging once the rest passes.

mocompute:
Brilliant, it's a much nicer interface in the terminal package, thank you!

Kelimion:
Some tests we really shouldn't run on the emulated NetBSD runner so we don't die of old age waiting for it.

Anyway, merged. Thanks!

PR #7162, opened 2026-07-30
handle spaces in args for 'odin run . -- <args>' on Windows

### LINKED ISSUE #4393: os.args parses \ incorrectly on Windows

BunterSchatten:
fixes #4393

also fixes using `odin run .` with an ouput path containing spaces.

`odin run . -- <args>` does not forward <args> correctly on Windows, if there are quoted spaces in them. For exmple `odin run . -- "one two"` gets forwarded as `one two`.
This is because the Odin cmpiler parses all argument into an array and Windows handles multiple args in `_spawnv` by just joining them together with spaces, which looses all information of the quotations.

I fixed that by detecting the `--` directly in the wchar command line parsing for windows and saving a pointer to the whole rest of the string. Then I changed `run_subprocess` on Windows to not use `_spawnv` anymore but instead directly call `CreateProcessW`. I also (unconditionally) qoute the exe path argument here, to make it work with paths containing spaces.

I'm not sure, if the last two commits are really necessary:
The first one just out-defines the now unused args array building on non-Windows.
The second one also returns the double_dash_pos from the windows parsing and asserts it against the later detection logic, so a mismatch there could be detected.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -640,7 +640,7 @@ gb_internal gb_inline f64 gb_sqrt(f64 x) {
 
 #if defined(GB_SYSTEM_WINDOWS)
 
-gb_internal wchar_t **command_line_to_wargv(wchar_t *cmd_line, int *_argc) {
+gb_internal wchar_t **command_line_to_wargv(wchar_t *cmd_line, int *_argc, wchar_t **_after_double_dash_raw) {
 	u32 i, j;
 
 	u32 len = cast(u32)string16_len(cast(u16 *)cmd_line);
@@ -649,6 +649,8 @@ gb_internal wchar_t **command_line_to_wargv(wchar_t *cmd_line, int *_argc) {
 	wchar_t **argv = cast(wchar_t **)GlobalAlloc(GMEM_FIXED, i + (len+2)*gb_size_of(wchar_t));
 	wchar_t *_argv = cast(wchar_t *)((cast(u8 *)argv)+i);
 
+	wchar_t *after_double_dash_raw = nullptr;
+
 	u32 argc = 0;
 	argv[argc] = _argv;
 	bool in_quote = false;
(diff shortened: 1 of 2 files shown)


src/common.cpp
```diff
@@ -657,6 +660,18 @@ gb_internal wchar_t **command_line_to_wargv(wchar_t *cmd_line, int *_argc) {
 	i = 0;
 	j = 0;
 
+	auto const check_double_dash = [&]() {
+		if (double_dash_pos == -1 &&
+			argc >= 1 &&
```
Kelimion, line 665: This appears to be missing a bounds check on `argv[argc - 1]`. It could be shorter than `--\0`.

BunterSchatten: I don't think so, every argv is null-terminated, so the comparison against '-' would fail early, but its probably still better to add an explicit bounds check.

Kelimion: Yeah, it's probably fine.

Kelimion:
Thanks!

PR #7164, opened 2026-07-30
Don't assert on -define: values with an exponent

### LINKED ISSUE #7163: Compiler assertion in big_int.cpp parsing a `-define:` value with a negative exponent (`-define:X=1e-5`)
## Context

* **Operating System & Odin Version:** macOS Tahoe 26.5.2 (arm64), Odin `dev-2026-07:819fdc7a8` (installed via Homebrew)
* **`odin report` output:**

```
      Odin:    dev-2026-07:819fdc7a8
      CPU:     Apple M4 Pro
      RAM:     24576 MiB
      Backend: LLVM 22.1.8
```

## Expected Behavior

A `-define:` value of `1e-5` is a valid float literal, so I'd expect it to parse as a float, or, if `-define:` doesn't accept floats, to fall back to being treated as a string. Either is fine.

For a genuinely malformed value such as `7e-`, I'd expect a diagnostic naming the bad `-define:`, not an abort.

## Current Behavior

Any `-define:` whose value looks like scientific notation with a negative exponent aborts the compiler with an assertion failure in `big_int.cpp`.

The value does not have to be referenced anywhere — no `#config` is needed, and `odin check` on an empty package is enough to trigger it.

The trigger is the pattern `<digits>` `e`/`E` followed by `-`:

| value | result |
|---|---|
| `1e-5` | **crash** (valid float literal) |
| `1e-x` | **crash** |
| `1E-x` | **crash** |
| `7e-` | **crash** |
| `781377e-dirty` | **crash** |
| `1e5` | ok — parses, warns `given -define:X is unused` |
| `1-a` | ok |
| `1.5-x` | ok |
| `1p-x` | ok |
| `12345` | ok |
| `a-b` | ok |
| `781377f-dirty` | ok |

It looks like the `-define:` value parser routes anything that begins lexing as a number into `big_int_from_string`, which doesn't handle a sign in the exponent.

### How this shows up in practice

A common build-stamp idiom is:

```sh
odin build src -define:BUILD=$(git describe --always --dirty --tags)
```

`git describe` emits an abbreviated commit hash and, on a dirty tree, appends `-dirty`. When the abbreviated hash happens to end in `e` — roughly 1 commit in 16 — the value becomes e.g. `781377e-dirty` and the build dies. Since it depends on the current commit hash *and* on whether the tree is dirty, it presents as an intermittent compiler error pointing at whichever file was edited most recently,
which is a long way from the actual cause.

Workaround: quote the value to force a string, and strip the quotes at the use site, since they arrive as part of it.

```sh
odin build src -define:BUILD="\"$(git describe --always --dirty --tags)\""
```

### Steps to Reproduce

1. Create a package containing nothing:

```odin
// main.odin
package main
main :: proc() {}
```

2. Run any of `odin check`, `odin build` or `odin run` with a `-define:` whose value has a negative exponent:

```
odin check . -define:X=1e-5
```

3. The compiler asserts. Note that `X` is never declared with `#config` and never referenced.

### Failure Logs

```
$ odin check . -define:X=1e-5
src/big_int.cpp(253): Assertion Failure: `text[i] != '-'`
```

Identical output from `odin build . -define:X=1e-5` and `odin run . -define:X=1e-5`.

For contrast, a positive exponent is accepted:

```
$ odin check . -define:X=1e5
Warning: given -define:X is unused in the project
      Suggestion: use the -show-defineables flag for an overview of the possible defines
```

jerksto:
Fixes #7163.

`-define:X=1e-5` aborts the compiler in `big_int_from_string`. The
exponent branch there asserts `base == 10` and `text[i] != '-'`, which
holds for anything the lexer produced but not for a `-define:` value,
which is arbitrary text. `-define:X=0b1e5` hits the first assert the
same way: `e` is not a base 2 digit, so the digit loop breaks into the
exponent path. `7e` read one byte past the end.

Both asserts become parse failures, and the exponent now has to have at
least one digit.

That alone would make `1e-5` an error rather than a crash, so the second
change is in `build_param_to_exact_value`, which only sent a value to
the float parser if it contained a `.`. It now also does so for a base
10 exponent, skipping prefixed literals since `e` is a hex digit.

```
-define:X=1e-5           1e-05 (f64)
-define:X=1e5            100000 (int)
-define:X=0x1e           30 (int)
-define:X=781377e-dirty  781377e-dirty (string)
-define:X=0b1e5          0b1e5 (string)
```

The last one is the case that started this: `git describe` output ending
in `e` no longer has to be quoted.

`tests/core/math` and `tests/core/strconv` pass, and `1e-5`, `1e5`,
`0x1e` and `1.5e+3` as source literals are unchanged.

### DIFF
--- a/src/big_int.cpp
+++ b/src/big_int.cpp
@@ -249,13 +249,28 @@ gb_internal void big_int_from_string(BigInt *dst, String const &s, bool *success
 	}
 	if (i < len && (text[i] == 'e' || text[i] == 'E')) {
 		i += 1;
-		GB_ASSERT(base == 10);
-		GB_ASSERT(text[i] != '-');
+		if (base != 10) {
+			// An exponent is only meaningful for a base 10 literal.
+			*success = false;
+			return;
+		}
+		if (i >= len) {
+			// Nothing follows the exponent marker.
+			*success = false;
+			return;
+		}
+		if (text[i] == '-') {
+			// A negative exponent is never an integer.
+			// The caller is expected to parse the value as a float instead.
+			*success = false;
+			return;
+		}
 		if (text[i] == '+') {
 			i += 1;
 		}
 
 		u64 exp = 0;
+		isize exp_digits = 0;
 		for (; i < len; i++) {
 			char r = cast(char)text[i];
 			if (r == '_') {
@@ -270,6 +285,11 @@ gb_internal void big_int_from_string(BigInt *dst, String const &s, bool *success
 			}
 			exp *= 10;
 			exp += v;
+			exp_digits += 1;
+		}
+		if (exp_digits == 0) {
+			*success = false;
+			return;
 		}
 
 		// NOTE(Jeroen): A valid integer can never have an exponent larger than 308 (per `max(f64)`).
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -537,6 +537,27 @@ gb_internal void add_flag(Array<BuildFlag> *build_flags, BuildFlagKind kind, Str
 	array_add(build_flags, flag);
 }
 
+// A value such as `1e-5` is a float, not an integer. Base prefixed literals are excluded because
+// `e` is a valid digit in bases 16 and above, e.g. `0x1e`.
+gb_internal bool build_param_looks_like_float(String const &param) {
+	if (string_contains_char(param, '.')) {
+		return true;
+	}
+	isize i = 0;
+	if (param.len > 0 && (param[0] == '-' || param[0] == '+')) {
+		i = 1;
+	}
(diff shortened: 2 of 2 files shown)


Kelimion:
This is a good interim solution that we'll adopt, thanks.

At some point before we hit 1.0, I'd like to see all of the `-define`s turned into untyped pseudo constants - for example by registering the the same was as e.g. `ODIN_OS`, except tagged so they're not available until paired with `#config(defined_ident)`, and only at that point have the compiler perform type checking based on the default value.

`#config(FOO, false)` would see `-define:FOO=true` treated as a bool, whereas the same define paired with `#config(FOO, "false")` keeps it as text. And we'd be able to get rid of all of that special casing for defines in `Exact_Value`.

PR #7170, opened 2026-07-31
docs: reorder switch-case concepts in demo.odin

dofi4ka:
I found this demo a bit confusing and it also had some duplication. I like to show `fallthrough` example right after its first mention. 

Also, as a non-native English speaker, I found phrase "case values need not be integers nor constants." a bit confusing. May be it is correct fluent usage of English, but I suppose simpler explanation: "case values do not have to be integers or constants"


dofi4ka:
thats frustrating

PR #7171, opened 2026-07-31
Add native Windows ARM64 support

### LINKED ISSUE #6686: Support Windows ARM64
Please add support for:
- Targeting Windows ARM64, and
- Provide a native Windows ARM64 compiler build.

I would like to avoid x64 emulation when using my own device.

LehMichael:
Fixes #6686.

## Summary

This adds Windows ARM64 as a supported Odin target and makes the compiler itself build and run natively on Windows ARM64.

- Register `windows_arm64` and use the `aarch64-pc-windows-msvc` LLVM triple.
- Detect native ARM64 Windows hosts and resolve ARM64 MSVC and Windows SDK tools/libraries.
- Use the canonical ARM64 calling convention for Windows `system`/stdcall declarations.
- Add the Windows ARM64 runtime support required by the compiler.
- Split Windows compiler and vendor binaries into architecture-specific directories, selecting them through `ODIN_ARCH`.
- Add native ARM64 vendor binaries where they are available or can be built.
- Add a DLL ABI test that exercises C calls and callbacks across a Windows DLL boundary.
- Add native Windows ARM64 CI and nightly builds.
- Keep published Windows archives architecture-specific so each archive does not carry the other architecture's binaries.

## Impact

Windows ARM64 users can build and run Odin natively without x64 emulation. Existing Windows x64 builds remain covered by the same test suite, including AddressSanitizer.

Some optional vendor packages still do not have upstream Windows ARM64 binaries. Their imports remain architecture-specific, so using one without the matching library fails normally at link time rather than silently selecting an x64 binary.

AddressSanitizer is not enabled for the ARM64 CI job. LLVM 20.1 does not currently provide a usable Windows ARM64 ASan runtime: building compiler-rt for ARM64 exposes unsupported Windows interception code and an incompatible shadow-memory layout. This is an upstream LLVM limitation; the existing Windows x64 ASan coverage is unchanged.

## Validation

- Full CI passed on both Windows x64 and native Windows ARM64, along with the other platform jobs: https://github.com/LehMichael/Odin/actions/runs/30622655926
- Nightly packaging passed for both Windows architectures: https://github.com/LehMichael/Odin/actions/runs/30620247367
- Native ARM64 compiler builds and runs on Windows ARM64.
- The Windows DLL ABI test passes natively on x64 and ARM64.
- The branch was rebased onto the current upstream `master`.

## Development disclosure

For transparency: this change was developed with OpenAI Codex under my direction, with iterative review and testing on native Windows ARM64 hardware.


gingerBill:
We CANNOT accept this PR for numerous reasons:

* It is vibe coded slop (not just LLM assisted)
* You changed bits of code which were not even related
* You added binaries and even changed some which you should not have done
  * Only core contributors are allowed to even do that because of trust
* In general, unless it is easy to procure, we are vetoing adding targets until at least 1 core contributor has hardware for it AND is willing to support it.

LehMichael:
Hey Bill,
thanks for the feedback. While i don't really agree with the first two points, I totally understand where you are coming from. I'll be maintaining the fork since i need it for my own projects so if you ever need any changes where this would become more acceptable just hit me up any time.
I'm sorry if I caused any additional workload.
BR

Kelimion:
I can understand a difference of opinion on the first statement, but the second one is indisputable, imo. The `core:crypto/rsa` change was unrelated and unnecessary.

But thanks for taking the feedback in good spirits.

LehMichael:
True, i missed that one.
This was a test while debugging that slipped through. The test_rsa_sign_pss fails when optimizations are enabled, and this "fixed" it somehow, even if it seems unrelated.
Probably some LLVM-optimizer weirdness and maybe related to the RISC-V issue with the same test.

PR #7172, opened 2026-07-31
Fix typo in constant name and a different comment

mocompute:


mocompute:
On further thought, it might be better to deprecate the mis-spelled constant and add a correctly spelled one. I will have a look at that today.

Kelimion:
Just rename. Argon2id is very new code. The handful of people who use that constant can adapt.

mocompute:
Fair enough, thank you!

Kelimion:
Thanks for the fix, btw.

PR #7174, opened 2026-07-31
Impl simd arm crc32

alexthed1rk:


Kelimion:
Will you close this one in a couple of minutes as well, or what's going on?

alexthed1rk:
@Kelimion no, I wouldn't close this one 🙃 
My perfectionism got in the way; there were some things I had overlooked that needed to be corrected.
I'm sorry for spamming you with requests.

Kelimion:
Ah, I thought this was an AI assistant gone rogue type thing.

If you do want to make further edits, just make a new commit and push it to the PR's topic branch. We can always squash merge into a single commit if necessary. Or you could force push to replace the commits. There's no need to cycle through pull requests if you want to end up with a neat commit history. :-)

PR #7176, opened 2026-07-31
Fix octal escape parsing by core:text/scanner

GPotoshin:
While parsing the Odin standard library with the scanner, I encountered an error on the line `_POSIX_VDISABLE :: '\377'` in `core/sys/posix/unistd.odin`. The issue was an off-by-one error: after reading the first of 3 octal digits, a length of 3 was passed for the remaining digits instead of 2.

### DIFF
--- a/core/text/scanner/scanner.odin
+++ b/core/text/scanner/scanner.odin
@@ -462,7 +462,7 @@ scan_string :: proc(s: ^Scanner, quote: rune) -> (n: int) {
 		if ch == '\\' {
 			ch = advance(s)
 			switch ch {
-			case quote, 'a', 'b', 'e', 'f', 'n', 'r', 't', 'v', '\\':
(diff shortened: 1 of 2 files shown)


Kelimion:
If you're not too busy, we don't have a `tests/core/text/scanner` yet either. ;-)

GPotoshin:
I can write tests for different escape sequences. I think I can make testing fail on error messages, by passing a custom error procedure

Kelimion:
That'd be wonderful, thank you.

GPotoshin:
There was another incorrectly handled case. In the file `core/odin/tokenizer/tokenizer.odin` on the line 364 we have
```odin
  case 'a', 'b', 'e', 'f', 'n', 't', 'v', 'r', '\\', '\'', '\"':
```
and scanner was failing on `'\"'`, because it considered only quotes of the same kind for an escape sequence, but it works in odin. I changed that, plus I wrote a testing helper function for scanner and few tests

PR #7196, opened 2026-08-03
Make soa builtins go vroom

corleypc:
This PR is the result of extensive review, iteration, testing and benchmarking of the #soa dynamic arrays over the course of a couple of weeks. It introduces significant performance optimizations and incidentally fixes #soa arrays from #packed structs.

The #soa builtins currently still use the approach of their original implementation (6-7 years old?), where every element operation loops through the element type's fields through runtime type info, copying each field with a mem_copy. Meanwhile the compiler gained #soa store/load lowering (lb_addr_soa_variable), which the builtins haven't adopted. This PR uses it for the single-element operations (append_soa_elem, unordered_remove_soa, inject_at_elem_soa) and adds typed vectorizer-friendly per-field copy passes for batch appends (for types with up to 16 fields). In my benchmark this is up to ~11× faster on single appends, batch appends and unordered removes and up to ~3x on single injections. The PR additionally fixes a correctness bug where the existing type-erased loops assume default struct layout, so #soa arrays of #packed structs read and write the wrong field offsets and corrupt data.

Result tables first, because a number is worth a thousand words. (not really :)

(All benchmarks here done on Win10, LLVM 20.1, i7-4770, -o:speed, default odin compiler target (which is SSE capable, but not AVX capable); benchmark code (a bit tidied up to be more presentable) in this [gist](https://gist.github.com/corleypc/868e3f3d14b3b3151fd1207ddcf7ab8e))

**Single element appends (array size = 4128)**

| elem type | SOA fields | bytes/elem | speedup |
|---|---|---:|---:|
| `Vec2_f64` | 2 x 8B | 16 | **1.92x** |
| `Vec3_f64` | 3 x 8B | 24 | **4.89x** |
| `Vec4_f32` | 4 x 4B | 16 | **4.94x** |
| `Mixed` | 3 (1/8/2B) | 11 | **5.17x** |
| `RGBA` | 4 x 1B | 4 | **3.98x** |
| `Particle` | 4 (12/12/16/4B) | 44 | **2.88x** |
| `Fat` | 8 x 4B | 32 |  **11.05x** |
| `Fatter` | 12 x 4B | 48 |  **10.06x** |
| `VeryFat` | 16 x 4B | 64 | **6.24x** |
| `Fattest` | 18 x 4B | 72 | **5.96x** |
| `[2]f32` | 2 x 4B | 8 | **1.93x** |
| `[4]u16` | 4 x 2B | 8 | **4.56x** |

**Single element appends (varying array size)**

| elem type | SOA fields | bytes/elem | n | speedup |
|---|---|---:|---:|---:|
| `Vec3_f64` | 3 x 8B | 24 | 64  | **5.02x** |
| `Vec3_f64` | 3 x 8B | 24 | 65,536 | **3.05x** |
| `Vec3_f64` | 3 x 8B | 24 | 1,048,576 | **2.67x** |

**Batch appends speedups (array size = 4128)**

| elem type | batch of 4 | batch of 8 | batch of 15 | batch of 32 | batch of 61 |
|---|---:|---:|---:|---:|---:|
| `Vec2_f64` | 4.32x | 5.48x | 5.39x | 7.83x | 8.23x |
| `Vec3_f64` | 1.36x | 1.46x | 1.41x | 2.05x | 1.95x |
| `Vec4_f32` | 3.82x | 5.74x | 4.66x | 9.82x | 9.11x |
| `Mixed` | 3.56x | 5.06x | 4.97x | 10.27x | 9.46x |
| `RGBA` | 2.98x | 4.41x | 5.02x | 9.42x | 8.78x |
| `Particle` | 2.99x | 3.32x | 3.85x | 4.43x | 4.68x |
| `Fat` | 4.15x | 5.81x | 4.55x | 9.74x | 8.92x |
| `Fatter` | 3.93x | 5.02x | 3.74x | 9.80x | 7.36x |
| `VeryFat` | 3.66x | 5.37x | 3.56x | 9.81x | 6.85x |
| `[2]f32` | 5.19x | 6.75x | 5.90x | 11.38x | 10.85x |
| `[4]u16` | 3.57x | 5.40x | 6.02x | 10.19x | 9.62x |

**Unordered removes (array size = 4128)**

| elem type | SOA fields | bytes/elem | speedup |
|---|---|---:|---:|
| `Vec2_f64` | 2 x 8B | 16 | **10.17x** |
| `Vec3_f64` | 3 x 8B | 24 | **2.65x** |
| `Vec4_f32` | 4 x 4B | 16 | **2.93x** |
| `Mixed` | 3 (1/8/2B) | 11 | **9.15x** |
| `RGBA` | 4 x 1B | 4 | **2.12x** |
| `Particle` | 4 (12/12/16/4B) | 44 | **2.06x** |
| `Fat` | 8 x 4B | 32 | **1.69x** |
| `Fatter` | 12 x 4B | 48 | **3.89x** |
| `VeryFat` | 16 x 4B | 64 | **3.58x** |
| `Fattest` | 18 x 4B | 72 | **3.45x** |
| `[2]f32` | 2 x 4B | 8 | **11.85x** |
| `[4]u16` | 4 x 2B | 8 | **2.77x** |

*Scroll down for a list of specific changes in the commit.*

**Where the gains come from**

As said in the prologue, the curremt implementation of the #soa builtins is now quite old, and in the meantime there's been compiler lowering imlemeneted through the LLVM backend SOA addressing (lb_addr_soa_variable), which is not used by the builtins. The builtins extensively use type-erased loops over the fields. These type-erased loops nevertheless get instantiated per element type, so they don't deliver type erasure's full size benefit of a single shared implementation. They read the RTTI on every invocation per every field, which particularly affects the single-element operations, where there is no batch to amortize over.

The other significant addition in the PR is typed copies for batch appends (append_soa_elems, non_zero_append_soa_elems), using expand_values. This sidesteps RTTI altogeter, cause everything is compile-time known. It is vectorizer friendly and is significantly faster than the compiler generated lowering. The downside is that code size scales at around ~80-85 asm instructions per field (x64, -o:speed, default target), so this is currently limited to 16 fields and conditioned on .ODIN_OPTIMIZATION_MODE >= .Speed. At >16 fields, the type erased loop is kept and at these field counts its field count invariant size is also smaller for .Size targets. (this is also the reason there is no row for the 18-field type in the batch appends table, it's pretty much the same code with minor tweaks)

<details>
<summary>Deeper dive for the curious soul</summary>
If the spread of speedups in the tables in relation to field count had you raise an eyebrow, congratulations, you rightly did so. There is no apparent monotonicity of speedups over field count. Currently, every #soa builtin copies each field with <code>mem_copy(..., ..., type.size)</code>, with type.size read from RTTI. The type info table is emitted as LLVM constant globals, so those loads are foldable in principle, but the index into it is the field loop counter, so the size only becomes a compile-time constant if that loop is fully unrolled first. Whether it unrolls is decided per instantiation by the unroller. When it unrolls, the copies lower to inline movs. When it doesn't, mem_copy stays a CRT memmove call, one per field in the single-element paths, and one per field per element in the batch ones, where the call is in the inner loop. And not-unrolling is where the largest speedups in the tables come from.

This isn't a property of the type layout. An isolated append_soa_elem unrolls up to 4 fields and doesn't unroll at 5+, but the same e.g. 3-field element may not unroll when append_soa_elem is in an actual program. Ditto for unordered_remove_soa, which doesn't even seem to have a threshold for unrolling. So the unrolling decision depends on the surrounding code rather than type layout.

The compiler #soa lowering this commit uses for the single element paths and the typed per-field copies it uses for append_soa_elems, have no such dependency. Offsets and sizes are compile-time constants, with no loop that has to be unrolled to expose them. Cost is monotonic in field count: usually ~2-4x against the current unrolled case, 8–12x against the current slower non-unrolled case.

Now, you may be wondering "Can't we force unroll the type erased loop so that the index is compile-time known all the time?". Yes, we can, using <code>#unroll for</code> with the compile-time known field count, and this was an intermediate step tried before arriving at the uniformly faster typed copies in the commit. Note that #unroll doesn't 100% reliably force unrolling, but boosts the odds of it happening significantly.

"Hmm, somehow it seems to fill my head with ideas, only I don’t exactly know what they are!" Ah, welcome. :)
While scratching my head, trying to squeeze the numbers, I got another bright idea. In the batched appends, the mem_copy is in the inner loop. What if we have a switch on type.size and do a simple fixed width copy for the common field sizes, falling back to mem_copy for uncommon field sizes? Here is a [gist](https://gist.github.com/corleypc/1845a592d1b0bf95472dc70f4aa44527).
So #unroll tries to force type.size itself as a compile-time constant, by unrolling so that the index is folded. It is still subject to the unroller's cost decisions (hence not 100% predictable) and code size is O(field count).
The switch leaves type.size as a runtime value and dispatches to one of several statically-sized copy loops, and code size is field count invariant.
The switch actually does better than #unroll for the common field sizes (it does vectorize), but still falls off the cliff for the "uncommon" ones (the default switch case). And still not close to the typed copies, but it remains a viable idea for the >16 fields types, if they ever come to need a faster treatment.
</details>

**Injections and ordered remove**

These didn't get the same attention as appends and unordered removes, because they should better be avoided altogether.

Single injection got a significant boost mostly due to the committed version avoiding redundant zeroing of the reused memory. Using compiler lowering for the injected copy also boosts performance, particularly for injecting near the end. Measured speedup up to ~2.7x for injecting within ~16 elements from the end.

Batch injections also benefit from avoiding zeroing of reused memory, but this is dominated by doing the copy when injecting larger batches, and the copy is still done with a type erased loop, cause I judged the added complexity wouldn't match the gain for an operation destined to be avoided. Ordered removes get (mostly) small gains (up to ~1.4x) due to the #no_alias changes and minor changes to the tails loops that hoist len above the loop. 

**Important note on array capacity and performance**

By design, each field is stored in contiguous memory with stride between fields of `capacity * size_of(field)`  (+alignment). When the stride is a multiple of 4KB, e.g. a power-of-2 capacity >= 1024 with 4-byte fields, every same-size field's cache line for a given index maps to the same L1 set, and per-element operations degrade (e.g. append_soa, inject_at_soa). It is worst past 8 same-size fields where the burst exceeds the common L1 8-way associativity and the fields collide (measured up to ~7-8x degradation on a type with 18 same-size fields; fields of different sizes advance through the sets at different rates, so they don't collide persistently). SoA iterative flows may also be affected, particularly when the loop touches multiple fields and other hot data aliases the same L1 set. And SoA iterative flows is what we use SoA for in the first place.

So, for types with several fields, code should pad such capacities by at least the number of fields that fit in a CPU cache line.
Modern cache lines are typically 64 (Intel/AMD/Arm64) or 128 bytes (Apple Silicon), so pad with e.g. `128 / size_of(field)` (for the smallest field when field sizes differ), that is, prefer 4128 instead of 4096 for 4-byte fields. Setting cap to e.g. 4097 is NOT enough (it shifts field memory by only 4 bytes (same cache line) and fields keep colliding). The discussion above mentions a 4KB stride and degradation at >8 fields. 4KB is the x86 L1 set period, and some newer Intel cores are 12-way (32–48KB/8–12-way), so the hit would come harder at >12 fields. Apple Silicon (128KB/8-way) has a 16KB period, so 4096 × 4 = 16KB stride is still pathological there and the blanket recommendation of a 128-byte pad fixes it.

**Single element appends (array cap 4096 vs array cap 4128)**

| elem type | SOA fields | slowdown at 4096|
|---|---:|---:|
| `Vec4_f32` | 4 x 4B | **1.77x** |
| `Fat` | 8 x 4B | **3.45x** |
| `Fattest` | 18 x 4B | **7.26x** |

This effect already exists with the current code, there appends for the 18 field struct are ~2x slower at 4096 cap compared to 4128 cap, but the leaner code in the commit exposes the problem much sooner and with a much higher impact on the gains. (In case you wonder, the patched code is still 2-3x faster than current at capacity of 4096 for Fat and Fattest).

Accordingly, the commit adds the power-of-2 warning and the padding recommendation to the doc comment of reserve_soa.

Ideally, this should be handled internally in reserve_soa by allowing allocated capacity to exceed requested capacity.
I didn't do it, since this would break current runtime tests. They explicitly test for the requested cap.

Also note that capacities produced by append growth all the way from empty do `2*cap + 8` per grow and are not affected. (But pre-reserving memory is the proper way.)

**Changes, more specifically**

* Code now takes advantage of the compiler implemented lowering of #soa stores and loads, particularly on operations that do single element stores (append_soa_elem, unordered_remove_soa, inject_at_elem_soa). 
* Typed vectorizer friendly store handling for batch appends (append_soa_elems, non_zero_append_soa_elems) for structs with up to 16 fields. This are enabled for ODIN_OPTIMIZATION_MODE >= .Speed, and, conversely, the compiler lowering is used for ODIN_OPTIMIZATION_MODE <= .Size.
* Tweaked type erased loops over fields, where these loops still remain. While doing this, I realised the existing loops also break silently for #packed structs and introduce corruption. They assume default struct layout. So I took extra care to fix this, with the new version also resulting in an extra 10-15% performance gain.
* Judicious use of #no_alias on the array parameter. In order to have the annotation flow freely, uintptr casts are replaced with multipointer indexing, wherever array is cast for access to fields or footer. uintptr lowers to LLVM's ptrtoint which is conservatively treated as opaque during LLVM alias analysis, multipointer indexing lowers to getelementptr and is transparent during alias analysis.
* Injection procedures (inject_at_elem_soa and inject_at_elems_soa) now call non_zero_resize_soa instead of resize_soa to avoid redundant zeroing, with an explicit mem_zero added for the potential gap that would appear when injecting beyond the last element.
* _reserve_soa() is now #force_no_inline. It is not on the hot path, but the inliner had no scruples inlining it, so this is now explicitly disabled.
* Doc comment on reserve_soa() warns against power-of-2 manual cap reservations.
* The PR adds an extensive test that does a sequence of appends, removes, injections and compares to a mirror reference AoS array at each step.
* The PR now imports runtime in tests/core/speed.odin so that runtime tests are also run at -o:speed. I beleive this should be done even without the changes in the PR, cause there was -o:speed conditioned code in the runtime package already (e.g. AoS dynamic arrays).

Potential improvements not done:
* Overallocating in _reserve_soa() on capacities for which cap * size_of(field) is multiple of 4096. This will prevent cache associativity related performance degradation and I recommend doing it. (And, of course, volunteer to do it.)
* Switch over type.size with compile-time known fixed sizes for append_soa_elems for types with >16 fields, if such field counts ever come to matter.
* Typed store arms for inject_at_elems_soa, similar to append_soa_elems. I deemed the complexity too big, considering this is not a core operation for SoA, and it should generally be avoided.

**Finale**

If you did read all of this, thank you for taking the time. I left most details of what-led-to-what out, as well as oher tried and discarded ideas, cause, really, this would be of interest to performance junkies only. But it still took significant effort to write it down and even bigger one to beautify it, so that the text is remotely readable. :)

### DIFF
--- a/base/runtime/core_builtin_soa.odin
+++ b/base/runtime/core_builtin_soa.odin
@@ -50,13 +50,18 @@ Raw_SOA_Footer_Dynamic_Array :: struct {
 	allocator: Allocator,
 }
 
+// Note: When casting array to access footer/fields
+// uintptr(array) lowers to LLVM ptrtoint and captures pointer provenance,
+// whcih defeats #no_alias on the array in any code following (including in callers).
+// Multipointer indexing lowers to GEP and doesn't capture, so prefer that throughout.
+
 @(builtin, require_results)
 raw_soa_footer_slice :: proc(array: ^$T/#soa[]$E) -> (footer: ^Raw_SOA_Footer_Slice) {
 	if array == nil {
 		return nil
 	}
-	field_count := uintptr(len(E) when intrinsics.type_is_array(E) else intrinsics.type_struct_field_count(E))
-	footer = (^Raw_SOA_Footer_Slice)(uintptr(array) + field_count*size_of(rawptr))
+	field_count := len(E) when intrinsics.type_is_array(E) else intrinsics.type_struct_field_count(E)
+	footer = (^Raw_SOA_Footer_Slice)(&([^]byte)(array)[field_count*size_of(rawptr)])
 	return
 }
 @(builtin, require_results)
@@ -64,8 +69,8 @@ raw_soa_footer_dynamic_array :: proc(array: ^$T/#soa[dynamic]$E) -> (footer: ^Ra
 	if array == nil {
 		return nil
 	}
-	field_count := uintptr(len(E) when intrinsics.type_is_array(E) else intrinsics.type_struct_field_count(E))
-	footer = (^Raw_SOA_Footer_Dynamic_Array)(uintptr(array) + field_count*size_of(rawptr))
+	field_count := len(E) when intrinsics.type_is_array(E) else intrinsics.type_struct_field_count(E)
+	footer = (^Raw_SOA_Footer_Dynamic_Array)(&([^]byte)(array)[field_count*size_of(rawptr)])
 	return
 }
 raw_soa_footer :: proc{
@@ -118,15 +123,13 @@ make_soa_aligned :: proc($T: typeid/#soa[]$E, #any_int length, alignment: int, a
 	}
 	new_data := raw_data(new_bytes)
 
-	data := uintptr(&array)
 	offset := 0
 	for i in 0..<field_count {
 		type := si.types[i].variant.(Type_Info_Multi_Pointer).elem
 
 		offset = align_forward_int(offset, max_align)
 
-		(^uintptr)(data)^ = uintptr(new_data) + uintptr(offset)
-		data += size_of(rawptr)
+		([^]rawptr)(&array)[i] = rawptr(uintptr(new_data) + uintptr(offset))
 		offset += type.size * length
 	}
 	footer.len = length
@@ -174,7 +177,7 @@ make_soa :: proc{
 
 
 @builtin
-resize_soa :: proc(array: ^$T/#soa[dynamic]$E, #any_int length: int, loc := #caller_location) -> Allocator_Error {
+resize_soa :: proc(#no_alias array: ^$T/#soa[dynamic]$E, #any_int length: int, loc := #caller_location) -> Allocator_Error {
 	if array == nil {
 		return nil
 	}
@@ -215,7 +218,7 @@ resize_soa :: proc(array: ^$T/#soa[dynamic]$E, #any_int length: int, loc := #cal
 }
 
 @builtin
-non_zero_resize_soa :: proc(array: ^$T/#soa[dynamic]$E, #any_int length: int, loc := #caller_location) -> Allocator_Error {
+non_zero_resize_soa :: proc(#no_alias array: ^$T/#soa[dynamic]$E, #any_int length: int, loc := #caller_location) -> Allocator_Error {
 	if array == nil {
 		return nil
 	}
@@ -225,6 +228,15 @@ non_zero_resize_soa :: proc(array: ^$T/#soa[dynamic]$E, #any_int length: int, lo
 	return nil
 }
 
+// `reserve_soa` will try to reserve memory of a passed SOA dynamic array to the requested element count (setting the `cap`).
+//
+// For maximizing performance it is recommended to avoid power-of-2 capacities
+// when manually reserving memory for SOA arrays, more so for sizes above 512.
+// For element types with several fields, prefer padding such capacities by at
+// least the number of fields that fit in a CPU cache line.
+// Modern cache lines are typically 64 (Intel/AMD/Arm64) or 128 bytes (Apple Silicon),
+// so pad requested capacity with e.g. `128 / size_of(field)` (for the smallest field when field sizes differ),
+// that is, prefer capacity of 4128 instead of 4096 for 4-byte fields.
 @builtin
 reserve_soa :: proc(array: ^$T/#soa[dynamic]$E, #any_int capacity: int, loc := #caller_location) -> Allocator_Error {
 	return _reserve_soa(array, capacity, true, loc)
@@ -235,7 +247,33 @@ non_zero_reserve_soa :: proc(array: ^$T/#soa[dynamic]$E, #any_int capacity: int,
 	return _reserve_soa(array, capacity, false, loc)
 }
 
-_reserve_soa :: proc(array: ^$T/#soa[dynamic]$E, capacity: int, zero_memory: bool, loc := #caller_location) -> Allocator_Error {
+
+// Note: capacity and performance
+// Each field is stored in contiguous memory with stride between fields capacity * size_of(field) (+ alignment).
+// When the stride is a multiple of 4KB, e.g. a power-of-2 capacity >= 1024 with 4-byte fields,
+// every same-size field's cache line for a given index maps to the same L1 set, and
+// per-element operations degrade (e.g. append_soa, inject_at_soa).
+// It is worst past 8 same-size fields where the burst exceeds the common
+// L1 8-way associativity and the fields collide (measured up to ~7-8x degradation
+// on a type with 18 same-size fields; fields of different sizes advance through
+// the sets at different rates, so they don't collide persistently).
+// Common SoA iterative flows may also be affected, when the loop touches multiple fields
+// and other hot data aliases the same L1 set.
+// So, for types with several fields, prefer padding such capacities by at
+// least the number of fields that fit in a CPU cache line.
+// Modern cache lines are typically 64 (Intel/AMD/Arm64) or 128 bytes (Apple Silicon),
+// so pad with e.g. 128 / size_of(field) (for the smallest field when field sizes differ),
+// so prefer 4128 instead of 4096 for 4-byte fields. Setting cap to e.g. 4097 is NOT enough
+// (it shifts field memory by only 4 bytes (same cache line) and fields keep colliding).
+// Note: 4KB stride in the discussion above is the x86 L1 set period (32–48KB/8–12-way);
+// Apple Silicon (128KB/8-way) has a 16KB period, so cap 4096 × 4 = 16KB stride
+// is still pathological there and the recommended 128-byte pad fixes it.
+// Note: struct size also plays a role, but the user facing padding recommendation
+// on reserve_soa() will handle up to 256 fields (of 4 bytes) decently on 8-way caches.
+// Note: Ideally, this should be handled internally by allowing allocated capacity to exceed requested capacity.
+// This would break current runtime tests though, they explicitly test requested cap.
+// Note: Capacities produced by append growth all the way from empty (2*cap + 8) are not affected.
+_reserve_soa :: #force_no_inline proc(array: ^$T/#soa[dynamic]$E, capacity: int, zero_memory: bool, loc := #caller_location) -> Allocator_Error {
 	if array == nil {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


base/runtime/core_builtin_soa.odin
```diff
@@ -50,22 +50,27 @@ Raw_SOA_Footer_Dynamic_Array :: struct {
 	allocator: Allocator,
 }
 
+// Note: When casting array to access footer/fields
+// uintptr(array) lowers to LLVM ptrtoint and captures pointer provenance,
+// whcih defeats #no_alias on the array in any code following (including in callers).
+// Multipointer indexing lowers to GEP and doesn't capture, so prefer that throughout.
```
gingerBill, line 56: Not a criticism of your changes in the slightest, rather it reminds me how old this code is that I used `uintptr` because multi-pointers did not exist yet when this was written.

base/runtime/core_builtin_soa.odin
```diff
@@ -399,35 +437,13 @@ _append_soa_elem :: proc(array: ^$T/#soa[dynamic]$E, zero_memory: bool, #no_broa
...
-
-			soa_offset  = align_forward_int(soa_offset, max_align)
-			item_offset = align_forward_int(item_offset, type.align)
-
-			dst := rawptr(uintptr(data) + uintptr(soa_offset) + uintptr(type.size * footer.len))
-			src := rawptr(uintptr(arg_ptr) + uintptr(item_offset))
-			mem_copy(dst, src, type.size)
-
-			soa_offset  += type.size * cap(array)
-			item_offset += type.size
```
gingerBill, line 430: Damn is this code that old 🤣

base/runtime/core_builtin_soa.odin
```diff
@@ -464,41 +480,267 @@ _append_soa_elems :: proc(array: ^$T/#soa[dynamic]$E, zero_memory: bool, #no_bro
...
+
+			src_base := uintptr(raw_data(args))
+			for i in 0..<uintptr(FIELD_COUNT) {
+				type := si.types[i].variant.(Type_Info_Multi_Pointer).elem
+
+				dst := uintptr(([^]rawptr)(array)[i]) + uintptr(type.size*offset)
+				src := src_base + se.offsets[i]
+				for j in 0..<arg_len {
+					mem_copy(rawptr(dst + uintptr(j*type.size)), rawptr(src + uintptr(j*size_of(E))), type.size)
+				}
```
gingerBill, line 524: The more I think about it, I think this should probably just be an intrinsic of some sort to guarantee mostly good code generation without the need for optimizations, and for this to work without RTTI too. 

The woes of Odin not having compile-time metaprogramming 😛, but I made the hard stance of saying now long ago.

gingerBill:
Bloody wonderful PR!

PR #7202, opened 2026-08-03
Impl simd arm sha big endian

alexthed1rk:
SHA512 is ARM64 only, according to ARM intrinsics webpage
also `enable_target_feature` here should be `neon,sha3` but i kept as is,
bc I don't have the hw to test it correctly, if it would break the code

### DIFF
--- a/core/simd/arm/sha.odin
+++ b/core/simd/arm/sha.odin
@@ -1,76 +1,218 @@
 #+build arm64,arm32
 package simd_arm
 
+import "core:simd"
+
+// SHA1 hash update accelerator, choose.
+//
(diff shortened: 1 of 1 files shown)


alexthed1rk:
I wish netbsd wouldn't break with error
`Error: Try using '-microarch:native' as Odin defaults to x86-64-v2 (close to Nehalem), and your CPU seems to be older.`
and is it fine to keep `_ :: simd` to surpress vet warning compiling under little endian?
or is it better to close in `when` statement `simd` import?

Yawning:
There are essentially 0 people using big endian ARM.  I would take this PR if it only fixed the sha512 intrinsic availability (changing the `enable_target_feature` is fine).

PR #7216, opened 2026-08-04
Impl simd arm neon bit manipulation

alexthed1rk:
`NEON` `ARM` Bit manipulation
- Count leading sign bits.
- Count leading zero bits.
- Population count per byte.
- Vector bitwise bit clear.
- Bitwise Select.

Everything except `fp16` ones.

### DIFF
--- a/core/simd/arm/neon.odin
+++ b/core/simd/arm/neon.odin
@@ -1,0 +1,596 @@
+#+build arm64,arm32
+package simd_arm
+
+import "core:simd"
+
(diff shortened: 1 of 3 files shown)


Kelimion:
You didn't need to close it as far as I'm concerned. Was there something missing?

alexthed1rk:
I just got overheated by the new heat wave and became depressed.
Nevermind, i would reopen the pr again.

PR #7227, opened 2026-08-05
Improve procedure group overload scoring

kalsprite:
**Draft For Review / Discussion** 

Changes scoring behavior of proc groups.

## Background

Procedure group resolution scores each candidate and picks the highest. Three things in that scoring make it pick the wrong overload, or refuse to pick at all when one candidate is clearly better.

`assign_score_function(d) = max(3*MAXIMUM_TYPE_DISTANCE² + 1 - d², 0)`, so with
`MAXIMUM_TYPE_DISTANCE = 10` a perfect match is worth **301** and the entire quality range across a parameter is ~100. Anything else added to the score at that magnitude swamps the argument signal.

### Change 1. The polymorphic adjustment is inverted

```cpp
// prefer non-polymorphic procedures over polymorphic
item.score += assign_score_function(1);
```

`assign_score_function(1)` is **+300**. The comment says prefer non-polymorphic; the code gives polymorphic candidates a bonus just short of a perfect match. The result is that a generic overload beats an exact concrete one:

```odin
concrete :: proc(x: int) { ... }   // scores 300
generic  :: proc(x: $T)  { ... }   // scores 300 + 300 = 600
g :: proc{concrete, generic}
g(1)   // calls generic
```

Changed to `item.score -= 1` so it tie-breaks instead of dominating.

### Change 2. A synthesised default argument is scored twice

Measured scores for `g(1)`:

| overload | score |
|---|---|
| `proc(x: int)` | 300 |
| `proc(x: int, y := 0)` | **901** |
| `proc(x: int, y := 0, z := 0)` | **1502** |

Each unprovided default is worth **+601** — `assign_score_function(1)` as a dummy-argument bonus, plus `assign_score_function(0)` when the synthesised operand is then scored as a perfect-match argument. So an overload that needs defaults filled in always beats one that matches the call exactly, and adding a defaulted sibling to a group silently steals calls from an existing member.

`dummy_argument_count` already exists for this and is incremented at two sites, but never read. This patch reads it and discounts both contributions.

### Change 3. Untyped constants ignore their default type

```cpp
case Basic_UntypedInteger:
    if (is_type_integer(dst) || is_type_rune(dst)) {
        return 1;      // identical for int, i64, u8, i8, rune, ...
    }
```

Every member of a family gets the same distance, so `int` and `i64` are indistinguishable for `g(1)`. `default_type()` already maps each untyped kind to the right answer; this uses it, giving three tiers: default type (1) < same family (2) < cross-family (3). The third tier is required with only two, `i64 vs f64` regresses into ambiguity.

## Results

40 proc-group resolution cases; `AMBIG` = compiler refused as ambiguous.

| case | baseline | proposed |
|---|---|---|
| `proc(x:int)` vs `proc(x:int, y:=0)`, `g(1)` | defaulted | **exact** |
| `proc(x:int)` vs `proc(x:int, y:=0, z:=0)`, `g(1)` | defaulted | **exact** |
| `proc(x:int, y:=0)` vs `proc(x:int, y:=0, z:=0)`, `g(1)` | more defaults | **fewer defaults** |
| `proc(x:int)` vs `proc(x:int, r:..int)`, `g(1)` | `AMBIG` | **non-variadic** |
| 5-member group, exact match + defaulted sibling | defaulted sibling | **exact** |
| `int` vs `i64`, `g(1)` | `AMBIG` | **int** |
| `f32` vs `f64`, `g(1.5)` | `AMBIG` | **f64** |
| `rune` vs `int`, `g('x')` | `AMBIG` | **rune** |
| `string` vs `cstring`, `g("hi")` | `AMBIG` | **string** |
| `bool` vs `b32`, `g(true)` | `AMBIG` | **bool** |
| `proc(x:int)` vs `proc(x:$T)`, `g(1)` | generic | **concrete** |
| `proc(x:int)` vs `proc(x:$T)`, typed arg | generic | **concrete** |
| `proc(x:string)` vs `proc(x:$T)`, `g("hi")` | generic | **concrete** |

**23/40 → 38/40** cases resolve to the expected overload in a local test suite. No case regressed. The entirety of core/vendor still check okay, but runtime testing has not yet been performed.

Deliberately still ambiguous: `^int` vs `[]int` for `g(nil)` — `nil` genuinely fits both, and
refusing is the right answer. The goal is not to remove ambiguity, only to stop reporting it when one candidate is better.

Two cases remain unfixed, both in polymorphic instantiation rather than scoring, and likely
overlapping #7208:

- a specialised `proc(x: $T/[]$E)` ties with a bare `proc(x: $T)`
- passing a polymorphic procedure to a group whose members take proc-typed parameters is reported ambiguous even when only one member can accept it

## Note on splitting

The three changes are independently implementable: 
- Change 1 is a one-line fix where the code contradicts its own comment and is the least contentious
- Change 2 is the most invasive
- Change 3 is the most user-visible

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -7603,24 +7603,37 @@ gb_internal CallArgumentData check_call_arguments_proc_group(CheckerContext *c, 
 				array_add(&proc_entities, data.gen_entity);
 				index = proc_entities.count-1;
 
-				// Order candidates: value-polymorphic > concrete > type-polymorphic
+				// Order candidates:
+				//   value-polymorphic > concrete > specialized generic > unconstrained generic
 				//
 				// `proc($S: string)` specialises on a compile-time *value*
-				// `proc(x: $T)` specialises on a *type* and is a fallback, so it should
-				// lose to an exact concrete overload
+				// `proc(x: $T)` specialises on a *type* and is a fallback, so it should lose
+				//  to an exact concrete overload
+				// `proc(x: $T/[]$E)` constrains that type, so it is the closer of the two
 				//
-				// Both are small tie-breaks on purpose: assign_score_function(1) is
+				// These are small tie-breaks on purpose: assign_score_function(1) is
 				// ~a full perfect-match unit and would swamp argument match quality.
 				bool has_polymorphic_constant = false;
+				bool has_specialized_generic = false;
 				if (pt->Proc.params != nullptr) {
 					for (Entity *param : pt->Proc.params->Tuple.variables) {
-						if (param != nullptr && param->kind == Entity_Constant) {
+						if (param == nullptr) {
+							continue;
+						}
+						if (param->kind == Entity_Constant) {
 							has_polymorphic_constant = true;
-							break;
+						}
+						Type *bt = base_type(param->type);
+						if (bt != nullptr && bt->kind == Type_Generic && bt->Generic.specialized != nullptr) {
+							has_specialized_generic = true;
 						}
 					}
 				}
-				item.score += has_polymorphic_constant ? +1 : -1;
+				if (has_polymorphic_constant) {
+					item.score += 2;
+				} else {
+					item.score += has_specialized_generic ? -1 : -2;
+				}
 			}
 
 			max_matched_features = gb_max(max_matched_features, matched_target_features(&pt->Proc));
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -7607,10 +7607,24 @@ gb_internal CallArgumentData check_call_arguments_proc_group(CheckerContext *c, 
 				array_add(&proc_entities, data.gen_entity);
 				index = proc_entities.count-1;
 
-				// Prefer non-polymorphic procedures over polymorphic. This must be a
-				// small tie-break: assign_score_function(1) is ~a full perfect-match
-				// unit, so adding it made a generic beat an exact concrete overload.
-				item.score -= 1;
+				// Order candidates: value-polymorphic > concrete > type-polymorphic
+				//
+				// `proc($S: string)` specialises on a compile-time *value*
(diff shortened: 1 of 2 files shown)


A1029384756:
for scoring, one case that you overloooked that i do believe should be considered (and is used in other libraries/bindings like clay) is the following:
```odin
foo :: proc {
  foo_static,
  foo_dynamic,
}

foo_static :: proc($S: string) {}
foo_dynamic :: proc(s: string) {}

// static variant
foo("string") 

// dynamic variant
str := "string"
foo(str)
```
at least based on your written description (and my knowledge of how the current scoring works), this scoring would be inverted — meaning the polymorphic, static variant would never be used. this would break some assumptions that current projects have and would eliminate what in my opinion is, a useful optimization

if there is a way to keep the current priority for `proc($S: string)` and invert it for `proc(val: $T)`, the behavior would be less intrusive (although still a breaking change in some edge cases afaict) without removing this useful functionality


### potential alternative
if breaking changes are being fully considered, a breaking but not uncommonly proposed change in discord (credit @Barinzaya) has been to instead just iterate over the variants in order, picking the first one that satisfies the type system. this would mean:
```odin
foo :: proc {
  foo_static,
  foo_dynamic,
}
```
would pick the static variant if possible and:
```odin
foo :: proc {
  foo_dynamic,
  foo_static,
}
```
would pick the dynamic variant every time. the benefit to this however is that it would allow the user to perform their own scoring/precedence, ordering the statements within the proc group so that their desired outcome is reached

kalsprite:
Scoring updated per your suggestion and it improved the specificity above my commit's; because its a behavior change still causes `test_type_inference_on_literals_with_default_args` to fail in CI.

gingerBill:
If the CI is failing but the test needs updating, update the test.

kalsprite:
Updated to include #7208 changes, metrics now work on the following ordering 
`value-polymorphic > concrete > specialized generic > unconstrained generic`

PR #7087, opened 2026-07-19
Implement fixes for constant unions

korvahkh:
Firstly, thanks to @tf2spi who has been of great help in discussion and ideas for fixes on this issue.

Fixes: #7036 #7083 #6699 #6896 #6897 #6895 #6100

There are a few known issues, which I have left for as an exercise for our future selves in the interest of getting the fixes I have merged sooner:
- #5206
- #7018
- #6122 & #6626, which seem to be a problem with `complex` and not unions specifically, as the they also occur when broadcasting to an array.
- Constant unions of procs e.g.
  ```odin
  f :: proc() {}
  FU : union{proc()} : f
  fu := FU
  fmt.println(fu.(proc()), fu())
  ```
  cause type system confusion about `fu`.
  (I don't know if an issue has been filed for this yet.)

PR #7233, opened 2026-08-05
llvm: preserve nested union source types

antisaling:
Tried to cover all the relevant cases on the repro test this time around.

Not sure if this is the most elegant fix however.


antisaling:
Closed in favor of #7087

PR #7256, opened 2026-08-08
Impl simd arm neon table lookup

alexthed1rk:
all the basic table lookups, not the extended ones


Kelimion:
Looks like github is about to have another outage. The diff won't load.

image

alexthed1rk:
it opens partially the diff for me, but i have to wait a bit

Kelimion:
I'll keep this tab open and will retry later. I'm sure it's fine like your previous PR's, but I do want to read it before merging even so. :D

Kelimion:
Looks like GH has recovered. Merged!

PR #7257, opened 2026-08-08
Return false for identical specialization types

### LINKED ISSUE #7012: `type_is_specialization_of(T, T)` is `true` when it should be `false`.
## Context

```
Odin:    dev-2026-07:a6ba977a3
OS:      Windows 11 Professional (version: 25H2), build 26200.8655
RAM:     32691 MiB
Backend: LLVM 20.1.0
```

## Expected Behavior

`type_is_specialization_of(T, T)` should be `false` since `T` is not a specialization of itself.

## Current Behavior

`type_is_specialization_of(T, T)` returns `true`.

### Steps to Reproduce

```odin
import "base:intrinsics"
Foo :: struct($T: typeid) {x: T}
#assert(intrinsics.type_is_specialization_of(Foo(int), Foo) == true)
#assert(intrinsics.type_is_specialization_of(Foo, Foo)      == false) // Error: Compile time assertion
#assert(intrinsics.type_is_specialization_of(i32, Foo)      == false)
```

michtesar:
Fixes #7012.

`intrinsics.type_is_specialization_of` now returns `false` when both arguments are identical types.

The specialization compatibility check remains unchanged. An explicit identity check is performed before calling it.

The test is registered in both the Unix and Windows issue test runners.

The focused regression test passes:
```sh
./odin check tests/issues/test_issue_7012.odin \
    -file \
    -no-entry-point \
    -vet \
    -strict-style \
    -ignore-unused-defineables
```

The new test also passes as part of tests/issues/run.sh. The full local suite later fails on an unrelated AddressSanitizer linker error in test_issue_5640 on macOS ARM64. Not sure what to with this...

Also, if this is approved and merged there is consecutive ticket about docs https://github.com/odin-lang/pkg.odin-lang.org/issues/74...I could even squeeze it this PR if possible.

### DIFF
--- a/src/check_builtin.cpp
+++ b/src/check_builtin.cpp
@@ -7370,7 +7370,11 @@ gb_internal bool check_builtin_procedure(CheckerContext *c, Operand *operand, As
 
 			operand->mode = Addressing_Constant;
 			operand->type = t_untyped_bool;
-			operand->value = exact_value_bool(check_type_specialization_to(c, s, t, false, false));
+			bool is_specialization = false;
+			if (!are_types_identical(s, t)) {
+				is_specialization = check_type_specialization_to(c, s, t, false, false);
+			}
+			operand->value = exact_value_bool(is_specialization);
 
 		}
 		break;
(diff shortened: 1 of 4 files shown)


Kelimion:
Thanks for this.

https://github.com/odin-lang/pkg.odin-lang.org/issues/74 has to be resolved here: https://github.com/odin-lang/pkg.odin-lang.org/blob/master/intrinsics.odin#L355-L363

Kelimion:
Bug, or was `intrinsics.type_is_specialization_of(Foo, Foo)` intended to be `true`, @gingerBill?

gingerBill:
I guess the question is whether the test is strict or not strict. If it's strict, then `(Foo, Foo) == false`. If it is not strict `(Foo, Foo) == true`. 

I think the more I think about it, it should be strict.

michtesar:
@Kelimion just for the visibility this is also ready https://github.com/odin-lang/pkg.odin-lang.org/pull/76

PR #7266, opened 2026-08-09
vendor: add Box3D support for WebAssembly

nic-vdwalt:
## Summary

- Link Box3D as a vendored WASM object for Odin's JS targets.
- Add Box3D to the JS vendor compilation check.

## Test plan

- Native and `js_wasm32` Box3D checks pass.
- JS vendor check passes with no `system:box3d` imports.

Browser builds currently run Box3D serially; threaded Web Worker support will be investigated separately. File-backed height-field loading remains unavailable in browser WASM.


Kelimion:
Not a bad idea, but in accordance with [vendor:README.md](https://github.com/odin-lang/Odin/blob/master/vendor/README.md), kindly delete the object file.

Also, please just add the build steps to `vendor/box3d/src/build.sh‎` instead of adding a `Makefile` and then calling it from the build script.

nic-vdwalt:
> Not a bad idea, but in accordance with [vendor:README.md](https://github.com/odin-lang/Odin/blob/master/vendor/README.md), kindly delete the object file.
> 
> Also, please just add the build steps to `vendor/box3d/src/build.sh‎` instead of adding a `Makefile` and then calling it from the build script.

Hi Kelimion, thanks for the feedback. I have made the changes as suggested.

Kelimion:
Looks good. If the CI passes doesn't object, I'll merge it and furnish the object file.

nic-vdwalt:
@Kelimion, Would you be open to a follow-up exploring threaded Box3D support for WebAssembly using Web Workers and shared memory, or would you prefer the vendor package remain serial on web?

Kelimion:
> @Kelimion, Would you be open to a follow-up exploring threaded Box3D support for WebAssembly using Web Workers and shared memory, or would you prefer the vendor package remain serial on web?

I use wasm rarely enough that I'll leave that to my other team members who regularly use it to consider.

PR #7270, opened 2026-08-09
Fixes chain indexing for array elements in soa containers

### LINKED ISSUE #5637: Assertion failures when using #soa on 2D arrays
```
package test

main :: proc() {
	foo: #soa[1][1]f32
	f:   f32

	BUG :: 0
	when BUG == 0 {
		f = foo[0][0]
	} else when BUG == 1 {
		for b in foo {
			f = b[0]
		}
	}
}
```

`BUG :: 0`
```
src/llvm_backend_expr.cpp(4550): Assertion Failure: `is_type_multi_pointer(type_of_expr(ie->expr))` [1]f32
```

`BUG :: 1`
```
src/llvm_backend_utility.cpp(1509): Assertion Failure: `is_type_array(st) || is_type_enumerated_array(st) || is_type_matrix(st)` #soa ^#soa[1][1]f32
```

`odin version dev-2025-08:2f0b018db`

corleypc:
Indexing into array elements doesn't work for soa containers.

```odin
package main
import "core:fmt"

main :: proc() {
    a: #soa[3][4]u16
    
    fmt.println(a[0][0])   // compiler error
    a[0][0] = 1            // can't assign error
    p := &a[1][2]          // can't take the pointer address of a[1][2]
    for &v in a {
        fmt.println(v[0])  // compiler error
    }
    
    nested: #soa[2][3][2]u8
    nested[1][0][1] = 42   // multidimensional doesn't compile either
}
```

This PR makes indexing a first class citizen. :)
The groundwork was laid down in https://github.com/odin-lang/Odin/pull/7259, so this is fairly small.

Also fixes https://github.com/odin-lang/Odin/issues/5637

### DIFF
--- a/src/llvm_backend_general.cpp
+++ b/src/llvm_backend_general.cpp
@@ -617,6 +617,12 @@ gb_internal lbAddr lb_addr_soa_variable(lbValue addr, lbValue index, Ast *index_
 }
 
 // pointer to the index element of the field_index component
+//
+// the returned pointer type depends on the soa kind (because the field types do):
+// ^T for StructSoa_Fixed (field is [N]T array), but [^]T for the slice and dynamic
+// kinds (field is the [^]T this offsets);
+// loads and stores work with either, but an lbAddr must not hold the multipointer,
+// so use lb_addr_soa_field_elem to build an lbAddr from this
 gb_internal lbValue lb_soa_field_elem_ptr(lbProcedure *p, lbValue soa_ptr, i32 field_index, lbValue index) {
 	Type *t = base_type(type_deref(soa_ptr.type));
 	GB_ASSERT_MSG(t->kind == Type_Struct && t->Struct.soa_kind != StructSoa_None, "%s", type_to_string(t));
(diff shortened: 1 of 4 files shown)


src/llvm_backend_general.cpp
```diff
@@ -628,6 +634,14 @@ gb_internal lbValue lb_soa_field_elem_ptr(lbProcedure *p, lbValue soa_ptr, i32 f
 	return lb_emit_ptr_offset(p, lb_emit_load(p, field), index);
 }
 
+// lbAddr over an lb_soa_field_elem_ptr pointer, retyped ^T when it came as [^]T
```
corleypc, line 637: `lb_addr_type` derefs with `type_deref(addr.addr.type)`, without allow_multi_pointer, unlike lb_emit_store which passes it; so an lbAddr holding [^]T would report [^]T where the caller expects T, and lb_addr_store would then convert the stored value into it ->  "Invalid type conversion: u16 to [^]u16"

I initally thought of simply doing `type_deref(addr.addr.type, true)` in `lb_addr_type`, and didn't see anything regressing, but ultimately chickened out.

Kelimion: It's lavishingly commented with design decisions. Thanks.

PR #7275, opened 2026-08-10
Impl simd arm neon extended table lookup

alexthed1rk:
all the extended table lookups


alexthed1rk:
@Kelimion could you please review this pr, so i would make
another one with all the logical neon simd?

PR #7277, opened 2026-08-11
Fixes range loop over array elements in soa containers

corleypc:
Range for over array elements doesn't work for soa containers. 

``` odin
package main
import "core:fmt"

main :: proc() {
	a: #soa[2][4]u16
	for v in a[0] {fmt.println(v)}  // compiler error
	for &v in a[0] { _ = v }   // not addressable error
	p := &a[0]
	for v in p^ { _ = v } 	// compiler error 
	for &v in p {fmt.println(v) } 	// compiler error 
}
```

The PR expands on https://github.com/odin-lang/Odin/pull/7270 and fixes this.

### DIFF
--- a/src/llvm_backend_expr.cpp
+++ b/src/llvm_backend_expr.cpp
@@ -5087,26 +5087,11 @@ gb_internal lbAddr lb_build_addr_soa_elem_index(lbProcedure *p, Ast *expr, lbAdd
 		return lb_addr_soa_field_elem(lb_soa_field_elem_ptr(p, soa_addr.addr, cast(i32)component, soa_addr.soa.index));
 	}
 
-	// for non-constant index do chain select between the component pointers;
-	// an element array holds at most 4 components, so this is at most 3 compares and 3 selects (branchless), 
-	// and should be folded if j turns out constant after inlining;
-	// TODO: more efficient codegen can be done, most definitely for fixed kind, possibly for slice/dynamic,
-	// (but this works for both kinds)
-	//
 	// Note: a temp holding the whole element would be simpler but would not be an lvalue,
 	// and writes go through here
 	lbValue index = lb_emit_conv(p, lb_build_expr(p, ie->index), t_int);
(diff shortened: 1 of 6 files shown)


src/llvm_backend_expr.cpp
```diff
@@ -5194,6 +5179,13 @@ gb_internal lbAddr lb_build_addr_index_expr(lbProcedure *p, Ast *expr) {
 		if (array_addr.kind == lbAddr_SoaVariable) {
 			return lb_build_addr_soa_elem_index(p, expr, array_addr, t);
 		}
+		// p[j], where p is an #soa element pointer (e.g. p := &soa[i])
+		if (is_type_soa_pointer(type_of_expr(ie->expr))) {
+			// build a soa variable from the soa ptr to get the address of the component
+			lbValue soa_ptr = lb_addr_load(p, array_addr);
+			lbAddr soa_addr = lb_addr_soa_variable_from_soa_ptr(p, soa_ptr);
+			return lb_build_addr_soa_elem_index(p, expr, soa_addr, t);
+		}
```
corleypc, line 5188: This went under the radar in https://github.com/odin-lang/Odin/pull/7270

src/llvm_backend_stmt.cpp
```diff
@@ -377,11 +377,22 @@ gb_internal void lb_build_when_stmt(lbProcedure *p, AstWhenStmt *ws) {
 
 gb_internal void lb_build_range_indexed(lbProcedure *p, lbValue expr, Type *val_type, lbValue count_ptr,
                                         lbValue *val_, lbValue *idx_, lbBlock **loop_, lbBlock **done_,
-                                        bool is_reverse, i64 unroll_count=0) {
+                                        bool is_reverse, i64 unroll_count=0, lbAddr const *soa_elem=nullptr) {
```
corleypc, line 380: Incidentally, unroll_count is never used in the body, and never passed by callers.

Kelimion:
Very nice work. Thanks again.

PR #7292, opened 2026-08-12
amd64/arm64: classify aggregates from the source type

kalsprite:
**Important:** darwin-arm64 is checked by signature only, I'd request that be run on physical hardware to verify my changes

### issue

Both classifiers ran over the lowered LLVM type, where Odin has already turned padding into an explicit [N x i8] member and a #raw_union into an opaque integer. 
- SysV §3.2.3 says padding contributes no class and a union merges its members' classes
- AAPCS64 §5.9.5 counts a union as a Composite Type

So an f32 alone in an eightbyte was demoted to `integer` by the padding beside it, and a union of floats never reached a floating-point register at all:

```odin
A :: struct { a: i64, b: f32 }
@(export) f :: proc "c" (s: A, next: f64) -> f64 { return next }
```
```
Odin  : define double @f({ i64, i64 }, double)
clang : define double @f(i64, float, double)
next: want 7, got 5.33607e-315
```
The source type still has both facts, so classify it instead, keeping the lowered walk for anything the source path does not handle. #packed, SoA, tagged unions, complex and SIMD are untouched and classify exactly as before.

Verified by linking Odin against clang-compiled callees and running them natively on x86-64 and under qemu-aarch64, in both argument and return position.

### DIFF
--- a/src/llvm_abi.cpp
+++ b/src/llvm_abi.cpp
@@ -599,11 +599,12 @@ namespace lbAbiAmd64SysV {
 	};
 
 	gb_internal void classify_with(LLVMTypeRef t, Array<RegClass> *cls, i64 ix, i64 off);
+	gb_internal void unify(Array<RegClass> *cls, i64 i, RegClass const newv);
 	gb_internal void fixup(LLVMTypeRef t, Array<RegClass> *cls);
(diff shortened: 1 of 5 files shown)


Kelimion:
It passes on real hardware. However, I'll let @gingerBill decide on the patch itself.

PR #7321, opened 2026-08-12
Fixes float to 128-bit int/uint conversion

corleypc:
As a followup of https://github.com/odin-lang/Odin/pull/7319, conversion to 128-bit integers also doesn't work, but apparently this was known (Bill's TODOs in internal_i128.odin)

```odin
package main

import "core:fmt"
import "core:os"

f64_to_u128 :: #force_no_inline proc(f: f64) -> u128 { return u128(f) }
f64_to_i128 :: #force_no_inline proc(f: f64) -> i128 { return i128(f) }
f32_to_u128 :: #force_no_inline proc(f: f32) -> u128 { return u128(f) }

main :: proc() {
	a64 := f64(len(os.args)) // prevent folding, 1.0 at runtime, not representable after addition below
	a32 := f32(len(os.args)) //

	f := f64(1 << 80) + a64  // 2^80, exact in f64, > max(u64)
	g := 1e19 + a64          // 10^19, exact in f64; > max(i64)
	h := f32(1 << 80) + a32  // 2^80, exact in f32

	fmt.printfln("u128(f64 2^80): %d, must be %d", f64_to_u128(f), u128(1) << 80)
	fmt.printfln("i128(f64 -2^80): %d, must be %d", f64_to_i128(-f), -(i128(1) << 80))
	fmt.printfln("i128(f64 1e19): %d, must be %d", f64_to_i128(g), i128(10_000_000_000_000_000_000))
	fmt.printfln("u128(f32 2^80): %d, must be %d", f32_to_u128(h), u128(1) << 80)
}
```

This implements the TODOs with direct port from clang (compiler-rt) [here](https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/fp_fixuint_impl.inc) and [here](https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/fp_fixint_impl.inc), with a small deviation in out of range behavior for signed conversion. I believe this one is more predictable, cause it saturates consistently, while clang's initially wraps on the first overflown bit, and just then saturates from the next bit ownwards.

Also adds a test.

### DIFF
--- a/base/runtime/internal_i128.odin
+++ b/base/runtime/internal_i128.odin
@@ -133,7 +133,7 @@ fixuint :: proc "contextless" ($U: typeid, a: f64) -> U where intrinsics.type_is
 // decompose the f64 into significand and exponent, shift the significand into place,
 // saturate out of range values
 @(private="file")
-fixint :: proc "contextless" ($T: typeid, a: f64) -> T where intrinsics.type_is_integer(T) && !intrinsics.type_is_unsigned(T) {
+fixint :: proc "contextless" ($T: typeid, a: f64) -> T where intrinsics.type_is_integer(T), !intrinsics.type_is_unsigned(T) {
 	BITS             :: 8 * size_of(T)
 	SIGNIFICAND_BITS :: 52
 	EXPONENT_BIAS    :: 1023
--- a/base/runtime/internal_i128.odin
+++ b/base/runtime/internal_i128.odin
@@ -92,9 +92,11 @@ floattidf_unsigned :: proc "c" (a: u128) -> f64 {
 
 
 // f64 -> unsigned integer conversion (truncating toward zero)
-// port of clang/compiler-rt's fp_fixuint_impl.inc;
 // decompose the f64 into significand and exponent, shift the significand into place,
 // saturate out of range values
+//
+// Uses parts of compiler-rt's [fp_fixuint_impl.inc](https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/fp_fixuint_impl.inc), licensed under the [Apache License v2.0 with LLVM Exceptions](https://github.com/llvm/llvm-project/blob/main/compiler-rt/LICENSE.TXT).
+// For a copy of the license, see `LICENSE-compiler-rt.txt`.
 @(private="file")
 fixuint :: proc "contextless" ($U: typeid, a: f64) -> U where intrinsics.type_is_unsigned(U) {
 	BITS             :: 8 * size_of(U)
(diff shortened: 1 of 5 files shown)


base/runtime/internal_i128.odin
```diff
@@ -173,45 +258,3 @@ divti3 :: proc "c" (a, b: i128) -> i128 {
 	return i128((udivmodti4(u128(an), u128(bn), nil) ~ u_s_a) - u_s_a) // negate if negative
 }
 
-
-@(link_name="__fixdfti", linkage=RUNTIME_LINKAGE, require=RUNTIME_REQUIRE)
-fixdfti :: proc "c" (a: u64) -> i128 {
```
corleypc, line 178: Minor bug here for sub-1 magnitude, so rewritten to match the polymorphic unsigned one.

base/runtime/internal_i128.odin
```diff
@@ -91,22 +91,107 @@ floattidf_unsigned :: proc "c" (a: u128) -> f64 {
...
+fixunsdfti :: proc "c" (a: f64) -> u128 {
+	return fixuint(u128, a)
 }
 
 @(link_name="__fixunsdfdi", linkage=RUNTIME_LINKAGE, require=RUNTIME_REQUIRE)
-fixunsdfdi :: #force_no_inline proc "c" (a: f64) -> i128 {
-	// TODO(bill): implement `fixunsdfdi` correctly
-	x := i64(a)
-	return i128(x)
+fixunsdfdi :: proc "c" (a: f64) -> u64 {
```
corleypc, line 186: This and fixdfdi aren't actually used anywhere yet, but here for completeness.

src/llvm_backend_expr.cpp
```diff
@@ -2538,8 +2538,8 @@ gb_internal lbValue lb_emit_conv(lbProcedure *p, lbValue value, Type *t) {
 			TEMPORARY_ALLOCATOR_GUARD();
 
 			auto args = array_make<lbValue>(temporary_allocator(), 1);
-			args[0] = value;
-			char const *call = "fixunsdfdi";
+			args[0] = lb_emit_conv(p, value, t_f64);
+			char const *call = "fixdfti";
```
corleypc, line 2542: Fixed to proper sign and wdith. The lb_emit_conv not strictly needed but added here, so that input formally matches.

base/runtime/internal_i128.odin
```diff
@@ -91,22 +91,107 @@ floattidf_unsigned :: proc "c" (a: u128) -> f64 {
 }
 
 
+// f64 -> unsigned integer conversion (truncating toward zero)
+// port of clang/compiler-rt's fp_fixuint_impl.inc;
```
Kelimion: We need to link/mention this, e.g.:
```
Uses parts of compiler-rt's [fp_fixuint_impl.inc](https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/fp_fixuint_impl.inc), licensed under the [Apache License v2.0 with LLVM Exceptions](https://github.com/llvm/llvm-project/blob/main/compiler-rt/LICENSE.TXT).

For a copy of the license, see `LICENSE-compiler-rt.txt`.
```
And then add `LICENSE-compiler-rt.txt` to `base:runtime`.

corleypc: I'll probably mess this up, so it's better you do it. :)

Kelimion: Done.

Kelimion: Joy!
image

PR #7323, opened 2026-08-12
core/strings: fix substring_from/substring boundary bug

### LINKED ISSUE #7322: `substring_from` fails at `rune_start == rune_count(s)`, and mishandles multi-byte runes there
## Context
```
odin:    dev-2026-08
OS:      Ubuntu 24.10, Linux 6.17.0-19-generic
Backend: LLVM 19.1.1
```

## Expected Behavior

`substring_from(s, rune_start)` should succeed with an empty result when `rune_start == rune_count(s)`, the same way `s[len(s):]` is a valid empty slice.

The doc comment says `ok` reports "whether the rune indexes where in bounds of the original string", which could be read as excluding the one-past-the-end index. But `substring` already treats that index as in bounds, `tests/core/strings/test_core_strings.odin` ([line](https://github.com/odin-lang/Odin/blob/fa342bb54087031fbee88c5ef29739cfe54778f7/tests/core/strings/test_core_strings.odin#L169)) pins `substring("Hello", len("Hello"), len("Hello")) == ("", true)` , so `substring_from` should be consistent with it.

## Current Behavior

For ASCII strings it returns `ok = false` instead of succeeding. For strings with multi-byte runes it's worse: it returns a wrong, non-empty, truncated substring instead of an empty one.

### Steps to Reproduce

1. Run the snippet below.
2. `substring_from("hello", 5)` returns `("", false)` — expected `("", true)`.
3. `substring_from("héllo", 5)` returns `("o", false)` — expected `("", true)`.

Live repro: https://godbolt.org/z/415EzMW1h

### Failure Logs

```odin
package main

import "core:strings"
import "core:fmt"

main :: proc () {
	fmt.println(strings.substring_from("hello", 5))   // want ("", true), got ("", false)
	fmt.println(strings.substring_from("héllo", 5))    // want ("", true), got ("o", false)
}
```

Related to #6189, which reported a similar bug in `substring`, that specific case is already fixed on master, but the same root cause is still present in `substring_from` and for non ASCII strings.

mororo18:
Fixes #7322

### DIFF
--- a/core/strings/strings.odin
+++ b/core/strings/strings.odin
@@ -3494,41 +3494,53 @@ levenshtein_distance :: proc(a, b: string, allocator := context.allocator, loc :
 @(private, require_results)
 internal_substring :: proc(s: string, rune_start: int, rune_end: int) -> (sub: string, ok: bool) #no_bounds_check {
 	sub = s
-	ok  = true
-
 	rune_i: int
 
 	if rune_start > 0 {
-		ok = false
+		start_byte := -1
 		for _, i in sub {
-			if rune_start == rune_i {
-				ok = true
-				sub = sub[i:]
+			if rune_i == rune_start {
+				start_byte = i
 				break
 			}
 			rune_i += 1
 		}
-		if !ok {
-			sub = sub[rune_i:]
-		}
-	}
-
-	if rune_end >= rune_start {
-		ok = false
-		for _, i in sub {
-			if rune_end == rune_i {
-				ok = true
-				sub = sub[:i]
-				break
+		if start_byte < 0 {
+			// One past the last rune is a valid, empty-tail start.
+			if rune_i != rune_start {
+				return "", false
 			}
-			rune_i += 1
+			start_byte = len(sub)
 		}
-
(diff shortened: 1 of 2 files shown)


core/strings/strings.odin
```diff
@@ -3494,41 +3494,53 @@ levenshtein_distance :: proc(a, b: string, allocator := context.allocator, loc :
...
-				sub = sub[:i]
-				break
-			}
-			rune_i += 1
-		}
+	if rune_end < 0 {
+		ok = true
+		return
+	}
+	if rune_end < rune_start {
```
Cammymoop, line 3522: Isn't this just a redundant check as `strings.substring` already tests for this and this proc is internal?

Cammymoop (commented):
Might not be necessary to rewrite the whole function into something longer when just a change from
```odin
sub = sub[rune_i:]
```
to
```odin
sub = sub[len(sub):]
ok = rune_start == rune_i
```
seems to do the trick, but I'm glad to find there's a pr fix for this regardless.

PR #7326, opened 2026-08-13
i386: correct System V aggregate alignment and byval stack slots

kalsprite:
i386 emits `byval ... align 8` for every aggregate argument. That is the amd64 stack slot; i386's is 4, and the over-aligned slot gets four bytes of padding in front of it, shifting the aggregate and every argument after it.

Second commit: 64 bit scalar alignment on i386 System V

`align_of(i64)` and `align_of(f64)` were 8 on every i386 target. The i386 System V
psABI says 4; Windows says 8.


kalsprite:
Replacing this PR with a systematic ABI test harness, same changes to be included.

PR #7334, opened 2026-08-15
riscv64: read a flattened aggregate's members from their real offsets

kalsprite:
The FP calling convention flattens an aggregate by dropping padding, and the dense type it builds is applied by reinterpreting the value's bits from offset zero. When an over-aligned member leaves a gap, that reads it from where the padding used to be, eg `struct #min_field_align(16){i8, f32}` passed its float from offset 4 instead of 16. 

Carry the real offsets and read each member from its own.

### DIFF
--- a/src/llvm_abi.cpp
+++ b/src/llvm_abi.cpp
@@ -16,6 +16,14 @@ struct lbArgType {
 	i64 byval_alignment;
 	bool is_byval;
 	bool no_capture;
+
+	// For RiscV (Optional for others): A `cast_type` is normally applied by reinterpreting the value's 
+	// bits from offset zero. Only correct when the two layouts agree. When an ABI flattens an aggregate 
+	// it drops padding, and the dense type it produces puts the surviving members at different offsets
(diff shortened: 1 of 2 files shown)


src/llvm_abi.cpp
```diff
@@ -16,6 +16,14 @@ struct lbArgType {
 	i64 byval_alignment;
 	bool is_byval;
 	bool no_capture;
+
+	// For RiscV (Optional for others): A `cast_type` is normally applied by reinterpreting the value's 
+	// bits from offset zero. Only correct when the two layouts agree. When an ABI flattens an aggregate 
+	// it drops padding, and the dense type it produces puts the surviving members at different offsets
+	// than they really have, eg `struct #min_field_align(16){i8, f32}` flattens to `{i8, float}`, moving 
+	// the float from offset 16 to offset 4. These give the real byte offset of each `cast_type` element.
+	i64 *coerce_offsets;
```
gingerBill, line 25: Why not just use `Slice<i64> coerce_offsets;`?

PR #7341, opened 2026-08-15
Makfile and Profile Guided Optimization

michalskocz:
Profile Guided Optimization is a clang mechanism that allows you to run your program several times before final compilation. 

These runs create files with metadata about the application's runtime, and optimizations are applied based on this data. For example, if a function has 30 if else statements, PGO checks how many times each case actually happened and, for example, changes the order of if else statements.

I also converted the script from the repo into an actual makefile. The project has two input files, so it's difficult to optimize the build for asynchronous execution...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -321,3 +321,7 @@ CMakeLists.txt
 sandbox/
 
 vendor/box3d/lib/box3d_wasm.o
+
+
+# PGO tmp profile data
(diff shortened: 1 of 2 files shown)


gingerBill:
We use bash scripts BECAUSE makefiles are not portable enough to be reliable.

PR #7351, opened 2026-08-16
Set console codepage to utf-8 (on Windows), and restore old one on exit

Kelimion:
Avoid mojibake.

### DIFF
--- a/base/runtime/core.odin
+++ b/base/runtime/core.odin
@@ -945,5 +945,25 @@ default_assertion_contextless_failure_proc :: proc "contextless" (prefix, messag
 			print_byte('\n')
 		}
 	}
+	when ODIN_CODEPAGE_MAGIC {
+		SetConsoleOutputCP(old_console_codepage)
+	}
 	trap()
+}
+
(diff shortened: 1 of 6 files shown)


blob1807:
Could you also make so the default assertion & oob procs restore the codepage.

I would prefer if Odin, the lang, didn't modify the environment at all.  It'll at least, in most cases, look like it doesn't if those two procs restore it as well.

Kelimion:
> Could you also make so the default assertion & oob procs restore the codepage.

Sure.

> I would prefer if Odin, the lang, didn't modify the environment at all. It'll at least, in most cases, look like it doesn't if those two procs restore it as well.

The only way to avoid that completely (and not have to set the codepage) is to swap out `WriteFile` with `WriteConsoleW` for the console, including a utf-8 to utf-16 conversion, which is objectionable in its own way.

blob1807:
> Sure

Thank you.

> The only way to avoid that completely (and not have to set the codepage) is to swap out `WriteFile` with `WriteConsoleW` for the console, including a utf-8 to utf-16 conversion, which is objectionable in its own way.

Ya I know.  It's change that is better for the general user (including myself).  I'm just just being a bit of a prune about it.  
Having a definable for it, is appreciated though.

PR #7353, opened 2026-08-16
Fixes the %% operator

corleypc:
The floor modulo (remainder) operator doesn't work correctly.

```odin
package main
import "core:fmt"
import "core:os"
main :: proc () {
    b := i8(len(os.args) - 1)   // 0, runtime dependent, so it doesn't fold
    a : i8 = i8(126 + b) %% i8(127)
    fmt.println(a) // prints -3, must be 126
}
```

The reason is, the operator is implemented as `remainder = ((x % y) + y) % y`   (or an llvm sequence of srem -> add -> srem). This is mathematically correct at arbitrary precision (so it works for constants), but the addition can overflow and wrap at finite precision.

Complications arise from the fact that Odin specs %, %% and / with no UB. See [here](https://odin-lang.org/docs/overview/#integer-operators), and note that there is a typo there, so read "most non-negative" as "most negative". In particular, it requires that `min(T) / -1 = min(T)`, `min(T) % -1 = 0`, `min(T) %% -1 = 0`, for any signed integer type T. _None of the three operations actually obeys the spec and they all exhibit UB at min(T)_. The complication comes from the fact, that LLVM's [srem instruction](https://llvm.org/docs/LangRef.html#srem-instruction) is UB when the quotient calculation overflows  (for signed T, min(T) / -1 must be max(T) + 1, so it is unrepresentable in T), and the purpose of srem is to lower to instructions that calculate both quotient and remainder. 

The PR implements %% using a single srem and bit operations for constant divisors (compile time known), and with an additional select for runtime divisors. It is branchless and vectorizable.
To get around the srem UB, the PR exploits the fact that x % -1 = x % 1 = 0 for any x, so it substitutes -1 to 1 for this case, which is defined for srem for any x.

The new version is 1.3x-1.6x faster for up to 32-bit ints (on i7-3520M, LLVM 22), which is expected. But, surprisingly, it runs 2.3x faster for 64-bit ints, which doesn't make any sense on an initial thought. The division/srem is the slowest part by far. You take srem+add+srem, turn it to srem+add+bunch of bit ops (+select) and get it more than 2x faster?! Now, the idiv instruction that srem lowers to on x86 does have excessive latency and disastrous throughput (according to the [Agner Fog tables](https://www.agner.org/optimize/instruction_tables.pdf) for this architecture: 28-103 cycles of latency and 26-88 cycles reciprocal throughput for r64 operands) and it seems that a dependent second idiv fires both of these through the roof and costs towards the bigger number in the range on both accounts.

I haven't fixed the / and % operators to make them respect the spec (no UB), because a fix would make both a bit slower and I'm hesitant to do it, cause these two are no doubt much wider used compared to %%. They both would need a trick similar to the one described above to get around srem's UB.

### DIFF
--- a/src/llvm_backend_expr.cpp
+++ b/src/llvm_backend_expr.cpp
@@ -322,6 +322,74 @@ gb_internal IntegerDivisionByZeroKind lb_check_for_integer_division_by_zero_beha
 }
 
 
+// implements %% (the remainder/floored mod operator) on signed integers;
+// this is branchless and vectorizable, so it also covers vectors
+gb_internal LLVMValueRef lb_emit_signed_floor_mod(lbProcedure *p, LLVMValueRef lhs, LLVMValueRef rhs) {	
+	// constants implement this as srem(srem(x, y) + y, y), which is mathematically correct,
+	// and works for arbitrary precision integers, but the add can wrap at finite precision
+
+	// the Odin spec mandates min(Integer_Type) %% -1 must be 0,
+	// but LLVM has srem(min(Integer_Type), -1) as UB and results in FP exception;
+	// since x %% -1 == 0 for every x, a constant rhs = -1 can fold,
+	// and a runtime -1 can be swapped with 1 (x srem 1 is 0 for every x, no exceptions)	
+	LLVMValueRef minus_one = LLVMConstAllOnes(LLVMTypeOf(rhs));
+	LLVMValueRef safe_rhs = rhs;
+	if (LLVMIsAConstantInt(rhs)) {
+		if (rhs == minus_one) {
+			return LLVMConstNull(LLVMTypeOf(rhs)); // the entire %% op folds to 0
+		}
+	} else {
+		// safe_rhs = (rhs == -1) ? 1 : rhs 
+		// vectorizable construction,
+		// build 1 as neg(-1), this folds for both scalars and vectors
+		LLVMValueRef one = LLVMBuildNeg(p->builder, minus_one, "");
+		LLVMValueRef is_minus_one = LLVMBuildICmp(p->builder, LLVMIntEQ, rhs, minus_one, "");
+		safe_rhs = LLVMBuildSelect(p->builder, is_minus_one, one, rhs, "");
+	}
+	LLVMValueRef r = LLVMBuildSRem(p->builder, lhs, safe_rhs, "");
+	// srem truncs to 0, so r needs a +rhs correction when the operands signs differ (and r != 0)
+	// so we implement
+	// r = lhs % rhs
+	// if (r != 0 && sign(lhs) != sign(rhs))
+	// 	 return r + rhs
+	// else return r
+
+	// the sign of a const rhs is compile time known, which simplifies the correction to a mask
+	LLVMValueRef zero = LLVMConstNull(LLVMTypeOf(lhs));
+	if (LLVMIsAConstantInt(rhs)) {
+		// both operands are constant, so this folds
+		LLVMValueRef rhs_sign = LLVMBuildICmp(p->builder, LLVMIntSLT, rhs, zero, "");
+		// always true in current LLVM, only here for future proofing
+		if (LLVMIsAConstantInt(rhs_sign)) {
+			bool negative_rhs = !LLVMIsNull(rhs_sign);
+			// r has the sign of lhs (trunc mod), we use it here instead of lhs,
+			// cause the r = 0 case works out cleanly (-> mask = 0);
+			// correction only for (neg rhs and pos r) OR (pos rhs and neg r),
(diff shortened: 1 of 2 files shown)


src/llvm_backend_expr.cpp
```diff
@@ -553,9 +619,7 @@ gb_internal bool lb_try_direct_vector_arith(lbProcedure *p, TokenKind op, lbValu
 				if (is_type_unsigned(integral_type)) {
 					z = LLVMBuildURem(p->builder, x, y, "");
 				} else {
-					LLVMValueRef a = LLVMBuildSRem(p->builder, x, y, "");
-					LLVMValueRef b = LLVMBuildAdd(p->builder, a, y, "");
-					z = LLVMBuildSRem(p->builder, b, y, "");
+					z = lb_emit_signed_floor_mod(p, x, y);
```
corleypc, line 622: This is behind an if(false) but updated for completeness.

kalsprite: I had the same bug in my work queue; I went ahead and tested the PR and found it to fix all observed errors I'd aggregated. 

Tests you might consider adding for completeness:

```odin
@(test)
modmod_vec_wide :: proc(t: ^testing.T) {
	// a wider array takes the other lowering: [4]i32 emits `srem <4 x i32>`, [16]i32 emits
	// scalar `srem i32`. This reaches the call site `modmod_vec` does not
	x: [16]i32
	y: [16]i32
	for i in 0..<16 {
		x[i] = i32(i) - 8
		y[i] = i % 2 == 0 ? max(i32) : -max(i32)
	}
	x = not_const(x)
	y = not_const(y)
	r := x %% y
	for i in 0..<16 {
		testing.expectf(t, r[i] == floor_mod(x[i], y[i]),
			"[16]i32 idx %v: %v %%%% %v == %v, want %v", i, x[i], y[i], r[i], floor_mod(x[i], y[i]))
	}
}

@(test)
modmod_assign :: proc(t: ^testing.T) {
	// %%= must agree with %%
	{
		x := not_const(i32(1))
		x %%= not_const(max(i32))
		testing.expect_value(t, x, 1)
	}
	{
		x := not_const(i32(-3))
		x %%= not_const(min(i32))
		testing.expect_value(t, x, -3)
	}
	{
		x := not_const(i8(-7))
		x %%= not_const(i8(3))
		testing.expect_value(t, x, 2)
	}
	{
		// vector form
		x := not_const([4]i32{1, 3, -3, 12})
		y := not_const([4]i32{max(i32), max(i32), 7, max(i32)})
		x %%= y
		testing.expect_value(t, x, [4]i32{1, 3, 4, 12})
	}
}

@(test)
modmod_i16_boundaries :: proc(t: ^testing.T) {
	vals := [?]i16{min(i16), min(i16) + 1, -32000, -300, -7, -3, -1, 1, 3, 7, 300, 32000, max(i16) - 1, max(i16)}
	for x in vals {
		for y in vals {
			if y == 0 { continue }
			if x == min(i16) && y == -1 { continue }
			got := x %% y
			want := floor_mod(x, y)
			testing.expectf(t, got == want, "%v %%%% %v == %v, want %v", x, y, got, want)
		}
	}
}

@(test)
modmod_unsigned_widths :: proc(t: ^testing.T) {
	// for unsigned types %% must match % at every width, including near the maximum
	check :: proc(t: ^testing.T, $T: typeid, loc := #caller_location) {
		vals := [?]T{1, 2, 3, 7, max(T) / 2, max(T) / 2 + 1, max(T) - 1, max(T)}
		for x in vals {
			for y in vals {
				if y == 0 { continue }
				a, b := not_const(x), not_const(y)
				testing.expectf(t, a %% b == a % b,
					"%v: %v %%%% %v == %v, want %v", typeid_of(T), a, b, a %% b, a % b, loc = loc)
			}
		}
	}
	check(t, u8)
	check(t, u16)
	check(t, u32)
	check(t, u64)
}

@(test)
modmod_endian :: proc(t: ^testing.T) {
	// endian-annotated types reach the same lowering through a conversion
	{
		x, y := not_const(i32le(1)), not_const(i32le(max(i32)))
		testing.expect_value(t, x %% y, 1)
	}
	{
		x, y := not_const(i32be(1)), not_const(i32be(max(i32)))
		testing.expect_value(t, x %% y, 1)
	}
	{
		x, y := not_const(i64le(-3)), not_const(i64le(min(i64)))
		testing.expect_value(t, x %% y, -3)
	}
}
```

corleypc: I don't see any setting but perhaps @Kelimion can give you permissions to add commits to the PR?

Kelimion: It's already open.
image

PR #7354, opened 2026-08-16
mem: Preserve alignment when reusing Dynamic_Arena blocks

samuelgozi:
## Summary

- Recompute alignment padding after selecting a recycled `Dynamic_Arena` block.
- Skip recycled blocks that cannot fit the allocation plus its padding.
- Add regression tests for reusable and incompatible recycled blocks.

## Why

After a reset, `unused_blocks` can contain blocks created for a smaller alignment. `_dynamic_arena_cycle_new_block` may select one for a later allocation with a larger alignment, but the allocation path assumed the recycled block was already aligned and reset the margin to zero. That could return misaligned memory.

The allocation path now recalculates the aligned address and margin for every selected block. If the block cannot fit the allocation after padding, it continues to the next block.

## Testing

`./odin test tests/core/mem -vet -vet-tabs -strict-style -vet-style -warnings-as-errors -disallow-do -define:ODIN_TEST_FANCY=false -define:ODIN_TEST_FAIL_ON_BAD_MEMORY=true`

### DIFF
--- a/core/mem/allocators.odin
+++ b/core/mem/allocators.odin
@@ -1813,7 +1813,7 @@ dynamic_arena_alloc_bytes_non_zeroed :: proc(a: ^Dynamic_Arena, size: int, align
 	}
 	memory := align_forward(a.current_pos, uintptr(actual_alignment))
 	margin := int(uintptr(memory) - uintptr(a.current_pos))
-	if a.bytes_left < margin + n {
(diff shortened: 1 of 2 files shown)


Kelimion:
Was this closed by accident?

LemonatedCat:
likely accidental purge of unused branches, thus closing it

PR #7398, opened 2026-08-20
backend: a union nested inside a union panicked with "unfound variant"

### LINKED ISSUE #6895: compiler: panic "unfound variant" when constructing union from distinct type wrapping a union
## Context

Odin:    dev-2026-06:47cf0d3f4
OS:      Arch Linux, Linux 7.0.9-arch2-1
RAM:     64219 MiB
Backend: LLVM 22.1.5

## Expected Behavior

Successful compilation.

## Current Behavior

odin build .
...
src/types.cpp(3516): Panic: unfound variant -> union {Wrapper_A, Wrapper_B} int

### Steps to Reproduce

Create File

```
package main

Base_Union :: union { int, f32 }
Wrapper_A  :: distinct Base_Union
Wrapper_B  :: distinct Base_Union
Outer_Union :: union { Wrapper_A, Wrapper_B }

main :: proc() {
	_ = Outer_Union(Wrapper_A(int(42)))
}
```

Run odin build .

kalsprite:
```odin
package main

Base_Union :: union { int, f32 }
Wrapper_A  :: distinct Base_Union
Wrapper_B  :: distinct Base_Union
Outer_Union :: union { Wrapper_A, Wrapper_B }

main :: proc() {
	_ = Outer_Union(Wrapper_A(int(42)))
}
```
```
src/types.cpp(3541): Panic: unfound variant -> union {Wrapper_A, Wrapper_B} int
```

Closes  #6895.

**Two defects stacked**, which is why the panic named a type that appears nowhere in the outer union.

**1. The peel over-shot.** `lb_build_expr_original_const_type` recursed for as long as the type was a union, so it stripped `Outer_Union(Wrapper_A(int(42)))` all the way down to `int`. The backend then looked for `int` among `Outer_Union`'s variants and panicked.

It now stops at the level that is a variant of the union being built:

```cpp
Type *arg_type = type_of_expr(arg);
if (arg_type != nullptr && union_is_variant_of(type, arg_type)) {
	return arg_type;
}
```

This is pre-existing before #7396.

**2. The inner union then hit the old panic.** With `Wrapper_A` correctly chosen, `lb_const_value`
builds its payload from the value alone, and at that depth the expression is gone, so a scalar whose type is the union has nothing left to peel.

For that case only, the variant is resolved from the value's kind, **and only when exactly one variant matches**:

```
nested, unambiguous inner    Outer(WA(int(7)))   -> 7
nested, ambiguous inner      O(W(i16(3)))        -> still panics, no guess
direct, ambiguous            W(i16(300))         -> 300, via the expression (#7396's path)
```

### DIFF
--- a/src/llvm_backend_const.cpp
+++ b/src/llvm_backend_const.cpp
@@ -791,6 +791,33 @@ gb_internal lbValue lb_const_value_bit_field(lbModule *m, Type *type, Ast *value
 }
 
 
+// The variant a scalar constant belongs to, when the value's kind names exactly one of them. Only
+// used where the expression is no longer available, eg nested unions, that the payload
+// lb_const_value builds from the value alone. Ambiguity is left to the caller
+gb_internal Type *lb_union_variant_from_value_kind(Type *bt, ExactValue const &value) {
+	GB_ASSERT(bt->kind == Type_Union);
+
+	Type *found = nullptr;
+	for (Type *vt : bt->Union.variants) {
+		bool matches = false;
+		switch (value.kind) {
+		case ExactValue_Integer:    matches = is_type_integer(vt);    break;
+		case ExactValue_Float:      matches = is_type_float(vt);      break;
+		case ExactValue_Bool:       matches = is_type_boolean(vt);    break;
+		case ExactValue_String:     matches = is_type_string(vt);     break;
+		case ExactValue_Complex:    matches = is_type_complex(vt);    break;
+		case ExactValue_Quaternion: matches = is_type_quaternion(vt); break;
+		}
+		if (matches) {
+			if (found != nullptr) {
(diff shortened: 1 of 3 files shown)


korvahkh:
I have some work on union constants in #7087 that should in theory address this.

Kelimion:
@korvahkh, do you think you can fix the merge conflicts on your patch? It looks like they were introduced when the asm templates were merged.

korvahkh:
> @korvahkh, do you think you can fix the merge conflicts on your patch? It looks like they were introduced when the asm templates were merged.

I should be able to have a look at it.

kalsprite:
I am going to close this one - we use #7087 instead. i think my impl has a bug in it to boot. #7397 and #7399 i think are complementary to #7087

PR #7401, opened 2026-08-20
nbio(windows): fix use-after-free, spurious poll readiness, and double completion

kalsprite:
Six items were found in the Windows backend tested under windows 10. Before this, the `core:nbio` suite could not complete reliably under the default 25-thread test runner, roughly half of full-suite runs either deadlocked for 60s or died silently with SIGILL. After implementing the fixes in this pr, **25/25 runs clean, zero crashes, zero hangs**.

### `op.l` re-read after the operation is published

`b029cf08a` ("core/nbio: fix use-after-free of op.l in cross-thread exec") fixed this in `exec()` and left a comment explaining it. That commit touched only `nbio.odin`; the identical pattern survived in the cross-thread completion path, where the target loop can complete and recycle the operation before `op.l` is read again.

### Recompute the timeout after the pre-wait drains

The drains can add timeouts, and `timeout_exec` only puts those in `l.timeouts` without posting anything, so the timeout computed before them could be `INFINITE` with a timeout already pending.

### `nbio(windows): clear the OVERLAPPED when arming a poll`

`poll_exec` was the only exec path that didn't reset it, so a recycled `Operation` fed stale `Internal`/`InternalHigh` to `get_result`.

### `nbio(windows): tear down poll waits before their event`

Unregister with `INVALID_HANDLE_VALUE` so an in-flight `wait_callback` can't touch an `Operation` that has already been recycled.

### `nbio(windows): clear the socket event record on poll completion`

`WSAEnumNetworkEvents` is what clears it. Left set, the next `WSAEventSelect` on that socket signals immediately from the stale record, completing a poll for a readiness that never happened and making the caller's send fail with `WOULDBLOCK`. **11/20 failures → 0/20.**

This is also what fixed the deadlocks. Same root cause, other direction: a registration that consumes the stale record never sees the readiness it was waiting for. Isolated against pristine master: `master + crash fix` 5/14 hangs, `+ this` 0/14.

### `nbio(windows): don't complete synchronously-failed overlapped ops twice`

`FILE_SKIP_COMPLETION_PORT_ON_SUCCESS` only elides the completion packet on success. When an overlapped op completed synchronously with a failure status the exec procs returned `.Done`, so the op was completed once inline and once by the packet that was still queued. The second one landing on an `Operation` already recycled into the pool, which tripped the `unreachable()` in the completion dispatch and killed the process with SIGILL. **7/12 crashed runs → 0/12.**

Applies to all seven exec paths; only `.Read` was observed crashing, the rest are the same defect fixed by inspection.

### `nbio(windows): handle queued work in the tick that was woken for it`

A wake, or another loop routing a completion to us, leaves work on the queues that the tick returned without touching. **87 such ticks over 8 runs → 0.**

### `nbio(tests): let wake_up tolerate a tick that returns without progress`

Another loop routing a completion to this one wakes its thread by APC, and the pre-wait drain may already have taken the work, so `tick()` can return having done nothing. Loop until the wake is observed; a lost wake still blocks and trips the fail timeout. Repro without it: `wake_up + client_and_server_send_recv`, 4/10.

### Dropped: the unconditional `_wake_up`

The first version of this PR also made `_wake_up` queue an APC unconditionally, on the theory that wakes sent while the loop was awake were lost. After addressing the lines above, it was an incorrect theorized fix, so reverted it. As @laytan pointed out in comments below, a wake only has to guarantee previously queued work is picked up, and the re-drain after storing `.Sleeping` already covers that. 

### poll sockets through AFD

`WSAEventSelect` is edge triggered, `FD_WRITE` is only recorded again after a send fails with `WOULDBLOCK`. Polling a socket that is already writable waited for an edge that never came and hung the loop. Neither `select` nor `WSAPoll` can stand in, both report a socket writable while its send buffer is full.

AFD is the driver underneath winsock and reports readiness that is already present, matching `poll(2)` and the Linux backend. It also completes on the IOCP, so a poll becomes an ordinary overlapped operation: the event, the thread pool wait, the network event record and the removal special case are all gone.

AFD delivers one poll per socket, which is documented on `poll`. Two polls on one socket never worked on Windows. `WSAEventSelect` silently cancelled the first. `remove_multiple_poll` now uses two sockets, which also makes it a stronger test: both become readable; the removed poll is actually given an event it shouldnt deliver.

### nbio(tests): accept a reset when a cancelled sendfile is closed

Removing a `sendfile` cancels the callback, not the transfer: `TransmitFile` has already queued bytes by the time `CancelIoEx` lands. Closing a socket with that cancelled transfer on it ends the connection with a reset often enough that the peer's `recv` sees `Connection_Closed` rather than a graceful close. I made it accept both.

### `nbio(tests): fill the socket until sending blocks`

The test assumed one 4 GB send returns `Would_Block`. How much a socket absorbs before blocking is not a fixed number, it scales with the size of the individual send request. Measured on one machine:

| request size | absorbed before blocking |
|---|---|
| 4 MiB chunks | 8 MiB |
| 16 MiB chunks | 32 MiB |
| one 4 GiB request | 4294967294 of 4294967296 bytes |

So the old assertion was not passing because 4 GB is a lot, it was passing because a single 4 GiB request landed two bytes inside the limit and only the final call blocked. On other machines those last bytes fit too, `_send_tcp` completes the buffer and returns nil, which is the CI failure. Keep sending until it actually blocks instead of assuming an amount that does.

### Not addressed

The spurious wake is inherent: waking a specific loop thread requires `QueueUserAPC`, and a spurious wake is indistinguishable from a legitimate bare `wake_up()`. Both are an APC with empty queues. Per-loop IOCPs would remove the routing path and the whole class, but would permanently bind each handle to one loop and I think that fundamentally changes the api; so not attempted.

### DIFF
--- a/core/nbio/impl_windows.odin
+++ b/core/nbio/impl_windows.odin
@@ -676,7 +676,7 @@ _remove :: proc(target: ^Operation) {
 
 	switch target.type {
 	case .Poll:
-		win.UnregisterWaitEx(target.poll._impl.wait_handle, nil)
+		win.UnregisterWaitEx(target.poll._impl.wait_handle, win.INVALID_HANDLE_VALUE)
 		target.poll._impl.wait_handle = nil
 
 		ok := win.PostQueuedCompletionStatus(
@@ -1500,9 +1500,14 @@ sendfile_callback :: proc(op: ^Operation) -> Op_Result {
 	return .Done
 }
 
+// Bit indices into `WSANETWORKEVENTS.iErrorCode`, corresponding to the `FD_*` masks.
+FD_READ_BIT  :: 0
+FD_WRITE_BIT :: 1
+
 @(require_results)
 poll_exec :: proc(op: ^Operation) -> Op_Result {
 	assert(op.type == .Poll)
+	op._impl.over = {} // Operations are recycled, clear stale state from a previous use.
 
 	events: i32 = win.FD_CLOSE
 	switch op.poll.event {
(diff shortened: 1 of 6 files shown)


laytan:
Hmm I'm trying to wrap my head around the problem (1). You say a bare wake does not enqueued anything, which is true. But the point of a wake is to tell the event loop "hey there is some work for you", so it only makes sense coupled with actually enqueueing some work.

In other words a wake should guarantee that previously queued work is picked up, nothing more iirc.

So, is this really a problem?

2 obviously is a problem and should be fixed.

laytan:
An example of a use that would go wrong here would help a ton.

Kelimion:
This fixes `nbio` for me. First time I've ever been able to run the test suite without failures or hangs.

image

kalsprite:
@Kelimion was able to cause resource exhaustion on an older laptop with smaller ram pool (8gb iirc). I think this is more the shape of the tests themselves, I'd propose the tests get their own once-over on a separate PR later, and we keep the fixes above limited to just fixes.

laytan:
fwiw even though it is already merged this looks good to me :)

PR #7408, opened 2026-08-20
[core:debug/trace] remove `os`, `fmt`, `slice`, and `strconv` imports

laytan:


laytan:
This may enable us to move the tracking allocator from core:debug/trace to core:mem

A1029384756:
i've given this a deeper look to see what things would look like from an import perspective if the tracking allocator in `core:mem` instead were updated to be the one from `core:debug/trace`. at least on linux, there would be four total (this includes transitive) dependencies on other `core` packages that would be introduced by merging those allocators:
- `debug/trace` (obviously)
- `strings`
- `bytes`
- `simd`

each of these dependencies themselves are relatively lightweight but this still would be an additional few dependencies for anyone importing `core:mem`. that being said, the more exotic targets (freestanding and wasm) already don't work with the `core:mem` tracking allocator so i don't particularly see this being a huge issue

core/debug/trace/allocator.odin
```diff
@@ -290,43 +289,49 @@ tracking_allocator_print_results :: proc(t: ^Tracking_Allocator, temp_allocator
 	ALLOCATOR_MAX_BACKTRACES :: 16
 
 	for _, leak in t.allocation_map {
-		fmt.eprintfln("%v leaked %m", leak.location, leak.size)
+		runtime.print_caller_location(leak.location)
+		runtime.print_string(" leaked ")
+		runtime.print_int(leak.size)
```
Kelimion: I wonder if we should steal the `%m` implementation from `fmt` for `leak.size` to keep that consistent with the current output.

laytan: That's probably nice yeah, let me do that

laytan:
@A1029384756 thanks for looking into it, for now I am planning to keep this package separate. Also keeping in mind the process spawning of addr2line on Linux and the "private" framework on MacOS. Might be too much to have everyone use without an opt-in to this package.

core/debug/trace/allocator.odin
laytan: hmm that does mean strconv probably comes back as a dep

laytan: maybe I can get around it

laytan: I added a more minimal version of the one in core:fmt so I don't need strconv.

A1029384756:
@laytan don't mean to necro this but `core:mem` (which this package depends on) already has a `core:strconv` dependency through a direct `core:sys/posix` dependency. though i guess that's only the case on non-windows platforms

laytan:
@A1029384756 core:sys/posix depends on core:strconv? I would not expect that.

I do see that core:strings depends on core:strconv so it is already in the dependency tree. I should have checked. But eh, it's fine.

PR #7409, opened 2026-08-20
Fix reordered compound literal comparison

### LINKED ISSUE #7260: Incorrect results in constant compound literal comparisons
Odin:    dev-2026-08:1a3d2dc09
	OS:      Arch Linux, Linux 7.1.5-zen1-2-zen
	CPU:     AMD Ryzen 7 2700X Eight-Core Processor
	RAM:     15906 MiB
	Backend: LLVM 22.1.8

## Repro

This program:

```odin
package main

import "core:fmt"

main :: proc() {
	A :: [2]int
	fmt.println(A{0=0, 1=1} == A{1=1, 0=0})
}
```

outputs `false` instead of `true`.

This is caused by `compare_exact_values_compound_lit` simply iterating the values in order, which doesn't correctly handle `field=value` compound literals.

michtesar:
Fixes #7260

The problem was in the compiler function that was comparing the elements based on the order not by its actual values.

Now we get the real index for each element. We sort them and then compare them.

This is my largest change so far so I would appretiate a code review especially on time complexity as I am not entirely sure.

As I saw the comparison parts, I think we could have compared an implicit zero, e.g.:

```odin
A{0 = 1}
A{0 = 1, 1 = 0}
```

and found out that the ranges are not covered:

```odin
A{0..<2, = 1}
A{0 = 1, 1 = 1}
```

These two topics are not in the PR. Not sure if we shall make a ticket(s) for it...

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -13035,6 +13035,48 @@ gb_internal bool is_exact_value_zero(ExactValue const &v) {
 
 
 
+struct IndexedCompoundLitElem {
+	i64  index;
+	Ast *elem;
+};
+
+gb_internal bool compound_lit_elem_index(Type *type, Ast *elem, isize positional_index, i64 *index) {
+	if (elem->kind != Ast_FieldValue) {
+		*index = cast(i64)positional_index;
+		return true;
+	}
+
+	ast_node(fv, FieldValue, elem);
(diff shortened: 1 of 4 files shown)


Kelimion:
Thanks, Michael

PR #7412, opened 2026-08-21
fix incorrect offset_of and size_of for unions with cyclic pointers

### LINKED ISSUE #7008: Incorrect Offsets for Reads and Writes to and From Struct Fields Following Unions Including Structs With Pointers to the Original Struct
## Context

	Odin:    dev-2026-07:a60a77f57
	OS:      Arch Linux, Linux 7.1.3-zen1-1-zen
	CPU:     AMD Ryzen 9 3900X 12-Core Processor            
	RAM:     31988 MiB
	Backend: LLVM 22.1.8

## Expected Behavior

Correct program output. Better description in the reproductions.

## Current Behavior

What is the current behavior?

Incorrect program output.

Offsets to fields in struct A after a field of a union containing a struct B with a pointer to struct A are incorrectly computed.

`offset_of` and `size_of` look correct, reflection also looks correct as confirmed by `printfln`, the original use case of json deserialization, and looking at a dump of the struct by aliasing it as a byte slice.

### Steps to Reproduce

1. step 1
2. step 2
3. you get it...

Some examples:

```odin
package main;

import "core:fmt";
import "core:slice";

Bar :: struct
{
	a: i32,
	b: ^Foo,
}

Foo :: struct
{
	a: Maybe(Bar),
	b: string
}

main :: proc ()
{
	foo: Foo;
	foo.b = "deadbeef"; // writes to an incorrect address
	fmt.printfln("%i", size_of(Foo));
	fmt.printfln("%i", offset_of(foo.b));
	fmt.printfln("%#v", foo);
	fmt.printfln("%s", foo.b); // reads from the same incorrect address
	fmt.printfln(
		"%02X",
		slice.bytes_from_ptr(&foo, size_of(Foo)) // display the actual memory of the struct
	);
}
```

```odin
package main;

import "core:fmt";
import "core:slice";

Bar :: struct
{
	a: i32,
	b: ^Foo,
}

Foo :: struct
{
	a: Maybe(Bar),
	b: string
}

main :: proc ()
{
	foo: Maybe(Foo) = Foo{ b = "deadbeef" }; // oddly enough, writes to the correct address
	fmt.printfln("%i", size_of(Maybe(Foo)));
	fmt.printfln("%i", offset_of(Foo, b));
	switch v in foo
	{
	case Foo:
		fmt.printfln("%#v", v);
		fmt.printfln("%s", v.b); // reads from the incorrect address
	case nil:
		fmt.println("Diddly squat");
	}
	fmt.printfln(
		"%02X",
		slice.bytes_from_ptr(&foo, size_of(Foo)) // display the actual memory of the struct
	);
}
```

Some counter-examples:

```odin
package main;

import "core:fmt";
import "core:slice";

Baz :: struct
{
	a: u64,
	b: u64,
	c: u64,
	d: u64,
	e: u64,
}

Bar :: struct
{
	a: i32,
	b: ^Baz, // another struct of the same size as Foo
}

Foo :: struct
{
	a: Maybe(Bar),
	b: string
}

main :: proc ()
{
	foo: Foo;
	foo.b = "deadbeef";
	fmt.printfln("%i", size_of(Foo));
	fmt.printfln("%i", offset_of(foo.b));
	fmt.printfln("%#v", foo);
	fmt.printfln("%s", foo.b);
	fmt.printfln(
		"%02X",
		slice.bytes_from_ptr(&foo, size_of(Foo)) // display the actual memory of the struct
	);
}
```

```odin
package main;

import "core:fmt";
import "core:slice";

Bar :: struct
{
	a: i32,
	b: ^[5]u64, // something that isn't even a struct at all
}

Foo :: struct
{
	a: Maybe(Bar),
	b: string
}

main :: proc ()
{
	foo: Foo;
	foo.b = "deadbeef";
	fmt.printfln("%i", size_of(Foo));
	fmt.printfln("%i", offset_of(foo.b));
	fmt.printfln("%#v", foo);
	fmt.printfln("%s", foo.b);
	fmt.printfln(
		"%02X",
		slice.bytes_from_ptr(&foo, size_of(Foo)) // display the actual memory of the struct
	);
}
```

```odin
package main;

import "core:fmt";
import "core:slice";

Baz :: distinct Foo; // with a distinct Foo

Bar :: struct
{
	a: i32,
	b: ^Baz,
}

Foo :: struct
{
	a: Maybe(Bar),
	b: string
}

main :: proc ()
{
	foo: Foo;
	foo.b = "deadbeef";
	fmt.printfln("%i", size_of(Foo));
	fmt.printfln("%i", offset_of(foo.b));
	fmt.printfln("%#v", foo);
	fmt.printfln("%s", foo.b);
	fmt.printfln(
		"%02X",
		slice.bytes_from_ptr(&foo, size_of(Foo)) // display the actual memory of the struct
	);
}
```

```odin
package main;

import "core:fmt";
import "core:slice";

Bar :: struct
{
	a: i32,
	b: ^Foo,
}

Foo :: struct
{
	a: Bar, // when Bar is not considered part of a union
	b: string
}

main :: proc ()
{
	foo: Foo;
	foo.b = "deadbeef";
	fmt.printfln("%i", size_of(Foo));
	fmt.printfln("%i", offset_of(foo.b));
	fmt.printfln("%#v", foo);
	fmt.printfln("%s", foo.b);
	fmt.printfln(
		"%02X",
		slice.bytes_from_ptr(&foo, size_of(Foo)) // display the actual memory of the struct
	);
}
```

### Failure Logs

The programs compile, no additional logs.

mocompute:
Fixes #7008

This was an interesting couple of hours, and gave me a chance to learn a bit of the compiler internals. I found that `lb_sizeof` will return 0 for the size of a variant block when the type is a struct that contains a pointer to a *forward* type, for reasons I don't fully understand. I can guess that it's simply a void type at that stage of the pipeline because of the forward reference. Here's a simple example:

```
Bar :: struct
{
	a: ^Foo,
}

Maybe_Bar :: union {
	Bar,
}

Foo :: struct {
	a: Maybe_Bar,
	b: string,
}
```

When `lb_type_internal_union_block_type` builds up the LLVM struct type for Maybe_Bar, it incorrectly calculates that additional padding is required, because it thinks the size of the union is 0 (before adding the size of the tag). If Bar.a were declared as `^int`, OR if the declaration of Foo is move BEFORE Bar, the code returns the correct non-zero size.

My proposed fix is to detect the 0 size condition, and fall back to using `type_size_of` on the first variant of the union.

While working with the code, I noticed that the special branch for the single variant case was duplicated code. The call to `lb_type_internal_union_block_type` at the top of the branch delegates to `lb_type` immediately, making the general branch identical to the single-variant branch. So I removed the latter.

Note that the fix must be applied in the general branch. For example, adding:

```
Baz :: struct {

}

Maybe_Bar :: union {
	Bar,
	Baz,
}
```

triggers the same 0-size bug, so it needs to be handled in the general case.

Oh, I forgot to mention: I verified the extra padding before the fix, and the correct padding afterwards, by looking at `-build-mode:llvm`. I ran the ABI tests locally, but I'm not exactly sure how to make a test for this. I'm happy to work on adding a test if I could get some guidance on it.

### DIFF
--- a/src/llvm_backend_general.cpp
+++ b/src/llvm_backend_general.cpp
@@ -2770,26 +2770,17 @@ gb_internal LLVMTypeRef lb_type_internal(lbModule *m, Type *type) {
 			if (is_type_union_maybe_pointer(type)) {
 				LLVMTypeRef variant = lb_type(m, type->Union.variants[0]);
 				array_add(&fields, variant);
-			} else if (type->Union.variants.count == 1) {
-				LLVMTypeRef block_type = lb_type(m, type->Union.variants[0]);
-
-				LLVMTypeRef tag_type = lb_type(m, union_tag_type(type));
-				array_add(&fields, block_type);
-				array_add(&fields, tag_type);
-				i64 used_size = lb_sizeof(block_type) + lb_sizeof(tag_type);
-				i64 padding = size - used_size;
-				if (padding > 0) {
-					LLVMTypeRef padding_type = lb_type_padding_filler(m, padding, align);
-					array_add(&fields, padding_type);
-				}
-				is_packed = true;
 			} else {
 				LLVMTypeRef block_type = lb_type_internal_union_block_type(m, type);
 
 				LLVMTypeRef tag_type = lb_type(m, union_tag_type(type));
 				array_add(&fields, block_type);
 				array_add(&fields, tag_type);
-				i64 used_size = lb_sizeof(block_type) + lb_sizeof(tag_type);
+				i64 block_size = lb_sizeof(block_type);
+				if (block_size == 0) {
+					block_size = type_size_of(type->Union.variants[0]);
+				}
+				i64 used_size = block_size + lb_sizeof(tag_type);
 				i64 padding = size - used_size;
 				if (padding > 0) {
 					LLVMTypeRef padding_type = lb_type_padding_filler(m, padding, align);
--- a/tests/issues/run.bat
+++ b/tests/issues/run.bat
@@ -40,6 +40,7 @@ set COMMON=-define:ODIN_TEST_FANCY=false -file -vet -strict-style -ignore-unused
 ..\..\..\odin test ..\test_issue_6753.odin %COMMON%  || exit /b
 ..\..\..\odin check ..\test_issue_6874.odin %COMMON% 2>&1 | find /c "Error:" | findstr /x "1" || exit /b
 ..\..\..\odin check ..\test_issue_6979.odin -no-entry-point %COMMON%  || exit /b
+..\..\..\odin test ..\test_issue_7008.odin %COMMON%  || exit /b
 ..\..\..\odin check ..\test_issue_7012.odin -no-entry-point %COMMON% || exit /b
 ..\..\..\odin build ..\test_issue_7037.odin %COMMON% -o:none  || exit /b
 ..\..\..\odin build ..\test_issue_7188.odin %COMMON%  || exit /b
--- a/tests/issues/run.sh
+++ b/tests/issues/run.sh
@@ -88,6 +88,7 @@ else
 	exit 1
 fi
 $ODIN check ../test_issue_6979.odin -no-entry-point $COMMON_CHECK
+$ODIN test ../test_issue_7008.odin $COMMON
 $ODIN check ../test_issue_7012.odin -no-entry-point $COMMON_CHECK
 $ODIN build ../test_issue_7037.odin $COMMON -o:none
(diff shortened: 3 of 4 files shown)


mocompute:
I have an idea for a failing Odin test to add: test that writing a byte field and then directly reading the byte using offset_of returns a different result. I will have a look at that today.

mocompute:
I've added a regression test which fails on master and demonstrates the bug. I've also changed the title of this PR to be more about the view from Odin rather than compiler internals.

PR #7419, opened 2026-08-22
core/hash: add murmur3 hash implementation

Lusori0:
This PR adds the murmur3 hash functions, as discussed with @Yawning and also adds tests for all 3 versions of the hash.
The implementaiton is ported directly from https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp

As a first time contributor to the project, any and all feedback is welcome :)

### DIFF
--- a/core/hash/murmur3.odin
+++ b/core/hash/murmur3.odin
@@ -1,0 +1,286 @@
+package hash
+
+rotl32 :: #force_inline proc "contextless" (x: u32, r: u8) -> u32 {
+  return (x << r) | (x >> (32 - r))
+}
+
+rotl64 :: #force_inline proc "contextless" (x: u64, r: u8) -> u64 {
+  return (x << r) | (x >> (64 - r))
+}
+
+fmix32 :: #force_inline proc "contextless" (h: u32) -> u32 {
+  h := h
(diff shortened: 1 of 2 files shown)


core/hash/murmur3.odin
```diff
@@ -0,0 +1,286 @@
...
+
+    h1 ~= k1
+    h1 = rotl32(h1, 13)
+    h1 = h1 * 5 + 0xe6546b64
+  }
+
+  // TAIL
+  k1 : u32 = 0
+  switch(len & 3) {
+    case 3: k1 ~= u32(data[nblocks*4+2]) << 16; fallthrough
```
Kelimion: Applies to the entire patch: We use tabs to indent and spaces to align.

Applies to all of the switches: `case` is aligned with `switch`.

It's also a bit strange to see `case: k1 ~= u32(data[nblocks*4])`, and then continue on the next line. It's a block pretending to be a single-line `case`... that then continues. Indent all of these statements under the `case`, please, not split like this.

Lusori0: Thank you for the feedback. I hope I understood all your points correctly and didn't miss anything.

PR #7428, opened 2026-08-23
add regression test for #7336

### LINKED ISSUE #7336: `[dynamic; N]matrix` missing tail padding, `offset_of` and codegen disagree, SIGSEGV on `movaps`
## Context

        Odin:    dev-2026-08:8412dc37a
        OS:      Debian GNU/Linux 12 (bookworm), Linux 7.0.6-1.qubes.fc37.x86_64
        CPU:     Intel(R) Core(TM) Ultra 7 155U
        RAM:     5810 MiB
        Backend: LLVM 19.1.7

and on

        Odin:    dev-2026-08:8412dc37a
        OS:      Debian GNU/Linux 13 (trixie), Linux 6.12.48+deb13-amd64
        RAM:     31841 MiB
        Backend: LLVM 19.1.7

also crashes on `Odin:    dev-2026-08:db635cab1 (master)`

## Expected Behavior

does not crash, `offset_of(S, b)` and the offset used by generated code should agree

## Current Behavior

segfaults, field placed after `[dynamic; 1]matrix[4, 4]f32` lands 24 bytes earlier than `offset_of` reports, that leaves it misaligned and `movaps` faults

```
package crash

import "core:fmt"

S :: struct {
	a: [dynamic; 1]matrix[4, 4]f32,
	b: matrix[4, 4]f32,
}

main :: proc() {
	s := new(S)
	fmt.println(offset_of(S, b))                            // 96
	fmt.println(uintptr(rawptr(&s.b)) - uintptr(rawptr(s))) // 72
	s.b *= s.b
}

```

```
> gdb crash.bin
...
Program received signal SIGSEGV, Segmentation fault.
0x000055555557ebd4 in crash::main () at main.odin:14
14              s.b *= s.b
(gdb) p &s.b
$1 = (struct matrix[4, 4]f32 *) 0x5555555b6320
(gdb) x/i $pc
=> 0x55555557ebd4 <crash::main+436>:    movaps 0x48(%rax),%xmm10
(gdb) p/x $rax + 0x48
$2 = 0x5555555b6308
(gdb)
```

```movaps``` requires alignment of 16 but doest get it (```s.b``` is misaligned) -> crash

debug symbols report `&s.b` at address ```...320``` but in reality its at ```...308```

i believe the ```[dynamic; 1]matrix[4, 4]f32``` is missing tail padding and causes `s.b` to be misaligned

Possibly related to #7303 

### Steps to Reproduce

```
> odin run crash
96
72
[1]    9188 segmentation fault  odin run crash
```

### PR DESCRIPTION
**Pull request by mocompute**
I was investigating another open bug that referred to problems with `offset_of`, #7336, and discovered that it had been fixed. The large merge #7327 seems to have addressed the issue. I've added a regression test to catch the specific misalignment cited in the issue.

Closes #7336.
Fixed by #7327.

### DIFF
--- a/tests/issues/test_issue_7336.odin
+++ b/tests/issues/test_issue_7336.odin
@@ -1,0 +1,20 @@
+// Tests issue #7336 https://github.com/odin-lang/Odin/issues/7336
+package test_issues
+
+import "core:testing"
+
+S :: struct {
+	a: [dynamic; 1]matrix[4, 4]f32,
+	b: matrix[4, 4]f32,
+}
+
+@(test)
+test_offset :: proc(t: ^testing.T) {
+	s := new(S)
+	defer free(s)
+
+	// 4*4 f32 + dynamic len == 72
+	expect: uintptr = 4 * 4 * 4 + size_of(int)
+	testing.expect(t, offset_of(S, b) == expect)
+	testing.expect(t, uintptr(rawptr(&s.b)) - uintptr(rawptr(s)) == expect)
+}
--- a/tests/issues/run.bat
+++ b/tests/issues/run.bat
@@ -44,6 +44,7 @@ set COMMON=-define:ODIN_TEST_FANCY=false -file -vet -strict-style -ignore-unused
 ..\..\..\odin check ..\test_issue_7012.odin -no-entry-point %COMMON% || exit /b
 ..\..\..\odin check ..\test_issue_7260.odin -no-entry-point %COMMON% || exit /b
 ..\..\..\odin test ..\test_issue_bool_comparison_truthiness.odin %COMMON%  || exit /b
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: tests/issues/test_issue_7336.odin [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+	b: matrix[4, 4]f32,
+}
+
+@(test)
+test_offset :: proc(t: ^testing.T) {
+	s := new(S)
+	defer free(s)
+
+	// 4*4 f32 + dynamic 8 == 72
+	testing.expect(t, offset_of(S, b) == 72)
```
gingerBill: This test will only work on 64-bit targets.

mocompute: I will fix.

PR #7440, opened 2026-08-24
Fixes -o:size

corleypc:
The -o:size optimization level currently doesn't work, it occasionally produces code bigger than o:speed.

This PR fixes it by adding the optsize attribute per procedure at this optimization level. Optsize is then consulted by most of the passes.

On Linux, x64, default target, -o:size, LLVM 22, compiling examples/demo.odin I get
| demo.odin      | code section   | .rodata | filesize   |
|---------------------------|--------:|--------:|--------:|
| before   | 331350 | 29248  | 540200 |
| after | 240342 | 26736  | 447832 |

or around 27% smaller code and around 17% smaller binary.

Note that one of the tests (union_in_aggregates in tests/internal/test_union_const.odin) segfaults at -o:size (Linux, x64, default target (SSE), LLVM 22) due to hitting one of the compiler's alignment bugs. On my pc the test is working by accident at o:speed, -o:minimal and -o:none (an align-1 alloca's (u in the test) field happens to land at alignment 16 (type alignment required and assumed for indirect args), which satisfies the movdqa instruction issued, but -o:size happens to change the stack layout and the same align-1alloca lands at alignment 8, causing the segfault). This bug is fixed in https://github.com/odin-lang/Odin/pull/7382 by copying underaligned indirect args to aligned temp before passing.

### DIFF
--- a/src/build_settings.cpp
+++ b/src/build_settings.cpp
@@ -464,6 +464,15 @@ enum IntegerDivisionByZeroKind : u8 {
 	IntegerDivisionByZero_AllBits,
 };
 
+// values of BuildContext.optimization_level;
+// matches Odin_Optimization_Mode in checker.cpp
+enum OptimizationLevel : i32 {
+	OptimizationLevel_None       = -1,
+	OptimizationLevel_Minimal    =  0,
+	OptimizationLevel_Size       =  1,
+	OptimizationLevel_Speed      =  2,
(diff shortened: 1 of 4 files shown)


corleypc:
Looks like Intel based CI test failures are unrelated and come from https://github.com/odin-lang/Odin/pull/7434

PR #7447, opened 2026-08-25
Investigate win wycheproof

mocompute:
Not for merging: I am testing a theory about the flaky wycheproof test.


mocompute:
Interesting: removing -microarch:native from the wycheproof tests seems to avoid the failure, and the remaining tests (still with -microarch:native) succeed.

The other changes I made, I suspect, are red herrings: ignoring non-fatal win32 exceptions, accepting Close and Break on the Control-C path.

The CI that passed, I also removed all non-windows CI. It would be pretty frustrating if that were a factor.

mocompute:
Clue: https://forums.developer.nvidia.com/t/nvc-ice-on-github-runners-zen-4-without-avx512-capability-when-lowering-vector-ldexp/379026

LLVM 20.1.0 *may* be detecting AVX512 support on this runner even though the VM doesn't support it

Kelimion:
Can you put the rest of the targets and tests back again? Just Windows and Wycheproof is a bit sparse.

mocompute:
I've converted the PR to draft to avoid confusion. Right now I'm just using this PR to test the GitHub runners. The code changes to the test runner should be innocuous, though. I can pull them out into a separate proper PR if you think they ought to be merged once fully tested.

Kelimion:
I think we can merge the Wycheproof `-microarch:native` removal (or replace it with `-microarch:x86-64-v3` (or `x86-64-v2`)), as well as the rest runner exception changes, so if you want to add that to a separate PR, that'd be lovely.

PR #7455, opened 2026-08-25
Remove microarch:native from 3 Windows tests and tweak Win test runner

mocompute:
Following #7447, we suspect an issue on LLVM instruction set discovery on certain GitHub test runners, namely the AMD EPYC 9V74. This PR removes the microarch:native build flag from three tests that have seen intermittent failures.

Further research should investigate whether microarch flags like `x86-64-v3` or `x86-64-v2` also mitigate the issue.

The PR also adds some additional checks to the Windows test runner to maintain correctness in case of unexpected signals.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -264,19 +264,18 @@ jobs:
       - name: Wycheproof tests
         shell: cmd
         run: |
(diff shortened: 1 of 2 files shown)


mocompute:
The Windows GH runner that just completed these tests successfully used AMD EPYC 9V74, so that's encouraging, but since these are intermittent failures, we won't know for sure until many more runs have completed.

Kelimion:
Thanks, Mo.

PR #7461, opened 2026-08-26
fill left-justified field with spaces

kalsprite:
core/fmt: fill a left-justified field with spaces, not '0'

`%-5d` of 42 printed `42000`. The padding for a left-justified field goes to the right of the digits, so a '0' fill is read back as part of the number rather than as filler.

`fmt_write_padding` chose its fill byte from `fi.space` alone and never consulted `fi.minus`, even though both flag parsers already record that `-` suppresses zero fill (`fi.minus = true; fi.zero = false`, and `fi.zero = !fi.minus`).

Right-justified zero fill is Odin's own convention and is unchanged.

## Before / after

| verb | value | before | after |
| --- | --- | --- | --- |
| `%-5d`    | `42`    | `42000`            | `42   ` |
| `%-6d`    | `-42`   | `-42000`           | `-42   ` |
| `%-8x`    | `255`   | `ff000000`         | `ff      ` |
| `%-10b`   | `5`     | `1010000000`       | `101       ` |
| `%-10.1e` | `300.0` | `3.0e+02000`       | `3.0e+02   ` |
| `%-14e`   | `3.0`   | `3.000000e+0000`   | `3.000000e+00  ` |
| `%-8.0m`  | `1 TiB` | `1tib0000`         | `1tib    ` |
| `%-5.0d`  | `0`     | `00000`            | `     ` |
| `%-8v`    | `[2]int{1,2}` | `[10000000, 20000000]` | `[1       , 2       ]` |

### DIFF
--- a/core/fmt/fmt.odin
+++ b/core/fmt/fmt.odin
@@ -1037,7 +1037,9 @@ fmt_write_padding :: proc(fi: ^Info, width: int) {
 	}
 
 	pad_byte: byte = ' '
-	if !fi.space {
+	if !fi.space && !fi.minus {
+		// a left-justified field pads to the right of the digits, where a '0' would read
+		// as part of the number rather than as filler
 		pad_byte = '0'
 	}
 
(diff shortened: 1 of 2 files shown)


LemonatedCat:
%-5.0d with a value of 0 printing *only* spaces seems rather problematic

PR #7462, opened 2026-08-26
This Fixes #5939 , where many characters like flags are reported with…

### LINKED ISSUE #5939: `core:unicode/utf8` grapheme_count reported width seems to be wrong for complex emojis
## Context

I was trying to add emojis support on my terminal editor when I faced something that I think might be an issue for the `grapheme_count` reported `width` for complex emojis like "🏳️‍🌈".

Odin:    dev-2025-11:e5153a937
OS:      NixOS 25.11 (Xantusia), Linux 6.17.7
CPU:     AMD Ryzen 9 5950X 16-Core Processor
RAM:     32008 MiB
Backend: LLVM 20.1.8

## Expected Behavior

I was expecting the result `1, 4, 2` for respectively `graphemes, runes, width`.

## Current Behavior

I got `1, 4, 1`.

It seems for be the same for `decode_grapheme_clusters`.
For the `decode_grapheme_iterate` it seems to be a bit different, the reported `grapheme` seems wrong, but if you check the `last_...` in the iterator struct you get (I think) the right values.

Thank you.

MoonlightChase:
… width 1 even though their width is 2. Country glags need additional special case also since RI letters are EAW neutral and not Wide.
i was working on [this](https://github.com/MoonlightChase/bellerophon) when i stumbled on this bug, so i did bunch of different characters to see which ones cause issue, hence test has many different options. I could make it either simpler or split it, as needed.

### DIFF
--- a/core/unicode/utf8/grapheme.odin
+++ b/core/unicode/utf8/grapheme.odin
@@ -46,7 +46,6 @@ Grapheme_Iterator :: struct {
 	last_rune:                  rune,
 	last_rune_breaks_forward:   bool,
 
-	last_width:                 int,
 	last_grapheme_count:        int,
 
 	bypass_next_rune:           bool,
@@ -58,6 +57,9 @@ Grapheme_Iterator :: struct {
 
 	current_grapheme:           Grapheme,
 	continue_grapheme:          bool,
+
+	current_cluster_width:      int,
+	current_cluster_ri_count:   int,
 }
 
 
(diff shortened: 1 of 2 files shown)


core/unicode/utf8/grapheme.odin
```diff
@@ -2,22 +2,22 @@ package utf8
...
-ZERO_WIDTH_JOINER                 :: unicode.ZERO_WIDTH_JOINER
-is_control                        :: unicode.is_control
-is_hangul_syllable_leading        :: unicode.is_hangul_syllable_leading
-is_hangul_syllable_vowel          :: unicode.is_hangul_syllable_vowel
-is_hangul_syllable_trailing       :: unicode.is_hangul_syllable_trailing
-is_hangul_syllable_lv             :: unicode.is_hangul_syllable_lv
-is_hangul_syllable_lvt            :: unicode.is_hangul_syllable_lvt
-is_indic_conjunct_break_extend    :: unicode.is_indic_conjunct_break_extend
-is_indic_conjunct_break_linker    :: unicode.is_indic_conjunct_break_linker
+ZERO_WIDTH_JOINER :: unicode.ZERO_WIDTH_JOINER
```
Kelimion: Please restore the original alignment.

core/unicode/utf8/grapheme.odin
```diff
@@ -100,12 +94,12 @@ Returns:
 decode_grapheme_clusters :: proc(
 	str: string,
 	track_graphemes := true,
-	allocator       := context.allocator,
+	allocator := context.allocator,
```
Kelimion: Please restore the original alignment.

core/unicode/utf8/grapheme.odin
```diff
@@ -117,8 +111,8 @@ decode_grapheme_clusters :: proc(
 	}
 
 	grapheme_count = it.grapheme_count
-	rune_count     = it.rune_count
-	width          = it.width
+	rune_count = it.rune_count
+	width = it.width
```
Kelimion: Please restore the original alignment.

core/unicode/utf8/grapheme.odin
```diff
@@ -222,7 +240,7 @@ decode_grapheme_iterate :: proc(it: ^Grapheme_Iterator) -> (text: string, graphe
 			// GB7:  (LV | V)  ×  (V | T)
 			// GB8: (LVT | T)  ×   T
 			if is_hangul_syllable_leading(this_rune) ||
-			   is_hangul_syllable_lv(this_rune)      ||
+			   is_hangul_syllable_lv(this_rune) ||
```
Kelimion: And again.

Basically please just redo the patch without touching the whitespace on lines you didn't change or add.

Kelimion:
It feels like there's some autoformatting going on, and apart from destroying intentional alignment, it also means that we have to review the entire file instead of the couple of lines that pertain to the bug fix.

MoonlightChase:
> It feels like there's some autoformatting going on, and apart from destroying intentional alignment, it also means that we have to review the entire file instead of the couple of lines that pertain to the bug fix.

Yes, let me check i think ols and helix are the reason, il try to disable formating all together the realign code and push it again. Is that okay?

Kelimion:
> > It feels like there's some autoformatting going on, and apart from destroying intentional alignment, it also means that we have to review the entire file instead of the couple of lines that pertain to the bug fix.
> 
> Yes, let me check i think ols and helix are the reason, il try to disable formating all together the realign code and push it again. Is that okay?

Sure thing.

MoonlightChase:
> > > It feels like there's some autoformatting going on, and apart from destroying intentional alignment, it also means that we have to review the entire file instead of the couple of lines that pertain to the bug fix.
> > 
> > 
> > Yes, let me check i think ols and helix are the reason, il try to disable formating all together the realign code and push it again. Is that okay?
> 
> Sure thing.

I hope its better now, i am not very familiar with contributing rules, might have rushed a bit. I changed a test bit to be more in line with other tests i saw in the file.
should have 🇺🇸🇬
🇺🇸🇬🇧 
🏳️‍🌈 and so on.

PR #7468, opened 2026-08-27
checker: Do not deduplicate cases in tagless switches

### LINKED ISSUE #7421: False duplicate case error when switching on type alias
## Context
See title

        Odin:    dev-2026-08-nightly:902106f
        OS:      Ubuntu 26.04 LTS, Linux 7.0.0-29-generic
        CPU:     Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
        RAM:     15403 MiB
        Backend: LLVM 20.1.8

## Expected Behavior

To be able to case a type at least once

## Current Behavior

What is the current behavior?
Only allows 2 unique cases. Anything over 2 unique cases produces a duplicate case error for each extra unique case.

### Steps to Reproduce

`package main

import "core:fmt"

FOO :: u16

main :: proc() {
	switch {
	case FOO == u16:
		fmt.println("FOO == u16")
	case FOO == i32:
		fmt.println("FOO == i32")
	case FOO == string:
		fmt.println("FOO == string")
	case:
		fmt.println("FOO == else")
	}
}`

### Failure Logs

/home/tausy/projects/odin_bug/main.odin(13:7) Error: Duplicate case 'FOO == string'
        previous case at /home/tausy/projects/odin_bug/main.odin(11:7)
        case FOO == string:

mac119:
## Summary

- skip duplicate-value tracking for tagless switches, whose cases are ordered predicates
- preserve duplicate-case diagnostics for explicitly tagged switches
- add positive and negative regression coverage to the Unix and Windows issue-test runners

## Root cause

The checker models a tagless switch with an implicit constant `true` tag. Distinct constant predicates can therefore fold to the same Boolean value and collide in the value-switch `SeenMap`, producing a false `Duplicate case` diagnostic.

Duplicate-value tracking now runs only when the source switch has an explicit tag.

## Testing

- `make release-native`
- positive regression: tagless type-comparison predicates compile and select the expected branch
- negative regression: a tagged Boolean switch still emits exactly one `Error: Duplicate case`
- `tests/issues/run.sh`: the new tests and all preceding issue tests pass; the final existing ASan test cannot link against the local macOS ARM64/LLVM 19 ASan runtime, and passes when run without ASan

Fixes #7421

### DIFF
--- a/src/check_stmt.cpp
+++ b/src/check_stmt.cpp
@@ -1167,6 +1167,8 @@ gb_internal void check_switch_stmt(CheckerContext *ctx, Ast *node, u32 mod_flags
 	ast_node(ss, SwitchStmt, node);
 
 	Operand x = {};
+	// Tagless switch cases are independent predicates, not case values.
+	bool check_duplicate_cases = ss->tag != nullptr;
 
 	mod_flags |= Stmt_BreakAllowed | Stmt_FallthroughAllowed;
 	check_open_scope(ctx, node);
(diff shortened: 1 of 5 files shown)


mac119:
Hi, Regarding the pr i created, do you have any questions to be clarified? @Kelimion

Kelimion:
> Hi, Regarding the pr i created, do you have any questions to be clarified? @Kelimion

No, it's fine. Hadn't had a chance to look at it in detail.
Thanks for the patch.

PR #7480, opened 2026-08-29
`core:net`: use explicitly passed allocators

### LINKED ISSUE #7453: `net.resolve*` procs hidden allocations
`resolve`, `resolve_ip4` and `resolve_ip6` in `core:net` may end up calling `get_dns_records_from_os` or `get_dns_records_from_nameservers`, both of which have an allocator argument. However, the resolve procs don't have an explicit allocator argument and use `context.temp_allocator` when calling the `get_dns...` procs:
https://github.com/odin-lang/Odin/blob/ee68d39c203d6fe0d10fcd55f4caf22685358af7/core/net/dns.odin#L112
https://github.com/odin-lang/Odin/blob/ee68d39c203d6fe0d10fcd55f4caf22685358af7/core/net/dns.odin#L114
(similar for `resolve_ip6`)

This behavior isn't documented and as far as I can tell goes against the usual "no hidden allocations in core" rule.

FourteenBrush:
Procedures were explicitly using the context.allocator, now allows passing in an allocator, and actually uses it
Fixes #7453

### DIFF
--- a/core/net/dns.odin
+++ b/core/net/dns.odin
@@ -620,7 +620,8 @@ validate_hostname :: proc(hostname: string) -> (ok: bool) {
 	return true
 }
 
-parse_record :: proc(packet: []u8, cur_off: ^int, filter: DNS_Record_Type = nil) -> (record: DNS_Record, ok: bool) {
+parse_record :: proc(packet: []u8, cur_off: ^int, filter: DNS_Record_Type = nil, allocator := context.allocator) -> (record: DNS_Record, ok: bool) {
+	context.allocator = allocator
 	record_buf := packet[cur_off^:]
 
 	srv_record_name, hn_sz := decode_hostname(packet, cur_off^, context.temp_allocator) or_return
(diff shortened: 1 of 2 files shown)


Kelimion:
`split_url` could also use the `split_iterator` and avoid the allocation altogether.

FourteenBrush:
@Kelimion Got rid of all temporary allocations (didn't notice these could use a context.temp_allocator in the first place).
Tests still pass, also tested with some additional urls.

Just noting this for completeness sake, but an url allows multiple values for each query component (as in ?x=y&x=z), while we consider the queries as a map[string]string, I suppose this "should" in fact be a map[string][]string?

Kelimion:
> @Kelimion Got rid of all temporary allocations (didn't notice these could use a context.temp_allocator in the first place). Tests still pass, also tested with some additional urls.

Thanks.
 
> Just noting this for completeness sake, but an url allows multiple values for each query component (as in ?x=y&x=z), while we consider the queries as a map[string]string, I suppose this "should" in fact be a map[string][]string?

I suppose it should be, but for now let's merge this. :-)

PR #7489, opened 2026-09-01
Impl simd arm neon shift left

alexthed1rk:
All neon shift left bindings


alexthed1rk:
@Kelimion could you please review this pr?

Kelimion:
> @Kelimion could you please review this pr?

Sure. Is it ready? I see you've just added some more to it.

alexthed1rk:
Yes, I've done all the left shifts and because 5 days have passed I decided to add some right shifts as well into pr.
Initially I only wanted to do left shifts only, because there is already a lot of code to review, and make an extra pr
with the right shifts next.

PR #7491, opened 2026-09-01
Add LLVM 23 compatibility

kuvaus:
Hi,

LLVM 23 was recently released and this PR adds Odin compatibility with the LLVM 23 backend.

In LLVM 23, branching `LLVMBr` was divided into two types: `LLVMUncondBr` and `LLVMCondBr`, depending on whether the branching was unconditional or conditional.

This PR checks if `LLVM_VERSION_MAJOR` is 23 or newer and handles both types. The `#else` clause also preserves backward compatibility with older LLVM versions.

### Changes:

In `src/llvm_backend_general.cpp`,  line 554:
```cpp
#if LLVM_VERSION_MAJOR >= 23
        case LLVMUncondBr:
        case LLVMCondBr:
#else
        case LLVMBr:
#endif
```

In `src/llvm_backend_stmt.cpp`,  line 2807:
```cpp
#if LLVM_VERSION_MAJOR >= 23
        auto opcode = LLVMGetInstructionOpcode(if_instr);
        GB_ASSERT(opcode == LLVMUncondBr || opcode == LLVMCondBr);
#else
        GB_ASSERT(LLVMGetInstructionOpcode(if_instr) == LLVMBr);
#endif
```

LLVM 23 compatibility was also added to `build_odin.sh`:
```sh
SUPPORTED_LLVM_VERSIONS="23 22 21 20 19 18 17"
SUGGESTED_LLVM_VERSION="23"
```

Though maybe we could stay in 22 for now.

### Testing:
After the small changes above, I successfully built the compiler on Intel macOS using Homebrew's LLVM 23.1.0. I tested using small odin program with `if` clauses (`CondBr`) and one with `for`, `while`  loops and `break`s (`UncondBr`). Both test programs worked as they should. I did not do a full test on everything but maybe this PR could serve at least as a starting point for LLVM 23 compatibility.


laytan:
Please look at previous update prs to see what else you need to add. Hint: microarch and optimization updates.

kuvaus:
Oh, it became quite a bit more involved than just a couple of lines.
I'll take a look, but I may be out of my depth here. And thanks a lot for pointing the way: I just saw the commit and your comments on the LLVM 22 compatibility pr.

PR #7501, opened 2026-09-02
Fix #6966

### LINKED ISSUE #6966: dev-2026-07 broke `using _: bit_field ...` initialization
## Context

        Odin:    dev-2026-07-nightly:ab0131c
        OS:      Ubuntu Stonking Stingray (development branch), Linux 7.0.0-15-generic
        CPU:     Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
        RAM:     31950 MiB
        Backend: LLVM 20.1.8

## Expected Behavior

Initialization of "using" bit_field fields should be initialized with given values (as worked before).

## Current Behavior

A "using" bit_field fields are zero when initializing them as part of whole struct.

Please consider following code:
```odin
package main

import "core:fmt"

main :: proc () {
    Strata :: enum i8 {
        background  = -1,
        base        = 0,
        high        = 1,
        overlay     = 2,
        tooltip     = 3,
    }

    View :: struct {
        flags: Flags,
        using _: bit_field u16 {
            strata  : Strata    | 4,
            level   : int       | 12,
        },
        size: [2] f32,
    }

    Flags :: bit_set [Flag; u32]
    Flag :: enum { debug }

    fmt.println(ODIN_VERSION, ODIN_VERSION_HASH)

    v := View { flags={.debug}, strata=.high, level=777, size={111,222} }
    fmt.println(v)

    v.strata = .overlay
    v.level = 888
    fmt.println(v)
}
```

I have run this code in new `dev-2026-07` and prev `dev-2026-06` releases, the results are different:
```
> odin run .
dev-2026-07 ab0131c
View{flags = bit_set[Flag; u32]{debug}, _ = bit_field{strata = "base", level = 0}, size = [111, 222]}
View{flags = bit_set[Flag; u32]{debug}, _ = bit_field{strata = "overlay", level = 888}, size = [111, 222]}

odin run .
dev-2026-06 7ab61e4
View{flags = bit_set[Flag; u32]{debug}, _ = bit_field{strata = "high", level = 777}, size = [111, 222]}
View{flags = bit_set[Flag; u32]{debug}, _ = bit_field{strata = "overlay", level = 888}, size = [111, 222]}
```

As you see, the difference:
- `bit_field{strata = "base", level = 0}` -- zero value in the new release
- `bit_field{strata = "high", level = 777}` -- correct value in the prev release

Extra observations:
- if we use named "using", e.g. `using x: bit_field u16` -- nothing changes (zero value stays)
- if we use named "using" (or don't use "using") and setup value as `x={strata=.high, level=777}` -- the value is correct (not zero), so this is probably some new issue with "using" + "bit_field" (?) as i tried "using" + "struct" (e.g. "using _: struct { ... }") and no issues found.

Kelimion:
Fix #6966

### DIFF
--- a/src/check_expr.cpp
+++ b/src/check_expr.cpp
@@ -10587,7 +10587,7 @@ gb_internal void check_compound_literal_field_values(CheckerContext *c, Slice<As
 						ft = bt->Array.elem;
 						break;
 					case Type_BitField:
-						// is_constant = false;
+						is_constant = false;
 						ft = bt->BitField.fields[index]->type;
 						break;
 					default:
--- a/tests/internal/test_union_const.odin
+++ b/tests/internal/test_union_const.odin
@@ -244,3 +244,38 @@ union_named_constants :: proc(t: ^testing.T) {
 		inner = {Promoted_Value(int(1)), 0, 0},
 	})
 }
+
+@(test)
+union_anonymous_using_constants :: proc(t: ^testing.T) {
+	S :: struct {
+		using a: bit_field u32 {
+			x: u32 | 32,
+		},
+	}
+	s := S{x = 1}
+	testing.expect_value(t, s.x, 1)
+
+	Strata :: enum i8 {
+		background  = -1,
+		base        = 0,
+		high        = 1,
+		overlay     = 2,
+		tooltip     = 3,
+	}
+
+	View :: struct {
+		flags: Flags,
+		using _: bit_field u16 {
+			strata  : Strata    | 4,
+			level   : int       | 12,
+		},
+		size: [2] f32,
+	}
+
+	Flags :: bit_set [Flag; u32]
+	Flag :: enum {debug}
+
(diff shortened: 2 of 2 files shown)


Kelimion:
This fixes the regression introduced in baef272bbd08c365fe961abd76084227fc91c0d4 and doesn't appear to break anything else. Do you know of a better way to do this, @gingerBill?

Edit: It can possibly be more narrowly scoped to set `is_constant = false;` only when dealing with a `using _: bf` bit_field, and left alone when it's named or there's no `using` involved. Either that or fix the regression on the backend side so it also works for anonymous using.

gingerBill:
Well the idea was to try and get bit fields to be constant, so making them non-constant kind of defeats the point.

Kelimion:
> Well the idea was to try and get bit fields to be constant, so making them non-constant kind of defeats the point.

True, but the regression it introduced also defeats the point, to be fair. :-D

PR #7536, opened 2026-09-10
Fix #7526 : Handle triple quote/tick string in odin/core

### LINKED ISSUE #7526: `core:odin` Doesn't support triple quote string.
## Context

```
Odin:    dev-2026-09:104048516
OS:      Windows 11 Home Basic (version: 25H2), build 26200.9168
CPU:     AMD Ryzen 7 6800HS with Radeon Graphics
RAM:     15608 MiB
Backend: LLVM 20.1.0
```

## Expected Behavior

Should be supported.

## Current Behavior

`core:odin` doesn't support triple quote/backtick strings. 

### Steps to Reproduce

1. Copy Paste
2. Run
3. Profit...

```odin
package missing_multiline_string_support

import "core:fmt"
import "core:odin/ast"
import "core:odin/parser"

main :: proc() {
    pkg, ok := parser.parse_package_from_path(".")
    assert(ok)
    fmt.println("ast.inspect")
    ast.inspect(&pkg.node, print)
}

print :: proc(n: ^ast.Node) -> bool {
    if n == nil { return false }

    Val: #partial switch val in n.derived {
    case ^ast.Value_Decl:
        #partial switch ident in val.names[0].derived {
        case ^ast.Ident:
            if ident.name != "str" && ident.name != "raw" {
                break Val
            }

            #partial switch lit in val.values[0].derived {
            case ^ast.Basic_Lit:
                fmt.printfln(
                    "\t%q %v: %q", 
                    ident.name, lit.tok.kind, lit.tok.text
                )
            }
        }
    } 
    return true
}

str :: """
       multiline string
       """
raw :: ```
       multiline raw string
       ```
```
The above will print:
```
main.odin(10:10): string literal was not terminated
main.odin(12:10): string literal was not terminated
ast.inspect
        "str" String: "\"\""
        "raw" String: "``"
```

tGautot:
Fix #7526 which I also ran into today

Update the tokenizer in core/odin to handle triple-quote/tick strings, also added one small test to cover this use-case

This is my first PR for Odin, nice to meet you all, let me know if I missed anything :)

### DIFF
--- a/core/odin/tokenizer/tokenizer.odin
+++ b/core/odin/tokenizer/tokenizer.odin
@@ -327,35 +327,81 @@ scan_identifier :: proc(t: ^Tokenizer) -> string {
 	return string(t.src[offset : t.offset])
 }
 
-scan_string :: proc(t: ^Tokenizer) -> string {
+scan_string :: proc(t: ^Tokenizer, quote: rune) -> string {
 	offset := t.offset-1
-	
-	triple_quoted := t.ch == rune('"') && peek_rune(t, 0) == rune('"')
-	
-	if triple_quoted {
-		advance_rune(t)
-		advance_rune(t)
-	}
 
-	for {
-		ch := t.ch
-		if (!triple_quoted && ch == '\n') || ch < 0 {
-			error(t, offset, "string literal was not terminated")
-			break
+	exit : {
+	if quote == '"' {
+		// Python-style triple-quoted string literal `"""..."""`.
+		if t.ch == '"' && peek_rune(t, 0) == '"' {
+			advance_rune(t) // consume the second opening `"`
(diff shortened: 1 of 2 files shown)


gingerBill:
I'd highly recommend just copying the logic of the C++ compiler DIRECTLY because it also gives good error message handling.

tGautot:
Thanks for the recommendation, went ahead with it. `scan_string` is now pretty much a Odin translated copy-paste of the compiler logic.

tGautot:
I went through the different commands ran by the CI checks and discovered the `tests/issues` folder. I am not sure what is most aligned with what is usually done on the project: leave the tests where I put them, move them to their own `tests/issues` test file, or duplicate them to be in both?

Issue #7305, opened 2026-08-12
M012 — Wide SIMD approximate reciprocal returns incorrect lanes


mat888:
# M012 — Wide SIMD approximate reciprocal returns incorrect lanes

## Context

`odin report` output:

```text
Odin:    dev-2026-08-nightly:6d5dbdce1
OS:      Arch Linux, Linux 7.1.4-arch1-1
CPU:     AMD Ryzen 7 7800X3D 8-Core Processor
RAM:     31179 MiB
Backend: LLVM 22.1.8
```

## Expected Behavior

approximately `1`, `0.5`, and `0.0078125`.

## Current Behavior

approximately `0.99975586`, `0.06248474`, and `0.06248474`.

## Failure Information

Wide SIMD approximate reciprocal produces incorrect values in selected lanes.

### Steps to Reproduce

1. `main.odin`:

```odin
package main

import "base:intrinsics"
import "core:fmt"

main :: proc() {
	x: #simd[8]f32 = {1, 2, 4, 8, 16, 32, 64, 128}
	y := intrinsics.simd_approx_recip(x)
	fmt.println(intrinsics.simd_extract(y, 0), intrinsics.simd_extract(y, 1), intrinsics.simd_extract(y, 7))
}
```

### Failure Logs

approximately `0.99975586`, `0.06248474`, and `0.06248474`.

### Compiler-option matrix

All of these invocations exit `0` and print approximately `0.99975586 0.06248474 0.06248474`:

```text
odin run main.odin -file -thread-count:1                  -> 0.99975586 0.06248474 0.06248474
odin run main.odin -file -thread-count:4                  -> 0.99975586 0.06248474 0.06248474
odin run main.odin -file -thread-count:1 -no-threaded-checker -> 0.99975586 0.06248474 0.06248474
odin run main.odin -file -thread-count:1 -o:none         -> 0.99975586 0.06248474 0.06248474
odin run main.odin -file -thread-count:1 -o:speed        -> 0.99975586 0.06248474 0.06248474
odin run main.odin -file -thread-count:1 -o:aggressive   -> 0.99975586 0.06248474 0.06248474
```

The wrong SIMD lanes are invariant across the tested options.

## Root cause

In the current `src/llvm_backend_proc.cpp` helper `lb_llvm_simd_bulk_op_unary`, the outer chunk loop and inner lane loop both use `i`:

```cpp
for (unsigned i = 0; i < parts_count; i++) {
	...
[...]

PR #7548, opened 2026-09-12
Fixes simd_approx_recip and simd_approx_recip_sqrt

corleypc:
simd_approx_recip and simd_approx_recip_sqrt build lane masks incorrectly when the lane count is either larger or smaller than the default intrinsic width. Minimal example for simd_approx_recip in https://github.com/odin-lang/Odin/issues/7305, which this also fixes.

### DIFF
--- a/src/llvm_backend_proc.cpp
+++ b/src/llvm_backend_proc.cpp
@@ -1522,17 +1522,13 @@ gb_internal bool lb_llvm_simd_bulk_op_unary(lbProcedure *p, lbValue arg, LLVMVal
 
 	LLVMValueRef val = arg.value;
 	unsigned count = cast(unsigned)vt->SimdVector.count;
-	Type *elem = base_type(vt->SimdVector.elem);
 
 	if (count < default_width) {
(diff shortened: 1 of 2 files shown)


src/llvm_backend_proc.cpp
```diff
@@ -1522,17 +1522,13 @@ gb_internal bool lb_llvm_simd_bulk_op_unary(lbProcedure *p, lbValue arg, LLVMVal
...
 	if (count < default_width) {
 		LLVMValueRef *grow_indices   = gb_alloc_array(temporary_allocator(), LLVMValueRef, default_width);
 		LLVMValueRef *shrink_indices = gb_alloc_array(temporary_allocator(), LLVMValueRef, count);
 
 		for (unsigned i = 0; i < count; i++) {
-			ExactValue idx = is_type_float(elem) ?
-				exact_value_float(cast(f64)i) :
-				exact_value_u64(i);
-			shrink_indices[i] = lb_const_value(p->module, elem, idx).value;
+			shrink_indices[i] = lb_const_value(p->module, t_u32, exact_value_u64(i)).value;
```
corleypc, line 1531: Mask is a list of indices, must be an integer type, not the vector element type.

PR #7551, opened 2026-09-12
core/sys/linux: Added O_TMPFILE as Open_Flags.TMPFILE.

CzechBlueBear:
The O_TMPFILE flag (added to Linux in kernel 3.11) is very useful for creating anonymous temporary files (without it, the programmer must do a ritual dance with mkstemp()). Added the flag to the core/sys/linux/bits.odin. Please consider this change - thanks! :)


CzechBlueBear:
Withdrawing this request for now - it seems there is something strange with the O_TMPFILE constant in some (older) versions of Linux, causing the call to always fail with EINVAL - investigating. Apologies...

PR #7557, opened 2026-09-13
Fixes _stderr_write in both WASM os specific files to use fd 2 instead of fd 1

thomasefbland:
This is a small two character change, across two files, changing the _stderr_write functions to use fd 2 instead of 1.

While compiling to js_wasm32, I noticed that using panic resulted in calling console.log instead of console.error.
I was not sure if this was intentional, so I asked in Discord, explaining the details, and was told a patch was welcome.

I have included the exact text contents of the first message below the images, which explains the details.

image

image

> Hey, I did not want to jump to filing a bug report when I am not certain if it is intentional or not, especially for something so small, so wanted to ask here in case anyone knew.
>
> On Linux, macOS, Windows, and BSDs panic gets sent to stderr, aka fd 2 (for all but Windows obv).
>
> But then on wasm32-js and wasi its `write(1, data)`([source permalink](<https://github.com/odin-lang/Odin/blob/17cbeb68cf2319c49100c099acd93ad77fa8811b/base/runtime/os_specific_js.odin#L13>)) and `fd_write(1, {data}, &n)`([source permalink](<https://github.com/odin-lang/Odin/blob/17cbeb68cf2319c49100c099acd93ad77fa8811b/base/runtime/os_specific_wasi.odin#L38>)) respectively.
> This is despite them being inside functions named `_stderr_write`
>
> Does anyone know if this is intentional, because of something related to WebAssembly? Assuming only one fd or something like that?
>
> I ran into it when using panic in wasm32-js and at first thought it was something wrong with my `odin_env.write`.

Hopefully this small patch is acceptable. Thanks.


Kelimion:
https://github.com/odin-lang/Odin/blob/master/core/sys/wasm/js/odin.js#L1474 will also want console.error then? It's called with `true`  when `fd == 2`.

thomasefbland:
I personally think console.error is more idiomatic than coloring it red with %c, so I would +1 that.

I opted to use console.error in my implementation of odin_env.

thomasefbland:
We should not merge in current state, the writeToConsole does a lot of css based-formatting that I need to review more carefully if we want to swap to .error, otherwise it is a bit pointless IMO

Kelimion:
I've merged the original commit 315725197f5bab85f677552a0368916988f31478.

PR #7595, opened 2026-09-19
Fix #7547: double-NUL-terminate the SHFileOperationW path list in os.remove_all

### LINKED ISSUE #7547: os.remove_all on Windows: path passed to SHFileOperationW is not double-NUL terminated (out-of-bounds read, intermittent crash)
## Summary

`os.remove_all` on Windows passes a **single-NUL-terminated** path string as `pFrom` to `SHFileOperationW`, but that API requires its path lists to be **double-NUL terminated**. The operation therefore reads past the end of the path buffer until it happens to find another zero `u16` — undefined behaviour that can show up as anything from silent success to bogus errors to a fatal access violation. We hit the fatal variant in a real project's test suite (Windows 11 x64, `dev-2026-07-nightly`); the crash disappeared as soon as we stopped using `os.remove_all`. The same code path is still present in current `master` (checked 2026-09-12).

## Location

`core/os/path_windows.odin`, `_remove_all`:

```odin
dir := win32_utf8_to_wstring(path, temp_allocator) or_return

empty: [1]u16

file_op := win32.SHFILEOPSTRUCTW {
	nil,
	win32.FO_DELETE,
	dir,                          // pFrom: n code units + ONE NUL, then heap garbage
	cstring16(&empty[0]),         // pTo: single-zero stack array (single NUL)
	win32.FOF_NOCONFIRMATION | win32.FOF_NOERRORUI | win32.FOF_SILENT,
	false,
	nil,
	cstring16(&empty[0]),
}
res := win32.SHFileOperationW(&file_op)
```

`core/os/file_windows.odin`:

```odin
win32_utf8_to_utf16 :: proc(s: string, allocator: runtime.Allocator) -> (ws: []u16, err: runtime.Allocator_Error) {
	// ...
	text := make([]u16, n+1, allocator) or_return
	// ...
	text[n] = 0
	for n >= 1 && text[n-1] == 0 {
		n -= 1
	}
	ws = text[:n]                  // exactly one NUL lives at text[n]
	return
}

win32_utf8_to_wstring :: proc(s: string, allocator: runtime.Allocator) -> (ws: cstring16, err: runtime.Allocator_Error) {
	ws = cstring16(raw_data(win32_utf8_to_utf16(s, allocator) or_return))
	return
}
```

So `pFrom` points at `n` code units, one `0`, and then unallocated/uninitialized memory. The Win32 documentation for `SHFileOperationW` treats `pFrom` as a NUL-separated path list that must end with **two consecutive NULs**, so the API keeps reading `text[n+1], text[n+2], ...` past the end of the allocation.

## Why it usually appears to work

Freshly committed heap pages are zero-filled, so the word right after the buffer is often `0x0000` and the scan stops immediately. When the following bytes are non-zero (reused heap chunks, adjacent allocations), the API parses the garbage as additional source paths — observed effects range from strange `Error`s to a fatal signal. That makes the bug look flaky/machine-dependent while actually being an unconditional out-of-bounds read.

## Suggested fix

`pFrom` (and `pTo`, if non-NULL) must be double-NUL terminated. The smallest fix that repairs every `win32_utf8_to_wstring` consumer is to give `win32_utf8_to_utf16` two trailing NULs:

```odin
text := make([]u16, n+2, allocator) or_return
// ...
text[n]     = 0
text[n + 1] = 0
```

The extra NUL is invisible to all existing single-NUL consumers. Alternatively, `_remove_all` can build its own double-terminated buffer explicitly. For `FO_DELETE`, `pTo` is not used and could simply be `nil`, which would also remove the single-NUL stack array.

## Environment

- Windows 11 (10.0.26200), x64
- `dev-2026-07-nightly` — crash observed here
- `master` (2026-09-12) — same code confirmed in `core/os/path_windows.odin` / `core/os/file_windows.odin`

chillenious:
Fixes #7547.

`_remove_all` on Windows passed the converted path to `SHFileOperationW` with the single NUL terminator produced by `win32_utf8_to_wstring`. `pFrom` is read as a *double*-NUL-terminated path list, so the shell operation scans past the buffer until it finds a zero u16 behind it — an unbounded out-of-bounds read. Fresh arena memory is zero-filled, which usually hides the scan; on recycled memory it can surface as a bogus error return, a fatal access violation, or a hang (see the issue for an access-violation reproduction in a project's test suite).

**Change:** build the `pFrom` buffer locally as the path followed by two NUL units, exactly as the API documents. `win32_utf8_to_utf16`'s slice excludes its terminator; the freshly `make`d trailing pair is zero, so the copy leaves both NULs in place.

**Test:** `tests/issues/test_issue_7547.odin` drives the documented failure scenario — repeated non-empty-tree removals from a test worker thread with path-shaped allocation churn in the thread's `core:os` temporary arena — registered in `tests/issues/run.bat`/`run.sh`. Because the observable fault of the unterminated scan depends on heap layout, the test asserts the post-fix contract (every removal reports success and deletes its tree) rather than a specific crash.

**Verification:** the new issue test is green, and the full `tests/core/normal.odin -all-packages` suite passes with the change (614 tests). Independently reproduced the fault (fatal access violation and livelock under `odin test` worker threads) on `dev-2026-08` and confirmed the identical code on master.


chillenious:
Closed because I don't understand the code base well enough to recommend.

bplu4t2f:
Looked plausible at first glance, although one could certainly do the conversion directly into a sufficiently large buffer, rather than copying it an additional time.

I might look into this over the next days...

chillenious:
My crashes only ever appeared in Odin test bodies deferring os.remove_all over non-empty temp trees so I wasn't able to create a minimal test case to prove.

PR #7602, opened 2026-09-20
Adds lifetime markers for named locals

corleypc:
This PR implements LLVM lifetime markers for named locals. The feature is opt-in with the -lifetime-markers compiler flag. It requires -o:size or higher.

Benefits
- Lifetime markers allow LLVM's StackColoring pass to reuse a local's stack slot once the local is dead. Largest benefit when big on-stack variables are used in non-overlapping scopes in a procedure. Leads to smaller stack frames, possibly less CPU cache thrashing. On Windows, stack frames above 4KB need a __chkstk() call inserted, so some procedures may shrink their stack frames below 4KB and avoid it.
- The dead store elimination pass can use the lifetime.end marker to remove dead stores.
- Future: lifetime markers are required if ASan's use-after-scope option gets enabled.

Details
- Scopes that put markers on named locals: procedure body, loop/if/else bodies, {}, do, switch cases.
- Locals NOT being marked: proc params and named results, compiler generated allocas, locals declared outside a statement (e.g. `for i :=0;...`, or `if a := f() {}`), locals being assigned const literal slices or structs containing a const literal slice.
- The feature needs explicit opt in. Packages in core do use a local's address outside the declaring scope, **which is a miscompile under this flag**. At least core:crypto and core:encoding/json do this. It is perhaps a good idea to fix these. (If asan gets the use-after-scope option enabled (the compiler has TODOs about doing it), these will trigger.)
Here is an example from `core:encoding/json`  (the address of key_cstr used outside of the declaring switch case scope):
```odin
			key_ptr: rawptr
			#partial switch tk in t.key.variant {
				case runtime.Type_Info_String:
                    ......
					key_cstr: cstring
					if reflect.is_cstring(t.key) {
						key_cstr = cstring(raw_data(key))
						key_ptr = &key_cstr
					}
                    ......
			}
			set_ptr := runtime.__dynamic_map_set_without_hash(raw_map, t.map_info, key_ptr, map_backing_value.data)
```

Here is an example from `core:crypto` (the address of tmp_ctx used outside its declaring scope):
```odin
	ctx := ctx
	if finalize_clone {
		tmp_ctx: T
		clone(&tmp_ctx, ctx)
		ctx = &tmp_ctx
	}
	defer(reset(ctx))
```

- The flag doesn't work with the -sanitize:address flag, cause it may interfere with the manual unpoisoning the compiler currently does under that flag.
- Lifetime ends are emitted in lb_emit_defer_stmts, after the defers are built. Also, a missing end (unlike incorrectly placed end) is not a miscompile, the local gets to live for the whole procedure.
- Anything promoted to registers by the optimizer obviously isn't on the stack and doesn't get markers.

Tested on Windows 10 / LLVM 20.1 and Linux / LLVM 22.1. Tests added to tests/issues/test_lifetime_markers.odin (set to run at -o:size and -o:speed).


Kelimion:
Interesting feature. Is it available from the lowest LLVM version we support?

corleypc:
Yes, LLVM 17 has them. 
These are also on by default in clang at -o1 and higher, since local access after its scope is UB in C/C++.

PR #7603, opened 2026-09-20
Impl simd arm neon zip unzip extract

alexthed1rk:
All the integer arm neon vector extract, zip and unzip.


alexthed1rk:
idk why git deletes and adds 1kloc, in commit history it is clear that I haven't changed a thing there.

PR #7606, opened 2026-09-21
Make `lookup_env_alloc` return `Error` (like `lookup_env_buf`)

bplu4t2f:
Laytan left some TODO notes regarding unhandled allocator errors in various `_lookup_env_alloc` code paths.

`lookup_env_alloc` is defined to return `(value: string, found: bool)`.

`lookup_env_buf` is defined to return `(value: string, err: Error)`.

There is also the easy, "I don't care about errors" `get_env` alternative, so it does feel like the `lookup_env` variant is supposed to be a little more precise.

I drafted this to see how far-reaching the changes would be if `lookup_env_alloc` also returned `Error`. It would break quite a bit in other `core` packages and examples, but nothing insurmountable.

Not sure if reasonable...


core/os/env_wasi.odin
```diff
@@ -67,15 +67,19 @@ delete_string_if_not_original :: proc(str: string) {
...
 	if err := build_env(); err != nil {
-		return
+		return "", .Env_Var_Not_Found
 	}
 
 	sync.shared_guard(&g_env_mutex)
 
-	value = g_env[key] or_return
-	value, _ = clone_string(value, allocator)
+	value, found = g_env[key] or_return
```
Kelimion: I think the `or_return` makes sense when the last return value is a `bool`, but here now assigned to `found` and checked on the next line.

bplu4t2f: @Kelimion Oh yeah good catch.

This is WIP anyway because various dependencies in `core` need fixing.

I don't know if I should invest more time in it - depends on what Bill thinks the API should be.

bplu4t2f:
I also made a decision on what `env_lookup("", ...)` does: It returns `"", .Env_Var_Not_Found`.

Previously, this was inconsistent between implementations.

PR #7610, opened 2026-09-21
Fix #7547 (double-null-terminate path in SHFileOperationW)

bplu4t2f:
- See: https://github.com/odin-lang/Odin/issues/7547
- See: https://github.com/odin-lang/Odin/pull/7595

This implementation performs the full conversion directly rather than copying an additional time.

Note that the `empty` string buffer also should have been double-null-terminated. However, MSDN says that the unused `pTO` parameter should explicitly be set to NULL (rather than a double-null-terminated empty string) if not used (which is the case here). The `lpszProgressTitle` parameter got the same treatment because it is also not used.

Maybe we can try hunting down some more deterministic reproduction code for the test...

Side note: Users of `core:sys/windows` may also want a way to convert to double-null-terminated strings to interact with certain APIs. Currently, that needs manual attention.


bplu4t2f:
I was trying to come up with a clean, deterministic regression test, but I see now what @chillenious meant by `My crashes only ever appeared in Odin test bodies deferring os.remove_all over non-empty temp trees so I wasn't able to create a minimal test case to prove.`

My guts tell me you can provoke this by calling `os.remove_all` in a weird scope exit scenario, or maybe re-entrantly from inside the result allocator callback somehow.

It's obviously possible if the `global_default_temp_allocator_arenas` get poisoned, and as far as I'm aware there is an ongoing debate whether arenas should free their memory on reset or alloc - that would also affect this.

For the time being I'll just leave it be.

Also @chillenious - I used parts of your commit but you had deleted your branch in your fork. So you're formally not in the git log. If you want, I can try rebasing on the deleted commit (providing github would let me do that?).

